PR #201, opened 2014-09-10
Fix headers

stweil:
This pull request combines two previous ones which are still pending (waiting for feedback from Goobi e.V. management). It adds a new patch which is needed because new files with the invalid FSF address were added recently.

Please consider pulling all three patches soon.

Regards
Stefan


stweil:
Anything new on this issue? The code base still contains more than 400 files which refer to launchpad.net (which was replaced by github), and more than 500 files show an invalid address of the Free Software Foundation. Both modification are not license changes, but update or fix wrong information.

I don't think the correction is necessary for 1.10.x, but the current version of goobi-production should be fixed. As soon as there is consensus on that, I can send an updated pull request.

Here some statistics:

    # goobi-production 1.10.x
    $ git grep -l Temple|wc
    525     525   28587
    # goobi-production 1.10.x
    $ git grep -l Temple|wc
    562     562   31492

So even new files get added with the wrong FSF address. Luckily the number of files which mention launchpad decreased by 3 percent. :-)

henning-gerhardt:
Sorry, still no changes possible.

sebastian-meyer:
The Goobi e.V. managing board has decided to not change anything about the license headers prior to a legal workshop which takes place in the beginning of April. After that we will consider merging your proposed changes as we feel confident that they are purely cosmetic from a legal point of view.

Thank you for your well appreciated contribution and patience!

henning-gerhardt:
Closed because pull request against old 1.10.x branch. New pull request as #336.

PR #307, opened 2015-03-06
Replace dubious MD5 helper code by the original code

stweil:
The code no longer is dubious as I found and contacted the author of the original code.
He also released the code under the LGPL which is compatible with our license.

Should I format the code in a 2nd step? Or remove unused parts of the code?

Regards
Stefan


claussni:
Thanks Stefan! Formatting and cleansing would be appreciated.
I'm not entirely sure about the licensing here. Can we just use the (yet wrong) standard header? Or does the file require it's own?

stweil:
Hello Ralf,

adding the current standard header seems wrong for imported code:
That code is not (c) 2014 Goobi. Digitalisieren im Verein e.V.
Only (major) modifications could get us the right to add our own copyright.

Regards
Stefan

henning-gerhardt:
Looks ok, but like for MD4 I will provide first a test for current implementation so that the new implementation works like old one.

stweil:
As both implementations are identical in the relevant parts (only code location, formatting and copyright are different), you will get identical results. For MD4 this also applies.

You can compare old and new implementation with "diff --ignore-all-space" to suppress at least some of the changes caused by code formatting.

claussni:
Implementations match, but I still prefer providing a Test. Are there tests in the original programs?

stweil:
The class includes a main method with a very simple test which calculates the MD5 of "Tests" and which is disabled in the original code by comment delimiters. It could be enabled, and you can improve it by calculating the MD5 of command line arguments (parameter args).

claussni:
Nah, that's not automated. Tests should run on ```ant test```. @henning-gerhardt will write a proper test.

@stweil We need to make sure that there is no main() method flying around. Could you remove this commented-out code in compliance to the license?

stweil:
@claussni I suggest first applying the new test code for MD4 and MD5. Then I must update my pull requests because the import statements in the tests need to be fixed for the new file locations. With this update I'll also add a patch which removes the main method.

claussni:
Absolutely. That's how we will do it.

claussni:
https://github.com/goobi/goobi-production/commit/79433a60f281a9f07bd33b86f4d9519c375ec074 adds MD5 test. Please adjust your solution.

stweil:
Now the updated patch with support for MD5 test is ready for pulling.

PR #308, opened 2015-03-06
Replace dubious MD4 helper code by the original code

stweil:
The code was taken from VirgoFTP. Downloadlink:
http://qftp.cvs.sourceforge.net/viewvc/qftp/VirgoFtp/src/edu/sysu/virgoftp/ftp/encrypt/MD4.java

Andrew Tridgell published it under GPL2 or later.

Fix all related import statements.


henning-gerhardt:
Looks ok, but I will provide first a test case for current implementation. It would be nice if you could afterwards change your pull request for this test. So we can prove that the new implementation works like the old implementation.

claussni:
https://github.com/goobi/goobi-production/commit/471c2c3afe542d34e0ebc895e92f408eb8508fed provides MD4 test. Please adjust your solution.

stweil:
The first patch was updated to support the new test code - now ready for pulling.

PR #318, opened 2015-03-17
Adding tests for MD4 calculation.

henning-gerhardt:


Goobi/test/src/de/sub/goobi/helper/encryption/MD4Test.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+ * Place, Suite 330, Boston, MA 02111-1307 USA
+ *
```
stweil: This address is still wrong - why use it for new code?

PR #322, opened 2015-03-19
Fix cause of possible NullPointerException

matthias-ronge:
Catching the NoSuchAlgorithmExceptionc and continuing would cause a NullPointerException when trying to get encryptMsg.length in for() loop in line 94 whose cause is hard to track back then. Better way to get rid of an unwanted checked exception is by wrapping it into a RuntimeException.


henning-gerhardt:
Thanks for catching this. I think that this should never happened that MD5 algorithm is not provided by JRE / JDK. So I would merge this in next release.

PR #323, opened 2015-03-19
Fix name of Goobi e.V. in headers

stweil:
Replace also the unneeded HTML character entities.

This typo is even added with new files. The best way to eliminate more copy-and-paste typos
seems to fix all occurrences.


henning-gerhardt:
Get an OK from management.

PR #325, opened 2015-03-25
Fix some Javadoc warnings

stweil:
These patches reduce the number of warnings from 147 to 96.
They also change the copyright from GDZ to Goobi e.V. (1st patch, can optionally be omitted).

### DIFF
--- a/Goobi/src/de/sub/goobi/beans/Batch.java
+++ b/Goobi/src/de/sub/goobi/beans/Batch.java
@@ -307,8 +307,7 @@ public class Batch {
 	 *
 	 * @param processes
 	 *            collection containing elements to be removed from this set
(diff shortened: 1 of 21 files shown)


matthias-ronge:
I think, empty `@param name` and `@throws NameOfException` lines, or those just repeating the variable name (`@param name the name`) don’t provide anything helpful and just increase the size of the source files. Wouldn’t it be better to just remove them?

stweil:
@matthias-ronge Those lines can later be enhanced by more documentation...

I updated the patches now after feedback from @henning-gerhardt.

stweil:
Now the pull request no longer updates the copyright, adds the author tags which were removed from some methods to the related class and contains one more patch for void functions.

Please pull.

Thanks,
Stefan

PR #328, opened 2015-03-27
Fixes “EmptyTask.clone() does not call super.clone()”

matthias-ronge:
The function is a misnomer. Theads are like matchsticks for one-time use only and cannot be cloned. A burnt down thread needs to be replaced, this is what the implementation already does, so the function was renamed to make clear its behaviour differs from the contract of java.lang.Object#clone().

### DIFF
--- a/Goobi/src/de/sub/goobi/helper/ScriptThreadWithoutHibernate.java
+++ b/Goobi/src/de/sub/goobi/helper/ScriptThreadWithoutHibernate.java
@@ -67,7 +67,7 @@ public class ScriptThreadWithoutHibernate extends EmptyTask implements INameable
 			function = "executeAllScriptsForStep";
 		} else if (step.isTypExport()) {
 			function = "executeDmsExport";
-		} else if (step.getStepPlugin() != null && step.getStepPlugin().length() > 0) {
+		} else if ((step.getStepPlugin() != null) && (step.getStepPlugin().length() > 0)) {
 			function = "executeStepPlugin";
 		}
 		List<String> parameterList = new ArrayList<String>(1);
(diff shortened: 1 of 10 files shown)


stweil:
s/Theads/Threads/ in the commit message.

Your patch adds a lot of additional unrelated changes which make the code more readable but which are not strictly necessary (adding parentheses). I'd put those in an extra commit.

Cheers
Stefan

matthias-ronge:
Scott Ambler’s *Coding Standards for Java*, which are referenced as binding from *Leitfaden für Goobi-Entwickler*, request them (cf. [section 2.4.7](http://www.ambysoft.com/downloads/javaCodingStandards.pdf#page=21)), so I enabled auto-setting of missing parentheses for Goobi.Production and they will be added as I touch any class file. Same is true for missing braces.

Adding the parentheses separately isn’t productive as—as far as I know—it would require to open and safe file by file manually.

henning-gerhardt:
It is a conflict between automatic code correction and a clean commit structure :( I don't like it too but I don't know productive way to solve this.

One big solution could be: apply and fix code style, merge all this changes and continue to work.

stweil:
I know @matthias-ronge 's problem because I have a similar problem in this and other projects. My personal approach how to solve it is like this:

If I know that I have to fix files a, b and c, then I first open these files and store them (depending on the editor, it may be necessary to do some dummy modification, save, revert the modification and save again). Then I commit any changes which were applied automatically by my editor.

Now I can add my fixes and other manual modifications and commit those changes, too.

This is a little bit more work for me, but makes code reviews much easier. Most open source communities won't accept patches with unrelated modifications. Usually nobody will complain if there are only one or two unrelated modifications which only change the code formatting, so common sense applies here.

The big solution mentioned by @henning-gerhardt is not needed here.

henning-gerhardt:
Maybe @stweil workflow is a doable way. I will merge this as I found nothing else to discuss

matthias-ronge:
Putting an extra commit in the same branch will not help as the final diff will be the same. Creating another branch for just fixing the braces, indent, … and whatever in two files, upload it, create pull request, do something else because I have to wait for it was merged, remember to come back to this, pull main project again, create a branch to fix the things I want to … seems too much for both developer & release management. And then there is a third file involved that I didn’t think would be and we are in the same mess again.

I think we need clear rules (parentheses, braces, line width, imports, `this.` prefix yes or no, …) and do a complete code clean up − and perhaps repackaging too.

henning-gerhardt:
In general I used both: single commits and final diff. 

Single commits should reflect changes in commit messages and hopefully nothing more. I comment most on commit statements, sometimes on single lines or block of code.

Final diff is for indent and end-of-line changes which are not visible through webinterface.

If first commit is code style cleanup and following commits are changes in this files too, it would be OK for me.

General code style clean up would be a total stopper for any pull request until clean up is done.

PR #330, opened 2015-03-27
Remove more unused jsp files

stweil:
Those files were removed with this short script:

for file in `find -name "*.jsp"`; do \
 jsp=`basename $file`; git grep --quiet $jsp || git rm $file; \
done


matthias-ronge:
*meminfo.jsp* does work standalone. I don’t use it but perhaps someone does.

henning-gerhardt:
``meminfo.jsp`` provide an easy access to memory usage. I didn't use much but sometimes if I had no direct access to the server and users complain about bad performance I can look after memory usage.

``version.jsp`` contains old information (version: 0.94 from 2006-01-04).

stweil:
@matthias-ronge, @henning-gerhardt, thanks for your comments. As this is not a high priority patch, we can simply wait for more feedback. In two or three weeks, all 9 files can either be removed, or I'll update my pull request to keep those which are still useful.

henning-gerhardt:
I tried to use / access ``/newpages/version.jsp`` or ``/newpages/version.jsf`` but I get only error messages / wrapped faces exceptions. I think some work must be done to get this a part of this files useable. So best way is to remove them before next release. If somebody needs or missing them in next release, this file or files could be restored.

matthias-ronge:
I can confirm that `version.jsp` doesn’t work here either. (Btw., URL must be like `…/newpages/version.jsf`, not `.jsp`.) However, the error message reads:
> This is not a bug. Please see: http://wiki.apache.org/myfaces/Upgrading_to_Tomahawk_1.1.3
I think this file should either be removed or fixed.

stweil:
The pull request was updated to be compatible with the latest version
(some of the removed files also had a header change which resulted
in merge conflicts).

henning-gerhardt:
@stweil : Can you please update this pull request? I could merge and fix this conflict manual but I don't want to break your signed commit.

stweil:
@henning-gerhardt The pull request is now up-to-date again.

henning-gerhardt:
@stweil Thank you. I will merge this pull request. If somebody needs `meminfo.jsp` / `meminfo.jsf` then this file could be re-added.

henning-gerhardt:
Maybe we must re-add `newpages/error.jsp`. Reason is that in `WEB_INF/web.xml` error page is defined for `newpages/error.jsf`, see https://github.com/goobi/goobi-production/blob/1.11.x/Goobi/WEB-INF/web.xml#L327-L331

PR #338, opened 2015-04-09
Add missing licence header

henning-gerhardt:
Add license header to two files. Taking over from old 1.8.x branch and from intranda fork.

@claussni and @stweil : can you please look over this?


stweil:
The result looks good. I did not compare the license texts with the original version, but believe that copy+paste works.

PR #346, opened 2015-04-28
Fix typos and case of English text

stweil:


henning-gerhardt:
@stweil: Please split this pull-request in three separate pull-requests. FindBugs fixed should be a standalone pull-request. Commit "helper: fix typo" and "Fix case of English text" could be an other. A third pull-request should contain fixes / corrects in css/jsp files (your commit "Fix character encoding (Umlaut)".

stweil:
See also http://www.odi.ch/prog/design/newbies.php#4.

stweil:
@henning-gerhardt: Updated pull request to include only patch 2 and 3. For the other two patches I'll send separate pull requests as suggested.

PR #358, opened 2015-05-27
Merge: Make eclipse compatible

matthias-ronge:
I finally succeeded to have both build/execute **and** git functionality in the same project in Eclipse. This requires a different `.project` file. If possible, this file should not be part of the repository and be listed as to be ignored to git. Some more ignored files are needed.

Please merge if this is okay for you.


henning-gerhardt:
Just a question: Inside of `Goobi` directory are some files and a directory which looks to belong for Eclipse:

- `.settings/` (directory, containing two files)
- `.classpath`
- `.fatjar`
- `.pmd`
- `.project`
- `.tomcatplugin`

They are still needed?

matthias-ronge:
I don’t need them. They seem to remain from the time when the Goobi/ subfolder was the root folder; but this former design made it impossible to run the project as a web project without lots of symbolic pointer acrobatics in Eclipse because the web content (jsp files root) *must* for Eclipse be in some subfolder, which now is the case. (Its much smoother now, auto-detecting library changes etc.) I **never** used these files (by not explicitly linking them).

henning-gerhardt:
So this files and one directory could be removed?

PR #360, opened 2015-05-27
Remove obsolete Eclipse project files.

henning-gerhardt:
Remove not used and obsolte Eclipse prject files.

@matthias-ronge : As a well known Eclipse user: can you prove / test that this files could be removed?


matthias-ronge:
Thank you for the hint. The files

    Goobi/.classpath
    Goobi/.project
    Goobi/.settings/*

seem to originate to some Eclipse, but do not make sense in a subfolder and currenty do not seem to have a meaning.

What these files are for I do not know:

Goobi/.fatjar
Goobi/.pmd
Goobi/.tomcatplugin

I do not have use for them. I would suggest to remove all these files.

henning-gerhardt:
Thank you. I will merge this and hope nobody complain about this removing :-)

PR #365, opened 2015-11-09
Merge Goobi/.gitignore into .gitignore

stweil:
* Remove duplicate entries
* Sort entries alphabetically
* Use leading slash for all entries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,9 @@
-/work
+# Ant.
 /build
-dist/
-/WEB-INF/classes 
-build.properties
-deploy.properties
-config-local
-/.project
-/.classpath
-/.settings/
+/build.properties
+/config-local
+/deploy.properties
+/dist
+
+# Eclipse (Goobi deployed as root of web application).
+/Goobi/META-INF/
--- a/Goobi/.gitignore
+++ b/Goobi/.gitignore
@@ -1,7 +1,0 @@
-/private
-/.project
-/dist
-config-local
-/work
-/build
-/META-INF/
--- a/Goobi/WEB-INF/.gitignore
+++ b/Goobi/WEB-INF/.gitignore
@@ -1,1 +1,0 @@
-/classes


henning-gerhardt:
I'm not so familiar with directory ignore. Some merged values looks strange as excluded directories will never exists in this place.

stweil:
Which values are you thinking of?

In my build environments, these entries are not needed and could be removed:
  * /.project
  * /.classpath
  * /.settings/
  * /WEB-INF/classes
  * /private
  * /work

I suggest doing this in a separate commit, because it is not related to the merge.

matthias-ronge:
Does not work for me. Directory `META-INF` in `Goobi` folder is not ignored.

stweil:
@matthias-ronge, how did you get Goobi/META-INF (which tool or process creates it)?
I don't get it when I clone goobi-production and run "ant all".

henning-gerhardt:
@stweil : Some directories are created and/or used depending on used IDE and how this IDE is setup. In my opinion should directories created and used by IDE set to ignore to global git ignore list and not to project git ignore list. But changing this is not easy as every developer setup / configured his IDE in a different way.

matthias-ronge:
I created it manually. When I run Goobi from within Eclipse I deploy the content of the folder `Goobi` as the root of the web application. This is the most simple way to get it running. However, Goobi expects a folder named `META-INF` to be there and to contain a file `MANIFEST.MF`. If it is missing, it will not start. (As far as I remember, the feature does not much aside from displaying one of the strings from that file somewhere on the start page.) See a picture which explains it better:

unbenannt

matthias-ronge:
Correction would be to replace the line

    /META-INF/

with

    /Goobi/META-INF/

The following more entries are files and folders created by Eclipse that should not be checked in:

    /.settings/
    /.classpath
    /.project

I do not know where these entries come up:

    /WEB-INF/classes
    /build.properties
    /config-local
    /deploy.properties
    /dist
    /private
    /work

stweil:
@henning-gerhardt, yes, files and directories which are created by private tools (anything which is not enforced by the project) can be ignored either in the user's gitignore or in .git/info/exclude.

I don't mind if we keep existing entries for such tools in .gitignore, but it would be good to add a comment. @matthias-ronge, for Eclipse, this addition in .gitignore should fix it:

    # Eclipse.
    /Goobi/META-INF

henning-gerhardt:
This discussion about Eclipse files reminds me on open pull-request #360 There are some Eclipse project files which could be removed but I did not use Eclipse I could not verify this.

henning-gerhardt:
Following directories are used by ant as build tool. Ant is used to build official packages (war and jar files) and this directories should ignoried in project git ignore file:

```
/build/ (directory; holds created class files)
/build.properties (file; build.properties.template is maintained in project)
/config-local/ (directory; used if other config files should be included in war file)
/deploy.properties (file; deploy.properties.template is maintained in project)
/dist/ (directory; holds created war and jar files)
```

stweil:
I pushed an updated patch (added comments, fixed entry for Eclipse).

henning-gerhardt:
After merging #360 following entries for Eclipse could be removed:

```
/.classpath
/.project
/.settings/
```

Directory `Goobi/WEB-INF' contains a `.gitignore` file with content:

```
classes/
```
File `Goobi/WEB-INF/.gitignore` should be removed too.

stweil:
@henning-gerhardt, @matthias-ronge: I now merged Goobi/WEB-INF/.gitignore as well, removed some more unused entries and updated the commit message. Please review.

henning-gerhardt:
Looks good. 

In my humble opinion entries `/private` and `/work` could be removed. Entry  `/WEB-INF/classes` should be renamed to `/Goobi/WEB-INF/classes` or removed too.

Just a questions: should directory entries contain a trailing slash or not? Directories `build`, `config-local` and `dist` are different handled as `Goobi/META-INF`.

stweil:
@henning-gerhardt: /dir matches files and directories, /dir/ only matches directories.
See http://stackoverflow.com/questions/17888695/difference-between-gitignore-rules-with-and-without-trailing-slash-like-dir-an

Technically it does not matter which variant is used for directories. Maybe using a trailing slash better documents that it is used for a directory.

stweil:
Patch updated. Removed additional entries as suggested by @henning-gerhardt.

henning-gerhardt:
@matthias-ronge : Any comments on current changes?

henning-gerhardt:
As long no comment is a good comment, I will merge this pull request.

matthias-ronge:
These enties were removed but should not have been:

    /.settings/
    /.classpath
    /.project

stweil:
@matthias-ronge, maybe you get those entries because of your local Eclipse configuration? Then you can add them to .git/info/exclude.

matthias-ronge:
Works for me.

Issue #367, opened 2015-11-12
Fix Apache Commons Collections


matthias-ronge:
Goobi may be affected by the [unserialize vulnerability in Apache Commons Collections](https://issues.apache.org/jira/browse/COLLECTIONS-580) which may allow arbitrary remote command execution on the host system.

Although no concrete attack scenario has come to our attention that could immediately be applied against a Goobi Production instance, the library should be patched.

matthias-ronge:
Merged as #368

stweil:
I suggest to re-open this issue as the current solution is only a hack which was good a year ago but not for the future. In the meantime, Apache has released new versions which should allow a clean fix for this issue.

matthias-ronge:
Doesn’t maven automatically pick the latest acceptable version?

matthias-ronge:
This fixes it for the 2.x branch. [Here](http://stackoverflow.com/q/34671338/1503237) is what should be considered for current master / Maven builds.

henning-gerhardt:
Fixed in both active branches. Closing.

stweil:
Is this sufficient? What about [encrypting the view state](https://www.alphabot.com/security/blog/2017/java/Misconfigured-JSF-ViewStates-can-lead-to-severe-RCE-vulnerabilities.html)?

stweil:
As encryption is enabled by default, this should be secure unless an installation disables it explicitly.

PR #368, opened 2015-11-12
Merge: Fix Apache Commons Collections (#367)

matthias-ronge:
Fix bug #367


henning-gerhardt:
From off-GitHub communication:
Class `InvokerTransformer` of package `org.apache.commons.collections.functors` are replaced by a empty body double which create a security exception if any method of this class is called. Modified class is attached as text file: [InvokerTransformer.java.txt](https://github.com/goobi/goobi-production/files/32696/InvokerTransformer.java.txt)

PR #371, opened 2015-11-12
Clean up code (Eclipse)

stweil:
When running "clean up" in Eclipse, missing annotations and superfluous type casts are two of the issues which get fixed. These patches contain the resulting code.


stweil:
I just added issue #373 for further discussion of Java versions.

PR #374, opened 2015-11-23
Fix Coverity CID 120038 (Dereference null return value)

stweil:

### DIFF
--- a/Goobi/test/src/de/sub/goobi/helper/encryption/MD5Test.java
+++ b/Goobi/test/src/de/sub/goobi/helper/encryption/MD5Test.java
@@ -64,7 +64,8 @@ public class MD5Test {
 		MD5 md5 = new MD5("");
 		for (String clearText: testData.keySet()) {
 			String encrypted = md5.getMD5(clearText);
-			assertTrue("Encrypted password doesn't match the precomputed one! ", encrypted.equals(testData.get(clearText)));
+			assertTrue("Encrypted password doesn't match the precomputed one!",
+				   encrypted != null && encrypted.equals(testData.get(clearText)));
 		}
 	}
 }


stweil:
I added one more patch which fixes Coverity errors.

henning-gerhardt:
Please don't mix different fixes. Instead offer more pull requests.

stweil:
@henning-gerhardt: Done.

henning-gerhardt:
Mhh, introducing white space as indent in a tab indent file.

stweil:
@henning-gerhardt, blanks are needed even in a tab indented file to align code (I aligned the arguments of assertTrue). Otherwise you can only indent with multiples of 8. That's one of the many reasons why I'd prefer banning tabs.

henning-gerhardt:
It depends on how you configure your preferred IDE or editor. Tabs could easily displayed as 4 space indents. 

This file is with tab indent so changes should not change indent style.

stweil:
@henning-gerhardt, yes, you can configure your editor to use tab indent with multiples of 4 spaces. That's one more reason why I don't like tabs in code: how do you tell everybody that they have to set their editors to tabwidth = 4? How is this done for the different kinds of editors for the single project (because other projects need a different setting)? Does it work with all editors? Does it also work with your printer? Does it also work with shell output? Or in the GitHub view? I'm afraid that the only answer for several of those questions is no.

If you use tabs, then the only common standard is setting tabwidth = 8, because that is the value it always had, so that's the only value which works with (nearly) any software and printer.

By the way, even with a tab width of 2 or 4, you cannot always align code with tabs only. "assertTrue(" is 11 characters, so you need 1 (tab width 2) or 3 blanks (tab width 4 or 8).

I only see two possible choices for indentation in Goobi.production:
* Use tabs with tab width 8, optionally followed by up to 7 spaces if needed for code alignment (that's what the Linux kernel uses).
* Don't use tabs. Indent with blanks, usually with multiples of 2 or 4 (that's what many other free projects use).

A blank followed by a tab is always wrong (I'm afraid that we have some of those nevertheless) and detected by Git.

matthias-ronge:
I would prefer to follow [Google Java Style](https://google.github.io/styleguide/javaguide.html#s2.3.1-whitespace-characters) guide here, which declares:

> […] the ASCII horizontal space character (0x20) is the only whitespace character that appears anywhere in a source file. This implies that:
> 1. All other whitespace characters in string and character literals are escaped.
> 2. Tab characters are not used for indentation.

I find this perfectly rational. Personally, I also prefer an indent of 4, not 8, and a maximum line length ≤ 110 because GitHub does not do well in horizontal scrolling.

stweil:
@matthias-ronge I agree.

henning-gerhardt:
I will merge this. Indent war should go to a separate issue and pull request.

PR #375, opened 2015-11-23
Fix Coverity CID 44602, CID 44700, CID 44719 (Explicit null dereferen…

stweil:
…ced)

Parameter stepOpenOnly expects a Boolean type,
so pass false instead of null.


henning-gerhardt:
Boolean vs. boolean (class versus type) Class Boolean accept null value but not type. Changing this should not change functionality.

PR #376, opened 2015-11-23
Revert "Fix Coverity CID 44602, CID 44700, CID 44719 (Explicit null dereferen…"

matthias-ronge:
As far as I see it, Coverity has failed here. The `Boolean` object is correcty null-checked and evaluated ([here](https://github.com/goobi/goobi-production/blob/1.11.x/Goobi/src/org/goobi/production/flow/statistics/hibernate/FilterHelper.java#L617-L626)). Changing this to `false` *is* a behavioural change, as it will add a condition for *istTemplate==false*.

Or am I missing something essential here?

Reverts goobi/goobi-production#375

### DIFF
--- a/Goobi/src/org/goobi/production/flow/statistics/hibernate/UserDefinedFilter.java
+++ b/Goobi/src/org/goobi/production/flow/statistics/hibernate/UserDefinedFilter.java
@@ -188,7 +188,7 @@ public class UserDefinedFilter implements IEvaluableFilter, Cloneable {
 		 * part was exported to FilterHelper so that other Filters could access
 		 * it * --------------------------------
 		 */
-		String message = FilterHelper.criteriaBuilder(session, inFilter, crit, null, myParameter, false, null, true);
+		String message = FilterHelper.criteriaBuilder(session, inFilter, crit, null, myParameter, null, null, true);
 		if (message.length() > 0) {
 			myObservable.setMessage(message);
 		}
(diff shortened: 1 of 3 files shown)


henning-gerhardt:
You are sure? The changed parameter is `stepOpenOnly` and not `isTemplate`. Value of `stepOpenOnly` is used in [FilterHelper class  line 609](https://github.com/goobi/goobi-production/blob/1.11.x/Goobi/src/org/goobi/production/flow/statistics/hibernate/FilterHelper.java#L609). In called method `limitToUserAssignedSteps`  line 116 and line 152 using method parameter `stepOpenOnly`.

stweil:
@henning-gerhardt, yes. @matthias-ronge, you looked at the wrong parameter.

matthias-ronge:
You are right, I got the wrong one. However, if this code could never be executed and never caused a problem … is ist perhaps completely unused? If so it could also be deleted.

henning-gerhardt:
Parameter `stepOpenOnly` is used. Look at class [UserDefinedStepFilter](https://github.com/goobi/goobi-production/blob/1.11.x/Goobi/src/org/goobi/production/flow/statistics/hibernate/UserDefinedStepFilter.java).

matthias-ronge:
Yes, the prameter is used, that is why the code cannot ever have been working. It must have been throwing nullpointerexception. because we didn’t come over that I think the code is probably unused?

(however I close the pull request)

henning-gerhardt:
Boolean class is nullable [1], so never a NullPointerException is thrown. I think null value is casted to false for a boolean condition.

[1] http://stackoverflow.com/questions/11185321/when-should-null-values-of-boolean-be-used

PR #380, opened 2015-11-30
Pull 365 addon

matthias-ronge:
Local ignores are possible. So remove the one remaining and add documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,11 +4,3 @@
 /config-local
 /deploy.properties
 /dist
-
-# Eclipse
-/.project
-/.classpath
-/.settings/
-
-# Eclipse: /Goobi/ deployed as root of web application
-/Goobi/META-INF/
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,10 @@
 /deploy.properties
 /dist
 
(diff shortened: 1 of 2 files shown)


henning-gerhardt:
Ignoring `.properties` and `.xml` files inside `Goobi\src` is not a good idea because you loose changes for bean hbm xml files or message translations. I did not found any `.txt` and `.xsl` files inside `Goobi\src` so ignoring could be dangerous for any future update. 

Is ignoring for this file extensions really necessary?

matthias-ronge:
@henning-gerhardt The file exclude patterns are not applied to subdirectories. None of these files are on the repository of course.

@stweil The directory seems not to be created when checking out the github branch using the Eclipse plugin (at least on my Windows). I completely agree that it does not create empty directories and files as they only eat up disk space (though little).

matthias-ronge:
local

henning-gerhardt:
@matthias-ronge: I added temporally your proposed settings to my local development environment and you are right your changes are not affecting subdirectory entries.

stweil:
I suggest combining all 4 patches in one single patch.

matthias-ronge:
I always do many commits. Just ignore and read only the result.

PR #384, opened 2015-12-10
Do null check on java.io.File list() results

matthias-ronge:
Fixes a handful of issues where the results of a list() call on a directory were not checked for `null` before object access.

### DIFF
--- a/Goobi/src/de/sub/goobi/beans/Prozess.java
+++ b/Goobi/src/de/sub/goobi/beans/Prozess.java
@@ -405,7 +405,7 @@ public class Prozess implements Serializable {
 		if (tifOrdner.equals("") && useFallBack) {
 			String suffix = ConfigMain.getParameter("MetsEditorDefaultSuffix", "");
 			if (!suffix.equals("")) {
-				String[] folderList = dir.list();
+				String[] folderList = FilesystemHelper.list(dir);
 				for (String folder : folderList) {
 					if (folder.endsWith(suffix)) {
 						tifOrdner = folder;
@@ -421,7 +421,7 @@ public class Prozess implements Serializable {
 	                File tif = new File(tifOrdner);
 	                String[] files = tif.list();
 	                if (files == null || files.length == 0) {
-	                    String[] folderList = dir.list();
+	                    String[] folderList = FilesystemHelper.list(dir);
 	                    for (String folder : folderList) {
 	                        if (folder.endsWith(suffix) && !folder.startsWith(DIRECTORY_PREFIX)) {
 	                            tifOrdner = folder;
@@ -466,7 +466,7 @@ public class Prozess implements Serializable {
 		if (testMe.list() == null) {
 			return false;
 		}
-		if (testMe.exists() && testMe.list().length > 0) {
+		if (testMe.exists() && FilesystemHelper.list(testMe).length > 0) {
 			return true;
 		} else {
 			return false;
@@ -487,7 +487,7 @@ public class Prozess implements Serializable {
 			};
 
 			String origOrdner = "";
-			String[] verzeichnisse = dir.list(filterVerz);
+			String[] verzeichnisse = FilesystemHelper.list(dir, filterVerz);
 			for (int i = 0; i < verzeichnisse.length; i++) {
 				origOrdner = verzeichnisse[i];
 			}
@@ -495,7 +495,7 @@ public class Prozess implements Serializable {
 			if (origOrdner.equals("") && useFallBack) {
 				String suffix = ConfigMain.getParameter("MetsEditorDefaultSuffix", "");
 				if (!suffix.equals("")) {
-					String[] folderList = dir.list();
+					String[] folderList = FilesystemHelper.list(dir);
 					for (String folder : folderList) {
 						if (folder.endsWith(suffix)) {
 							origOrdner = folder;
@@ -511,7 +511,7 @@ public class Prozess implements Serializable {
 					File tif = new File(origOrdner);
 					String[] files = tif.list();
 					if (files == null || files.length == 0) {
-						String[] folderList = dir.list();
+						String[] folderList = FilesystemHelper.list(dir);
 						for (String folder : folderList) {
 							if (folder.endsWith(suffix)) {
 								origOrdner = folder;
@@ -1049,7 +1049,7 @@ public class Prozess implements Serializable {
 	private void removePrefixFromRelatedMetsAnchorFilesFor(String temporaryMetadataFilename) throws IOException {
 		File temporaryFile = new File(temporaryMetadataFilename);
 		File directoryPath = new File(temporaryFile.getParentFile().getPath());
-		for (File temporaryAnchorFile : directoryPath.listFiles()) {
+		for (File temporaryAnchorFile : FilesystemHelper.listFiles(directoryPath)) {
 			String temporaryAnchorFileName = temporaryAnchorFile.toString();
 			if (temporaryAnchorFile.isFile()
 					&& FilenameUtils.getBaseName(temporaryAnchorFileName).startsWith(TEMPORARY_FILENAME_PREFIX)) {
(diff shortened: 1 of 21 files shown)


stweil:
This replaces OO programming by a procedural style (static functions). You could also create a new class SafeFile based on File but with overloaded methods and use that new class.

What about the changes of import statements? Are they related to the commit subject, or should they go into a separate patch?

matthias-ronge:
I tried the `SafeFile` approach but findbugs didn’t recognize it as fix. I don’t know why. Give it a try if you want, here is the file:

	import java.io.*;
	import java.net.URI;
	import java.util.*;
	
	public class SafeFile extends File {
	
		private static final SafeFile[] NO_FILE = new SafeFile[0];
		private static final String[] NO_STRING = new String[0];
	
		public static SafeFile createTempFile(String prefix, String suffix) throws IOException {
			return new SafeFile(File.createTempFile(prefix, suffix));
		}
	
		public SafeFile(String pathname) {
			super(pathname);
		}
	
		public SafeFile(File path) {
			super(path.getPath());
		}
	
		public SafeFile(URI uri) {
			super(uri);
		}
	
		public SafeFile(String parent, String child) {
			super(parent, child);
		}
	
		public SafeFile(File parent, String child) {
			super(parent, child);
		}
	
		private SafeFile[] createAll(File[] files) {
			SafeFile[] result = new SafeFile[files.length];
			for (int i = 0; i < files.length; i++)
				result[i] = new SafeFile(files[i]);
			return result;
		}
	
		public static List<SafeFile> createAll(List<File> files) {
			ArrayList<SafeFile> result = new ArrayList<SafeFile>(files.size());
			for (File file:files){
				result.add(new SafeFile(file));
			}
			return result;
		}
	
		@Override
		public String[] list() {
			String[] unchecked = super.list();
			return unchecked != null ? unchecked : NO_STRING;
		}
	
		@Override
		public String[] list(FilenameFilter filter) {
			String[] unchecked = super.list(filter);
			return unchecked != null ? unchecked : NO_STRING;
		}
	
		@Override
		public SafeFile[] listFiles() {
			File[] unchecked = super.listFiles();
			return unchecked != null ? createAll(unchecked) : NO_FILE;
		}
	
		@Override
		public SafeFile[] listFiles(FilenameFilter filter) {
			File[] unchecked = super.listFiles(filter);
			return unchecked != null ? createAll(unchecked) : NO_FILE;
		}
	
		@Override
		public SafeFile[] listFiles(FileFilter filter) {
			File[] unchecked = super.listFiles(filter);
			return unchecked != null ? createAll(unchecked) : NO_FILE;
		}
	}

If you find this working (working here means that the issue is considered solved by coverity/findbugs, too) we could use it also.

Import statemants are automatically created from Eclipse. Does it need fixing?

stweil:
On a 2nd though, adding the null checks where needed might be the best solution.

matthias-ronge:
They are needed in all of these places − for the case that the directory does not exist or is not readable, or …

It’s a bit unfortunate that the `list()` functions return any `null` at all.

henning-gerhardt:
- I would prefer the `SafeFile` approach if possible. Its looks a lot nicer and cleaner.
- Clean up of import statements are handled by used IDE and how you configure your used IDE to collapse / summarize import statements. As there are no [performance issues](http://stackoverflow.com/questions/7128348/performance-difference-between-a-wild-card-import-and-the-required-class-import) between wildcard and explicit import statements, it should be ok to use wildcard import statements.

matthias-ronge:
I would prefer it too. Here is the branch where I tried it:
https://github.com/matthias-ronge/goobi-production/tree/io-file-npe-issues
But FindBugs here does not stop to report the issues. Or am I missing something?

henning-gerhardt:
Can not compile your branch with ant:

```
compile:
    [javac] Compiling 345 source files to <path>/goobi-production/build/WEB-INF/classes
    [javac]  <path>/goobi-production/Goobi/src/de/sub/goobi/beans/Prozess.java:30: error: cannot find symbol
    [javac] import org.goobi.io.SafeFile;
    [javac]                    ^
    [javac]   symbol:   class SafeFile
    [javac]   location: package org.goobi.io
... (a lot more compile errors)
```

henning-gerhardt:
After adding your `SafeFile` class inside correct directory and updating findbugs to version 3.0.1 I got  same findbugs complaints.

matthias-ronge:
That is what I found, too. Cannot say why.

henning-gerhardt:
Maybe findbug use hard coded error reporting for any list() call of a File class neverless if a extended class fix this behaviour?

henning-gerhardt:
I would go with the SafeFile class usage and ignore (maybe) wrong messages from findbugs.

stweil:
Why not add the missing null checks? I'd write a patch for that, but my FindBugs does not report those warnings. Are they also in the Coverity list?

henning-gerhardt:
Which version of findbugs did you use and how did you configure findbugs? I use 3.0.1 and I did not enable any filter (All bug ranks, All bug priorities, All issues, All time).
No reports on Coverity for this or I did not found any :(

matthias-ronge:
> Why not add the missing null checks?

This is what my patch does. Just not everywhere, but only one centraliseld method. ([Don’t repeat yourself.](https://de.wikipedia.org/wiki/Don%E2%80%99t_repeat_yourself))

matthias-ronge:
I am using the findbugs plugin from Eclipse, v. 3 0.1.20150306-5afe4d1

stweil:
@henning-gerhardt, after an update from Debian's FindBugs 2.0.3 to the latest FindBugs from GitHub, the number of reports increased from 444 to 1082. Among the new ones, I also get those related to the pull request here.

Goobi/src/de/sub/goobi/beans/Prozess.java
```diff
@@ -487,7 +487,7 @@ public boolean accept(File dir, String name) {
 			};
 
 			String origOrdner = "";
-			String[] verzeichnisse = dir.list(filterVerz);
+			String[] verzeichnisse = FilesystemHelper.list(dir, filterVerz);
 			for (int i = 0; i < verzeichnisse.length; i++) {
 				origOrdner = verzeichnisse[i];
```
stweil: It looks like more changes are needed here. The for loop looks very strange.

Goobi/src/de/sub/goobi/beans/Prozess.java
```diff
@@ -466,7 +466,7 @@ public Boolean getTifDirectoryExists() {
 		if (testMe.list() == null) {
 			return false;
 		}
-		if (testMe.exists() && testMe.list().length > 0) {
```
stweil: What about moving this if block into the try block?

matthias-ronge: And then…? Add a `catch (NullPointerException`? Not the nearest solution either, is it?

Goobi/src/de/sub/goobi/beans/Prozess.java
matthias-ronge: Could be simplified to `origOrdner = verzeichnisse[verzeichnisse.length];` but that is a different pull request then.

Goobi/src/de/sub/goobi/beans/Prozess.java
stweil: No, just to limit the scope of `testMe`.

Goobi/src/de/sub/goobi/beans/Prozess.java
stweil: Was that the intention? If so, this would be the correct code:

    if (verzeichnisse != null) {
        origOrdner = verzeichnisse[verzeichnisse.length - 1];
    }

matthias-ronge: You are right. I forgot the `-1`.

Goobi/src/de/sub/goobi/beans/Prozess.java
matthias-ronge: The problem in this line is not that `testMe` could be null, it can’t because of line 456, but the problem is that the result of `testMe.list()` could be null, so querying `lenght` on the result of `list()` could cause a `NullPointerException`.

matthias-ronge:
It seems that I’ve got the https://github.com/matthias-ronge/goobi-production/tree/io-file-npe-issues branch to work now. The problem was that I used a subclass, which still “is a” superclass. This was now changed. If you agree to further inspect the object oriented approach i would close this pull request later.

matthias-ronge:
Pull request for object orientend approcach: #397

PR #385, opened 2015-12-10
Allow ProjectTask to be serializable

matthias-ronge:
ProjectTask is referenced from Serializable class ProjectStatusDataTable.

### DIFF
--- a/Goobi/src/org/goobi/production/chart/ProjectTask.java
+++ b/Goobi/src/org/goobi/production/chart/ProjectTask.java
@@ -1,4 +1,7 @@
 package org.goobi.production.chart;
+
+import java.io.Serializable;
+
 /**
  * This file is part of the Goobi Application - a Workflow tool for the support of mass digitization.
  * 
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
Which kind of error did you try to fix?

henning-gerhardt:
Ok, but one last question: Should
```java
private static final long serialVersionUID = 1L;
```
contain an other value as `1L`? As far as I know this values are generated in some way.

henning-gerhardt:
I found some more serializeable classes:
- Goobi/src/org/goobi/production/model/bibliography/course/Course.java
- Goobi/src/de/sub/goobi/metadaten/copier/DataCopyrule.java
- Goobi/src/de/sub/goobi/persistence/BatchDAO.java
- Goobi/src/de/sub/goobi/helper/exceptions/GUIExceptionWrapper.java
- Goobi/src/de/sub/goobi/helper/ProjectHelper.java

with the `1L` value. So fixing this should occur for all this places.

PR #386, opened 2015-12-10
Remove always true condition

matthias-ronge:
Condition is known to be always true in current implementation. Was replaced by assertion which will not execute unless the flag is explicitly set.

### DIFF
--- a/Goobi/src/de/sub/goobi/helper/tasks/EmptyTask.java
+++ b/Goobi/src/de/sub/goobi/helper/tasks/EmptyTask.java
@@ -481,11 +481,8 @@ public class EmptyTask extends Thread implements INameableTask {
 	 *            the tasks progress
 	 */
 	public void setProgress(int progress) {
-		if (progress >= 0 || progress <= 100) {
-			this.progress = progress;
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
It is now needed to run Goobi application with -ea flag to get assertion error message?

matthias-ronge:
Yes, you need the flag `-ea` within the JVM start options to enable assertions. The idea is to have all of these checks (that can only be fixed by modifying the source code if they fail) only executing during development, and being skipped in productive environment.

henning-gerhardt:
I'm unsure with this change. There are no tests and only current usage of this class provides no error. 

If somebody use this method with wrong values at runtime there is no error thrown and maybe other code depends on correct values (in this case between 0 and 100).

@stweil and / or @claussni : any comments on this?

matthias-ronge:
Percentage is generated within code. There is no way to use the method with a wrong value aside from programming something new.

henning-gerhardt:
As nobody else complain, I will merge this pull request as I know now that this change will hurt us in future.

PR #387, opened 2015-12-10
Add hashCode() function for HistoryEvent

matthias-ronge:
HistoryEvent defines `equals()` but lacks a corresponding `hashCode()`.


henning-gerhardt:
At least Batch class and maybe more bean classes needs a hashCode() method too.

henning-gerhardt:
Fast grep on bean classes offer only Batch class with missing `hashCode()` method if `equals()` method is existing. Could be fixed in a separate pull request or this should be renamed.

matthias-ronge:
To be fixed separately…

PR #388, opened 2015-12-10
Correct lazy initialisation in readAllModulesFromConfiguration()

matthias-ronge:
Fix possible exposure of a partially initialized object


henning-gerhardt:
Are both `null` checks for `modulemanager` in line 105 and line 107 still necessary?

matthias-ronge:
Yes. See [here](https://de.wikipedia.org/wiki/Doppelt_%C3%BCberpr%C3%BCfte_Sperrung#Doppelt_.C3.BCberpr.C3.BCfte_Sperrung_in_Java).

Issue #65, opened 2014-01-31
[IP] Remove or repair swap functionality


sebastian-meyer:
Goobi provides an option to swap out finished processes for archiving. This function is broken as it triggers the restoration of the processes immediately after swapping them out.

matthias-ronge:
The swap function was removed in 3.x. The ticket can then be closed.

matthias-ronge:
I have given the <kbd>question</kbd> badge, because it has to be clarified whether this should still be implemented, and if so, which is the desired solution: repair function or removeit?

PR #391, opened 2015-12-13
Suppress compiler warnings (deprecation)

stweil:

### DIFF
--- a/Goobi/src/de/sub/goobi/beans/Prozess.java
+++ b/Goobi/src/de/sub/goobi/beans/Prozess.java
@@ -85,6 +85,7 @@ import de.sub.goobi.metadaten.MetadatenSperrung;
 import de.sub.goobi.persistence.BenutzerDAO;
 import de.sub.goobi.persistence.ProzessDAO;
 
+@SuppressWarnings("deprecation")
 @XmlAccessorType(XmlAccessType.NONE)
 // This annotation is to instruct the Jersey API not to generate arbitrary XML
 // elements. Further XML elements can be added as needed by annotating with
--- a/Goobi/src/de/sub/goobi/export/dms/AutomaticDmsExportWithoutHibernate.java
+++ b/Goobi/src/de/sub/goobi/export/dms/AutomaticDmsExportWithoutHibernate.java
@@ -65,6 +65,7 @@ import de.sub.goobi.persistence.apache.ProcessObject;
 import de.sub.goobi.persistence.apache.ProjectManager;
 import de.sub.goobi.persistence.apache.ProjectObject;
 
+@SuppressWarnings("deprecation")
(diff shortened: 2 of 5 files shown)


matthias-ronge:
Deprecations should be a means of documenting that there is something that needs to be fixed. For example, [Prozess.java, l. 573](https://github.com/goobi/goobi-production/blob/master/Goobi/src/de/sub/goobi/beans/Prozess.java#L573) produces the warning
> The method setShowMessages(boolean) from the type LongRunningTask is deprecated.

because `ProcessSwapInTask` is using a deprecated API that could be considered to be fixed along with resolving issue #65. It is however an acceptable decision to add `@SuppressWarnings("deprecation")` to indicate that it is planned to restrain from changing this in the future.

henning-gerhardt:
Why should this suppression done class wide? 

Deprecation warnings should be raised and hint the developer to remove old and / or deprecated API (methods and / or classes). Hiding this won't fix this automatically.

stweil:
I'd consider two different cases:

(1) Using a deprecated method from some external source. This should be fixed as soon as possible by replacing the method because the method won't be supported by future versions.

(2) Using a deprecated method from our own code. Then it is up to us when we will remove the method and fix the calling code.

As long as we don't decide to fix deprecated method calls, we can as well remove the compiler warning. The code which calls deprecated methods can still be found by a simply search for "deprecated".

Of course we should strive to get code with neither "deprecated" suppressions nor with such warnings.

henning-gerhardt:
Your cases are good but I did not see why hiding this messages could help. Searching for hidden messages are a wrong way imho: except persons who wrote this hiddings nobody know how to find them. 

There are so many hidden messages (undocumented configuration parameters for example) or features (mass import from an excel file, module manager, ocr url service, ...) so adding more such stuff would not help anybody.

A better way is 1) open an issue, 2) fix or repair such things, 3) close issue.

stweil:
Currently deprecation warnings are not explicitly enabled. The Java compiler only reports a summary like this:

    [javac] Note: Some input files use or override a deprecated API.
    [javac] Note: Recompile with -Xlint:deprecation for details.

I enabled those warning in my local build and would like to enable them by default.

matthias-ronge:
If the code uses a deprecated internal class it is highlighted accordingly so you want to fix it. At least Eclipse does:

high_light

If you put `@SuppressWarnings("deprecation")` for the class the highlighting is removed and the code looks okay. That is why I would vote against this.

henning-gerhardt:
Enabling this deprecation messages would be ok (maybe confusing for someone) but as @matthias-ronge  mentioned add suppress warning annotations on class level should be avoided or not done.

stweil:
I close this PR because there is no consensus to suppress deprecation warnings in the code.
There is consensus that compiler warnings should be fixed. Therefore such warnings should be addressed as an open issue (I'll create one).

PR #393, opened 2015-12-14
Add 'static' attribute to inner classes without reference to parent o…

stweil:
…bject

This fixes warnings from FindBugs.

### DIFF
--- a/Goobi/src/de/sub/goobi/forms/CalendarForm.java
+++ b/Goobi/src/de/sub/goobi/forms/CalendarForm.java
@@ -81,7 +81,7 @@ public class CalendarForm {
 	 * 
 	 * @author Matthias Ronge &lt;matthias.ronge@zeutschel.de&gt;
 	 */
-	public class Cell {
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
Interesting. 

My findbugs (3.0.1) found 3 classes
1) Cell inside CalenderForm
2) IssueOption inside Calenderform
3) Parameters inside UserdefindedFilter
and Coverity reports only for two 2) CID 44972 and 3) CID 44990

stweil:
As I noticed recently, newer versions of FindBugs are much better in finding problems (the number of findings increased more that twice when I switched from Debian's version 2 to latest FindBugs).
Obviously Coverity scan is conservative and does not use the latest version.

henning-gerhardt:
Do you want add the third class to this pull request or should maybe missing class fixed in an other pull request?

stweil:
I'd prepare another pull request to fix the remaining cases, so you can pull this one.

Issue #289, opened 2014-12-19
PDF creation failed


henning-gerhardt:
Florian Wille reported following error message on Goobi Community mailing list:

```
ERROR 2014-12-18 16:32:17,952 (GetMetsPdfAction.java:run:328)
         error during pdf generation (java.io.FileNotFoundException)

java.io.FileNotFoundException: 
/usr/local/goobi/metadata/13/images/usr/local/goobi/metadata/13/images/blawvon_672764172_tif/00000004.TIF 
(Datei oder Verzeichnis nicht gefunden)
     at java.io.FileInputStream.open(Native Method)
```

GetMetsPdfAction class is part of delivered WEB-INF/ics-1.1.5.jar file.

henning-gerhardt:
I tried PDF export with a very old Goobi process and its works for me. The main difference between this very old process and new stored processes is that inside mets filesection the path to image files is relative to image directory in old process while new processes contains a absolute path to image files. 

If I manipulate manual path information to relative in new (stored) processes then PDF exports works for this processes too.

Maybe the bug for this is in @matthias-ronge posted source code above: In [line 258](https://github.com/goobi/goobi-production/blob/1.11.x/Goobi/src/de/sub/goobi/export/download/ExportMets.java#L258) are image base path (containing general meta base path including process id and "images" string) and value of stored image path of meta.xml concated to a absolute path.

If image base path is f.e. ``/home/goobi/work/daten/<processid>/images`` and stored value inside meta.xml is ``/home/goobi/work/daten/<processid>/images/<processtitle>/<imagefile>`` then the resulting and later as error message announced string is ``/home/goobi/work/daten/<processid>/images/home/goobi/work/daten/<processid>/images/<processtitle>/<imagefile>`` like @matthias-ronge already showed above. If stored value but only a relative path like ``./<processtitle>/<imagefile>" then is resulting string ```/home/goobi/work/daten/<processid>/images/<processtitle>/<imagefile>`` which works for PDF export.

One question remains: why does this (part of code)[https://github.com/goobi/goobi-production/blob/1.11.x/Goobi/src/de/sub/goobi/export/download/ExportMets.java#L250-260] works correct for any other usage scenario?

frank-ulrich-weber:
In my installation the fileGrp within the meta.xml of a new process contains file-paths like the following:
file:/D:\Goobi_Data\metadata\68825\images\749696575_76787661X_0110_FUW_tif\00000001.tif
[...]

PR #394, opened 2015-12-14
Merge:  PDF creation failed (#289)

matthias-ronge:
Fix bug #289


henning-gerhardt:
There is a similar code block in [ExportMetsWithoutHibernate](https://github.com/matthias-ronge/goobi-production/blob/issue_289/Goobi/src/de/sub/goobi/export/download/ExportMetsWithoutHibernate.java#L226-L236) class. Should this be fixed too?

PR #399, opened 2015-12-15
Fix possible null pointer dereference

stweil:
FindBugs reports this warning:
	Possible null pointer dereference in method on exception path

Fix it by moving the relevant code into the try block.

### DIFF
--- a/Goobi/plugins/opac/PicaPlugin/org/goobi/production/plugin/CataloguePlugin/PicaPlugin/Catalogue.java
+++ b/Goobi/plugins/opac/PicaPlugin/org/goobi/production/plugin/CataloguePlugin/PicaPlugin/Catalogue.java
@@ -88,25 +88,24 @@ class Catalogue {
 	 */
 	private void parseIktList(String iktList) {
 		InputSource iktSource = new InputSource(new StringReader(iktList));
-		DocumentBuilder docBuilder = null;
-		Document document = null;
 		try {
-			docBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+			DocumentBuilder docBuilder =
+				DocumentBuilderFactory.newInstance().newDocumentBuilder();
 			//get xml-docment
-			document = docBuilder.parse(iktSource);
+			Document document = docBuilder.parse(iktSource);
+
+			//get all keys
+			NodeList keys = document.getElementsByTagName("KEY");
+
+			//go through all keys and put them into their hashmap
+			for (int i = 0; i < keys.getLength(); i++) {
+				picaToKey.put(((Element) keys.item(i)).getAttribute("mnemonic"),
+					keys.item(i).getFirstChild().getNodeValue());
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
It is not a fix to move code inside a try catch block to catch a possible null pointer exception.

stweil:
It is. The null pointer occurred because a local variable was set to null and accessed after the try / catch block, even if its value was not changed because an exception had occurred. Now it is only used in the try block where its value is set before use.

henning-gerhardt:
Whole method need some love:
- Every exception is catched but only `ParserConfigurationException`, `IOException`, `SAXException` and `IllegalArgumentException` are thrown.
- catched exception prints only a stack trace instead of logged error
- after catched exception code goes on and a possible null pointer exception occurs

IMHO if an exception are catched method should return and not change any value

henning-gerhardt:
Simple and correct fix is to add an`return` statement inside catch block.

stweil:
@henning-gerhardt, that's also a possible solution, but your fix requires an additional code line while my patch was able to remove two code lines (the null assignments).

henning-gerhardt:
I looked into java doc description of `parse()` method of `DocBuilder` class
```java
/**
     * Parse the content of the given URI as an XML document
     * and return a new DOM {@link Document} object.
     * An <code>IllegalArgumentException</code> is thrown if the
     * URI is <code>null</code> null.
     *
     * @param uri The location of the content to be parsed.
     *
     * @return A new DOM Document object.
     *
     * @throws IOException If any IO errors occur.
     * @throws SAXException If any parse errors occur.
     * @throws IllegalArgumentException When <code>uri</code> is <code>null</code>
     *
     * @see org.xml.sax.DocumentHandler
     */

    public Document parse(String uri)
        throws SAXException, IOException {
        if (uri == null) {
            throw new IllegalArgumentException("URI cannot be null");
        }

        InputSource in = new InputSource(uri);
        return parse(in);
    }
```

Used method returning a new Object or throw an exception in error case. If you catch the exceptions and let your method return in this case no future check is necessary. Or did I oversee something?

henning-gerhardt:
IMHO: try catch statements should be small as possible and only catch thrown exceptions and not programmer errors.

henning-gerhardt:
After analysing usage of this method with IDEA this method could be removed. `picaToKey` and `picaToDescription` are never used. I don't know if this an old feature or a never correct implemented feature.

stweil:
@henning-gerhardt, that's correct, the community edition does not use it. The Intranda edition does - that's why I hesitate to remove the code.

henning-gerhardt:
I see there no problem: If some changes as prober pull-request from Intranda edition come back to community edition this removed lines come back too. If changes go from community edition to Intranda edition this lines will be removed too.

Should we really look into every single branch / fork every time when we change something and cancel this change because somewhere something got broken which is maintained by somebody else? I don't think so.

stweil:
Closed because this patch is superseeded by #406.

PR #411, opened 2015-12-17
Performance: Don’t build strings if logging is off

matthias-ronge:
Check if logging is *on* for the given log level before starting to build the log message strings in memory.

### DIFF
--- a/Goobi/plugins/import/PicaMassImport/de/intranda/goobi/plugins/PicaMassImport.java
+++ b/Goobi/plugins/import/PicaMassImport/de/intranda/goobi/plugins/PicaMassImport.java
@@ -158,7 +158,9 @@ public class PicaMassImport implements IImportPlugin, IPlugin {
 
 		currentIdentifier = data;
 
-		logger.debug("retrieving pica record for " + currentIdentifier  + " with server address: " + this.getOpacAddress());
+		if (logger.isDebugEnabled()) {
+			logger.debug("retrieving pica record for " + currentIdentifier  + " with server address: " + this.getOpacAddress());
+		}
 		String search = SRUHelper.search(currentIdentifier, this.getOpacAddress());
 		logger.trace(search);
 		try {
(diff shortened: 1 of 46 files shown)


stweil:
That's a patch series with nicely ordered commits. :-)

Are there performance problems which justify this change? If yes: can you provide some timing data? One drawback is that the conditional code is only executed if logging is enabled, so potential run time error won't be detected (and fix). This phenomen is called [bit rotting](https://en.wikipedia.org/wiki/Software_rot). Another drawback is the code which becomes more complex, of course.

claussni:
I agree with @stweil's comment. There is no significant performance gain and you might never trigger errors in log message construction code.

Besides, to make it nicer and readable I prefer using String.format() wrapped in a small logging method of the particular class:

```
private void logIfDebug(String message, Object... objects) {
        if (log.isDebugEnabled()) {
            log.debug(String.format(message, objects));
        }
    }
```
"retrieving pica record for " + currentIdentifier  + " with server address: " + this.getOpacAddress());
...which can be called via:
```
logIfDebug("retrieving pica record for %s with server address: %s", currentIdentifier, this.getOpacAddress());
```

It even nicely handles NULL values.

matthias-ronge:
I think you are wrong. Usually the development environment is in DEBUG mode (at least ours) and the productive servers are in ERROR mode. So errors in debug message construction will rise while developing, which is okay. On the other hand, do you want a productive server which crashes with a stack trace just because a debug log message that is not even not logged, but just thrown away, cannot be created?

No, I didn’t mesasure the time. This is my intention of good programming style. Why do you want to
* create a String buffer of 16 bytes
* write some bytes in it
* copy the array because it is to small
* call a toString() function on an object
* copy some of the bytes
* copy the array because it is to small
* …

just to pass the result to a *logger* call which decides that it should just do *nothing* with this data? Perhaps even [inside a loop](https://github.com/goobi/goobi-production/blob/master/Goobi/src/de/sub/goobi/export/download/Multipage.java#L82-L83).

henning-gerhardt:
@matthias-ronge : Did you read @claussni method approach? Flooding the code with many if statements did not improve reading of code and increase possibility of errors compared to a single place / method call.

matthias-ronge:
He can do that if he wants to.

henning-gerhardt:
I will merge this and maybe next year do some clean up for debug statements.

stweil:
@henning-gerhardt, I'm a little bit surprised that this pull request was merged. I still don't think that it is an improvement.

@matthias-ronge, your patch does not match the "don't repeat yourself" criteria.

I suggest reverting this merge commit.

henning-gerhardt:
I think this is an improvement but an ugly one. First step, done by Matthias, was do eliminate unnecessary string concats on runtime if you don't run application in debug log level mode. Next step should be to simplify log statements like Ralfs suggestion or something similar.

stweil:
Ugliness > Improvement => Don't apply.

claussni:
Yep. Goobi code is full of unnecessary and ugly things. It didn't get prettier.

The cost of useless string concatenation in non-debug environments is IMHO very low. But it's done now. Let's consider it an improvement of some sort and stop arguing about a couple of extra IFs for now.

PR #418, opened 2016-01-28
Adds the missing files sRGB.icc an errorfile.jpg

frank-ulrich-weber:


henning-gerhardt:
I have an other solution: extracting errorfile.jpg and sRGB.icc from ics.jar file while creating war file. 

See https://github.com/henning-gerhardt/goobi-production/commit/99a502b6d670fd33c542c0e4b59528ad0fd0e079

PR #428, opened 2016-02-04
Add missing call of close method

stweil:
This fixes warnings from FindBugs.


henning-gerhardt:
Good catch - don't look into [line 560 - 562](https://github.com/goobi/goobi-production/blob/master/Goobi/src/de/sub/goobi/helper/ldap/Ldap.java#L560-L562).

PR #429, opened 2016-02-04
Fix data copier breaking export

matthias-ronge:
The design idea of data copy rules is that rules that cannot be applied for a certain process are skipped silently. If a rule cannot be applied an exception will be thrown. These exceptions must be caught on a per-rule basis.

### DIFF
--- a/Goobi/src/de/sub/goobi/metadaten/copier/DataCopier.java
+++ b/Goobi/src/de/sub/goobi/metadaten/copier/DataCopier.java
@@ -42,6 +42,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import org.apache.commons.configuration.ConfigurationException;
+import org.apache.log4j.Logger;
 
 /**
  * A data copier is a class that can be parameterised to copy data in goobi
@@ -50,6 +51,8 @@ import org.apache.commons.configuration.ConfigurationException;
  * @author Matthias Ronge &lt;matthias.ronge@zeutschel.de&gt;
  */
 public class DataCopier {
+	
+	private static final Logger LOG = Logger.getLogger(DataCopier.class);
 
 	/**
 	 * Holds the rules this data copier can apply to a set of working data.
@@ -82,7 +85,13 @@ public class DataCopier {
 	 */
 	public void process(CopierData data) {
 		for (DataCopyrule rule : rules) {
-			rule.apply(data);
+			try {
+				rule.apply(data);
+			} catch (RuntimeException notApplicable) {
+				if (LOG.isInfoEnabled()) {
+					LOG.info("Rule not applicable for \"" + data.getProcessTitle() + "\", skipped: " + rule);
+				}
+			}
 		}
 	}
 
--- a/Goobi/src/de/sub/goobi/metadaten/copier/CopierData.java
+++ b/Goobi/src/de/sub/goobi/metadaten/copier/CopierData.java
@@ -149,6 +149,21 @@ public class CopierData {
 	}
 
 	/**
+	 * Returns the process title.
+	 * 
+	 * @return the process title
+	 */
+	public String getProcessTitle() {
+		if (process instanceof Prozess) {
+			return ((Prozess) process).getTitel();
+		} else if (process instanceof ProcessObject) {
+			return ((ProcessObject) process).getTitle();
+		} else {
+			return String.valueOf(process);
+		}
+	}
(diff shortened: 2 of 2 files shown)


Goobi/src/de/sub/goobi/metadaten/copier/DataCopier.java
```diff
@@ -82,7 +85,14 @@ public DataCopier(String program) throws ConfigurationException {
...
 	public void process(CopierData data) {
 		for (DataCopyrule rule : rules) {
-			rule.apply(data);
+			try {
+				rule.apply(data);
+			} catch (RuntimeException nA) {
+				if (LOG.isInfoEnabled()) {
+					LOG.info("Metadata copy rule »" + rule + "« not applicable for process »" + data.getProcessTitle()
+							+ "«. Rule skipped.");
+				}
```
stweil: Using »« looks nice and is innovative (first time it is used in goobi-production), but I suggest to use the old fashioned "" instead. :-)

matthias-ronge: I use it because the rules contain lots of quotes themselves to make the log mor readable

henning-gerhardt: I agree in this case with @matthias-ronge : It is more readable with »« as with "". Compare yourself:

```
Metadata copy rule »/NewspaperYear[*]/NewspaperMonth[*]/NewspaperDay[*]/PeriodicalIssue@ProcessID =format $process.id« not applicable for process »ArtiNo_425390144-1824103102«. Rule skipped.
```

or

```
Metadata copy rule "/NewspaperYear[*]/NewspaperMonth[*]/NewspaperDay[*]/PeriodicalIssue@ProcessID =format $process.id" not applicable for process "ArtiNo_425390144-1824103102". Rule skipped.
```

As I see now one part of copy rules is missing as original copy rule was

```
/NewspaperYear[*]/NewspaperMonth[*]/NewspaperDay[*]/PeriodicalIssue@ProcessID \=format "%1s" $process.id;
```

stweil:
I compared and see no advantage in using chevrons. Please stick to the English quotation marks, either `""` or `''`. Or use something like `=> RULES <=` for the rules (but not for the process title). IMHO diagnostic messages should try to avoid non ASCII characters. See [Wikipedia](https://en.wikipedia.org/wiki/Quotation_mark) for more information on quotation marks.

matthias-ronge:
[Google Java Style](https://google.github.io/styleguide/javaguide.html#s2.3.3-non-ascii-characters) recommends:
> Never make your code less readable simply out of fear that some programs might not handle non-ASCII characters properly. If that should happen, those programs are **broken** and they must be **fixed**.

I'd like to go with that one.

stweil:
@matthias-ronge, did you notice that those chevron characters are not used as quotation marks in English texts? That's why I included the Wikipedia link.

And we are not talking about readability of code. We are talking about text which is shown in a debug message.

henning-gerhardt:
English quotation marks are good for things like process title because a process title had never any quotation marks inside. For any other output which could contain quotation marks an other character should be used to mark this outputs.

stweil:
Yes, I agree. If the text in question could contain both `""` and `''`, we should not use the normal quotation marks. But neither should we use quotation marks which are uncommon in English (and most people won't be able to enter a chevron on their keyboard without any help). There remain other possibilities which are commonly used to mark text, like `*** TEXT ***` or `=> TEXT <=` or even `>> TEXT <<`.

henning-gerhardt:
I have no preferences which other quotation mark charachter should be used. It would be cool if same character would be used instead of many different for this case. Entering chevron characters is not easy - at least on my keyboard.

henning-gerhardt:
@stweil & @matthias-ronge : would it be ok to merge this pull request and shift the decision over right quotation mark character to an other point?

matthias-ronge:
Variant:
`Rule not applicable for process "ArtiNo_425390144-1824103102", skipped: /NewspaperYear[*]/NewspaperMonth[*]/NewspaperDay[*]/PeriodicalIssue@ProcessID =format "%s" $process.id`

No confusion of any kind. I change it, just wait …

henning-gerhardt:
Looks good for me.

stweil:
That's a good solution, thank you @matthias-ronge.

PR #430, opened 2016-02-04
Fix issues reported by FindBugs

stweil:


stweil:
@stweil only removed unused code. The removed block is executed conditionally, and the condition is never true, so the whole block is dead code.

stweil:
I just got a message that Travis was activated for goobi-presentation. So I triggered a Travis check now by closing and re-opening this pull request.

henning-gerhardt:
Travis integration is only activated for Goobi.Production and not for Goobi.Presenation! I enabled it only for testing and not yet for any productive usage as I'm not  yet convinced and familiar with Travis. 

@stweil: you must not close and reopened any of your pull requests - caused a lot of new mails in my inbox.

stweil:
Sorry for the typo, I meant goobi-production of course. And yes, you'll get more e-mails because Travis sends each build result per e-mail. Until now I did not find a configuration setting to disable that, so you will need some filter in your mail reader.

Issue #431, opened 2016-02-04
After running batch export an error message for closed connection appear


henning-gerhardt:
After exporting one or a few newspaper batches and switch to my task or searching for a volume an error message for a closed connection appear:

You can't operate on a closed Connection!!!  

and following stack trace (for searching for a volume):

```
java.lang.NullPointerException
    at com.mchange.v2.c3p0.impl.NewProxyConnection.getAutoCommit(NewProxyConnection.java:985)
    at org.hibernate.engine.jdbc.internal.LogicalConnectionImpl.isAutoCommit(LogicalConnectionImpl.java:392)
    at org.hibernate.engine.transaction.internal.TransactionCoordinatorImpl.afterNonTransactionalQuery(TransactionCoordinatorImpl.java:195)
    at org.hibernate.internal.SessionImpl.afterOperation(SessionImpl.java:598)
    at org.hibernate.internal.SessionImpl.list(SessionImpl.java:1630)
    at org.hibernate.internal.CriteriaImpl.list(CriteriaImpl.java:374)
    at de.sub.goobi.forms.SearchForm.(SearchForm.java:121)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
    at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
    at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
    at java.lang.Class.newInstance(Class.java:383)
    at org.apache.myfaces.shared_impl.util.ClassUtils.newInstance(ClassUtils.java:277)
    at org.apache.myfaces.shared_impl.util.ClassUtils.newInstance(ClassUtils.java:268)
    at org.apache.myfaces.config.ManagedBeanBuilder.buildManagedBean(ManagedBeanBuilder.java:49)
    at org.apache.myfaces.el.VariableResolverImpl.resolveVariable(VariableResolverImpl.java:314)
    at org.apache.myfaces.custom.redirectTracker.RedirectTrackerVariableResolver.resolveVariable(RedirectTrackerVariableResolver.java:41)
    at org.apache.myfaces.custom.security.SecurityContextVariableResolver.resolveVariable(SecurityContextVariableResolver.java:45)
[...]

PR #433, opened 2016-02-05
Remove unused code

matthias-ronge:
Inspection of #431 revealed some code that is not in use. It can be deleted. It can still be restored from VCS if needed.


matthias-ronge:
_The Travis CI build passed._ First time I see this working. :+1:

henning-gerhardt:
Actual it is dead code. But could switching to use this code solve this issue in#431 or be a part of solving this?

henning-gerhardt:
If we need this code again, we could retrieve it from history.

PR #434, opened 2016-02-05
Replace inefficient Boolean constructor

stweil:
This fixes a warning from FindBugs.

### DIFF
--- a/Goobi/src/org/goobi/webapi/beans/Field.java
+++ b/Goobi/src/org/goobi/webapi/beans/Field.java
@@ -79,7 +79,7 @@ public class Field {
 		this.key = toCopy.key;
 		this.option = toCopy.option != null ? new ArrayList<Label>(toCopy.option) : null;
 		this.required = toCopy.required;
-		this.ughbinding = toCopy.ughbinding != null ? new Boolean(toCopy.ughbinding) : null;
+		this.ughbinding = toCopy.ughbinding != null ? Boolean.valueOf(toCopy.ughbinding) : null;
 	}
 
 	public static List<Field> getFieldConfigForProject(Projekt project) throws IOException {


matthias-ronge:
Why not `this.ughbinding = toCopy.ughbinding;`? Am I missing something here?

stweil:
@matthias-ronge, we need a conversion from boolean to Boolean here. See also https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html.

matthias-ronge:
`toCopy` is a `Field`, so `toCopy.ughbinding` cannot be something else than `Boolean`. Moreover Java would do autoboxing if it was `boolean`. Or am I now barking up the wrong tree?

stweil:
When looking closer on the code, I think that you are right. It looks like a simple assignment would be sufficient:

    this.ughbinding = toCopy.ughbinding;

henning-gerhardt:
After reading [javadoc comment](https://github.com/goobi/goobi-production/blob/master/Goobi/src/org/goobi/webapi/beans/Field.java#L66-L72) and try to interpret [lines of code](https://github.com/goobi/goobi-production/blob/master/Goobi/src/org/goobi/webapi/beans/Field.java#L77-L82), maybe following was goal of implementation: hard copying values of given object as cloning object duplicate any other references too. Main goal could be preventing any miss-usage in web context.

If maybe exploding of this objects is not possible or I miss-interpret something wrong, lines [80](https://github.com/goobi/goobi-production/blob/master/Goobi/src/org/goobi/webapi/beans/Field.java#L80) and [82](https://github.com/goobi/goobi-production/blob/master/Goobi/src/org/goobi/webapi/beans/Field.java#L82) should be changed.

stweil:
@matthias-ronge, maybe you can say more on the original code, the comment and whether my simplification is valid or not.

matthias-ronge:
Line 82 is not needed, because the internal meaning of `Boolean.TRUE` or `Boolean.FALSE` will not change. If for a field `ughbinding` state should be changed the field will no longer point to the `Boolean.TRUE` object in memory, but to the `Boolean.FALSE` object.
_If Bob and Alice both point with their finger onto a red apple, and Bob changes his mind and then points his finger on a green apple, Alice can still point to the red apple._ (The apples here are **immutable**. They do _not change_ their colour.)

It is different for the list. If for one field the options are changed, the _internal_ state of the list changes, so it would influence on another object which points to the same list in memory. Therefore a copy of the list is created to prevent this.
_If Bob and Alice both point with their finger onto the same table, and Bob puts three more apples on the table, Alice does suddanly find three more apples on the table she points her finger to._ (The table here is **mutable**. It _changes_ its state.)
_We need to give Alice her own table to prevent her from surprises._ → That is why the list copy happens.

matthias-ronge:
The copy constructor is never used from within the code. It can be deleted.

henning-gerhardt:
It is not used, removing should not harm anybody.

stweil:
It wouldn't harm now, but might harm later if anybody adds code which implies using the copy constructor. What about adding a private copy constructor (with no code)? Then the Java compiler will prevent accidental usage.

matthias-ronge:
Removing it will be the same, you could no longer compile a class calling a non-existent constructor. I think it is a good point to remove unused code from the project.

henning-gerhardt:
+1 for removing unused code.

henning-gerhardt:
If this constructor is removed, then other class constructor is needed too? Comment in lines [56 to 60 ](https://github.com/stweil/goobi-production/blob/findbugs/Goobi/src/org/goobi/webapi/beans/Field.java#L56-L60) could be interpret as this constructor is only needed for second constructor.

stweil:
Good catch. The commit now removes all constructors.

PR #435, opened 2016-02-06
Simplify code further (FindBugs)

stweil:
While commit b0846ddc5061384a6618cc6fc90ff982034726f8 fixed a warning
from FindBugs, it introduced a new different warning.

It looks like the code which was added by commit
b818a41f03efdd84c9396798dc66f610d922afa1 can be simplified further.


henning-gerhardt:
@stweil : please read my (comment](https://github.com/goobi/goobi-production/pull/434#issuecomment-181249524) in former issue / pull-request.

stweil:
Updated commit and removed the copy constructor function as suggested in discussion #434.

stweil:
Updated commit. Now all constructors are removed.

PR #437, opened 2016-02-08
Remove hardcoded filename from code (FindBugs)

stweil:
Report from FindBugs:
  Code contains a hard coded reference to an absolute pathname

Replace the filename by a local default filename and allow any filename
to be passed as an argument.


henning-gerhardt:
`main` method in this class is more or less dead code and was only used while implementation this class or prove usability of this class - as known as a test. I would prefer to change this into a test or to remove this method.

stweil:
How is a mass import done from the command line without that main method?

matthias-ronge:
You can use the [Active MQ web services](https://github.com/goobi/goobi-production/blob/master/doc/developer/activemq_jms_api.md).

henning-gerhardt:
@stweil : Mass import does not work from command line - only reading and displaying an Excel file. From command line you have no access to database nor are processes are created. If you want to mass import a file, you must work with mass import form where you can upload an Excel file.

henning-gerhardt:
I will think over it to merge this change. Reason beyond is that this change made it possible to test an Excel file in any path before importing it. 

If this undocumented and maybe hard to use feature should be removed then this should be an other pull request.

PR #438, opened 2016-02-09
Fix duplicate "to" and English grammar issues

stweil:


Goobi/src/de/sub/goobi/export/dms/ExportDms.java
```diff
@@ -74,7 +74,7 @@
 
 	/**
 	 * The field exportDmsTask holds an optional task instance whose progress
-	 * will be updated and whom errors will be passed to to be visible in the
+	 * will be updated and whose errors will be passed to be visible in the
```
matthias-ronge: Your change appears to me to reverse the meaning. The errors will be passed into the the `exportDmsTask` instance. If I read ‘whose errors’ it more likely seems that the errors come from the the `exportDmsTask` which is not the case of useage here.

Maybe I got the translation wrong, but the intended meaning of the description is: _Das Feld exportDmsTask hält eine optionale Task-Instanz, deren Fortschritt aktualisiert wird und **der** Fehler übergeben werden um in der Anzeige des Task Managers sichtbar zu sein, wenn eine verfügbar ist._

matthias-ronge:
Should the files from `doc/**/*.md` be moved to the wiki?

henning-gerhardt:
My opinion: files from doc directory could be transferred to wiki but should removed from doc directory and any reference to doc directory should be updated.

stweil:
@matthias-ronge, thanks for the explanation. I updated the text in the pull request accordingly.

stweil:
> Should the files from doc/**/*.md be moved to the wiki?

Technically, the current doc directory is no burden for git (only 48 KiB) and all files are pure text, so this is not a criterion which requests a change. If there were large PDF or DOC files and files with lots of images, this would be a different situation.

Files which document user interfaces and where we expect updates from users are better kept in a wiki.

Files which only document developer interfaces or which contain information which is very specific for a certain software version are better maintained with the software, so I'd keep such files in `doc`. As soon as software and documentation are maintained separately, the danger that both diverge greatly increases.

matthias-ronge:
The text now says what it was intended to say.
It is a good point to keep programmers-targeted information with the code.

PR #439, opened 2016-02-09
workaround for #413

henning-gerhardt:
This pull request does not fix the issue mentioned in #413 but made search form more solid for a hibernate error after a batch export. On error case one of the select fields contain only default value and no other possible values from database. In this case you should reload the search form to get the missing entries of this particular select field.


matthias-ronge:
:+1: looks good to me

PR #444, opened 2016-02-19
MIT license for PaginationCriteria class

henning-gerhardt:
PaginationCriteria class was accidentally licensed to GPL2+ but without asking the author Steve Liang (@steventux). He suggests to use [MIT license](https://opensource.org/licenses/MIT) for his original code.


stweil:
Looks good.

Issue #445, opened 2016-02-19
LazyInitializationException while starting new batch export


henning-gerhardt:
While further investigations of #431 I started a second newspaper batch export while first was still running. After starting second export inside taskmanager window folling error message appear on previous started batch:

```
Fehler (LazyInitializationException while exporting ArtiNo_425390144-18220117: failed to lazily initialize a collection of role: de.sub.goobi.beans.Projekt.filegroups, no session or session was closed)
```

Both batches does not include same process twice. One batch contains processes of year 1822 and second batch contains processes of year 1826.

henning-gerhardt:
Issue can also be trigged while batch export is running, switching to batch form and switching back to task manager window.

henning-gerhardt:
Fixed by #452

PR #452, opened 2016-03-01
Pass only the ID of the newspaper batch to the worker thread

matthias-ronge:
Asynchronous work should be passed the ID of the object only. Found
here: http://blog.schauderhaft.de/2008/09/28/hibernate-sessions-in-two-tier-rich-client-applications/

> The answer is actually fairly easy once found: […] Background tasks
get their own session. The critical point is the transfer of objects
from one frame to another. It is tempting to just pass the object but
then you have a object from the wrong session in the frame. Instead pass
just the Id (the primary key) and use that to load the object in the new
session. This approach solved most of our problems nicely.


henning-gerhardt:
Looks great. At least issue #445 is not reproducible any more.

PR #454, opened 2016-03-03
Remove some unused exception code

stweil:
This also fixes a warning from FindBugs.


henning-gerhardt:
What about to remove this dubious class? As far as I can see there is no usage of this class except some other exception classes use this class instead of general `Exception` class.

stweil:
That's even better - I see no obvious reason why we should keep `AbstractGoobiException`, so go on with your pull request #455 (if nobody else complains). Hopefully we can replace `DesEncrypter`, too, then `Goobi/src-dubious` would be history.

henning-gerhardt:
Replacing `DesEncrypter` is more difficult as this class is used to store password in database. In my opinion there must be a good and better algorithm as replacement and a migration tool to change current passwords with new generated ones. As this is a major change this change should be done in next major release.

stweil:
This PR is superseeded by PR #455.

PR #459, opened 2016-03-06
Remove unneeded null check (fixes FindBugs warning) and optimize code

stweil:
The local variable meineListe cannot be null.

Re-ordering the assignment for local variable zaehler optimizes the
code a little bit.


henning-gerhardt:
I think you can simplify this even more. Conditon check `zaehler > 0` could be removed as inside `for each` iteration is an check of empty list or `meineListe.size() == 0`.

stweil:
Good hint - I updated the PR to remove the if statement completely.

PR #463, opened 2016-03-13
Remove redundant nullcheck

stweil:


henning-gerhardt:
Really difficult to read as there are many indent changes. Without this indent changes difference is easy to read:

```patch
--- a/Goobi/plugins/import/PicaMassImport/de/intranda/goobi/plugins/PicaMassImport.java
+++ b/Goobi/plugins/import/PicaMassImport/de/intranda/goobi/plugins/PicaMassImport.java
@@ -173,7 +173,6 @@ public class PicaMassImport implements IImportPlugin, IPlugin {
                        }
                        pica = addParentDataForVolume(pica);
                        Fileformat ff = SRUHelper.parsePicaFormat(pica, prefs);
-                       if (ff != null) {
                        DigitalDocument dd = ff.getDigitalDocument();
                        boolean multivolue = false;
                        DocStruct logicalDS = dd.getLogicalDocStruct();
@@ -345,10 +344,6 @@ public class PicaMassImport implements IImportPlugin, IPlugin {
                        }
 
                        return ff;
-                       } else {
-                               logger.error("pica record for " + currentIdentifier + " is empty");
-                               throw new ImportPluginException("pica record for " + currentIdentifier + " is empty");
-                       }
                } catch (ReadException e) {
                        logger.error(this.currentIdentifier + ": " + e.getMessage(), e);
                        throw new ImportPluginException(e);

```

stweil:
@henning-gerhardt, in cases like this one it helps to apply the changes in a local git copy and then run `git diff --ignore-space-change HEAD^` (that's how I looked at my modification before sending the pull request).

henning-gerhardt:
Yes, I did a similar git diff command. See diff output in my previous comment as result of this command.

PR #467, opened 2016-04-04
Change Hibernate schema update to schema validation

henning-gerhardt:
Up to now it is not possible to run Goobi.Production in productive environments with only Hibernate schema validation. You must Hibernate allow to change your database structure on every initialisation.

This pull request change fix this: you can now run your productive environment without automatic database schema update through Hibernate. You must then update database schema manual. 

But you can still allow Hibernate to update your database schema. For this you must change hibernate property `hbm2ddl.auto` from `validate` to `update`.


stweil:
These changes look reasonable, but I won't be able to test them in the next few months.

PR #471, opened 2016-05-29
Fix legal notice for test installations

stweil:
Test installations should not display the legal notice of SUB Göttingen.


Goobi/src/messages/messages_de.properties
```diff
@@ -600,8 +600,9 @@ importplugin=Plugin f\u00FCr den Import
...
 inAuswahllisteAnzeigen=In Auswahlliste anzeigen
```
henning-gerhardt: This new key `impressumTextSUBGoettingen` is never used. Why did you add this?

henning-gerhardt:
Regarding translating messages like `impressumTextDemo` or `impressumText` which should contain local / instute information: you can add your own message files in `/usr/local/goobi/messages` (or the path you configured) which are used for display.

matthias-ronge:
Why should Production at all contain an Impressum? Nowadays, all institutions have some kind of web presentation which contains an Impressum and the common behaviour is to link there. Imho, the best solution would be to have an Impressum _link_ only which is configurable in the config properties file.

henning-gerhardt:
Impressum / legal notice message was only an example for overwriting standard message values.

Goobi/src/messages/messages_de.properties
stweil: Only to preserve the text, so anybody who wants to use it can do so in the local configuration. We could as well remove it (and also `impressumText`) completely, if that is preferred.

henning-gerhardt: For preserving text we have a git based history, so anybody how need this could look into the history of this file. Removing this entries would result in an empty page which should avoided. Instead of this a notice should be added to installation guide to add a institute based legal notice in a custom translation message file.

stweil: So I'll update the PR to remove the unused `impressumText` and `impressumTextSUBGoettingen`. The default configuration will still point to `impressumTextDemo`, so the default installation (e. g. Docker container) will install the new text.

henning-gerhardt: possible to use a general legal notice? If so then this text could be added to `impressumText`, `impressumTextDemo` could be deleted and inside of `goobi_config.properties` file configuration property `ApplicationImpressumMsg` is changed to `impressumText`. What did you think?

stweil:
Updated PR, now removed texts for Göttingen's university libraries.

PR #473, opened 2016-07-01
Change Goobi to Kitodo

sebastian-meyer:
Changes Goobi to Kitodo in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,14 @@
-Goobi.Production
-================
+Kitodo.Production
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
You missed a rename from Goobi.Production to Kitodo.Production in "Vanilla buld with Ant" paragraph.

sebastian-meyer:
Thanks for the hint! I fixed it.

PR #483, opened 2016-07-27
Logo image adaption

gnuj:
Adapt logo images an minor changes to references and some styles attributes.


stweil:
What about using SVG instead of PNG, at least for the larger logo?

gnuj:
SVG is not a bad idea so far. i changed that.

PR #484, opened 2016-09-02
Fix the new coverity issues

matthias-ronge:
Fixes three issues by e-mail from Fr 02.09.2016 10:29 o’clock.

### DIFF
--- a/Goobi/plugins/opac/PicaPlugin/org/goobi/production/plugin/CataloguePlugin/PicaPlugin/OpacResponseHandler.java
+++ b/Goobi/plugins/opac/PicaPlugin/org/goobi/production/plugin/CataloguePlugin/PicaPlugin/OpacResponseHandler.java
@@ -73,11 +73,11 @@ class OpacResponseHandler extends DefaultHandler {
 		}
 
 		if (localName.equals("SET")) {
-            String hits = atts.getValue("hits");
-            if (hits == null) {
-                throw new NumberFormatException("null");
-            }
-            numberOfHits = Integer.parseInt(hits);
+			String hits = atts.getValue("hits");
+			if (hits == null) {
+				throw new NumberFormatException("null");
+			}
+			numberOfHits = Integer.parseInt(hits);
 		}
 
 		if (localName.equals("SHORTTITLE")) {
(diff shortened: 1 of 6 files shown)


Goobi/plugins/opac/PicaPlugin/org/goobi/production/plugin/CataloguePlugin/PicaPlugin/OpacResponseHandler.java
```diff
@@ -73,7 +73,11 @@ public void startElement(String namespaceURI, String localName, String qName, At
 		}
 
 		if (localName.equals("SET")) {
-			numberOfHits = Integer.valueOf(atts.getValue("hits")).intValue();
+            String hits = atts.getValue("hits");
+            if (hits == null) {
+                throw new NumberFormatException("null");
+            }
+            numberOfHits = Integer.parseInt(hits);
 		}
```
henning-gerhardt: Indent of changes are wrong: you put space indent but it must tab indent.

Goobi/src/de/sub/goobi/metadaten/Paginator.java
```diff
@@ -99,9 +99,20 @@ private void assertSelectionIsNotNull() {
...
+	 * @throws NumberFormatException
+	 *             if `paginationStartValue` isn’t valid
+	 */
 	private void assertValidPaginationStartValue() {
 		// arabic numbers
 		if (paginationType == Paginator.Type.ARABIC) {
+			/* coverity[USELESS_CALL]
+			 * Integer.parseInt() throws NumberFormatException if
+			 * paginationStartValue cannot be parsed to an `int`, and this is
+			 * what we want here. */
```
stweil: Would adding `throws NumberFormatException` to the function declaration also fix the coverity warning (instead of muting coverity by this comment)?

matthias-ronge: No idea, I cannot test it anyway. However, Joshua Bloch [writes](https://jtechies.blogspot.de/2012/07/item-62-document-all-exceptions-thrown.html) in Effective Java: “Use the Javadoc @throws tag to document each unchecked exception that a method can throw, but do _not_ use the throws keyword to include unchecked exceptions in the method declaration.” I followed his recommendation.

henning-gerhardt: This change is ok for me. It is a ugly fix of an input validation

matthias-ronge: I don’t find it even too ugly, I first thought about (and tried to implement correctly) manually checking. [Here is an example, but it does not take into account numeric overflow and does not iterate over code points, btw.](http://stackoverflow.com/a/237204/1503237)—and it is a lot of code. Second point is that `RomanNumeral.setValue(String)` (l. 121) behaves the same and also throws NumberFormatException if the input is incorrect. So both methods behave the same. It’s still a bit smelly, yes.

PR #485, opened 2016-09-06
Use a final object to lock on

matthias-ronge:
Java locks on _objects_ that are held in variables. It does _not_ lock on the variable. If the variable is changed in the `synchronized` block, a concurrent thread will lock on the other object, making the locking not working as expected.

Fix for Coverity IDs 44438, 44440

### DIFF
--- a/Goobi/src/de/sub/goobi/persistence/HibernateUtilOld.java
+++ b/Goobi/src/de/sub/goobi/persistence/HibernateUtilOld.java
@@ -37,8 +37,6 @@ import org.hibernate.Transaction;
 import org.hibernate.auction.exceptions.InfrastructureException;
 import org.hibernate.cfg.Configuration;
 
-import com.sun.org.apache.bcel.internal.generic.NEW;
-
 //TODO: Fix for Hibernate-Session-Management, replaced with older version, 
 // the newer version follows on bottom  of this class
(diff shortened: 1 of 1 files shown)


Goobi/src/de/sub/goobi/persistence/HibernateUtilOld.java
```diff
@@ -37,6 +37,8 @@
 import org.hibernate.auction.exceptions.InfrastructureException;
 import org.hibernate.cfg.Configuration;
 
+import com.sun.org.apache.bcel.internal.generic.NEW;
+
```
henning-gerhardt: Is this import statement necessary?

henning-gerhardt:
Did you search for similar (wrong) usage in project?

matthias-ronge:
No. I fixed it because coverity reported it. If I saw it right, it reported only these two locations of the problem.

Goobi/src/de/sub/goobi/persistence/HibernateUtilOld.java
matthias-ronge: No. I guess it came in by the Eclipse auto-complete feature when typing **new** Object()….
It’s ugly, sometimes.

PR #489, opened 2016-09-10
export: Fix some warnings (FindBugs)

stweil:
FindBugs warnings:
* Method may fail to clean up stream or resource

Fix it by using try-with-resources statements (supported since Java 7).


stweil:
See http://docs.oracle.com/javase/7/docs/technotes/guides/language/try-with-resources.html for a description of the try-with-resource statement.

PR #493, opened 2016-09-13
ConnectionManager: Close database resources on exception (FindBugs)

stweil:
FindBugs warnings:
* Method may fail to close database resource on exception
* Method may fail to clean up stream or resource on checked exception

Fix it by using try-with-resources statements (supported since Java 7).

### DIFF
--- a/Goobi/src/de/sub/goobi/persistence/apache/ConnectionManager.java
+++ b/Goobi/src/de/sub/goobi/persistence/apache/ConnectionManager.java
@@ -183,7 +183,7 @@ public class ConnectionManager {
 			logger.error(ex.toString());
 		} catch (Exception e) {
 			if(logger.isDebugEnabled()){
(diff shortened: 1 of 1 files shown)


stweil:
@henning-gerhardt, maybe it would be good to update the [Coverity scan for Kitodo](https://scan.coverity.com/projects/kitodo-kitodo-production) to get a better overview about the remaining issues.

stweil:
Updated PR because I noticed a word duplication in the debug message, so fix that (unrelated) issue, too.

henning-gerhardt:
@stweil:  If you have more similar changes in development or pending then this changes should be merged first.

stweil:
This one was the last in my current queue. I want to do more work, and a new scan would help.

henning-gerhardt:
Ok, I will initiate a new Coverity run.

PR #494, opened 2016-09-13
Fix some resource leaks

stweil:


Goobi/src/de/sub/goobi/persistence/apache/MySQLHelper.java
```diff
@@ -65,6 +65,8 @@ public Connection getConnection() throws SQLException {
 			return connection;
 		}
 
+		connection.close();
+
```
henning-gerhardt: I'm unsure about this and lower close() calls. So far as I know: database connection close() calls should not be needed in a connection pool managed environment as pool management handling connection to database.

stweil:
The commit for MySQLHelper is now removed from this PR, so we can discuss it separately.

Issue #373, opened 2015-11-12
Update Java version


stweil:
The goobi-production code still is based on Java 1.5. According to Wikipedia, J2SE 5.0 was released in September 2004.

The current Debian GNU Linux stable uses openjdk-7-jre (I assume this is the same as 1.7), and Oracle officially only supports a free version 8, so 7 and 8 are a must.

As version 7 is now 4 years old, I think support for older versions can be dropped, and version 7 / 1.7 can be used as the new base version.

claussni:
It's in principle not necessary to raise the source code language level unless we use >1.5 language constructs. But I agree to raise to byte code language level to adopt to more current JavaVM operations and avoid any issues when including third party libraries.

By the way, [(Oracle) Java 7 is EOL](http://www.infoq.com/news/2015/05/Oracle-Ends-Java-7Public-Updates) already.

stweil:
Are there servers running Goobi.Production with Oracle Java? This might be rather expensive because the prizes which Oracle requests for Java licenses are not cheap (maybe affordable for universities). Don't reply here if you know such servers - tell them to be careful.

matthias-ronge:
Java 7+ would provide ways to handle file and directory permissions without running scripts. However this is not just done by increasing the bytecode level but would require to rewrite the code accordingly. In addition, it might require the Tomcat to run with sufficient permissions to be allowed to perform these changes, which might be undesired in the sense of server security.

Java 8 comes with a cool feature for wrapping in-line defined runnables on the code level, called “lambda expressions”. Functional programming can inprove performance, but again it would require to rewrite the code accordingly. This is not a real gain in functionality, because implementing runnables is already possible in lower Java versions, it is just much less verbose the new way.

henning-gerhardt:
Improving code quality is independent from Java language level but a higher language level could help. Changing code without tests which prove that changes are correct are dangerous and introducing new bugs. 
Removing existing functionality (f.e. shell script support as our instance used heavy modified scripts to run ( f.e. processes directories are distributed to different storage locations)) without a proper replacement could stop using new versions.

Changing language level support is something for major release 1.12.x and not a minor update 1.11.1 which could / should released this month or at least this year.

stweil:
@matthias-ronge, don't the current shell scripts in FileSystemHelper.java also run with Tomcat permissions? Then only the code would need a rewrite.

claussni:
[...]

PR #495, opened 2016-09-14
Rewrite try-with-resources to try-catch blocks

matthias-ronge:
Recent changes introduced try-with-resources statements that do not compile with Java 1.6. Since the content server does not work with Java > 1.6 under certain circumstances, this needs to be fixed.

@henning-gerhardt Do the build scripts create byte code compliant to Java 1.6? I fear not, since the Travis CI build should have failed for the branches that introduced the try-with-resources statements. Could you please check for this and fix it if necessary?

### DIFF
--- a/Goobi/src/de/sub/goobi/export/dms/DmsImportThread.java
+++ b/Goobi/src/de/sub/goobi/export/dms/DmsImportThread.java
@@ -31,13 +31,12 @@ import java.io.BufferedReader;
 import java.io.File;
 
 import org.goobi.io.SafeFile;
-import java.io.FileReader;
+import java.io.IOException;
 
 import org.apache.log4j.Logger;
 
 import de.sub.goobi.beans.Prozess;
 import de.sub.goobi.config.ConfigMain;
-import de.sub.goobi.helper.Helper;
 
 public class DmsImportThread extends Thread {
 	private static final Logger myLogger = Logger.getLogger(DmsImportThread.class);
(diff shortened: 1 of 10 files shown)


henning-gerhardt:
In [ant build file](https://github.com/kitodo/kitodo-production/blob/master/build.xml#L38-#L39) Java src and target are defined for Java 1.7. As far as I know travis is using this settings. 

Under which conditions does the content server not work? Maybe fixing the content server is a better approach then reverting back to Java 1.6 / 1.5 compatible binaries.

stweil:
We dropped support for Java < 7 with commit 17fdc65be33fa6081daf35231db45d6f11e8612e. I suggest working on the content server to support new Java versions.

stweil:
See also issue #373 where I suggested recently to even go to Java 8 ([Java 7 already reached end of life](https://www.infoq.com/news/2015/05/Oracle-Ends-Java-7Public-Updates)).

matthias-ronge:
As far as I know, there are no source files for the content server.

henning-gerhardt:
I can't reproduce with current source code any failures of content server. Converting tif images to png files for displaying in meta data editor works and I don't know any further usage of this content server.

henning-gerhardt:
On Launchpad are source files existing for a version of a content server: https://launchpad.net/goobi-contentserver I have no further information about this project and it is not maintained since 2012.

joerg-liebenow:
If I remember correctly is the sourcecode for gcs now been replaced 
by the ics-1.1.5.jar .

stweil:
`ics-1.1.5.jar` is included in Kitodo, and indeed it looks like code for a content server. A short comparison with the code on https://launchpad.net/goobi-contentserver shows the same directory and file structure.

@henning-gerhardt, what about moving the project from Launchpad to GitHub (and renaming it to `kitodo-contentserver`)? IMHO having sources for each jar file is very important, not only because that might be needed for legal reasons but also because without sources and active development, no quality improvements are possible.

Who is currently using the content server? We did not find any documentation for it. Therefore we run our Kitodo installation without that content server.

henning-gerhardt:
@matthias-ronge : Can you provide an example for an error and how somebody else could reproduce this error? Jar files compiled for 1.5 should even work with 1.7 except they are using deprecated Java interfaces in 1.5 which removed later. But should compiling with travis or manually not fail? 

@stweil: 
- file and directory structure may be the same but I don't know if the launchpad version works to on any point of time and more important point is what are the differences between this versions.
- content server is used internally to convert tif image files to png files and scaling images in metadata editor. Maybe methods of this content server used somewhere else for any kind of image processing (maybe while creating pdf files with images included).

stweil:
I suggest to discuss the content server related  topics in a new issue (I'll create one).

@matthias-ronge, I'd rather find another solution for your problem than see your PR applied. IMHO we should not add code complexity to support a really old Java version.

stweil:
@matthias-ronge, do you think that this PR can be closed?

PR #496, opened 2016-09-14
Fix FindBugs warnings (Method uses the same code for two switch clauses)

stweil:

### DIFF
--- a/Goobi/src/de/sub/goobi/export/dms/AutomaticDmsExportWithoutHibernate.java
+++ b/Goobi/src/de/sub/goobi/export/dms/AutomaticDmsExportWithoutHibernate.java
@@ -138,9 +138,6 @@ public class AutomaticDmsExportWithoutHibernate extends ExportMetsWithoutHiberna
 				break;
 
 			case METS_AND_RDF:
-				newfile = new RDFFile(this.myPrefs);
-				break;
-
 			default:
 				newfile = new RDFFile(this.myPrefs);
 				break;
--- a/Goobi/src/de/sub/goobi/export/dms/ExportDms.java
+++ b/Goobi/src/de/sub/goobi/export/dms/ExportDms.java
@@ -202,9 +202,6 @@ public class ExportDms extends ExportMets {
 				break;
 
 			case METS_AND_RDF:
-				gdzfile = new RDFFile(this.myPrefs);
-				break;
-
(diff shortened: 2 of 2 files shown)


Goobi/src/de/sub/goobi/export/dms/AutomaticDmsExportWithoutHibernate.java
```diff
@@ -138,9 +138,6 @@ public boolean startExport(ProcessObject process) throws DAOException, IOExcepti
 				break;
 
 			case METS_AND_RDF:
```
henning-gerhardt: The unused case statement "case METS_AND_RDF:" could be removed too.

Goobi/src/de/sub/goobi/export/dms/ExportDms.java
```diff
@@ -202,9 +202,6 @@ public boolean startExport(Prozess myProzess, String inZielVerzeichnis, DigitalD
 				break;
 
 			case METS_AND_RDF:
```
henning-gerhardt: The unused state case statement "case METS_AND_RDF:" could be removed too.

Goobi/src/de/sub/goobi/export/dms/AutomaticDmsExportWithoutHibernate.java
stweil: I did not remove it because it matches the action taken here (`new RDFFile(this.myPrefs)`). The correct solution would be to name all cases which need this action and to throw an exception in the default case (which should never occur). But that solution would require more code reviews and good testing.
Therefore I suggest to apply the PR as it is.

Goobi/src/de/sub/goobi/export/dms/ExportDms.java
stweil: See my other comment.

Goobi/src/de/sub/goobi/export/dms/AutomaticDmsExportWithoutHibernate.java
henning-gerhardt: If somebody still use RDF as store and / or export format then he should get a proper deprecation message. One or two versions later this format should be removed and not supported formats should throw an error message in the gui.

PR #497, opened 2016-09-14
CopyProcess: Fix some warnings from FindBugs

stweil:
* Unread field: should this field be static?
  * The constant atstsl is obviously not used as intended.
    Maybe some implementation is missing.
  * The boolean constant updateData is always false,
    so a lot of dead code could be removed.
    That dead code was the only user of private function addCollections.

* Redundant nullcheck of value known to be non-null

* Method names should start with a lower case letter
  * Fix it for private methods. Rename also German names to English.

* Useless objected created
  * It is not necessary to create a temporary object for eigenschaften.


henning-gerhardt:
Please, do not change many different things in one big commit. Your changes could be split into two or three different commits in one pull request.

PR #499, opened 2016-09-15
SessionForm: Fix bugs in browser identification

stweil:
FindBugs reports these warnings:
    Method checks to see if result of String.indexOf is positive

The checks were wrong and would not find a user agent like "Safari x.y".
Fix the checks and optimize the code a little bit.

### DIFF
--- a/Goobi/src/de/sub/goobi/forms/SessionForm.java
+++ b/Goobi/src/de/sub/goobi/forms/SessionForm.java
@@ -103,25 +103,19 @@ public class SessionForm {
 				mybrowser = "-";
 			}
 			map.put("browser", mybrowser);
-			if (mybrowser.indexOf("Gecko") > 0) {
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
As long any browser identification string starts with "Mozilla/..." code is correct. 

Current Safari browser using agent string `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12) AppleWebKit/602.1.50 (KHTML, like Gecko) Version/10.0 Safari/602.1.50` on current macOS.

I will merge this as nobody knows how browsers identification string is in future.

henning-gerhardt:
Mhh, my fault. Your changes from many if statements to one if statements with else conditions does not work: Firefox browser now displayed as Gecko browser.

stweil:
Fixed with PR #506.

PR #500, opened 2016-09-16
MySQLHelper: Fix some resource leaks

stweil:
This fixes CID 44575, CID 44611, CID 44631 and CID 44661.

### DIFF
--- a/Goobi/src/de/sub/goobi/persistence/apache/MySQLHelper.java
+++ b/Goobi/src/de/sub/goobi/persistence/apache/MySQLHelper.java
@@ -65,6 +65,8 @@ public class MySQLHelper {
 			return connection;
 		}
 
+		connection.close();
+
 		for (int i = 0; i < MAX_TRIES_NEW_CONNECTION; i++) {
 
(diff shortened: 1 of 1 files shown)


stweil:
From #494:

> I'm unsure about this and lower close() calls. So far as I know: database connection close() calls should not be needed in a connection pool managed environment as pool management handling connection to database.

Two points why I think that `close` is needed:

1. Usually Coverity is rather good in finding bugs, and here it says that `close` is needed.
2. Even more important: The old code already uses lots of `close` or the equivalent `closeConnection` calls – IMHO it just missed some more locations where it is needed.

By the way: we have to restart our Tomcat server with Kitodo.Production from time to time because it runs out of resources (usually memory according to the Tomcat log files). Sometimes it only works less than 24 hours, sometimes more than a week. Therefore I think that fixing leaks is very important.

henning-gerhardt:
Fixing resource leaks is important and correct.The added database close calls are IMHO unnecessary as no connection to database exists and a new should be opened. On the other side this close calls on a closed connection should not hurt.

PR #501, opened 2016-09-16
Exclude empty packages from build (fix javac warnings)

stweil:
javac reports that it is creating empty classes. Remove them.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -247,6 +247,7 @@
 			debuglevel="lines,vars,source"
 			classpathref="compile.classpath">
             <compilerarg value="${build.javac.args}" />
+            <exclude name="**/package-info.java"/>
         </javac>
     </target>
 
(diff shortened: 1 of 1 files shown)


stweil:
@matthias-ronge, perhaps you can review this PR.

matthias-ronge:
The files called `package-info.java` [are there to document packages](http://docs.oracle.com/javase/specs/jls/se7/html/jls-7.html#jls-7.4.1-310). They are recognized by javadoc processors:

mro_0001

The naming as `*.java` files is a bit odd. These files should not contain any code and they do not need to be compiled. Correct solution for this should be to exclude any `package-info.java` files in the build script.

matthias-ronge:
I do not use Ant to build Kitodo Production, so I cannot test it, but [here is the syntax how to exclude files from the `javac` task in ant](https://ant.apache.org/manual/Tasks/javac.html#srcdirnote).

henning-gerhardt:
As this two files provides some documentation they should not be removed. They should be excluded from building.

stweil:
I updated the PR. Now the files are not removed, but excluded.

PR #505, opened 2016-09-19
Browser identification

henning-gerhardt:
Split big browser identification statement into smaller one to identify used browser.


stweil:
Breaking the if statement is not the right solution. The correct solution would be reordering the statements.

henning-gerhardt:
Reordered statements from old "general to specific" to new "specific to general".

stweil:
I suggest to apply PR #506, close PR #505 and create a new PR which replaces `indexOf` and fixes Opera (2 commits).

henning-gerhardt:
Closing this PR as recommend.

PR #506, opened 2016-09-19
Fix browser check (order of detection)

stweil:
Commit 07b3e0d5fc93d2c956c90e66dfe9fe17eb9053b7 replaced separated
if statements by a sequence of if / else if statements.

This changed the resulting browser icon when the user agent matched
several strings.

Reversing the order fixes this and restores the old behaviour while
keeping the optimized code.


henning-gerhardt:
Ok. Browser identification does not work for current IE versions too. Is `Trident` must be added too.

PR #507, opened 2016-09-19
Fix more warnings from FindBugs

stweil:

### DIFF
--- a/Goobi/src/org/goobi/production/properties/ProcessProperty.java
+++ b/Goobi/src/org/goobi/production/properties/ProcessProperty.java
@@ -59,9 +59,12 @@ public class ProcessProperty implements IProperty, Serializable {
 	private boolean currentStepDuplicationAllowed = false;
 	
 	public ProcessProperty() {
+		this.container = 0;
+		this.value = "";
 		this.possibleValues = new ArrayList<String>();
 		this.projects = new ArrayList<String>();
 		this.showStepConditions = new ArrayList<ShowStepCondition>();
+		this.prozesseigenschaft = new Prozesseigenschaft();
 	}
 
 	/*
@@ -291,9 +294,6 @@ public class ProcessProperty implements IProperty, Serializable {
 	 */
 
 	public void save(Schritt step) {
-		if (this.prozesseigenschaft != null) {
-
-		}
 	}
 
 	/*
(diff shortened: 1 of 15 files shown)


Goobi/src/org/goobi/production/properties/ProcessProperty.java
```diff
@@ -59,9 +59,12 @@
 	private boolean currentStepDuplicationAllowed = false;
 	
 	public ProcessProperty() {
+		this.container = new Integer(0);
```
henning-gerhardt: IntelliJ IDEA tolds me to remove this unnecessary boxing.

henning-gerhardt:
For commit: https://github.com/kitodo/kitodo-production/pull/507/commits/cb89fd05a3d14c0d91047e35b7caf6c930b1e418
You can not remove this private methods which are used by Hibernate:
```
Could not find a getter for datentyp in class de.sub.goobi.beans.Prozesseigenschaft
```
The other removed methods are maybe used too. So please test your changes in a proper test environment.

Goobi/src/org/goobi/production/properties/ProcessProperty.java
```diff
@@ -291,9 +294,6 @@ public boolean isValid() {
 	 */
 
 	public void save(Schritt step) {
-		if (this.prozesseigenschaft != null) {
-
-		}
```
henning-gerhardt: save method is not used nor defined in interface. Safe to remove?

Goobi/src/org/goobi/production/properties/ProcessProperty.java
stweil: I think we have more of that kind, so they can be handled in a new commit.

Goobi/src/org/goobi/production/properties/ProcessProperty.java
stweil: Yes, I assume so. I added a new commit which removes it.

stweil:
PR update:
* Removed commit which removed private methods which were reported to be unused but which are obviously used nevertheless.
* Added commit which removes `save` method.

Goobi/src/org/goobi/production/properties/ProcessProperty.java
henning-gerhardt: Introducing a known bad usage to change / fix it later is not a good way. We need no more bad usage which must be fixed sometime.

stweil: Sorry, that was a misunderstanding on my side. It's fixed now with an updated PR.

stweil:
New PR, addressed @henning-gerhardt comment.

PR #509, opened 2016-09-19
Remove unused private methods

stweil:
This is an extract of an earlier patch which wrongly also removed methods needed for hibernate.


Goobi/src/org/goobi/production/flow/statistics/hibernate/UserDefinedFilter.java
```diff
@@ -325,11 +325,6 @@ protected void setStepDone(Integer exactStepDone) {
 			this.exactStepDone = exactStepDone;
 		}
 
-		@SuppressWarnings("unused")
-		private Boolean getCriticalQuery() {
-			return flagCriticalQuery;
-		}
-
```
stweil: The method `setCriticalQuery` is currently used in another file. Removing that method and the member variable `flagCriticalQuery` should be addressed separately.

PR #523, opened 2016-09-29
Clear reimplementation of DesEncrypter class

henning-gerhardt:
Reimplementation of DesEncrypter algorithm with current used parameters.


stweil:
That's great news - the last code in `Gobbi/src-dubious` is finally eliminated.

One thought: wouldn't it be better to place new classes under a new directory `Goobi/src/de/kitodo`, so here the new code could be in `Goobi/src/de/kitodo/helper/encryption/DesEncrypter.java`?

henning-gerhardt:
A good suggestion @stweil ! I updated my PR.

PR #527, opened 2016-10-05
Replace some old path and entry names.

henning-gerhardt:
For new installation guide of upcoming release some path informations are changed.

### DIFF
--- a/Goobi/config/goobi_config.properties
+++ b/Goobi/config/goobi_config.properties
@@ -36,24 +36,24 @@
 # Note: Several, but not all configuration files are read from that directory.
 #       You may want to decide to point this path to the directory where
 #       the servlet container will extract the configuration files to (like
-#       webapps/goobi-ce/WEB-INF/classes) in order to make sure they are found.
-KonfigurationVerzeichnis=/usr/local/goobi/config/
+#       webapps/kitodo/WEB-INF/classes) in order to make sure they are found.
+KonfigurationVerzeichnis=/usr/local/kitodo/config/
 
 # Absolute path to the directory that the rule set definition files will be
 # read from. It must be terminated by a directory separator ("/").
-RegelsaetzeVerzeichnis=/usr/local/goobi/rulesets/
+RegelsaetzeVerzeichnis=/usr/local/kitodo/rulesets/
 
 # Absolute path to the directory that XSLT files are stored in which are used
 # to transform the "XML log" (as visible from the XML button in the processes
 # list) to a downloadable PDF docket which can be enclosed with the physical
 # binding units to digitise.
 # The path must be terminated by a directory separator ("/").
-xsltFolder=/usr/local/goobi/xslt/
+xsltFolder=/usr/local/kitodo/xslt/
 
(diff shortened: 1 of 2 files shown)


Goobi/config/log4j.properties
```diff
@@ -31,7 +31,7 @@ log4j.appender.stdout.layout.ConversionPattern=%-5p %d %C.%M(%F:%L)%n        %m%
 
 # ein Rolling-File-Appender (10 Backups)
 log4j.appender.rolling=org.apache.log4j.RollingFileAppender
-log4j.appender.rolling.File=/usr/local/logs/goobi.log
+log4j.appender.rolling.File=/usr/local/kitodo/logs/goobi.log
```
stweil: What about switching from `goobi.log` to `kitodo.log`? That also requires an update in `build.xml`.

Goobi/config/goobi_config.properties
```diff
@@ -115,20 +115,20 @@ useOrigFolder=true
 # to have sufficient permissions to do whatever they should do.
 #
 # For examples of these scripts, see:
-# https://github.com/kitodo/kitodo-production/tree/1.11.x/Goobi/scripts
+# https://github.com/kitodo/kitodo-production/tree/2.0.0/Goobi/scripts
```
henning-gerhardt: Linking to master / development branch is a bad idea because scripts could be change in development, changed to be incompatible, moved or deleted. I will add a notice to look for current stable tags.

Goobi/config/log4j.properties
henning-gerhardt: Name of log file should kitodo.log. I missed that. `build.xml` contains log file entries? I think this entries could be removed as no log file should be created inside source tree.

Goobi/config/goobi_config.properties
```diff
@@ -115,20 +115,24 @@ useOrigFolder=true
 # to have sufficient permissions to do whatever they should do.
 #
 # For examples of these scripts, see:
-# https://github.com/kitodo/kitodo-production/tree/1.11.x/Goobi/scripts
+#
+# https://github.com/kitodo/kitodo-production/tree/2.0.0/Goobi/scripts
+#
+# for script files of release 2.0.0. Never versions could be found under
+# appropriate release tag.
```
stweil: s/Never/Newer/. The link currently does not work, because there is no 2.0.0 tree.

For now, I suggest this comment:

    # For examples of these scripts, see:
    # https://github.com/kitodo/kitodo-production/tree/master/Goobi/scripts
    # Replace "master" by your version of Kitodo to see examples matching that version.

For the future, we should consider adding those scripts to the distribution and use those default scripts if there is no special configuration. Are there installations which override the scripts? If not, then why not remove that nuisance of requiring an installation in /usr/local?

henning-gerhardt: Your suggestion for distribution and using this script files are good and should be regard in future development.

henning-gerhardt: If we distribute this basic shell scripts, we should distribute SQL schema and database default entries (schema.sql and default.sql) too.

PR #531, opened 2016-10-07
Distribute sql files within war file.

henning-gerhardt:
Distribute sql files (schema.sql and default.sql) within generated war file for easier installation process.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -309,7 +309,7 @@
 
     <!-- Build of war file. -->
     <target name="war"
-            depends="compile, copy-config, copy-scripts, create-plugin-jar-files">
+            depends="compile, copy-files, create-plugin-jar-files">
 
         <war destfile="${build.dist.war}" webxml="${src.dir}/WEB-INF/web.xml">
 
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -196,6 +196,8 @@
         <mkdir dir="${test.report.dir}"/>
     	<mkdir dir="${build.dir.plugins.import}/PicaMassImport"/>
         <mkdir dir="${build.dir.plugins.opac}/PicaPlugin"/>
+        <mkdir dir="${build.dir}/scripts"/>
```
stweil: This is not needed in my ANT build.

stweil:
Is it possible that more setup related files would be useful in the future, for example scripts which help during a migration from an old release to a newer one? Then I'd keep the name `setup`.

henning-gerhardt:
I don't think that migration files should be provided in the war file. I accepted your PR and created this PR for easier new installations. 

Migration of any kind of data and structure should not applied automatically.

stweil:
Migration scripts were only an example. We could also add scripts which create the SQL database or which check whether all required software packages are installed and install those that are missing, so there are lots of use cases for scripts needed for new installations.

PR #533, opened 2016-10-07
Replace InfrastructureException class

henning-gerhardt:
Old class contains unused methods and was licensed by LGPL 2.1 only.


stweil:
Why not simply remove the unused methods? LGPL should not pose a problem, as it is [compatible with GPL 2 and GPL 3](https://www.gnu.org/licenses/license-list.de.html#GPLCompatibleLicenses).

I think that `serialVersionUID` is still needed.

henning-gerhardt:
I did the rewrite because for only one method there is no stand alone package needed. This class origin is a tutorial - we don't need unused and unneeded tutorial code in the project.

`serialVersionUID` is only needed for serializable classes and I don't think this is needed for an exception class.

stweil:
Nevertheless you should preserve the original header!

henning-gerhardt:
I don't think so that I have to preserve original headers on a rewrite. If a rewrite could not contain same class names and functional behaviour to prevent bigger refactoring processes then no rewrite could be done.

PR #534, opened 2016-10-10
Fix CID 137042: do not rely on default encoding

henning-gerhardt:


Goobi/src/org/kitodo/encryption/DesEncrypter.java
```diff
@@ -106,7 +106,7 @@ public String encrypt(String messageToEncrypt) {
 		try {
 			byte[] utf8 = messageToEncrypt.getBytes("UTF8");
 			byte[] enc = encryptionCipher.doFinal(utf8);
-			return new String(Base64.encodeBase64(enc));
+			return new String(Base64.encodeBase64(enc), "UTF8");
```
stweil: Must be "UTF-8".

stweil:
Please fix subject: either "Fix CID 137042: no reliance on default encoding" or "Fix CID 137042: do not rely on default encoding"

PR #535, opened 2016-10-11
Replace UTF8 with UTF-8 as correct charset name

henning-gerhardt:


stweil:
Maybe it would be even better to use the existing constant UTF_8, see https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html.

stweil:
Currently the code also uses org.apache.commons.lang.CharEncoding UTF_8 and org.apache.commons.compress.utils.Charsets UTF_8, so we have a mixture of different solutions for the same requirement. I suggest to use UTF_8 from the Java standard (see link above) were possible.

henning-gerhardt:
Changing to UTF_8 constant should be done after merging both PR (#534 and #535).

PR #536, opened 2016-10-11
Use standard charset UTF_8

stweil:
This implies that the UnsupportedEncodingException can no longer occur,
so the exception handling must be removed.

### DIFF
--- a/Goobi/plugins/import/PicaMassImport/com/googlecode/fascinator/redbox/sru/SRUClient.java
+++ b/Goobi/plugins/import/PicaMassImport/com/googlecode/fascinator/redbox/sru/SRUClient.java
@@ -25,6 +25,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -230,12 +231,9 @@ public class SRUClient {
      */
     public Document parseXml(String xmlData) {
         try {
-            byte[] bytes = xmlData.getBytes("utf-8");
+            byte[] bytes = xmlData.getBytes(StandardCharsets.UTF_8);
             ByteArrayInputStream in = new ByteArrayInputStream(bytes);
             return saxReader.read(in);
-        } catch (UnsupportedEncodingException ex) {
-            log.error("Input is not UTF-8", ex);
-            return null;
         } catch (DocumentException ex) {
             log.error("Failed to parse XML", ex);
             return null;
@@ -481,7 +479,7 @@ public class SRUClient {
         String response = null;
         try {
             byte[] bla = get.getResponseBody();
-			response = new String(bla, "utf-8");
+			response = new String(bla, StandardCharsets.UTF_8);
         } catch (IOException ex) {
             log.error("Error accessing response body: ", ex);
             return null;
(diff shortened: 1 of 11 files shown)


stweil:
This first example of using a standard character set shows a positive side effect: in some cases the code can be simplified because the Java compiler gets additional information.

henning-gerhardt:
Your change looks good. Can you add similar changes as commits to this PR? I don't think that it necessary to open more PR for this kind of changes.

stweil:
I think that all UTF-8 encodings should be fixed now. There remain some code location which use java.io.UnsupportedEncodingException, but I am not sure whether those lines can be removed.

henning-gerhardt:
I found one location with "utf-8" which should replaced too: [SRUClient.java#L484](https://github.com/kitodo/kitodo-production/blob/master/Goobi/plugins/import/PicaMassImport/com/googlecode/fascinator/redbox/sru/SRUClient.java#L484).

Regarding remaining `java.io.UnsupportedEncodingException` lines: I will look into this and add a comment.

stweil:
Thanks. I added that location now as well.

stweil:
At least for the `mdfour` method I see no reason why it might throw `UnsupportedEncodingException`.

henning-gerhardt:
Not the `mdfour` method is throwing `UnsupportedEncodingException` but inner method `getBytes` called on `inPassword`in 

- [Ldap](https://github.com/kitodo/kitodo-production/blob/master/Goobi/src/de/sub/goobi/helper/ldap/Ldap.java#L495)
- [LdapUser](https://github.com/kitodo/kitodo-production/blob/master/Goobi/src/de/sub/goobi/helper/ldap/LdapUser.java#L144)

is throwing this exception.

Note for `ProductionDataImport` class: `UnsupportedEncodingException` could be removed method declaration of methods `main` and `importData`, javadoc block of `main` method and `java.io.UnsupportedEncodingException` could be removed from import statements.

henning-gerhardt:
I did found an other suspect location: [Seperator class line 115](https://github.com/kitodo/kitodo-production/blob/master/Goobi/src/de/sub/goobi/metadaten/Separator.java#L115) 
In this class package `org.apache.commons.compress.utils.Charsets` is used but must this not be package `java.nio.charset.StandardCharsets`?

stweil:
PR updated to fix class `Separator`, too.

stweil:
It looks like `UnicodeLittleUnmarked` is the old name for `UTF-16LE` (see http://docs.oracle.com/javase/1.5.0/docs/guide/intl/encoding.doc.html). I suggest to address replacing the old by the new name (with StandardCharsets) in a separate PR. IMHO that would eliminate more `UnsupportedEncodingException` cases.

stweil:
> Note for ProductionDataImport class: UnsupportedEncodingException could be removed method declaration of methods main and importData, javadoc block of main method and java.io.UnsupportedEncodingException could be removed from import statements.

Should we address this in a separate PR?

henning-gerhardt:
Regarding class `ProductionDataImport`: Remove this exception declarations. Method `importData` is a private method and method `main` does not handle this exception nor anybody is calling this `main` method in regular usage.

henning-gerhardt:
Note to `UnicodeLittleUnmarked` and `"US-ASCII` usages: this could changed into this or a separate PR. Searching for `getBytes` is useful for finding this usages.

stweil:
> Regarding class ProductionDataImport: Remove this exception declarations. Method importData is a private method and method main does not handle this exception nor anybody is calling this main method in regular usage.

Done. The commit now removes that code parts.

henning-gerhardt:
Is this PR ready for merging or do you want to add more commits?

stweil:
There are no more commits planned from my side for this PR.

PR #538, opened 2016-10-12
ProzessverwaltungForm: Fix FindBugs warning (RuntimeException)

stweil:
FindBugs report: Exception is caught when Exception is not thrown

Now runtime exceptions will no longer be suppressed.

### DIFF
--- a/Goobi/src/de/sub/goobi/forms/ProzessverwaltungForm.java
+++ b/Goobi/src/de/sub/goobi/forms/ProzessverwaltungForm.java
@@ -1737,6 +1737,8 @@ public class ProzessverwaltungForm extends BasisForm {
 				out.flush();
 				facesContext.responseComplete();
 
+			} catch (RuntimeException e) {
+				throw e;
 			} catch (Exception e) {
 			}
 		}


henning-gerhardt:
Why not replace general `Exception` by `IOException` and `DocumentException` which are thrown in this block? At least every `catch` block should contain a warning log message and maybe a visible message for the user.

stweil:
It's simply a lot of work to decide which exceptions are thrown in more than 50 blocks. But I agree that naming the single exceptions would be the ideal solution.

Adding a log message to some catch blocks would be good, but I don't think it should be done for every catch block, simply because we don't need log messages for normal program flow with expected exceptions.

henning-gerhardt:
Using a general `Exception` to catch all exceptions is a lazy solution of the past which should be avoided and replaced by more specific exceptions. A good IDE could help to get all correct specific exceptions when replacing a general exception.

Empty catch statements should be avoided as they are bad practice. On catching an exception should result in reacting of this catched exception or is this not be possible as nothing could be done then at least a log message should be generated. If this log message appears on many times then it is for me a sign of bad code.

stweil:
I can provide patches like this one which fix the FindBugs report, but I am afraid that I won't have the time to replace the general catch by the correct specific ones. So what do you suggest to do? Postpone these issues? Use the intermediate solution which I can contribute? Or would you like to take over and make the specific catch blocks?

henning-gerhardt:
I think we should postphone this general exception issues as my time is limited too.

stweil:
Why not use the intermediate solution? It would not prevent a better solution later, as `RuntimeException` can be easily found, and it would allow finding and fixing such runtime exceptions now (besides fixing the FindBugs reports).

henning-gerhardt:
Intermediate solutions tends to permanent solutions. With your change or changes you made only FindBugs happy but you hide the real problem for everyone. If there is no time for a real fix / solution we should stand on current code so FindBugs could report this issue in future every time and we don't waste time in a intermediate solution which must be changed in future.

stweil:
Your concerns can be addressed by adding a new issue or by code comments, then that information does not get lost. Most IDEs can create open issue lists from such comments.
The important point is that the current code hides potential problems with runtime exceptions (that's the reason why FindBugs complains). The intermediate solution would allow to discover such problems. Therefore I don't think it is a waste of time, and I already said that I'd do the jobs as soon as I know that it will be accepted by the release management.

matthias-ronge:
This method is used to generate a PDF download. Therefore, the servlet response is modified. After that point, it is not sipmly possible to throw anything, because the Tomcat will output/append the error HTML page to the download, resulting in downloading a corrupted PDF and no error being shown to the user. One solution seems that the [catch block changes the content type back to `text/html`](http://stackoverflow.com/a/5598266/1503237); however, this will not work if part of the response is already sent to the user. To me, the best solution looks like:

1. Rewrite the method to do as much as possible of potentially critical things prior to changing the servlet response (outside the try-catch block)
1. After changing the servlet response, even catch `Throwable`, and write it to the log file, because the output stream is no longer available. `facesContext.responseComplete();` should appear in the catch block as well. The PDF will probably still be corrupted, but chances are that it is 0 bytes then, and having a 0 bytes download in the browser’s download area usually gives a clue that something didn’t work.

A highly sophisticated solution would be [to execute the download asynchronously through Javascript returned conditionally by an AJAX request](http://stackoverflow.com/a/37520406/1503237).

henning-gerhardt:
Thank you for this clarification. It is more complicated then I thought.

I want to vote to ignore / close this specific pull request as the changes does not help in any way without more work to correct the potential issues.

stweil:
I still think that detection of runtime errors would improve the situation a little bit, but don't insist on doing that in the official version. We can run that detection in our local installation. Therefore I close this PR.

PR #545, opened 2016-10-15
Fix #540: Use class loader to get resource file

### LINKED ISSUE #540: Tomcat Folder with spaces
Due to the fact, that goobi uses two different data source connections, it is possible that hibernate works, but jdbc doesn't. Unfortunately it isn't that obvious, because login for example is working.
A special szenario (which i ran into) occures on windows. If goobi wants to use jdbc, but the tomcat is located in the programms folder "Program Files", there are spaces in the folder name. due to the wrong coding of the space, the tomcat folder is not found.
In my special case, the hibernate.config wasn't found, so the default values from src/de/sub/goobi/persistence/apache/SqlConfiguration were used which leads to an "access denied for user ..." error, which wasn't exactly the problem.

henning-gerhardt:
Old fault implementation used log4j loader mechanism to locate hibernate configuration file. This mechanism does not work with at least white spaces in path. 

New implementation is using default class loader and URL to URI mapping to proper replace special characters in path.

### DIFF
--- a/Goobi/src/de/sub/goobi/persistence/apache/SqlConfiguration.java
+++ b/Goobi/src/de/sub/goobi/persistence/apache/SqlConfiguration.java
@@ -29,10 +29,11 @@ package de.sub.goobi.persistence.apache;
  */
 import java.io.File;
 import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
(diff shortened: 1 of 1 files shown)


KHeindl:
Tried it on Windows - works like a charm :)

PR #552, opened 2016-10-20
Remove unneeded calls of method setCreationDate

stweil:
They all pass the current date which is already set in the constructors.


henning-gerhardt:
Your PR is good but made the existence of this method in different classes marked as "unused" and could be removed. Remaining question is: should this date be set from outside or should it be set hidden on construction of the class?

stweil:
I noticed that, too, but simply don't know the answer to your question. As we saw recently, it is also sometimes difficult to see whether a method is really unused, because the reflection API of Java allows calling any method, and at least hibernation makes heavy use of that.

I can add a commit which removes the `setCreationDate` methods if you think that we should do that.

henning-gerhardt:
I will merge this request even if this method is unused afterwards. As a part of a public API could plugins and other extensions / modules relay on this method.

PR #554, opened 2016-10-20
Fix several build warnings reported by Eclipse

stweil:
This patch series fixes three different kinds of warnings reported by builds in Eclipse.


Goobi/src/org/goobi/production/flow/statistics/enums/TimeUnit.java
```diff
@@ -239,6 +239,8 @@ private DateTimeFormatter getFormatter(){
 		case quarters:
 			// has to be extended by the calling function
 			return DateTimeFormat.forPattern("yyyy");
+		case simpleSum:
+			break;
```
henning-gerhardt: Constant simpleSum could even removed as this constant is never used outside this class.

stweil:
Good idea. I updated the PR – it now removes `TimeUnit.simpleSum`.

PR #556, opened 2016-10-20
FilterHelper: Use German locale for method toLowerCase

stweil:
This fixes a warning from FindBugs:
    Consider using Locale parameterized version of invoked method

Optimize also the code by converting the string to lower case only once.

### DIFF
--- a/Goobi/src/org/goobi/production/flow/statistics/hibernate/FilterHelper.java
+++ b/Goobi/src/org/goobi/production/flow/statistics/hibernate/FilterHelper.java
@@ -31,6 +31,7 @@ package org.goobi.production.flow.statistics.hibernate;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 
 import org.apache.commons.lang.text.StrTokenizer;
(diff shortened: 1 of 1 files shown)


matthias-ronge:
Is it really necessary to have that hard-coded German locale in the code? To me, the `FilterString` constants feel more like a “programming language” than any natural language, and I would vote against a locale here.

stweil:
De facto there exist currently English and German filter strings, and neither of them uses Umlauts. So the English or US locale would have worked as well. It simply does not matter. I had chosen German to reflect that some of the strings are German.

matthias-ronge:
I fully understand your intention, but however it may feel uncomfortable to other developers that may join us in future - from other countries…

henning-gerhardt:
Question for me is: must we provide language depending filter words? If we must / should provide this then we must have a simpler way to solve this as language depending enums are usable for 2 or 3 languages but not for more.

stweil:
Maybe we should stick to the English filter strings for the future. It is kind of programming language, and IMHO such languages don't need translations.

There remain more than 20 other instances where `toLowerCase` or `toUpperCase` is called without an explicit locale. How should we handle those?

PR #559, opened 2016-10-21
Updating ugh and ics library from Kitodo repositories

henning-gerhardt:
Using current source code to create updated versions of ugh and ics libraries.


stweil:
IMHO we need new version numbers and/or different names (maybe `kcs` instead of `ics`) - otherwise it's not clear what was changed.

henning-gerhardt:
This (new name and version number) will be done before releasing next major Kitodo.Production version.

stweil:
Then the first commit of this PR is not needed and should be removed.

PR #561, opened 2016-10-23
GoobiScript: Fix wrong method called

stweil:
Coverity report: CID 44624 (#1 of 1): Wrong method called (WRONG_METHOD)

### DIFF
--- a/Goobi/src/de/sub/goobi/helper/GoobiScript.java
+++ b/Goobi/src/de/sub/goobi/helper/GoobiScript.java
@@ -153,7 +153,7 @@ public class GoobiScript {
         } else if (this.myParameters.get("action").equals("exportDms")) {
             exportDms(inProzesse, this.myParameters.get("exportImages"), true);
         } else if (this.myParameters.get("action").equals("export")) {
-            exportDms(inProzesse, this.myParameters.get("exportImages"), Boolean.getBoolean(this.myParameters.get("exportOcr")));
+            exportDms(inProzesse, this.myParameters.get("exportImages"), Boolean.valueOf(this.myParameters.get("exportOcr")));
         } else if (this.myParameters.get("action").equals("doit")) {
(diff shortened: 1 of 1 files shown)


stweil:
It would be interesting to know who is using GoobiScript and see some examples.

henning-gerhardt:
Function and usage is described in [Wiki](https://github.com/kitodo/kitodo-production/wiki/GoobiScript).

We use this on a daily base to export meta data updates without images.

stweil:
Thank you for the link to the Wiki article. I assume that you did not use `exportOcr` as it is not part of that documentation. By the way: wouldn't `exportOCR` be nicer? I don't think that we'll have a compatibility issue here if we rename it.

henning-gerhardt:
It depends on your export usage.
- If we have pagination errors we use `action:exportDms exportImages:true` to re-export everything including OCR data if OCR data is available. 
- If we fix typos in meta data we only use `action:exportDms exportImages:false`. In this case exporting OCR data again does not harm any external additional processing.

As `action:export` is not documented, it should be documented and listed as possible GoobiScript.

Adding a additional command `exportOcr` is possible too.

stweil:
Just to avoid confusion: `exportOcr` is currently a parameter for `action:export`. I did not suggest an additional command, but a renaming of that parameter.

henning-gerhardt:
Just document this action with all parameters but do not rename the parameter.

stweil:
What about renaming `GoobiScript` to `KitodoScript`?

stweil:
> Just document this action with all parameters but do not rename the parameter.

Done.

henning-gerhardt:
Content of GoobiScript wiki page should be transferred to a new KitodoScript page after Kitodo.Production 2.0 is available. Old site (GoobiScript) should contain a link to new content site with an explanation of content move.

PR #565, opened 2016-11-07
Add example ruleset files

henning-gerhardt:
Adding ruleset files from SLUB Dresden and SUB Hamburg as example ruleset files.


stweil:
`Goobi/rulesets/` is a new directory? Then I suggest to ommit the first part and use simply `rulesets/`. In a next step we could get rid of the `Goobi/` directory by moving its content to the top level directory.

henning-gerhardt:
I don't want to introduce a new directory structure before next release so I used the current structure for this change.

stweil:
I see. Then I think this PR is fine.

PR #566, opened 2016-11-09
Update the UGH library

matthias-ronge:
Pull request for new UHG binary, related to https://github.com/kitodo/kitodo-ugh/pull/45


stweil:
Updating that binary to fix a misbehavior is fine, but I'd never replace a binary without changing the version. Of course it requires a little more efforts to make a bug fix release of Kitodo.UGH, but I think it is worth the trouble. This would be the third file called `ugh-1.11.1.jar`. How should users know how they differ and whether they have installed the right one?

matthias-ronge:
Whenever someone will do an update he or she will check out the whole new Kitodo Production JAR with all of its JAR files, so there should be no problem. The build version is encoded into the JAR file as well by the build process, [see the build script](https://github.com/kitodo/kitodo-ugh/blob/master/ugh/build.xml#L67-L99). If you really need insight you can visit the `class Version`.

I would prefer not to have any UGH file at all, but instead to migrate the relevant classes right into the package hierarchy of Production. But that’s another discussion.

henning-gerhardt:
I see even no problem at all: internal version get rised on every commit (which get stored inside internal and exported mets/mods file) only the file name change after big changes or to reflect changes on Kitodo.Production.

PR #575, opened 2016-11-16
Add .gitignore paths required to integrate in Eclipse

matthias-ronge:
Integrating kitodo-production as an Eclipse project requires some additional files and directories which should be ignored by Git.


matthias-ronge:
Forgot local config. Pull request closed.

PR #576, opened 2016-11-17
Add slf4j Logger to use in the future

Kathrin-Huber:


stweil:
Shouldn't the version of `slf4j-api` match the version of `slf4j-log4j12`? Then either using version 1.5.8 would be better, or the jar file for slf4j-log4j12 needs an update.

Kathrin-Huber:
I removed the slf4j-log4j12.jar with this pull request.

stweil:
I see, sorry for the noise.

PR #578, opened 2016-11-17
Integrationtests

Kathrin-Huber:
Depends on PR #576


stweil:
s/seperate/separate/ (that's a very common typo)

Otherwise nice commit, it's good to see improvements of tests.

PR #580, opened 2016-11-21
Remove libraries from lib folder, add pom dependencies

Beacze:
Remove libraries from lib folder, add pom dependencies


stweil:
@Beacze, could you please squash those 3 commits? That means combining all 3 commits into a single one:

    # From your working directory:
    git rebase -i HEAD~3
    git commit -f

Otherwise the official Git would include two commits which fail to build (and which break `git bisect`).

PR #577, opened 2016-11-17
Remove jar files which are no longer needed

stweil:
The new Maven build does not need those files.

Signed-off-by: Stefan Weil <sw@weilnetz.de>

PR #587, opened 2016-11-26
Replace local library by library from mvn repository

stweil:
Version 1.6.5 from mvn repository does not work, but 1.6.3 can be used.
There are also newer (and even very new) versions which no longer come
from opensymphony but from org.quartz-scheduler.


stweil:
I think it would be good to base the build on libraries from mvn repository as much as possible and to remove local jar files. PR #577 tried this, but was incomplete because the new pom configuration needs additional work as it is shown in this example.

henning-gerhardt:
Which kind of error occur if you tried version 1.6.5 as maven dependency?

stweil:
Done.

The [mvn repository](https://mvnrepository.com/artifact/opensymphony/quartz-all) also has versions 1.6.5 and 1.6.6, but they come from a repository called Jahia instead of Central, and obviously Jahia does not provide pom information required by Maven (builds with those versions fail).

Basti890:
You have to add the Jahia Repository to the pom.xml than it can find the jar and all the pom information.
I have already done this in a branch of mine (https://github.com/Basti890/kitodo-production/tree/remove_old_libs) where we removed many unused libraries and replaced also many libraries with dependencies.

stweil:
Great, thanks - it's always nice to learn for me. I saw that you already handle `quartz-all` in your branch, so it is sufficient to use those commits and I can close my PR.

Issue #377, opened 2015-11-23
Find a consensus on how to indent code


stweil:
Currently we have a mix of different indentation styles in goobi-production: tabs versus blanks, different indent width (4, 8, maybe others). We should find a consensus for new / modified code, maybe also for existing code.

As far as I know there are currently at least two different opinions:
- Use tabs (with width 4?) for indentation. If this variant is chosen, more questions follow:
  - Tab width 8 (standard for many editors, shell, printers, GitHub, ...), 4 or other value?
  - Should tabs only be used for indentation at the beginning of a line, or may they also be used later in the line?
  - Are blanks after tabs allowed to align code?
- Forbid tabs (only exception: files which require tabs technically, e. g. Makefiles - not uses in goobi-production). Indent with blanks (width 4).
  - Indent steps 2, 4 or 8?

Another question is line length, but I suggest to focus the discussion here on indentation and whitespace usage.

Readings
- Discussion for pull request #374 
- [Google Java Style](https://google.github.io/styleguide/javaguide.html#s2.3.1-whitespace-characters)
- More? Suggestions welcome.

henning-gerhardt:
Existing indent of a file should be used and tab indent is preferred according coding guide line as specified in http://www.goobi.org/fileadmin/groups/goobi/dokumentation/20140730_Leitfaden_fuer_GoobiEntwickler.pdf

stweil:
@henning-gerhardt, indentation is not part of the approved "Leitfaden". Citation from 2014-07-30: 

"Da die Bestimmung offensichtlich gegen unsere am 16.7. erklärte Absicht verstößt, in Version 2 „weitgehend nur Änderungen formaler Art“ vorzunehmen, würden wir zusätzlich gern auch die Frage „Leerzeichen oder Tabs“ vertagen, also die Aussage „Einrückungen erfolgen mit Tabs“ nicht aufnehmen."

stweil:
There is a related issue for Goobi.Presentation (https://github.com/goobi/goobi-presentation/issues/80).

henning-gerhardt:
Closing as solved / consensus found.

PR #592, opened 2016-12-01
Run code formatter

Beacze:
For legal reasons (new coding guideline is not yet accepted) files are formatted with tabs, not white spaces.


stweil:
> For legal reasons (new coding guideline is not yet accepted) files are formatted with tabs, not white spaces.

There is currently neither an accepted rule to use tabs nor to use spaces!

Can you adjust the code formatter so that it won't increase line length? IMHO it is not reasonable to reformat a nice formatted comment with a line length of about 80 characters into something looking worse. The proposed coding style does not require long lines but simply sets an upper limit, doesn't it.

Beacze:
Adjusting of comments was necessary to achieve the same look for all of them.

Example:
https://github.com/kitodo/kitodo-production/pull/592/commits/eaebd2168cd21c2d3e2d8303c287558df7ab4f47#diff-001a2a1cb4bc1bbc92f4b47bf3dc070e in line 49 or 52 comments were very long (ca. 150)
https://github.com/kitodo/kitodo-production/pull/592/commits/eaebd2168cd21c2d3e2d8303c287558df7ab4f47#diff-689789a50eacd4aaeaf65f2d30b4e10e in line 43 and next ones comments were short (ca. 80)

stweil:
Thanks for the explanation.

I personally still prefer shorter lines, but that's a matter of taste, so no need for more discussion from my side.

Kitodo/src/main/java/org/goobi/mq/MapMessageObjectReader.java
```diff
@@ -42,127 +42,113 @@
...
+	 * @param key the name of the set to return
 	 * @return the set requested
-	 * @throws IllegalArgumentException
-	 *             in case that getObject returns null, the returned object is
-	 *             not of type Collection, any of the content elements are not
-	 *             of type String or the collection is empty at all.
-	 * @throws JMSException
-	 *             can be thrown by MapMessage.getObject(String)
+	 * @throws IllegalArgumentException in case that getObject returns null, the returned object is not of type
+	 * 									Collection, any of the content elements are not of type String or
```
stweil: Here tabs are not used to indent the code, but in a comment. The result looks horrible.

Kitodo/src/main/java/org/goobi/mq/processors/CreateNewProcessProcessor.java
```diff
@@ -78,8 +77,8 @@
  * 			<dd>Fields to be populated manually.
  * 		</dd>
  * </dl>
- * 
- * @author Matthias Ronge <matthias.ronge@zeutschel.de>
+ *
+ * @author Matthias Ronge 	&lt;matthias.ronge@zeutschel.de&gt;
```
Beacze: There are two reasons to use html entities.
1. Checkstyle treats it as a errors <error line="5" severity="warning" message="Javadoc comment at column 58 has parse error. Details: no viable alternative at input &apos;&lt;contact@&apos; while parsing HTML_ELEMENT" source="com.puppycrawl.tools.checkstyle.checks.javadoc.NonEmptyAtclauseDescriptionCheck"/>
2. It is necessary for correct generation of JavaDoc by generators.

Beacze: Tab was wrong.

Beacze: Yes, it is good idea.

stweil:
What about the question of using tabs or spaces? For legal reasons, currently no changes are allowed (see issue #377) – if you want to wait for the Mitgliederversammlung. I suggest a preliminary quick vote from all members who are contributing code to settle this question.

Kathrin-Huber:
We are going to announce the new guidelines soon. We have an internal meeting today, where we decide on the new document. After that it's going to the mailinglist to vote.

stweil:
Be brave - try openness and transparency:

* Publish the new guidelines as a pull request for public discussion.
* Invite all interested parties to contribute.
* Vote after the discussion.

Kathrin-Huber:
I already put the guidelines to vote on the mailing list, got some comments, inserted them, and now would like to send them again as "final version", to decide on them. 
I actually don't know, how to make a pull request, for a File thats not in git, and I'm not sure, if we should add it to the project.

stweil:
Yes, @matthias-ronge and I answered on the mailing list, but it was not a real discussion as we did not get replies.

What about adding a new project, for example https://github.com/kitodo/kitodo-documentation, and using it for documents like the coding guidelines, but also for other Kitodo e. V. related documents? They would have to be in some text form (maybe markup) of course to support comments and pull requests. Showing changes would be trivial then. The mailing list would still be useful for announcements and also for people who are less familiar with GitHub or English.

Kitodo/src/main/java/de/sub/goobi/beans/Benutzer.java
```diff
@@ -176,12 +175,18 @@ public void setPasswort(String inpasswort) {
 		this.passwort = inpasswort;
 	}
 
+	/**
+	 * @return add description
```
stweil: Some IDEs can show a list of open tasks based on keywords. Eclipse for example would list all comments containing `TODO`. Maybe that keyword could be used here, too (either instead of `add description` or additionally).

Beacze: add
 After doing it we would get about 2 000 (maybe even more) TODOs.

stweil: I understand - a task list of that size is not useful. But you could add one TODO comment per file, something like `TODO: fix all "add description" comments`. Or you write an issue on GitHub, so maybe some volunteers see where they can help. All that is not necessary of course if you already have plans to fix the comments.

Issue #590, opened 2016-11-29
Cleanup web-inf/lib directory


henning-gerhardt:
An other clean up of web-inf/lib directory: replace jar files of web-inf/lib directory with maven dependencies and remove unneeded pom entries

Part of DFG based development and sprint 3.

stweil:
See PR #587 for a related pull request.

henning-gerhardt:
PR #594 merged.

PR #594, opened 2016-12-01
Cleanup webinf/lib directory

henning-gerhardt:
As mentioned in #590 is this a pull request to clean up this directory. I did not squash the commits as possible regressions of this changes could not easy fetched with `git bisect`


Kitodo/pom.xml
```diff
@@ -23,6 +23,28 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     </properties>
 
+    <repositories>
+        <!-- Lets first add the default Maven central repo,
```
stweil: Should be either "Let us" or "Let's". Maybe also replace "maven" by "Maven" in following line.

PR #598, opened 2016-12-09
Move local libraries to custom Maven repository

Basti890:
Moved all custom libraries to the Slub-Nexus Maven repository


stweil:
s/libraies/libraries/ in the subject line.

stweil:
Is it possible to replace all (or at least most) http URLs by https?

Basti890:
The Maven-Central repository supports https. (testet)
Jiha doesn't support https. (testet)
Our Slub-Nexus may support https, but just changing the url doesn't work. There for i asked the colleague who manages the Nexus-Server.

Basti890:
At the moment only maven-central supports https. I added this change to the pull request.

stweil:
Thanks a lot. It would be great to add https to the slub repo as well in the future, but for now it's fine.

PR #612, opened 2016-12-20
Remove table schritteeigenschaften from migration v.1 and v.2, add de…

Beacze:
…leting of the table if exists in v.2

Solve problem with enum type.


Beacze:
Adding of foreign keys doesn't work because of problem with data in many to many relationship's tables. To fix.

PR #613, opened 2017-01-06
Fix migration script

Beacze:
Fixing according to live database.


funkyfuture:
are the changes from german plural to english singular intended in `V2_0__Translate_tables_names_English.sql`? the plural forms seem semantically more appropriate for a table name.

on that occasion, is there a specification, technical report or alike on the database schema?

Beacze:
https://github.com/kitodo/kitodo-production/wiki/Vereinbarungen-des-DFG-Projekts

Issue #619, opened 2017-01-13
Toughen Kitodo.Production up to handle RDA meta-data


matthias-ronge:
**Approach:** RDA meta-data shall be treated as meta-data groups. Using meta-data groups is already possible. PICA data can be imported, and they can be exported to METS&lt;MODS>-XML. However, in some records the RDA meta-data is not—or not to its whole—contained in the title record, just an identifier PPN is given which has to be retrieved separately from the catalogue. Exporting to METS&lt;MODS> cannot yet write attributes which is required for the GND identifier.

**Requirements:**

- [x] Toughen the UGH library up to export fields of meta-data groups to MODS XML attributes
- [x] Port norm data import to work with meta-data groups
- [x] Toughen the PICA plug-in up to resolve norm data records included by their PPN only

The development of this feature is funded by Staatsbibliothek zu Berlin − Preußischer Kulturbesitz.

henning-gerhardt:
Is this feature for Kitodo.Production 2.x or 3.x (DFG based development)?

matthias-ronge:
2.x − developed in behalf of Staatsbibliothek zu Berlin − Preußischer Kulturbesitz

henning-gerhardt:
Thank you. I added general Kitodo.Production 2.x milestone and change this milestone after merged pull request or before next 2.x release.

PR #631, opened 2017-01-31
Import authority records into meta-data groups

matthias-ronge:
This is part of issue #619


henning-gerhardt:
I'm missing license headers in
- Goobi/src/org/kitodo/production/lugh/Mods.java 
- Goobi/src/org/kitodo/production/lugh/Mets.java

henning-gerhardt:
A question that I rediscovered from your previous  pull requests:

There a different used `NodeReferences` which points to URI's like `http://names.zeutschel.de/GraphPath`. Using this URI's results into dns errors as domain `names.zeutschel.de` does not exists. If this URI's should / could not really usable then we could / should switch the URI's to `http://names.kitodo.org/GraphPath`.

If there any other usage scenario of this URI's then please explain this scenario.

matthias-ronge:
The primary use of URIs in linked data is the same as packages in Java: To prevent accidental duplicate definitions of a name with different semantics—and to define equality of semantics in case that the same URI is reused. At the moment, these URIs are neither imported from elsewhere, nor make it into the METS/MODS documents, so there is no need to publish their definitions. However, if this would ever change in the future it would make sense to set up a domain accordingly and publish them in a machine-readable format. This is what you should have in mind.

`http://names.kitodo.org/̀  seems a good domain to go for. Often, the namespace subdomain is just called `http://ns.whatever.example/` but me, personally, I don’t like that because of the connotation of that two-letter abbreviation in context of German history.

henning-gerhardt:
Thank you for this explanation and changed URI's.

PR #636, opened 2017-02-02
Remove the unused parameter "iktlist"

matthias-ronge:
The parameter is passed half way into constructors of the Pica plug-in, but, in the end, it is never assigned to a variable, nor used otherwise.


matthias-ronge:
Please wait to merge, there is more to clean up…

PR #637, opened 2017-02-02
Non-functional changes (code clean-up) to Pica plug-in

matthias-ronge:


henning-gerhardt:
IKTList is designed as a good feature which was never get implemented or was remove in the past :(

Instead of hard coded (or now self configurable) search keys you can get current and / or changed search keys at runtime as long as search OPAC supports this feature.

matthias-ronge:
This was never implemented. And example usages of the non-functional iktlist tag rather point to an obscure file being read from the hard disk, it seems the idea was not no fetch the search fields at runtime, but to save the iktlist to a file and store it locally.

henning-gerhardt:
If this feature is needed in future then it should implemented properly.

PR #640, opened 2017-02-06
Spell 'cannot' in one word

matthias-ronge:
Though both spellings are acceptable, spelling “cannot” in one word is far more common.


henning-gerhardt:
"can not" or "cannot" - I would personally prefer "can not" as this spelling is more familiar for me.

There are other opinions? 

@stweil : what did you think?

matthias-ronge:
I checked [Oxford dictionary](https://en.oxforddictionaries.com/usage/cannot-or-can-not) before I crafted this pull request, because I wasn’t sure either. [Here is another article](http://www.dailywritingtips.com/cannot-or-can-not/) that draws the same conclusion.

I think it is a good move in regards to an international community to not just use understandable, but also _common_ English.

stweil:
The standard form is _cannot_. As Matthias has written, there are some (rare) exceptions.

henning-gerhardt:
Ok, I see I must update my preferred spelling to "cannot" :)

PR #642, opened 2017-02-08
Fix some javadoc

matthias-ronge:
This commit fixes some javadoc comments.

### DIFF
--- a/Goobi/src/org/kitodo/production/lugh/ld/AccessibleObject.java
+++ b/Goobi/src/org/kitodo/production/lugh/ld/AccessibleObject.java
@@ -11,6 +11,7 @@
 
 package org.kitodo.production.lugh.ld;
 
+import java.nio.BufferOverflowException;
 import java.util.NoSuchElementException;
 
 /**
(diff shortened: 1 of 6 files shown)


Goobi/src/org/kitodo/production/lugh/ld/AccessibleObject.java
```diff
@@ -11,6 +11,7 @@
 
 package org.kitodo.production.lugh.ld;
 
+import java.nio.BufferOverflowException;
```
stweil: How is this modification related to the subject?

henning-gerhardt: `BufferOverflowException` is used inside a [JavaDoc comment](https://github.com/matthias-ronge/kitodo-production/blob/483fefbf78c872002dedcf2cf8996731df13ce3e/Goobi/src/org/kitodo/production/lugh/ld/AccessibleObject.java#L30) and a good IDE mentioned the programmer that this symbol `BufferOverflowException` could not be resolved. Matthias try to fix that and as this it is related to pull request title.

PR #645, opened 2017-02-09
Clean up logger calls

matthias-ronge:
**The mess before:** Logger variable is 72× spelled `logger` ([preferable spelling](http://stackoverflow.com/q/1417190/1503237)), 35× `myLogger`, 1× `LOGGER`, 1× `modsLogger`. 3× loggers are declared `protected`, 3× subclasses use the logger of the parent class (logging the wrong class name, cannot be configured separately). 1× logger calls were qualified with `this.` prefix.

**After clean-up:** 112× `private static Logger logger`


henning-gerhardt:
Good catch. 

What did you think about different logger class names and used classes like in [CopyPrrocess](https://github.com/kitodo/kitodo-production/blob/2.x/Goobi/src/org/goobi/production/cli/helper/CopyProcess.java#L75) (there are more like this usages).

matthias-ronge:
I agree, this is another sort of wrong usage.

Issue #290, opened 2014-12-19
ICS: missing file errorfile.jpg


henning-gerhardt:
While investigation  issue #289 in 1.11.x branch I found first IOException:

```
ERROR 2014-12-19 09:05:26,789 (ContentServer.java:doGet:159)
        An error occured
java.io.FileNotFoundException: /var/lib/tomcat7/webapps/goobi/WEB-INF/classes/errorfile.jpg (No such file or directory)
        at java.io.FileInputStream.open(Native Method)
        at java.io.FileInputStream.<init>(FileInputStream.java:146)
        at java.io.FileInputStream.<init>(FileInputStream.java:101)
        at org.goobi.presentation.contentServlet.controller.GetMetsPdfAction.run(GetMetsPdfAction.java:344)
        at de.unigoettingen.sub.commons.contentlib.servlet.controller.ContentServer.doGet(ContentServer.java:156)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:621)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:722)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:305)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:210)
        at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:225)
        at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:123)
        at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:472)
        at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:168)
        at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:98)
        at org.apache.catalina.valves.AccessLogValve.invoke(AccessLogValve.java:927)
        at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:118)
        at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:407)
        at org.apache.coyote.ajp.AjpProcessor.process(AjpProcessor.java:200)
        at org.apache.coyote.AbstractProtocol$AbstractConnectionHandler.process(AbstractProtocol.java:579)
        at org.apache.tomcat.util.net.JIoEndpoint$SocketProcessor.run(JIoEndpoint.java:310)
[...]

PR #641, opened 2017-02-07
Adds a feature to import delegate records during PICA import

matthias-ronge:
By an additional entry in opac.xml, a “resolve rule” can be defined. If a field in the imported PICA hit matches the rule’s `tag` and `subtag`, the value from that subtag is used to query the catalogue again. In a mapping, values from the result of that second query can be added as (fictious) subfields to the title field that the query was started from.

Example mapping:

    <catalogue title="…">
        
        <resolve tag="028C" subtag="9" searchField="12">
            <map tag="003U" subtag="a" asSubtag="g" />
        </resolve>
    
Example field on a title record:
    
    <field tag="028C">
        <subfield code="d">Dirk</subfield>
        <subfield code="a">Fahlenkamp</subfield>
        <subfield code="9">69749487X</subfield>
        <subfield code="8">Fahlenkamp, Dirk *1952-*</subfield>
    </field>
    
Example result after successful resolving:
    
    <field tag="028C">
        <subfield code="d">Dirk</subfield>
        <subfield code="a">Fahlenkamp</subfield>
        <subfield code="9">69749487X</subfield>
        <subfield code="8">Fahlenkamp, Dirk *1952-*</subfield>
        <subfield code="g">http://d-nb.info/gnd/172559227</subfield>
    </field>

This is part of issue #619.

**Additional edit:** German javadoc comments in `PicaPlugin.java` have been translated into English.

Issue #291, opened 2014-12-19
ICS: missing file sRGB.icc


henning-gerhardt:
While investigation issue #289 and after applying dirty solution mentioned in #290 next error message appear:

```
ERROR 2014-12-19 09:05:26,685 (GetMetsPdfAction.java:setPdfManagerDefaults:446)
        No ICC profile given!
```

dirty solution was (similar to #290):
- unzip ics-1.1.5.jar to temp directory
- copying  sRGB.icc to WEB-INF/classes

frank-ulrich-weber:
The missing file sRGB.icc prevents the integrated content server to create PDF/A conform PDF files.
Default Setting in contentServerConfig.xml 
<defaultPdfConfig ... writeAsPdfA="true" .../>
The PDF file will be created anyway but not PDF/A conform. No information to the user about that fact beside the "No ICC profile given!" entry within the log file.

henning-gerhardt:
@frank-ulrich-weber : Does it help you if you unzip `ics-1.1.5.jar` to a temporary directory and copy `sRGB.icc` to `<goobi-installation>/WEB-INF/classes` directory?

frank-ulrich-weber:
Yes, that solves the problem.

henning-gerhardt:
Then should both files `sRGB.icc` and `errorfile.jpg` (from #290) should be added to `WEB-INF/classes` directory.

henning-gerhardt:
Fixed with pull request #418 (https://github.com/goobi/goobi-production/commit/1688ffc588492bb23a35d5612c0c5a3be64588c8)

PR #647, opened 2017-02-10
Rename configuration files

henning-gerhardt:
Renaming configuration files could break a lot of systems (developer, test or productive systems) or pending developments (pull request #641 for example), so I'm unsure if we could or should rename this files.


matthias-ronge:
Renaming of configuration files should be feature of DFG development, which requires a different setup of the programming environment anyway (Maven…). Git should be able to handle pull requests through file renames.

But I’ve got some other points:

 - All configuration files should be named by one `String` in the class `org.goobi.production.constants.FileNames` and should only be referenced from there (_except_ plugins which should use their own constants, of course). If something breaks it is a good hint to find locations where this is not the case.
 - Why still have the `kitodo_` prefix at all? It is annoying when configuring a system from the shell (just additional typework) and why have both files written with underscores and camel-case? An even better solution would also point out which files have to stay in the `classes/` directory, and which can go to the `KonfigurationVerzeichnis` directory?
 - Perhaps rename `KonfigurationVerzeichnis` and `xsltFolder` to something of the same pattern, and both in English?
- XML files should be named like their root element. Renaming them should include renaming the root element (in most cases, it is not evaluated in the source code anyway).
 - The file `goobi_opac.xml` is read from both the core and the plugin, but both read other parts of it only. It should be split up into two files, each of them containing only the information read in that place.
 - Delete `webapi` config file. We don’t have a plugin using it.
 - Config files from the `WEB-INF/` directory could then move into the config tree as well.
 - By the way, what do `errorfile.jpg` and `iRGB.icc` in the `config/` dir? Wouldn’t they be better in `Kitodo/WEB-INF/classes/` directory [to be created]?

On a throw:

    config
     + classes
     |  + c3p0.properties
     |  + config.properties [goobi_config.properties]
     |  + contentServerConfig.xml
     |  + ehcache.xml
     |  + hibernate.cfg.xml
     |  + log4j.properties
     |  + login_blacklist.txt [goobi_loginBlacklist.txt]
     |  + modules.xml (rename root element, currently `ModuleDesc`)
     |  + quartz.properties
     + config_directory
     |  + collections.xml [goobi_digitalCollections.xml]
     |  + doctypes.xml [goobi_opac.xml] (*)
     |  + projects.xml [goobi_projects.xml]
     |  + properties.xml [goobi_processProperties.xml]
     |  + representations.xml [goobi_metadataDisplayRules.xml]
     + plugins
     |  + opac
     |     + mods_plugin
     |     |  + mods_plugin.xml [kitodo_mods_opac.xml]
     |     + pica_plugin
     |        + diactitics.txt [goobi_opacUmlaut.txt]
     |        + language-codes.txt [goobi_opacLanguages.txt]
     |        + pica_plugin.xml [goobi_opac.xml] (*)
     + WEB-INF
     |  + faces-config.xml
     |  + metadata-editor-config.xml
     |  + reverse-proxy.xml
     |  + web.xml
     + xslt_directory
        + docket_multipage.xml
        + docket.xml

henning-gerhardt:
My goal with this pull request is to end explanations and discussions about this goobi prefix as the product is called Kitodo which I do since a couple of months for new Kitodo users. The DFG based developement has some other approaches to solve this which I could not include into the 2.x branch.

If have no problems with removing this file name prefix. I have no opinion regarding camelCase or some other naming schema of configuration files as I can work with all of them.

Your other improvements and questions should be done separatly from renaming this configuration files. You can open a pull request on each of your suggestions :)

I will answer only to a few:

- `webapi`as long as we has this interface and all buisines logic for this kind of webapi usage, this configuration file should stay as this file is a white list access file

- config files from `WEB-INF` directory: this files are located there as they needed for running the application. We could move them inside source code to an other location but move them on creating the WAR file back to this location. One other thing is that Eclipse (and maybe other IDE's) expecting this files inside `WEB-INF` directory for running on local development systems

- `errorfile.jpg` and `iRGB.icc` files: mechanis inside Kitodo.Contentserver for finding / detecting this files is wrong for the use case inside of an application. Logic look into the whole class path of the application and not into the class path of the jar file. See closed issues #290 and #291.

matthias-ronge:
My gut instinct tells me _not to_ rename the files in the 2.x branch for the reasons you already listed. We also still have the `de.sub.goobi` and `org.goobi` packages here for the same reasons.

henning-gerhardt:
I let this PR open if the community or more people request renamed configuration files.

henning-gerhardt:
Closing as outdated and obsolete.

PR #648, opened 2017-02-13
Remove e-mail address

matthias-ronge:
Primary means of contact for issues with the code should be the GitHub issue tracker. E-mail address  `contact@kitodo.org` is provided in the header as well. Having e-mail addresses in the generated code documentation may be misleading.

To indicate unknown authors, [there](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@author) is the string constant `unascribed`.


henning-gerhardt:
This PR needs a rebase as not mergeable with current 2.x branch.

henning-gerhardt:
Personally I would remove all author tags as they are not needed and are a relic of non or very old version control systems.

PR #657, opened 2017-02-21
Remove outdated and outcommented test classes

henning-gerhardt:
Classes are outdated, unused and pagination behaviour is tested by `PaginatorTest`.


henning-gerhardt:
I added the removing of `ImportOpacTest` class in this pull request (previous in PR #658) as this removing belongs to this PR.

PR #658, opened 2017-02-21
Remove main methods

henning-gerhardt:
`main()` methods could only called while development and did not replace good tests as they maybe suggest. Removing this methods set a lot of other methods and classes into a not used state. This methods and classes are removed too.


henning-gerhardt:
Two removed tests `BenutzerTest` and `BenutzergruppenTest` classes stay as suggested by @Kathrin-Huber and removed `ImportOpacTest` class is removed in PR #657 .

Issue #650, opened 2017-02-16
Use the same sort of white space everywhere


matthias-ronge:
Leaving tab indent in existing files and using four spaces indent in newly created files (in 2.x branch) is a mess when editing the project with Eclipse. The spacing is configured in the “Code formatter” settings and cannot be defined on a per-file basis. This means that, whatever file you are currently editing, you always have to go to the project settings and change the code formatter used, because the setting also influences code automatically generated (such as Javadoc comment templates). You cannot safely cut/paste code accross files any more, nor use “save all” button any more.

Most files use tab in 2.x branch. This should also be applied to new files.

matthias-ronge:
… or the other way ‘round. It really exists both, but editing files is hell the way it is. I don’t have a best solution for it.

henning-gerhardt:
Stefan Weil made a statistic about indent usage and with his numbers there are more files with space indent as tab indent. New coding guidelines (not official released yet) prefer space indent. So solution should be that all files of 2.x branch should be converted to space indent. 

Issue is related to #377

matthias-ronge:
GitHub unified diff is capable to ignore white space differences, which will be necessary if we proceed here. You have to add **?w=1** to the URL in the browser to do so. Found [in the blog](https://github.com/blog/967-github-secrets).

PR #665, opened 2017-02-23
Use the same sort of white space everywhere

matthias-ronge:
Proposed fix for #650 

Uses (up to) 4 spaces instead for each tab character. Diff with URL option **&w=1** whill show _Whitespace-only changes_ everywhere.

**Note:** This pull request only replaces tabs by spaces. This was scripted. It does **not** correct wrong indentation which still might be necessary in the one or other file.


henning-gerhardt:
I found 3 xsl files with tab characters inside
- `Goobi/config/docket.xsl`
- `Goobi/config/docket_multipage.xsl`
- `Goobi/plugins/opac/ModsPlugin/org/kitodo/production/plugin/CataloguePlugin/ModsPlugin/kalliope2kitodo.xsl`
I think this files should converted too.

In general: should the `css` files converted? What is there the common indent style?

matthias-ronge:
I did only run the script for files with extensions `.xml`, `.html`, `.java`, `.jsp`, and `.properties`. We have binary files in there, why I did not specify ‘all’. `.xsl` and `.css` feels good. I think we should have only-spaces policy for all text files. However, I consider vector images as binary since I generate and edit them with a paint program, not with a text editor.

henning-gerhardt:
I think converting the `.xsl` and `.css` files should be enough. `.svg` and `.js` files should not be converted.

matthias-ronge:
`.js` should also because it is source code as well (Javascript), or why not?

henning-gerhardt:
It would look like that we maintain and / or modifying this files but we using them only (and really old versions of them) as they are.

If there are other opinions I will respect them and then they should converted too.

matthias-ronge:
Fixed `.xsl` and `.css` files. We should inspect what is up with that javascript stuff. I added another issue for that.

Issue #659, opened 2017-02-21
Create multiple structural elements at once


matthias-ronge:
Adding the same type of structural elements over and over again manually is tedious work. This shall be automatized in future.

**Approach:** The number of structual elements to create can be chosen arbitrarily. A meta-datum can be chosen to be automatically added, and support for a counter should be provided.

The development of this feature is funded by Landratsamt Esslingen.

matthias-ronge:
Development is for 2.x branch

henning-gerhardt:
Can this issue be closed as PR #676 is merged?

PR #676, opened 2017-03-02
Multiple structural elements

matthias-ronge:
Solution for #659


Goobi/src/de/sub/goobi/metadaten/TreeInsertionMode.java
```diff
@@ -0,0 +1,44 @@
...
+            return AFTER_ELEMENT;
+        case 3:
+            return AS_FIRST_CHILD;
+        case 4:
+            return AS_LAST_CHILD;
+        default:
+            throw new IllegalArgumentException("For string: " + i);
+        }
+    }
+}
```
henning-gerhardt: Missing new line at end of file

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -1067,141 +1071,123 @@ public String KnotenDelete() throws IOException {
...
-            for (Iterator<DocStruct> iter = alleDS.iterator(); iter.hasNext();) {
-                parent.addChild(iter.next());
+                // Set the new children on the edited element
+                for (DocStruct child : newChildren) {
+                    edited.addChild(child);
+                }
             }
+            return createdElements.iterator().next();
+        } catch (UGHException iek) {
+            throw new RuntimeException(iek.getMessage(), iek);
```
henning-gerhardt: It is a good idea to throw a `RuntimeException` which never get catched and is displayed to the user?

Goobi/pages/incMeta/Strukturdaten.jsp
```diff
@@ -90,11 +93,45 @@
...
+            <htm:td styleClass="eingabeBoxen_row3">
+                <h:commandLink action="#{Metadaten.addNodesClick}" value="#{msgs.strukturelementHinzufuegen}"
+                    target="links" style="float:right;" rendered="#{not Metadaten.addServeralStructuralElementsMode}" />
+                <htm:script rendered="#{Metadaten.addServeralStructuralElementsMode}">
+                    <h:outputText value="    function checkAddServeralStructuralElements(){
+        var inputBox = document.getElementById('formular2:elementsCount'); 
+        var count = inputBox.value;
+        if (count != '' && !isNaN(count)) {
+            var value = parseInt(count);
+            if(value > 28) {
```
henning-gerhardt: How did you come to a value of `28` or more as to many elements for inserting?

matthias-ronge: Our use case is to have years in the meta-data element. I mistyped the year in the count field and created 1982 structure elements. So there should be a warning when you are adding many elements. How much is _many_? I just guessed that 29 could be a good trade-off. I think it’s rare that books have so many chapters, and if it ever happens, well, it’s just a warning, and you can still confirm to do so.

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
matthias-ronge: It will be shown as a stack trace the same for checked and unchecked exceptions. In my eyes there are some unfortunate design decisions in the Ugh and that UGHException and its subclasses are _checked_ exceptions is one of them. If _docstruct has no type_ ever happens the user cannot handle it anyway, it’s a programming error then.

Goobi/pages/incMeta/Strukturdaten.jsp
henning-gerhardt: Time will tell us if this value of `28` is a good value or not.

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
henning-gerhardt: You are aware of this problem and it is a problem which must be solved inside UGH.

PR #680, opened 2017-03-07
Add OracleJDK and OpenJDK to travis build

henning-gerhardt:
Without defining a JDK, travis is using his default JDK which could OracleJDK8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@
 
 language: java
 
+jdk:
+  - oraclejdk7
+  - openjdk7
+
 addons:
   apt:
     packages:


stweil:
Why both oraclejdk7 and openjdk7? Wouldn't one of them be sufficient? I'd avoid using any Oracle JDK for legal reasons.

henning-gerhardt:
Oracle JDK is still used on Microsoft Windows Systems or on Enterprise Linux Systems (SuSE, RedHat, ...), so we should use both for testing.

stweil:
Thanks for the explanation. Maybe someone is interested to know that Red Hat also supports [OpenJDK for Enterprise Linux and Windows](https://developers.redhat.com/products/openjdk/overview/). I assume that SuSE also provides OpenJDK.

Using Oracle JDK can become really expensive if you don't restrict that usage to your PC, because all other usage is not covered by Oracle's license. I don't know whether using Oracle JDK on a continuous integration server is allowed by the license.

henning-gerhardt:
Travis is using as default JDK OracleJDK - according to their [documentation](https://docs.travis-ci.com/user/languages/java/#Overview) and [tests results](https://travis-ci.org/kitodo/kitodo-production/builds/208601954). I did only add OpenJDK to test builds to discover maybe existing differences. 

I pinned JDK to Java 7 as this is current minimal used and supported version for all branches on any Kitodo.Production related project except master branch of Kitodo.Production. Reason for this is, that master branch contains some changes which only available with Java 8 (ElasticSearch Rest Interface).

PR #686, opened 2017-03-14
Define flyway phase as a parameter

Beacze:
Flyway migration has to be executed only once but profile must be defined in parent pom.xml and every child pom.xml file. To achieve it, phase name as parameter was introduced.


OpacPica-Plugin/pom.xml
```diff
@@ -32,6 +32,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <parent.basedir>${basedir}/../</parent.basedir>
         <checkstyle.config.location>${parent.basedir}/config/checkstyle.xml</checkstyle.config.location>
+        <phase.prop>none</phase.prop>
```
Kathrin-Huber: you don't need this property in the child poms, because you already defined it in the parent pom. Only overwrite it, where it should be executed.

Beacze: done

Beacze:
Done

PR #705, opened 2017-03-27
Split REST Client implementation class

Beacze:
Preparation for search package

PR #710, opened 2017-03-29
Configure elasticsearch

Kathrin-Huber:
Depends on #705 and #704


Kitodo-DataManagement/src/main/java/org/kitodo/config/ConfigMain.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.commons.configuration.reloading.FileChangedReloadingStrategy;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+
+public class ConfigMain {
+    private static final Logger myLogger = Logger.getLogger(ConfigMain.class);
+    private static volatile PropertiesConfiguration config;
+    private static final String CONFIG_FILE = "elasticsearch.properties";
+
+    public static PropertiesConfiguration getConfig(String configFile) {
```
henning-gerhardt: Missing JavaDoc

henning-gerhardt:
- Until now there is only one big configuration file called `goobi_config.properties`. Should this file split up into several independent configuration files? If not the configuration for ElasticSearch should go into this file.
- `ConfigMain` as base (?) configuration class should use `goobi_config.properties` or none file but not `elasticsearch.properties` file.

Maybe this is a better solution: 
- define a base (abstract (?)) `ConfigMain` configuration class which provide all necessary general functionality and maybe implement a `ConfigurationInterface`. This class should not define where a configuration file should be found (class path or somewhere else into the file system)
- `ConfigCore` class extends the `ConfigMain` class and define `goobi_config.properties` as its used configuration file.
- `ElasticSearchConfig` extends  `ConfigMain` class too and use a `elasticsearch.properties` file. 
- Every configuration class could extends or use `ConfigCore` class if this configuration class needs some values from `ConfigCore` class.

Kitodo/src/main/resources/goobi_config.properties
```diff
@@ -533,3 +533,7 @@ superadminpassword=098f6bcd4621d373cade4e832627b4f6
 useSwapping=false
 # Pfad fuer Auslagerung
 swapPath=/usr/local/kitodo/swap
+# Elasticsearch properties
+elasticsearch.host=localhost
+elasticsearch.port=9200
+elasticsearch.protocol=http
```
henning-gerhardt: Please add a new line at end of file

PR #712, opened 2017-04-03
Checkstyle and findbugs logs

Kathrin-Huber:
remove checkstyle log errors from console. Only show critical errors for findbugs.


henning-gerhardt:
I did not see any findbug reports inside `target/findbugreports` with a `mvn clean package` command line call. What did I wrong?

Issue #470, opened 2016-05-26
Migration from Goobi.Production to Kitodo.Production


stweil:
This issue was opened to collect requirements, coordinate activities and discuss different options needed for the migration from Goobi.Production to Kitodo.Production.

Some selected questions (numbered for later reference):
1. Rename directory `Goobi/` to `Kitodo`?
2. Rename `goobi` subdirectories below `Goobi`? Which ones?

matthias-ronge:
Renaming the packages could go together with generally reorganizing them, giving up the duplicate branches `de.sub.` and `org.`. Here is an approach how package structure could better represent the contents of the app:

![kitodo_packages](https://cloud.githubusercontent.com/assets/3040657/15652352/e0d50498-2685-11e6-9d3d-593477092355.png)

I would encourage to import the relevant *UGH classes* as well.

henning-gerhardt:
Regarding general renaming from Goobi to Kitodo: In a short time period only visible `Goobi` names appearance should be replaced. No internal and only for a developer visible name should rename. In a long time period internal and for developer visible names should be replaced as it could break a lot of stuff.

Regarding package name renaming and in my humble opinion as developer: in a short time period there should no changes to current packages names as it breaks every existing third party plugin / module. In a med time period there should be an official interface declaration (API / ABI / SDK?) which should be used and changed only in major revisions.

Integration of UGH classes: should be discussed in general module concept discussion. Now UGH classes are in modular and separated from other source code and can be reused in other JAVA based projects.

stweil:
@henning-gerhardt, what about making https://github.com/kitodo/kitodo-production the primary source (and https://github.com/goobi/goobi-production the fork)? The same change is needed for https://github.com/goobi/goobi-presentation, too.

henning-gerhardt:
@stweil : I will forward your suggestion to @sebastian-meyer.

sebastian-meyer:
Since the Goobi GitHub account will most likely be transfered to Intranda in the future, I don't think we need to worry about which repository to be the fork. Sooner or later the Goobi repository won't exist anymore and therefore the Kitodo repository will become the main repository anyway.

Aside from that, I think the Kitodo repository should be the fork, because essentially this is what it is: the Goobi repository is older and the Kitodo repository is forked from it - not the other way round.

stweil:
[...]

PR #713, opened 2017-04-03
Rename goobi to kitodo

Kathrin-Huber:


stweil:
s/varaibles/variables/ in commit message.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/apache/SqlConfiguration.java
```diff
@@ -26,9 +26,9 @@
 public class SqlConfiguration {
 
     private String dbDriverName = "com.mysql.jdbc.Driver";
-    private String dbUser = "goobi";
+    private String dbUser = "kitodo";
```
stweil: Maybe something like `kitodo_db_user` would be even better here.

Kitodo/setup/migrate_index/add_new_index_1.11.x.sql
```diff
@@ -1,5 +1,5 @@
 --
--- Add new indexes for Goobi.Production CE newer than 1.11.2
+-- Add new indexes for Kitodo.Production CE newer than 1.11.2
```
stweil: I think `CE` can be removed for Kitodo (also in other places).

stweil:
Wouldn't it be better to always use `org.kitodo` instead of `de.sub.kitodo` if directories and class hierarchies are renamed?

Kitodo/setup/migrate_index/add_new_index_1.11.x.sql
henning-gerhardt: This comment should be renamend to "Add new indexes for Kitodo.Production 2.0.0" and the file name to "add_new_index_2.0.0.sql" as in 2.x branch.

henning-gerhardt:
I'm unsure about the whole renaming of classes and methods which contain the word "goobi". This change would make it nearly impossible or with a lot of more work to migrate some new features from 2.x to master branch.

stweil:
> I'm unsure about the whole renaming of classes and methods which contain the word "goobi".

So I am, too.

Kitodo/src/main/resources/kitodo_metadataDisplayRules.xml
```diff
@@ -12,8 +12,8 @@
  *
 -->
 
-<!--<displayRules xmlns="http://meta.goobi.org/displayRules" xmlns:tns="http://meta.goobi.org/displayRules" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://meta.goobi.org/displayRules http://meta.goobi.org/displayRules.xsd">-->
-<displayRules xmlns="http://meta.goobi.org/displayRules" xmlns:tns="http://meta.goobi.org/displayRules">
+<!--<displayRules xmlns="http://meta.kitodo.org/displayRules" xmlns:tns="http://meta.kitodo.org/displayRules" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://meta.goobi.org/displayRules http://meta.goobi.org/displayRules.xsd">-->
+<displayRules xmlns="http://meta.kitodo.org/displayRules" xmlns:tns="http://meta.kitodo.org/displayRules">
```
matthias-ronge: Be careful when renaming name spaces. It may add an additional layer of complexity when updating legacy installations. Changing the name space to one which is under our control is acceptable. We should keep in mind that it might turn out to be helpful to publish the semantics of the objects in the future, so the URL should be chosen carefuelly. We already introduced the domain base `http://names.kitodo.org/` for [GPath](https://github.com/kitodo/kitodo-production/blob/2.x/Goobi/src/org/kitodo/production/lugh/ld/GraphPath.java#L114-L137)s in `2.x`, [see discussion](https://github.com/kitodo/kitodo-production/pull/631#issuecomment-276664018). I would like to recommend to use the following name space for legacy display rule definitions:

    xmlns="http://names.kitodo.org/legacy/displayRules"

matthias-ronge:
`de.sub.kitodo` should be changed to `org.kitodo`. However, renaming Text occurrences is a different thing than cleaning up the package structure, which should perhaps be separated. I had proposed a possible [package tree](https://github.com/kitodo/kitodo-production/issues/470#issuecomment-222507851) some time ago.

Kathrin-Huber:
We are starting with renaming goobi to kitodo.
Changing the package structure is an evolving process, which we already startet. We put new stuff and cleaned stuff in org.kitodo, so i didn't want so mash it all together. There are already some files moved from org.goobi to org.kitodo wich are now mixed up with the new classes, but i didn't wanted to make an "old.kitodo" package or whatever.

All semantic changes like package structure or dbUser names should be done in another pullRequest.

PR #726, opened 2017-04-06
Clean index after test classes

Beacze:


Kitodo/src/test/java/org/kitodo/services/data/BatchServiceIT.java
```diff
@@ -36,8 +37,9 @@ public static void prepareDatabase() throws DAOException, IOException, ResponseE
     }
 
     @AfterClass
-    public static void cleanDatabase() {
+    public static void cleanDatabaseAndIndex() throws IOException, ResponseException{
```
henning-gerhardt: How can a no-op method throw an Exception and why is this code comment out? If this method is not really used or not providing any functionality then it should be removed. Method place holders are confusing.

Beacze: It is preparation for upcoming. This pull request can wait for next ones which will use this functionality.

henning-gerhardt: It is better to introduce new functionality if this functionality is used. Prepared functionality got sometimes lost and a future review everybody could not explain this. I think it is better remove this and re-add this if this is needed.

Beacze: Fixed, now method is only used by classes which currently make operations on ElasticSearch index.

Issue #723, opened 2017-04-06
"mvn clean package" fails on freshly checked out project


Ishiido:
there are tests that fail as they probably require preconditions

```
Results :

Failed tests:   shouldAddType(org.kitodo.data.elasticsearch.index.IndexRestClientTest): Add of type has failed - document id 1!

Tests in error: 
  org.kitodo.data.elasticsearch.search.SearchRestClientTest: Connection refused
  org.kitodo.data.elasticsearch.search.SearcherTest: Connection refused
  shouldAddDocument(org.kitodo.data.elasticsearch.index.IndexRestClientTest): Connection refused
  shouldDeleteType(org.kitodo.data.elasticsearch.index.IndexRestClientTest): Connection refused
  shouldDeleteDocument(org.kitodo.data.elasticsearch.index.IndexRestClientTest): Connection refused

Tests run: 28, Failures: 1, Errors: 5, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Kitodo Production .................................. SUCCESS [  3.746 s]
[INFO] OpacPica-Plugin .................................... SUCCESS [  7.035 s]
[INFO] Kitodo - Data Management ........................... FAILURE [  9.814 s]
[INFO] Kitodo - API ....................................... SKIPPED
[INFO] Kitodo - Core ...................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
```

Beacze:
It is necessary to have running ElasticSearch server to build the project.

Ishiido:
Ah - I see. Then https://github.com/kitodo/kitodo-production/blob/master/README.md gives incorrect instructions on how to run a build process. Should the relevant tests be ignored by annotation?

I have no idea on setting up ElasticSearch.

henning-gerhardt:
README in master branch  got new instructions for ElasticSearch (version, installation, ...). I hope this helps you.

stweil:
I got the same build error.

Is it possible to skip tests which cannot work because of missing preconditions (for example no ElasticSearch server) and report that only without a build failure? I appreciate that tests are run by default, but it would also be good to allow simple builds with incomplete tests.

henning-gerhardt:
So far as I know is there no way to skip the test execution if a necessary dependency is not available.

Ishiido:
[...]

PR #728, opened 2017-04-07
Add information about ElasticSearch to README.md

Beacze:
Resolve issue https://github.com/kitodo/kitodo-production/issues/723

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ and packaging of the application is [Maven](https://maven.apache.org/).
 ### Dependencies
 
 Available dependencies are fetched from Maven Central. Further dependencies (not available in Maven Central) are located in Kitodo/src/main/webapp/WEB-INF/lib.
-The only thing that is not shipped is a Tomcat specific servlet API. This should come along with the Tomcat distribution you are building against.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ and packaging of the application is [Maven](https://maven.apache.org/).
 ### Dependencies
 
 Available dependencies are fetched from Maven Central. Further dependencies (not available in Maven Central) are located in Kitodo/src/main/webapp/WEB-INF/lib.
-The only thing that is not shipped is a Tomcat specific servlet API. This should come along with the Tomcat distribution you are building against.
+Two things that are not shipped are a Tomcat specific servlet API and an ElasticSearch server. First thing should come along with the Tomcat distribution you are building against. Second, you need to download by yourself from [ElasticSearch Download Page](https://www.elastic.co/downloads/past-releases/elasticsearch-2-3-1) and start before Maven build. On the Windows system ElasticSearch starts after open elasticsearch.bat file from bin folder.
```
henning-gerhardt: You are linking to version ElasticSearch 2.3.1 but inside Maven dependencies and TravisCI we are using ElasticSearch 5.1.1  and installation guide for 3.x mentioned ElasticSearch 5.x. Should the download link not point to at least ElasticSearch 5.1.1 ?

Beacze: Here every version since 2.0 is suitable. They are compatible. I'm not sure but I would suspect that every version in range 2.0 - 5.1.1 will be suitable for user. Problematic can be higher.

henning-gerhardt: If a running ElasticSearch 2.x instance is a minimum requirement then we should could two things: downgrade ElasticSearch inside TravisCI tests to minimum supported ElasticSearch version and update installation guide for 3.x even to ElasticSearch 2.x. Or we should raise ElasticSearch to a 5.1.1 version everywhere (README.md, installation instruction of 3.x, ...) .

Beacze: We use everywhere ElasticSearch 5.1.1. Travis already uses it. Installation instruction will be changed.

PR #730, opened 2017-04-07
Add response filtering for asynchronus adding of documents to index - addType() method

Beacze:


Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/index/IndexRestClient.java
```diff
@@ -120,7 +121,15 @@ public boolean deleteIndex() throws IOException, ResponseException {
         return processStatusCode(indexResponse.getStatusLine()) == 200;
     }
 
-    private int processStatusCode(StatusLine statusLine)  throws ResponseException {
+    private void filterAsynchronousResponses(ArrayList<String> responses) throws ResponseException {
+        for (String response : responses) {
+            if (response.contains("HTTP/1.1 400")) {
```
henning-gerhardt: What about other error messages from server side (like 404, 500, ...) ?

Beacze: Solved! Now it will throw exception if status code is not 200 or 201.

PR #732, opened 2017-04-08
Fix two small issues reported by Codacy

stweil:


stweil:
Codacy was recently added to Kitodo.Presentation. It might also be useful for Kitodo.Production – both for cleaning style issues, but also for detection of real bugs.

stweil:
There are about 200 lines of code using `size() > 0`, `size() == 0` or `size() != 0`. For some kinds of collections, `isEmpty()` would be faster. I am not aware of performance problems with Kitodo.Production, but it might be a good idea to replace those patterns nevertheless, for example with Coccinelle.

henning-gerhardt:
We have already many tools inside and outside build process like TravisCI, Coverity, Findbugs and checkstyle checker. So I don't think that adding a new tool which discover no new defects or errors is useful.

Beacze:
I think that it is reasonable to fix those problems especially if else boolean and size checking. If it goes about unnecessary return statements. In fact, not every time they are not necessary, even if it reports so. For some cases they need to stay.

I agree with Henning that adding one more tool for build process doesn't really make sense.

stweil:
@Beacze, @henning-gerhardt, feel free to use the [Codacy results](https://www.codacy.com/app/UB-Mannheim/kitodo-production/dashboard) anyway.

PR #744, opened 2017-04-19
Remove double dependancy - org.apache.myfaces.tomahawk

Beacze:
It exists already in Kitodo Data Management - pom.xml line 122.


henning-gerhardt:
Why did you not remove this dependecy from the DataManagement module as it is not needed?

PR #756, opened 2017-04-25
Move functionality to file service

Kathrin-Huber:
renaming will be fixed in second pull request (because of some unneccesary conflicts)


Kathrin-Huber:
This is one of three pull requests (I didn't wanted to make them too big)
The last one will consider checkstlye, codestyle ect. So please focus on content changes.

Kitodo/src/main/java/de/sub/goobi/config/ConfigCore.java
```diff
@@ -147,4 +151,22 @@ public static Duration getDurationParameter(String inParameter, TimeUnit timeUni
...
+        return getParameter(METADATA_DIRECTORY);
+    }
+
+    /**
+     * Get Kitodo config directory.
+     *
+     * @return String
+     */
+    public static String getKitodoConfigDirectory() {
+        return ConfigCore.getParameter(CONFIG_DIR);
```
henning-gerhardt: This could be simplified to `return getParameter(CONFIG_DIR);` or did I miss something?

Kitodo/src/main/java/org/goobi/io/SafeFile.java
```diff
@@ -315,4 +315,8 @@ public URI toURI() {
     public URL toURL() throws MalformedURLException {
         return delegate.toURL();
     }
+
+    public File getDelegate() {
```
Beacze: What is purpose of this method?

henning-gerhardt: This is needed as the moved `start` method inside the `FileService` class need a `File` object instead of former `SafeFile` object. 

Maybe this method signature change should even reverted back.

henning-gerhardt:
This pull request needs a rebase as there is a merge conflict with a deleted file (`Kitodo/src/main/java/de/sub/goobi/forms/AdministrationForm.java`).

Kathrin-Huber:
I have rebased the branch.

Kitodo/src/main/java/org/goobi/io/SafeFile.java
Kathrin-Huber: I will not change the method signature back, as it will be followed by a huge trace of using the SafeFile class in the FileService, which I cleaned up properly.
The SafeFile class will be removed in the next PullRequest

henning-gerhardt: Moved `start` method does not use any of the `list` or `listFiles` methods from `java.io.File` which are known returning `null` values on errors instead of throwing exceptions. `SafeFile` class was introduced to wrap this `null` values to empty list return values to get ride of ugly not null checks every where a `list` or `listFiles` methods are used.

PR #760, opened 2017-04-26
Replace MD4 with org.bountycastle library

Beacze:


Kitodo/src/test/java/de/sub/goobi/helper/encryption/MD4Test.java
```diff
@@ -13,19 +13,18 @@
...
 
-import edu.sysu.virgoftp.ftp.encrypt.MD4;
+import org.junit.Test;
 
 public class MD4Test {
     static HashMap<String, byte[]> testData;
 
     static {
-        testData = new HashMap<String, byte[]>();
+        testData = new HashMap<>();
```
stweil: What is the reason for this change?

Kitodo/src/main/java/de/sub/goobi/helper/ldap/LdapUser.java
```diff
@@ -77,6 +76,7 @@ public LdapUser() {
      */
     public void configure(User inUser, String inPassword, String inUidNumber)
             throws NamingException, NoSuchAlgorithmException, IOException, InterruptedException {
+        JDKMessageDigest.MD4 digester = new JDKMessageDigest.MD4();
```
henning-gerhardt: `JDKMessageDigest` could be removed or did I miss something why it should used in this way?

Kitodo/src/test/java/de/sub/goobi/helper/encryption/MD4Test.java
```diff
@@ -38,16 +37,12 @@
     }
 
     @Test
-    public void encryptTest() {
+    public void encryptTest() throws Exception {
```
henning-gerhardt: Why did you not use more specific `UnsupportedEncodingException` instead of the really general `Exception`?

Beacze: I think for tests using more specific exception is not necessary. In this way method is shorter.

Kitodo/src/test/java/de/sub/goobi/helper/encryption/MD4Test.java
Beacze: It is not necessary to write type of key and value in this HashMap. It is already given in line 24.

Kitodo/src/main/java/de/sub/goobi/helper/ldap/LdapUser.java
Beacze: Fixed

Kitodo/src/test/java/de/sub/goobi/helper/encryption/MD4Test.java
henning-gerhardt: It should be OK for tests, but not somewhere else.

stweil: ... as long as static code analyzers don't complain.

Beacze: It doesn't compalin

PR #766, opened 2017-05-04
Fix problem with failing DBConnectionTestIT

Beacze:
Move inserting of the process to MockDatabase.

### DIFF
--- a/Kitodo/src/test/java/test/DBConnectionTestIT.java
+++ b/Kitodo/src/test/java/test/DBConnectionTestIT.java
@@ -11,27 +11,31 @@
 
 package test;
 
-import org.junit.Assert;
-import org.junit.Test;
-import org.kitodo.data.database.beans.Process;
-import org.kitodo.services.ServiceManager;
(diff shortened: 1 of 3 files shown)


Kitodo/src/test/java/test/DBConnectionTestIT.java
```diff
@@ -11,27 +11,31 @@
...
-        serviceManager.getProcessService().save(test);
 
         long counted = serviceManager.getProcessService().count("from Process");
         Assert.assertNotNull("No Process found", counted);
         Assert.assertEquals(4, counted);
 
         String title = serviceManager.getProcessService().find(1).getTitle();
-        Assert.assertEquals("First process", title);
-
+        Assert.assertEquals("DBConnectionTest", title);
```
henning-gerhardt: This test could not work, as you are search for first process but your test data is inserted as 4th.

Beacze: sure, wrong id, fixed

PR #768, opened 2017-05-05
Switch to JDK8

henning-gerhardt:
Use JDK8 (implemented by OpenJDK8 and OracleJDK8) as new base for further development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,12 @@ language: java
 
 cache:
 
+dist: trusty
+
+jdk:
+  - oraclejdk8
+  - openjdk8
+
 sudo: required
 
 env: DB=mysql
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,6 +4,12 @@ language: java
 
 cache:
 
+dist: trusty
```
Kathrin-Huber: What does dist:trusty do?

henning-gerhardt: Trusty is the name of used Ubuntu distribution. Without a named distribution the default from TravisCI is used and this is at the moment Precise. Precise is since May 4th deprecated in TravisCI and does not support Java8. So I switched explicit to next distribution to get current build processes run.

PR #769, opened 2017-05-05
Use TravisCI caching

henning-gerhardt:
With switching to Maven as build tool a lot of data is downloaded every time on every build run. With a cache directory which store downloaded files the build process should go faster and with less lines on build log.

This pull request depends on #768 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@
 language: java
 
 cache:
+  directories:
+    - $HOME/.m2
 
 dist: trusty
 


stweil:
It looks like the OpenJDK build did not reuse the downloads from the OracleJDK build.

henning-gerhardt:
This is expected according to documentation of [Travis Caching ](https://docs.travis-ci.com/user/caching/):
```
There is one cache per branch and language version/ compiler version/ JDK version/ Gemfile location/ etc.
```
As we using two different JDK versions, we have to different caches.

Every build after merging this pull request should benefit from this caches.

Kathrin-Huber:
I think travis should test the "all-new" szenario. It should act like it is installed on a very new machine, new Database etc...
Caching would shadow problems with changig or removed external libs.

Can the cache be emptied (for example) every day?

henning-gerhardt:
Cached files are only the files downloaded by Maven like everything in `$HOME/.m2` directory on a developer system. Everything else is "new".

stweil:
So the cache could be cleared (manually) when the Maven configuration changes. I don't expect this will happen daily.

PR #788, opened 2017-05-16
Additional exclusions of not used / conflicted dependencies

Beacze:
Depends on https://github.com/kitodo/kitodo-production/pull/787

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -115,6 +115,12 @@
             <groupId>commons-beanutils</groupId>
             <artifactId>commons-beanutils</artifactId>
             <version>1.7.0</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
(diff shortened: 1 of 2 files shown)


Kitodo/pom.xml
```diff
@@ -401,20 +449,32 @@
             <artifactId>kitodo-contentserver</artifactId>
             <version>2.0.0</version>
         </dependency>
-        <dependency>
-            <groupId>org.richfaces.framework</groupId>
-            <artifactId>richfaces-api</artifactId>
-            <version>3.1.6.GA</version>
-        </dependency>
```
henning-gerhardt: Removing this `richfaces-api` cause a regression on running the application. Reason: ClassNotFound exception for `org.ajax4jsf.application.ViewHandlerWrapper`. Better solution could be an exclusion of every dependency.

Beacze: done

PR #790, opened 2017-05-17
Fix adding of properties in ProzessverwaltungForm

Beacze:


henning-gerhardt:

PR #792, opened 2017-05-17
Remove not used save and delete property methods from AktuelleSchritt…

Beacze:
…eForm


henning-gerhardt:

PR #793, opened 2017-05-17
Fix adding of properties in AktuelleSchritteForm

Beacze:
Fixed in the same way like ProzessverwaltungForm: https://github.com/kitodo/kitodo-production/pull/790

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
@@ -28,6 +28,7 @@ import de.unigoettingen.goobi.module.api.exception.GoobiException;
 import java.io.File;
 import java.io.IOException;
 import java.text.SimpleDateFormat;
+import java.util.Arrays;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


henning-gerhardt:

Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -1355,6 +1340,19 @@ public void saveCurrentProperty() {
         loadProcessProperties();
     }
 
+    private boolean isValidProcessProperty() {
+        if (this.processProperty.isValid()) {
+            return true;
+        } else {
+            List<String> param = new ArrayList<String>();
+            param.add(processProperty.getName());
+            String value = Helper.getTranslation("propertyNotValid", param);
```
matthias-ronge: What about
`String value = Helper.getTranslation("propertyNotValid", Arrays.asList(new String[]{ processProperty.getName() }));`

Beacze: good point, it looks like better solution :)

PR #801, opened 2017-05-19
Hibernate 5 Migration

Beacze:
- update dependencies
- fix transaction management
- fix HibernateUtil
- add BaseIndexedBean


Kitodo/src/main/java/de/sub/goobi/helper/RefreshObject.java
```diff
@@ -78,8 +78,8 @@ public static void refreshProcessGUI(int processID) {
             session = Helper.getHibernateSession();
             if (session == null || !session.isOpen() || !session.isConnected()) {
                 logger.debug("session is closed, creating a new session");
-                HibernateUtilOld.rebuildSessionFactory();
-                session = HibernateUtilOld.getSessionFactory().openSession();
+                //HibernateUtil.rebuildSessionFactory();
```
henning-gerhardt: If this line is not needed any more, remove it.

Beacze: done

PR #807, opened 2017-05-31
Catch LazyInitializationException as workaround to display bugged pages

solth:
Until the reason for the LazyInitializationException has been found and fixed, this workaround prevents the pages "MeineAufgaben", "Vorgänge" and "Produktionsvorlagen" from running into an error (but still shows empty lists on the later two).


henning-gerhardt:
This is fixing only read access. If you try to add a new user defined filter you still get a `LazyInitializationException`. So if you really work around this then the writing access should be wrapped even.

The root of this problem is that the user object for this logged in user which is [fetched from database](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java#L81) are wrong initialised. Why this is happened must be discovered.

henning-gerhardt:
After some discussions this pull request should be merged as without this interim fix a lot more work could not be done. On goal of next development sprint is to solve the underlaying problem of not correct initialised user object.

PR #808, opened 2017-06-01
Change long to int in count methods

Kathrin-Huber:
A count method should always return int, because the amount of entries in a database is always an integer.


Kathrin-Huber:
I was confused about the value range and closing this PR again.

PR #813, opened 2017-06-08
Fix wrong uri creation

Kathrin-Huber:
Uri creation was wrong in some places.


solth:
These changes fix the problems I encountered that prevented me from logging into the system.

Kathrin-Huber:
I have no idea, I have removed it.

PR #817, opened 2017-06-12
Navigation menu update

solth:
These changes resolve the 'one-url-behind' problem that Kitodo.Production suffers from because h:commandLink and h:commandButton elements are used for normal page to page navigation. These command components are translated into POST requests (postbacks), instead of GET requests and are responsible for said problem.
The new h:link and h:button components introduced in JSF 2, which are recommended for page to page navigation because they produce GET requests, are used in these commits to replace the command components in the navigation menu.

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
@@ -451,7 +451,7 @@ public class ProzessverwaltungForm extends BasisForm {
             FilterAktuelleProzesse();
         }
         else {
-            System.err.println("ERROR: did not recognize link ID '"+originLink+"'.");
+            logger.error("ERROR: did not recognize link ID '" + originLink + "'.");
         }
         setFilter("");
     }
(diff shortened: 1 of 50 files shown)


Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
```diff
@@ -442,6 +435,28 @@ public String FilterVorlagen() {
...
+            NeuenVorgangAnlegen();
+        }
+        else if (Objects.equals(originLink, "templates")) {
+            FilterVorlagen();
+        }
+        else if (Objects.equals(originLink, "allProcesses")) {
+            FilterAktuelleProzesse();
+        }
+        else {
+            System.err.println("ERROR: did not recognize link ID '"+originLink+"'.");
```
henning-gerhardt: Do not use `System.err.println` - use a log statement for it

solth:
> I got this warnings but not sure if they are related to it:
> WARNUNG: HTML nesting warning on closing td: element h3 not explicitly closed
> Jun 15, 2017 9:52:07 AM org.apache.myfaces.shared.renderkit.html.HtmlResponseWriterImpl endElement
> WARNUNG: HTML nesting warning on closing h3: element td not explicitly closed

I think I saw these warnings in another branch as well and that they are unrelated to the current changes. Do you remember which page you were navigating to when these warning occurred?

Beacze:
Yes, it was from LDAP-Gruppen to Taskmanager.

solth:
> Yes, it was from LDAP-Gruppen to Taskmanager.
> 
> Also when I go from from other pages to Taskmanager, so it must be Taskmanager.

Yes, it looks that way. But the changes in this PR are definitely not responsible for the warning. I just verified that they happen in `master` as well. I will try to find and fix the cause in a separate PR.

PR #826, opened 2017-06-19
Replace log4j2.properties files with more advised log4j2.xml

Beacze:


henning-gerhardt:
Travis is failing because of missing directory `/usr/local/kitodo/logs`. I see at least two solutions:

1. create `/usr/local/kitodo/logs` inside section `before_script` in `.travis.yml` file
2. adjust used `log4j2.xml` for running tests to not use a file for logging

PR #829, opened 2017-06-20
Exclude js and css resources from requestControlFilter

chrizzor:
This is important for using primefaces.
Resources from primefaces are supplied with the extension “.jsf” and the requestControlFilter matches for files with this extension.
Without excluding, the files wont be delivered.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/web.xml
+++ b/Kitodo/src/main/webapp/WEB-INF/web.xml
@@ -124,6 +124,10 @@
     <filter>
         <filter-name>requestControlFilter</filter-name>
         <filter-class>de.sub.goobi.helper.servletfilter.RequestControlFilter</filter-class>
+        <init-param>
+            <param-name>excludePattern</param-name>
+            <param-value>.*\.(js|css)\.jsf</param-value>
+        </init-param>
     </filter>
 
     <filter>


Kitodo/src/main/webapp/WEB-INF/web.xml
```diff
@@ -124,6 +124,10 @@
     <filter>
         <filter-name>requestControlFilter</filter-name>
         <filter-class>de.sub.goobi.helper.servletfilter.RequestControlFilter</filter-class>
+        <init-param>
+            <param-name>excludePattern</param-name>
+            <param-value>.*\.(js|css)\.jsf</param-value>
```
henning-gerhardt: This pattern looks a bit strange. There should no file name ending of `.js.jsf` or `.css.jsf` if I understood this pattern correct.

solth:
> This pattern looks a bit strange. There should no file name ending of .js.jsf or .css.jsf if I understood this pattern correct.
> 
> I think you would only exclude .js and .css files. Then should the added \.jsf to be removed.

It looks strange but is the correct pattern to exclude for the `requestControlFilter`. Javascript and CSS resource files that come bundled with component libraries like PrimeFaces result in resource names like  `primefaces.js.jsf` in Facelets that use tags from the corresponding libraries.

Requests to these 'resources' should definitely not be processed by the `requestControlFilter`, because it often prevents their loading after a few calls to the same page, disabling any Javascript functionality and rendering the whole page unusable.

Individual Javascript files that don't come as part of a JSF component or tag library are not affected by this (for example the `toggle.js` file in `head.xhtml` included via a normal `<script>` tag), can therefore safely be processed by the filter and don't lead to the resource error described above.

Kitodo/src/main/webapp/WEB-INF/web.xml
solth: > This pattern looks a bit strange. There should no file name ending of .js.jsf or .css.jsf if I understood this pattern correct.
> 
> I think you would only exclude .js and .css files. Then should the added \.jsf to be removed.

It looks strange but is the correct pattern to exclude for the `requestControlFilter`. Javascript and CSS resource files that come bundled with component libraries like PrimeFaces result in resource names like  `primefaces.js.jsf` in Facelets that use tags from the corresponding libraries.

Requests to these 'resources' should definitely not be processed by the `requestControlFilter`, because it often prevents their loading after a few calls to the same page, disabling any Javascript functionality and rendering the whole page unusable.

Individual Javascript files that don't come as part of a JSF component or tag library are not affected by this (for example the `toggle.js` file in `head.xhtml` included via a normal `<script>` tag), can therefore safely be processed by the filter and don't lead to the resource error described above.

henning-gerhardt: Thank you for clarification this. Without this context information the exclude pattern looks strange. If this change where inside a primefaces pull request and not a stand alone pull request I would not asked for this as currently there are no or only a few files with this kind of file ending.

PR #837, opened 2017-06-27
Code improvements

Beacze:
- simplify if statements
- remove redundant conditional expressions
- simplify conditional expressions
- remove unused imports
- replace for with foreach loops
- replace indexOf with contains
- remove unnecessary boxing/unboxing
- replace explicit type
- remove redundant variables
- replace for loops with addAll
- remove String.toString()
- remove pointless boolean operations
- join together catch blocks
- use equals instead of ==

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
@@ -108,7 +108,7 @@ public class AktuelleSchritteForm extends BasisForm {
     private Boolean flagWait = false;
     private final ReentrantLock flagWaitLock = new ReentrantLock();
     private BatchStepHelper batchHelper;
-    private Map<Integer, PropertyListObject> containers = new TreeMap<Integer, PropertyListObject>();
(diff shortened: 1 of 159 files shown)


Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -170,7 +170,8 @@ public String filterAlleStart() {
     }
 
     /**
-     * This method initializes the task list without any filter whenever the bean is created.
+     * This method initializes the task list without any filter whenever the
```
Kathrin-Huber: are you using the right code formatter? 
formatting stuff shouldn't happen if the line is not touched.

Beacze: I use our code formatted, and it formats whole file on which it is applied, not only touched lines.

Beacze: Sometimes it can happen that previous new part was not formatted ;)

PR #842, opened 2017-07-03
Replace a4j status tag and enclose JS in CDATA sections

solth:
- replace `a4j:status` tag with `p:ajaxStatus` tag from PrimeFaces in `Metadaten2rechts.xhtml`
- enclose inline JS in Facelets in CDATA sections to prevent errors caused by characters that would be illegal in XML (like `&` or `<` for example)

### DIFF
--- a/Kitodo/src/main/webapp/pages/Metadaten3links.xhtml
+++ b/Kitodo/src/main/webapp/pages/Metadaten3links.xhtml
@@ -188,7 +188,7 @@
                     return true;
                     //	}
                 } catch (err) {
-                    parent.rechts.location.href = "Metadaten2rechts.jsf";
+                    parent.rechts.location.href = "Metadaten2rechts.xhtml";
                 }
             }
 
--- a/Kitodo/src/main/webapp/pages/Metadaten3links.xhtml
+++ b/Kitodo/src/main/webapp/pages/Metadaten3links.xhtml
@@ -169,64 +169,65 @@
 
         </h:body>
     </f:view>
-    <script type="text/javascript"><!--
-    // Funktion, die dynamisch das Stylesheet zuweist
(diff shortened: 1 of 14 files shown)


Kitodo/src/main/webapp/pages/Metadaten3links.xhtml
```diff
@@ -169,64 +169,65 @@
...
+                    // Daten sind gespeichert, also den Klick weitergeben
+                    var galleryLinks;
+                    galleryLinks = document.getElementsByTagName('a');
+                    for (var i = 0; i < galleryLinks.length; i++)
+                        galleryLinks[i].className = "document";
+                    element.className = "documentSelected";
+                    return true;
+                    //	}
+                } catch (err) {
+                    parent.rechts.location.href = "Metadaten2rechts.jsf";
```
henning-gerhardt: Is `Metadaten2rechts.jsf` correct or should it be `Metadaten2rechts.xhtml` ?

solth: Before JSF 2 it was required to call the templates with the `*.jsf` extension, even though the actual physical files had extensions like `*.jsp` or `*.xhtml`. The `FacesServlet`, configured in the `web.xml` deployment descriptor, would then map requests to `*.jsf` files to the corresponding physical templates with their real extensions. Calling a respective template with it's real extension directly would result in the source code of the template (for example in xhtml) be displayed in the browser. 

Since JSF 2, `*.xhtml` is also mapped to the `FacesServlet`, so both, calling pages with `*.jsf` or `*.xhtml` extension should work.

To keep the coding style in line with the other files, I will change the extension to `*.xhtml` here as well. Thanks for pointing it out.

PR #850, opened 2017-07-04
Fix metadata editor call

Beacze:


Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java
```diff
@@ -928,10 +928,8 @@ public Fileformat readMetadataFile(Process process) throws ReadException, IOExce
         }
 
         Fileformat ff = determineFileFormat(type, process);
-
         try {
-            // TODO: this is not working anymore because uri is not full path
-            ff.read(metadataFileUri.toString());
+            ff.read(new File(metadataFileUri).toString());
```
henning-gerhardt: Could this `new File(...)` call inside a method call create a resource leak?

Beacze: I don't think so... It works...

henning-gerhardt: It is not about the general working or not more a thing for watching of leaking resources. If this call is done per day 1.000 ore more times then we got maybe some got real problems about non-collectable resources.

Kathrin-Huber: I think, thats the responsibility of the Java Garbage collector. We don't need to care about unused ressources.

Kitodo/src/main/java/org/kitodo/services/file/FileService.java
```diff
@@ -564,44 +564,40 @@ public URI getProcessBaseUriForExistingProcess(Process process) {
...
      * @param processSubType
      *            The subType.
-     * @param id
+     * @param folderName
      *            the id of the single object (e.g. image) if null, the root
      *            folder of the sublocation is returned
      * @return The URI of the requested location
      */
-    public URI getProcessSubTypeURI(Process process, ProcessSubType processSubType, String id) {
+    public URI getProcessSubTypeURI(Process process, ProcessSubType processSubType, String folderName) {
```
Kathrin-Huber: it's not always a folder name. can be an imagename for example (as stated in the javadoc)

Beacze: so better name would resourceName?

Kathrin-Huber: yes, that sounds better

Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java
henning-gerhardt: Hopefully did this the garbage collector. In the past we had a lot of resource leaks and one solution was to use local variables which are collected by the garbage collector.

PR #851, opened 2017-07-04
Add method isCurrentImageCorrectImage to Metadaten

Beacze:


henning-gerhardt:
Naming of new method could be corrected as this method proves if a wanted image the image in the list. This must be not necessary every time the first image.

PR #852, opened 2017-07-04
Remove multiple FilenameFilter

Beacze:
Separate  redundant override accept method for FilenameFilter interface. Move them to separated classes, which implement this interface. Use those classes.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/filters/FileBeginEndFilter.java
+++ b/Kitodo/src/main/java/org/kitodo/filters/FileBeginEndFilter.java
@@ -1,0 +1,43 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.filters;
+
+import java.io.File;
+import java.io.FilenameFilter;
+
+/**
+ * Implementation of FileNameFilter for checking the begin and the end of files.
+ */
+public class FileBeginEndFilter implements FilenameFilter {
+
+    private String begin;
+    private String end;
+
+    /**
+     * Filter files by given begin and end of file.
+     *
+     * @param end
+     *            String
+     */
+    public FileBeginEndFilter(String begin, String end) {
+        if (begin == null || end == null) {
+            throw new IllegalArgumentException("No filter for file begin or end given.");
+        }
+        this.begin = begin;
+        this.end = end;
+    }
+
+    @Override
+    public boolean accept(File dir, String name) {
+        return (name.endsWith(end) && name.startsWith(begin));
+    }
+}
--- a/Kitodo/src/main/java/org/kitodo/filters/FileBeginEndFilter.java
+++ b/Kitodo/src/main/java/org/kitodo/filters/FileBeginEndFilter.java
@@ -25,12 +25,17 @@ public class FileBeginEndFilter implements FilenameFilter {
     /**
      * Filter files by given begin and end of file.
      *
+     * @param begin
+     *            file name should begin with this String
      * @param end
-     *            String
+     *            file name should end with this String
+     * @throws IllegalArgumentException
+     *             it is thrown in case one or both parameters are null or empty
+     *             Strings
      */
     public FileBeginEndFilter(String begin, String end) {
-        if (begin == null || end == null) {
-            throw new IllegalArgumentException("No filter for file begin or end given.");
+        if (begin == null || begin.equals("") || end == null || end.equals("")) {
+            throw new IllegalArgumentException("No filter or empty filter for file begin or end is given.");
         }
         this.begin = begin;
         this.end = end;
--- a/Kitodo/src/main/java/org/kitodo/filters/FileDirectoryFilter.java
+++ b/Kitodo/src/main/java/org/kitodo/filters/FileDirectoryFilter.java
@@ -1,0 +1,26 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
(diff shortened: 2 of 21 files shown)


Kitodo/src/main/java/org/kitodo/filters/FileBeginEndFilter.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Filter files by given begin and end of file.
+     *
+     * @param end
+     *            String
+     */
+    public FileBeginEndFilter(String begin, String end) {
+        if (begin == null || end == null) {
+            throw new IllegalArgumentException("No filter for file begin or end given.");
```
henning-gerhardt: - Should this compare not include empty string argument and not only null?
- I'm a little bit unsure about this not declared IllegalArgumentException. Maybe this exception should descriped in the JavaDoc part.

Kitodo/src/main/java/org/kitodo/filters/FileDirectoryFilter.java
```diff
@@ -0,0 +1,26 @@
...
+import java.io.FilenameFilter;
+
+/**
+ * Implementation of FileNameFilter for checking if file is a directory.
+ */
+public class FileDirectoryFilter implements FilenameFilter {
+
+    @Override
+    public boolean accept(File dir, String name) {
+        return (new File(dir, name).isDirectory());
```
henning-gerhardt: One more unsure thing: Would this new File declaration not left a unused resource left after this method is closed?

Kitodo/src/main/java/org/kitodo/filters/FileEndFilter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Filter files by given file end.
+     *
+     * @param end
+     *            String
+     */
+    public FileEndFilter(String end) {
+        if (end == null) {
+            throw new IllegalArgumentException("No filter for file end given.");
```
henning-gerhardt: Same questions like above.

Kitodo/src/main/java/org/kitodo/filters/FileEndNotBeginFilter.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Filter files by given begin and end of file.
+     *
+     * @param end
+     *            String
+     */
+    public FileEndNotBeginFilter(String notBegin, String end) {
+        if (notBegin == null || end == null) {
+            throw new IllegalArgumentException("No filter for file begin or end given.");
```
henning-gerhardt: Same like above.

Kitodo/src/main/java/org/kitodo/filters/FileDirectoryFilter.java
Beacze: I don't think so.

Kitodo/src/main/java/org/kitodo/filters/FileBeginEndFilter.java
Beacze: - Yes, comparing to empty String is a good idea.
- I can add it do JavaDoc.

Kitodo/src/main/java/org/kitodo/services/file/FileService.java
```diff
@@ -637,14 +638,9 @@ public boolean deleteProcessContent(Process process) {
      */
     public URI getSourceDirectory(Process process) {
         URI dir = getProcessSubTypeURI(process, ProcessSubType.IMAGE, null);
-        FilenameFilter filterVerz = new FilenameFilter() {
-            @Override
-            public boolean accept(File dir, String name) {
-                return (name.endsWith("_" + "source"));
-            }
-        };
+        FilenameFilter filterDirectory = new FileEndFilter("_source");
```
Kathrin-Huber: the variable name does not match the FileEndFilter. It is a filter FOR the directory. if you want to reflect that, then name it filterForDirectory

Kathrin-Huber:
I don't like the classnames
FileEndNotBeginFilter.java - I get it from the context, but the classname is pretty confusing.
same for
FileListFilter - is it filtered by a list?

FileDirectoryFilter - get all files in a directory?

You see, it's not uniquely. Maybe some verbes would help, like FileIsDirectoryFilter.

Beacze:
FileListFilter - I suppose (it is not my class) that it filters list of the files with given name

Which names you would suggest for FileEndNotBeginFilter?

henning-gerhardt:
There is one merge conflict. Please resolve this conflict.

henning-gerhardt:
Some name suggestions
- `FileDirectoryFilter` should be renamed to `DirectoryFilter` as this filter only checks if a given `File` object is a directory (alternative name could be  `IsDirectoryFilter`)
- `FileListFilter` is used for filtering files inside a directory with a general match as is could be called `FilennameMatchFilter`
- `FileEndFilter` could be renamed to `FilenameEndFilter`
- `FileNotEndFilter` could be renamed to `FilenameNotEndFilter`
- `FileBeginEndFilter` could be renamed to `FilenameBeginAndEndFilter`

What did you think @Beacze and @Kathrin-Huber ?

Kathrin-Huber:
Henning wrote: 

> Some name suggestions
> * FileDirectoryFilter should be renamed to DirectoryFilter as this filter only checks if a given File object is a directory (alternative name could be IsDirectoryFilter)
> * FileListFilter is used for filtering files inside a directory with a general match as is could be called FilennameMatchFilter
> * FileEndFilter could be renamed to FilenameEndFilter
> * FileNotEndFilter could be renamed to FilenameNotEndFilter
> * FileBeginEndFilter could be renamed to FilenameBeginAndEndFilter

I would really urge to use verbs
So my suggestions:

* IsDirectoryFilter
* MatchesFileNameFilter
* FileNameEndsWithFilter
* FileNameDoesNotEndWithFilter
* FileNameBeginsWithAndEndsWithFilter


Those are pretty stupid names and they are long... why do we need this classes? Can't we just make methods in the FileService?

henning-gerhardt:
@Kathrin-Huber : I tried to avoid long class names but I would prefer your suggestions.

We must use filter classes as long we are using the `listFiles()` method from `Files` class. Maybe from newer `java.nio.files` package we can use other possibilities.

henning-gerhardt:
In my opinion this PR is good now. 

@Kathrin-Huber: what did you think?

PR #860, opened 2017-07-06
Manage user usergroup dependencies

Beacze:


Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/index/type/BaseType.java
```diff
@@ -50,7 +50,7 @@
     @SuppressWarnings("unchecked")
     <F extends BaseBean> JSONArray addObjectRelation(List<F> objects) {
         JSONArray jsonArray = new JSONArray();
-        if (objects != null) {
+        if (objects != null && objects.size() > 0) {
```
henning-gerhardt: Why did you add this empty check? I think it is not necessary as an empty list will do nothing in later for statement.

henning-gerhardt:
PR has some merge conflicts. Please resolve them.

Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/index/type/BaseType.java
Beacze: It looks I had some problem and I have added this additional check, now it is not necessary anymore.

PR #862, opened 2017-07-06
Fix image list for metadata editor

Beacze:
Should be rebased after #863 is merged

PR #863, opened 2017-07-07
Convert tests which use hibernate to integration tests

Beacze:
Should be merged before #862


henning-gerhardt:
If you are remove the proposed change in #862 then it should not needed that both PR depending on each other.

PR #864, opened 2017-07-07
Separate manage dependencies methods for TemplateService and WorkpieceService

Beacze:


Kitodo/src/main/java/org/kitodo/services/data/TemplateService.java
```diff
@@ -73,19 +75,41 @@ public void saveToIndex(Template template) throws CustomResponseException, IOExc
...
-        if (template.getProcess() != null) {
-            serviceManager.getProcessService().saveToIndex(template.getProcess());
+        manageProcessDependenciesForIndex(template);
+        managePropertiesDependenciesForIndex(template);
+    }
+
+    private void manageProcessDependenciesForIndex(Template template) throws CustomResponseException, IOException {
+        Process process = template.getProcess();
+        if (template.getIndexAction() == IndexAction.DELETE) {
+            if (process != null) {
```
henning-gerhardt: A template should never exists without a corresponding process. It could be possible that a process has no template.

Kitodo/src/main/java/org/kitodo/services/data/WorkpieceService.java
```diff
@@ -73,19 +75,40 @@ public void saveToIndex(Workpiece workpiece) throws CustomResponseException, IOE
...
-        if (workpiece.getProcess() != null) {
-            serviceManager.getProcessService().saveToIndex(workpiece.getProcess());
+        manageProcessDependenciesForIndex(workpiece);
+        managePropertiesDependenciesForIndex(workpiece);
+    }
+
+    private void manageProcessDependenciesForIndex(Workpiece workpiece) throws CustomResponseException, IOException {
+        Process process = workpiece.getProcess();
+        if (workpiece.getIndexAction() == IndexAction.DELETE) {
+            if (process != null) {
```
henning-gerhardt: A workpiece should never exists without a corresponding process. It could be possible that a process has no workpiece.

Beacze: Null check is removed, excluding possibility of save without process must be handle in the future.

PR #865, opened 2017-07-07
Fix search form

Kathrin-Huber:
This fixes the Problm in the search form and prepares for the usage with index.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/PropertyDAO.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/PropertyDAO.java
@@ -13,8 +13,13 @@ package org.kitodo.data.database.persistence;
 
 import java.util.List;
 
+import javax.persistence.Table;
+
+import org.hibernate.Query;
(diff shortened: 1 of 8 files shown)


Beacze:
I don't know if I would go so deep with changes if it is only temporal approach before index is used for querying this values.

henning-gerhardt:
I could understood but I tried it with migrated data from our productive system (3.4 mio entries in property table) and after 10 minutes of waiting for displaying this form I stopped the approach. 

With this change this site / form will be usable for every one who has some data inside his database.

Without this change the form is displaying wrong data after a few seconds (3 or 4 on first run).

Kathrin-Huber:
@henning-gerhardt please test, if the search faster now

henning-gerhardt:
It is faster now except for step titles. Querying this values takes a long time.

General note: titles was sorted before and are now unsorted.

Kathrin-Huber:
@henning-gerhardt you can test it again.
I've added the ordering, and it actually IS ordering, but the translation comes after that, so it seems to be unorderd. We will fix that, when we use the index.

PR #868, opened 2017-07-11
Use correct URI creation

Beacze:
Correct way to create URIs is use Paths.get(...).toUri()

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java
@@ -252,17 +252,16 @@ public class LoginForm implements Serializable {
 
     private void AlteBilderAufraeumen() throws IOException {
         /* Pages-Verzeichnis mit den temporären Images ermitteln */
-        URI myPfad = serviceManager.getFileService()
-                .getInternUri(new File(ConfigCore.getTempImagesPathAsCompleteDirectory()).toURI());
+        URI path = ConfigCore.getTempImagesPathAsCompleteDirectory();
 
         /* Verzeichnis einlesen */
         FilenameFilter filter = new FileNameEndsWithFilter(".png");
-        ArrayList<URI> dateien = serviceManager.getFileService().getSubUris(filter, myPfad);
+        ArrayList<URI> files = serviceManager.getFileService().getSubUris(filter, path);
 
         /* alle Dateien durchlaufen und die alten löschen */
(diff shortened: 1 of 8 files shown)


Kitodo/src/main/java/de/sub/goobi/helper/WebDav.java
```diff
@@ -121,118 +125,100 @@ public void uploadFromHome(Process myProcess) {
...
-        nach = nach.resolve(myProcess.getTitle() + " [" + myProcess.getId() + "]");
-
-        /* Leerzeichen maskieren */
-        nach = URI.create(nach.toString().replaceAll(" ", "__"));
-        URI benutzerHome = nach;
-
-        fileService.deleteSymLink((benutzerHome));
+        after = Paths.get(new File(after).getPath(), getEncodedProcessLinkName(process)).toUri();
+        URI userHome = after;
+        fileService.deleteSymLink((userHome));
```
henning-gerhardt: Could not this 3 lines
```
after = Paths.get(new File(after).getPath(), getEncodedProcessLinkName(process)).toUri();
URI userHome = after;
fileService.deleteSymLink((userHome));
```
simplified to
```
after = Paths.get(new File(after).getPath(), getEncodedProcessLinkName(process)).toUri();
fileService.deleteSymLink(after);
```

Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java
```diff
@@ -252,17 +252,16 @@ public String BenutzerkonfigurationSpeichern() {
...
         FilenameFilter filter = new FileNameEndsWithFilter(".png");
-        ArrayList<URI> dateien = serviceManager.getFileService().getSubUris(filter, myPfad);
+        ArrayList<URI> files = serviceManager.getFileService().getSubUris(filter, path);
 
         /* alle Dateien durchlaufen und die alten löschen */
-        if (dateien != null) {
-            for (URI aDateien : dateien) {
-                URI file = myPfad.resolve(aDateien);
+        if (files != null) {
+            for (URI aFiles : files) {
```
henning-gerhardt: If you found a better variable name as `aFiles` (previous `aDateien` was even not better) then rename this variable.

henning-gerhardt:
I found a issue: symbolic links from image directory to user directory could not be deleted as path is a URI instead of a file path. WebDav class in both methods `uploadFromHome` and `removeAllFromHome`.

PR #870, opened 2017-07-13
Use more general Kitodo directory mapping

Beacze:
1. Separate map to Kitodo URI on three different types of mapping:
- config
- data
- root (apache)
2. Handle cases when return value was ignored
3. Add MappingType enum
4. Order methods in FileService

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/services/file/FileService.java
@@ -80,7 +80,8 @@ public class FileService {
     }
 
     /**
-     * Creates a directory at a given URI with a given name.
+     * Creates a directory at a given URI with a given name without
+     * mapping/unampping - actually with default data mapping.
      *
      * @param parentFolderUri
      *            the uri, where the directory should be created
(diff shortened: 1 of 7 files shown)


Kitodo/src/main/java/org/kitodo/services/file/FileService.java
```diff
@@ -215,24 +248,35 @@ public URI renameFile(URI fileUri, String newFileName) throws IOException {
...
-                    count += getNumberOfFiles(aChildren);
-                } else {
-                    count += 1;
+            if (isDirectory(directory)) {
+                for (URI aChildren : children) {
+                    if (isDirectory(aChildren)) {
+                        count += getNumberOfFiles(aChildren);
+                    } else {
+                        count += 1;
+                    }
```
henning-gerhardt: Both iterations through the list of URIs looks similar: why did you not move this counting to its own method? This would even reduce the level of nested `if` statements.

Kitodo/src/main/java/org/kitodo/services/file/FileService.java
```diff
@@ -191,7 +223,8 @@ public URI renameFile(URI fileUri, String newFileName) throws IOException {
                 }
                 millisWaited += SLEEP_INTERVAL_MILLIS;
             }
-        } while (!success && millisWaited < MAX_WAIT_MILLIS);
+        }
```
henning-gerhardt: Is this format change indented or valid with the used format rules?

Kitodo/src/main/java/org/kitodo/services/file/FileService.java
```diff
@@ -241,15 +285,17 @@ public Integer getNumberOfFiles(URI directory) {
      */
     public Integer getNumberOfImageFiles(URI directory) {
         int count = 0;
-        if (isDirectory(directory)) {
-            /*
-             * die Images zählen
-             */
+        if (directory.isAbsolute()) {
+            if (isDirectory(directory)) {
+                count = getSubUris(Helper.imageNameFilter, directory).size();
```
henning-gerhardt: Why did you add the number of subdirectories as a start value for the number of image files? Looks a bit strange.

Like in method `getNumberOfFiles` both code lines of the `if`statement looks similar and could be separated.

Kitodo/src/main/java/org/kitodo/services/file/FileService.java
matthias-ronge: Looks like lost code. A `do { … }` without `while ( … )` shouldn’t even compile.

henning-gerhardt: The corresponding `do` is in line 175 / 206.

Beacze: actually it is interesting case, auto formatter says:
` } while`
and checkstyle says
```
} 
while
```

Kathrin-Huber: I like the 
} while
better. But i don't care, we should just decide for one!

Beacze: so I will adjust checkstyle to accept this :P

Kitodo/src/main/java/org/kitodo/services/file/FileService.java
Beacze: I didn't add it, I was not changing logic of the method:
https://github.com/Beacze/kitodo-production/commit/e0b8a5e341e691cb7bb2a39a9f3088ea480bd871

henning-gerhardt:
@Beacze : your changes looks good.
@Kathrin-Huber: do you want to review this pull request?

Kathrin-Huber:
I already reviewed it! :)

PR #878, opened 2017-07-19
Checkstyle fixes

Kathrin-Huber:
Fixes some checkstyle errors.
Changed the config from

```
}
while
```

to

`} while`

### DIFF
--- a/Kitodo/src/main/java/org/goobi/production/chart/ProjectStatusDraw.java
+++ b/Kitodo/src/main/java/org/goobi/production/chart/ProjectStatusDraw.java
@@ -14,11 +14,7 @@ package org.goobi.production.chart;
 import de.intranda.commons.chart.results.DataRow;
 import de.intranda.commons.chart.results.DataTable;
 
-import java.awt.BasicStroke;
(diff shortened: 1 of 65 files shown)


Kitodo/src/main/java/org/goobi/production/chart/ProjectStatusDraw.java
```diff
@@ -14,11 +14,7 @@
 import de.intranda.commons.chart.results.DataRow;
 import de.intranda.commons.chart.results.DataTable;
 
-import java.awt.BasicStroke;
-import java.awt.Color;
-import java.awt.FontMetrics;
-import java.awt.GradientPaint;
-import java.awt.Graphics2D;
+import java.awt.*;
 import java.awt.geom.Line2D;
```
Beacze: it is actually against checkstyle

henning-gerhardt: What is the correct behaviour for existing checkstyle? How many similar import statements are needed before collapsing it with an star / asterix?

Issue #880, opened 2017-07-19
Additional fields for newspapers could not be used in firstchild docstruct


henning-gerhardt:
Additional fields on the new process dialog / form won't be filled up with data from used catalog on media / document newspaper if this additional field should be used in the first child of the document.

In method `fillFieldsFromMetadataFile` of class `ProzesskopieForm` in line [568](https://github.com/kitodo/kitodo-production/blob/kitodo-production-2.1.0/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java#L568) is a null pointer exception is thrown, while accessing the first child element as the list of child elements is empty.

In Kitodo.Production 2.0.0 was the behaviour that this null pointer exception was [catched and ignored](https://github.com/kitodo/kitodo-production/blob/kitodo-production-2.0.0/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java#L556-L559).

henning-gerhardt:
Closing as pull request was merged.

PR #881, opened 2017-07-19
Fix null pointer exception on accessing not existing list entry

henning-gerhardt:
Instead of ignoring a possible null pointer exception like in previous versions check for amount of child elements and log a debug message in case that there are no child elements.

### DIFF
--- a/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java
+++ b/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java
@@ -569,8 +569,12 @@ public class ProzesskopieForm {
                         if (allChildren != null) {
                             myTempStruct = allChildren.get(0);
                         } else {
-                            logger.debug("There is no child element for document type " + this.docType
-                                                 + " and meta data field " + field.getTitel());
+                            List<String> translationFields = new ArrayList<>();
+                            translationFields.add(field.getTitel());
+                            translationFields.add(this.docType);
+                            String translation = Helper.getTranslation("ErrorAdditionalFieldsNoChildForDocType", translationFields);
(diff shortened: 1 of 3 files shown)


henning-gerhardt:
This should fixes #880.

henning-gerhardt:
In the past no error is logged (silent ignored) and now it should displayed to the frontend user? I'm unsure about this change. Frontend user can nothing do and even a system administrator could nothing do as this is a program error.

So far as I can see into the code execution this error is happen because the structure of media type newspaper is generated later and not like others (f.e. monograph, periodical) in front of this. If you are switching from periodical to newspaper first [createNewFileFormat()](https://github.com/kitodo/kitodo-production/blob/kitodo-production-2.1.0/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java#L1117-L1182) is called and later [fillFieldsFromMetadataFile()](https://github.com/kitodo/kitodo-production/blob/kitodo-production-2.1.0/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java#L560-L612) and if you create the process method [NeuenProzessAnlegen()](https://github.com/kitodo/kitodo-production/blob/kitodo-production-2.1.0/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java#L775-L976). Inside this method in lines [838 until 856](https://github.com/kitodo/kitodo-production/blob/kitodo-production-2.1.0/Goobi/src/de/sub/goobi/forms/ProzesskopieForm.java#L838-L856) the real format structure is created if this structure is not created before (like for media type newspaper).

henning-gerhardt:
@matthias-ronge I added a translatable message for the frontend user. Its ok now?

PR #883, opened 2017-07-20
Fix used URI to get norm data

### LINKED ISSUE #882: Can not load norm data from DNB
With old 2.0.0 and even new 2.1.0 version it is not possible any more to load normed data from DNB.

gnd_error_001

If you try to call the URI f.e. http://d-nb.info/gnd/132775026/about/rdf in your web browser then you are redirected to http://d-nb.info/gnd/132775026/about/lds.rdf

henning-gerhardt:
This is one way to fix the "moved permanent" error message. Other way would be to find a possibility to follow redirects.

This could fix #882 .

### DIFF
--- a/Goobi/pages/Metadaten2rechts.jsp
+++ b/Goobi/pages/Metadaten2rechts.jsp
@@ -41,7 +41,7 @@
          --%>
             function toAjaxUrl(url){
                 return url.replace(new RegExp("^.*?://[^/]+/(.*)$", ""), function($0, $1){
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
My solution is a fast solution. 

It would be a better solution if this used URI is configurable. Big parts (`authority.default` and `authority.gnd.dataUrlTail`) of this URI are already inside the `goobi_config.properties` file only the dynamic part (GND identifier) is inserted by a user.

If somebody will change it in this way, I would reject my solution.

PR #898, opened 2017-08-01
Kitodo fix xhtml format

tim-boerner:
Added .editorconfig file in root folder for HTML-Codestyle in IntelliJ.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# top-most EditorConfig file
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


stweil:
Which editors should be supported by the project? There is a large number of editors, and many of them support their own configuration file, so we could end in having lots of editor configuration files.

henning-gerhardt:
I did not like the idea of adding IDE specific files. Is there no other way?

PR #899, opened 2017-08-02
Fixed Umlaute

tim-boerner:
Added
<?xml version="1.0" encoding="ISO-8859-1"?>
to xhtml file

### DIFF
--- a/Kitodo/src/main/webapp/newpages/ProjekteBearbeiten.xhtml
+++ b/Kitodo/src/main/webapp/newpages/ProjekteBearbeiten.xhtml
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
 <!DOCTYPE html>
 <!--
  *
@@ -79,7 +80,7 @@
                                                     infoClass="text_blue" showDetail="true" showSummary="true"
                                                     tooltip="true"/>
 
-                                        <!-- Box fÃ¼r die Bearbeitung der Details -->
+                                        <!-- Box für die Bearbeitung der Details -->
                                         <table cellpadding="3" cellspacing="0" width="100%"
                                                    class="eingabeBoxen">
 
@@ -95,7 +96,7 @@
                                                 </td>
                                             </tr>
 
-                                            <!-- Formular fÃ¼r die Bearbeitung der Texte -->
+                                            <!-- Formular für die Bearbeitung der Texte -->
                                             <tr>
                                                 <td class="eingabeBoxen_row2" colspan="2">
 
@@ -148,7 +149,7 @@
 
                                             </tr>
                                         </table>
-                                        <!-- // Box fÃ¼r die Bearbeitung der Details -->
+                                        <!-- // Box für die Bearbeitung der Details -->
 
                                     </td>
                                 </tr>


Kitodo/src/main/webapp/newpages/ProjekteBearbeiten.xhtml
```diff
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
stweil: I don't see that line in latest git master.

Kathrin-Huber: It didn't work with UTF-8.
The problem is fixed with this pull request. I'm open for other suggestions, otherwise I'm going to merge this version.

stweil: What about my other remarks?

Kathrin-Huber: Well of course it is based on the latest git master. Probably it will need a rebase now, but it had the latest changes. I have no idea, why this line is not in the master on other places, probably it got lost, when changing from jsp to xhtml.

stweil: Sorry, the remark about git master was a confusion on my side.

So you say that adding `<?xml version="1.0" encoding="UTF-8"?>` and keeping the UTF-8 encoded umlauts does not work? Do you have a simple test scenario so I can see what does not work, maybe a test case?

I'd try hard to avoid adding files with non UTF-8 encoding to the code base, because that makes maintenance more difficult (think of editor settings or spell checkers for example).

Kathrin-Huber: Our test scenario is the following:
Go to a project with an Umlaut in a field (for example the title), or change the title to one with an umlaut. then go to another tab, than go back to the tab with the umlaut -> the umlaut is not correctly formatted. it even changes again, if you switch again to another tab and back.

stweil: And does that test scenario work with non ISO-8859-1 characters (try Russian, Greek and Chinese) after that PR was applied?

Kitodo/src/main/webapp/newpages/ProjekteBearbeiten.xhtml
```diff
@@ -79,7 +80,7 @@
                                                     infoClass="text_blue" showDetail="true" showSummary="true"
                                                     tooltip="true"/>
 
-                                        <!-- Box für die Bearbeitung der Details -->
+                                        <!-- Box f�r die Bearbeitung der Details -->
```
stweil: The simplest solution for this and all other comments is using English without umlauts (as required by the coding conventions).

Beacze: But we don't talk here about comments inside the code files. We say about editing of project data inside the application.

stweil: The file `ProjekteBearbeiten.xhtml` is obviously used inside the application, isn't it? And although its comments should be ignored when that file is used, I know that this is not always true. There are parsers which get confused by unexpected characters in comments.

Kathrin-Huber: ah! I was wondering, why this comment is marked as a change, but of course, it's the encoding.

stweil: Yes. GitHub is too smart when showing that diff - it shows the same umlaut, no matter whether it is encoded in ISO-8859-1 or UTF-8.

matthias-ronge:
We should use the same kind of encoding at all layers of the application to avoid such problems. From [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html): “Never make your code less readable simply out of fear that some programs might not handle non-ASCII characters properly. If that should happen, those programs are **broken** and they must be **fixed**.”

Here in specially, the XML prolog has nothing to do with the HTTP request/response encoding. Solution may involve fixing the configuration and implementing a `Filter`. See [this posting](https://stackoverflow.com/a/33977270/1503237) for details.

Kathrin-Huber:
@tim-boerner 
Please investigate further and try the suggested solutions.

PR #914, opened 2017-08-11
Add command modul

tim-boerner:


Kitodo-Command/pom.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tim-boerner: done

Kitodo-Command/src/main/java/org/kitodo/command/Command.java
```diff
@@ -0,0 +1,12 @@
+package org.kitodo.command;
```
tim-boerner: done

Kitodo-Command/src/main/java/org/kitodo/command/Command.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.kitodo.api.command.CommandInterface;
+import org.kitodo.api.command.CommandResult;
+
+public class Command implements CommandInterface {
+
+    @Override
+    public CommandResult runCommand(Integer id, String command) {
+        return null;
```
henning-gerhardt: Returning `null` should be avoided even the implementation is not done yet. In my opinion I see two solutions:
- return a `CommandResult` object which indicate that the implementation is not done
- throw a RuntimeException with a message that the implementation is missing.

Kathrin-Huber: throwing a runtime exception would require an adjustment of the api, so i would not recommend that ;)

henning-gerhardt: Even for the implementation phase?

Kathrin-Huber: I just realised, that RuntimeException is unchecked, so the signature can stay as it is and everything is fine. :)

PR #917, opened 2017-08-14
Use sort in form methods

Beacze:


Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -177,23 +179,28 @@ public void initializeTaskList() {
...
         if (this.sortierung.equals("prozessAsc")) {
-            order = Order.asc("proc.title");
+            sort += ", " + SortBuilders.fieldSort("process").order(SortOrder.ASC).toString();
         }
         if (this.sortierung.equals("prozessDesc")) {
-            order = Order.desc("proc.title");
+            sort += ", " + SortBuilders.fieldSort("process").order(SortOrder.DESC).toString();
         }
-        if (this.sortierung.equals("batchAsc")) {
+        /*if (this.sortierung.equals("batchAsc")) {
```
henning-gerhardt: If your out-commented block is not used anymore or obsolete, then remove this. If this should be fixed later, then add a comment / TODO annotation.

Beacze: done

PR #930, opened 2017-08-17
Java Docs cleanup

Beacze:


Kitodo/src/main/java/org/goobi/production/export/ExportXmlLog.java
```diff
@@ -95,7 +95,6 @@ public void startExport(Process p, File dest) throws IOException {
      *            the process to export
      * @param os
      *            the OutputStream to write the contents to
-     * @throws IOException
```
henning-gerhardt: Why did you remove this entry as this exception is still thrown.

Beacze: restored

PR #935, opened 2017-08-21
Convert user visible attribute to boolean

Beacze:


Kitodo-DataManagement/src/main/resources/db/migration/V2_10__Change_visible_from_varchar_to_tinyint.sql
```diff
@@ -0,0 +1,23 @@
...
+
+-- 1. Update column visible to store correct tinyint values
+
+UPDATE user SET visible = '1' WHERE trim(visible) != 'deleted';
+UPDATE user SET visible = '0' WHERE trim(visible) = 'deleted';
+
+-- 2. Change column visible to store correct tinyint values
+
+ALTER TABLE user
+  MODIFY visible TINYINT(1) NOT NULL;
```
Kathrin-Huber: The script is not working for me.
I needed to add:
UPDATE user SET visible = '1' WHERE visible IS NULL;
Because null values didn't change.

henning-gerhardt: The first `UPDATE` query must be replaced with the suggested `UPDATE` from Kathrin.

Beacze: So previous version with '' and coalesce was better as it worked for me.

PR #939, opened 2017-08-22
Convert user visible attribute to deleted boolean

Beacze:


Kitodo-DataManagement/src/main/resources/db/migration/V2_10__Change_visible_from_varchar_to_tinyint.sql
```diff
@@ -0,0 +1,23 @@
...
+-- For the full copyright and license information, please read the
+-- GPL3-License.txt file that was distributed with this source code.
+--
+
+--
+-- Migration: Change column visible in user table from varchar to tinyint
+
+-- 1. Update column visible to store correct tinyint values
+
+UPDATE user SET visible = '0' WHERE visible IS NULL;
```
Kathrin-Huber: what if its not null and not deleted?

henning-gerhardt: So far as I know no other values like `null` and `deleted` are possible except somebody changed the values direct in the table.

Beacze: I wanted use this coalesce(trim(visible), '')

Kathrin-Huber: We can not exclude, that there MAY BE a change in the database. so we need to check for null, deleted and not deleted.

henning-gerhardt:
There are merge conflicts in four files. Please rebase this branch.

PR #924, opened 2017-08-16
Template folder

solth:
Relocated all templates to folder "pages"

PR #946, opened 2017-08-25
Fix remaing newpages usage

henning-gerhardt:
There are a few entries are left over of renaming `newpages` to `pages`.


solth:
These occurences of "newpages" were already replaced in https://github.com/kitodo/kitodo-production/pull/924/commits/232c54d9d0ff81eccdb4c5b88abf41594a787ab2 but were reverted in https://github.com/kitodo/kitodo-production/commit/bfca818abf8b789d3276942d06ef2fede7a8c6c7 because that later commit was made before the former was merged into the master.

PR #949, opened 2017-08-25
Fix deletion of user filters

Kathrin-Huber:
The user needed to be refreshed, after the removal of the filter.


henning-gerhardt:
There is a merge conflict in `Kitodo/src/main/java/org/kitodo/services/data/UserService.java` - please fix this conflict.

Kathrin-Huber:
Rebased

PR #967, opened 2017-09-11
Fix defects found out by Coverity

Beacze:


Kitodo-Docket/src/main/java/org/kitodo/docket/Docket.java
```diff
@@ -29,9 +29,14 @@ public File generateDocket(DocketData docketData, URI xslFileUri) throws IOExcep
         File file = new File("docket.pdf");
         FileOutputStream fileOutputStream = new FileOutputStream(file);
 
-        exportDocket.startExport(docketData, fileOutputStream, new File(xslFileUri));
-        fileOutputStream.flush();
-        fileOutputStream.close();
+        try {
```
henning-gerhardt: Please use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) statement instead of handling this manually.

Kitodo-Docket/src/main/java/org/kitodo/docket/Docket.java
```diff
@@ -43,9 +48,14 @@ public File generateMultipleDockets(Collection<DocketData> docketData, URI xslFi
         File file = new File("docket_multipage.pdf");
         FileOutputStream fileOutputStream = new FileOutputStream(file);
 
-        exportDocket.startExport(docketData, fileOutputStream, new File(xslFileUri));
-        fileOutputStream.flush();
-        fileOutputStream.close();
+        try {
```
henning-gerhardt: Please use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) statement instead of handling this manually.

Kitodo-Docket/src/test/java/org/kitodo/docket/ExportDocketTest.java
```diff
@@ -39,9 +39,14 @@ public void testStartExport() throws IOException {
         FileOutputStream fileOutputStream = new FileOutputStream(file);
         File xslFile = new File("src/test/resources/docket.xsl");
 
-        exportDocket.startExport(docketDataGenerator.createDocketData("processId", "signature", "doctype"),
-                fileOutputStream, xslFile);
-        fileOutputStream.close();
+        try {
```
henning-gerhardt: Please use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) statement instead of handling this manually.

Kitodo/src/main/java/de/sub/goobi/forms/GranularityForm.java
```diff
@@ -262,7 +262,11 @@ public Long getNumberOfPages() {
      * @return the total number of pages of the digitization project
      */
     public Long getNumberOfPagesPerIssue() {
-        return numberOfPages != null ? numberOfPages / course.countIndividualIssues() : null;
+        if (course.countIndividualIssues() != 0) {
+            return numberOfPages != null ? numberOfPages / course.countIndividualIssues() : null;
+        } else {
+            return null;
```
henning-gerhardt: We would never return `null` as this value cause null pointer exceptions or add a lot of null value checks.

Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java
```diff
@@ -1569,7 +1572,9 @@ private void writeToOutputStream(FacesContext facesContext, File file, String fi
         response.setHeader("Content-Disposition", "attachment;filename=\"" + fileName + "\"");
 
         ServletOutputStream outputStream = response.getOutputStream();
-        byte[] bytes = IOUtils.toByteArray(new FileInputStream(file));
+        FileInputStream fileInputStream = new FileInputStream(file);
+        byte[] bytes = IOUtils.toByteArray(fileInputStream);
+        fileInputStream.close();
         outputStream.write(bytes);
         outputStream.flush();
```
henning-gerhardt: `outputStream` would never be closed after and let a open handle back.

Kitodo/src/main/java/de/sub/goobi/forms/GranularityForm.java
Beacze: so what should be return here? 0?

henning-gerhardt: Yes, this should be ok and correct value.

Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java
henning-gerhardt: Please ignore my comment. I did not see that `outputStream` is taken from `response` object and `outputStream` should not closed here.

Kitodo-Docket/src/main/java/org/kitodo/docket/Docket.java
```diff
@@ -27,11 +27,12 @@ public File generateDocket(DocketData docketData, URI xslFileUri) throws IOExcep
...
 
         File file = new File("docket.pdf");
-        FileOutputStream fileOutputStream = new FileOutputStream(file);
 
-        exportDocket.startExport(docketData, fileOutputStream, new File(xslFileUri));
-        fileOutputStream.flush();
-        fileOutputStream.close();
+        try (FileOutputStream fileOutputStream = new FileOutputStream(file)) {
+            exportDocket.startExport(docketData, fileOutputStream, new File(xslFileUri));
+        } catch (IOException e) {
```
stweil: This looks strange. You are catching an IOException, just to throw it again? Which CID do you want to fix here?

Beacze: Yes, it is more elegant way of ensure that fileOutputStream will be closed. We rethrown it as we have convention that all exceptions are going to be passed to Kitodo Core, not caught in modules.

henning-gerhardt:
A few days ago I created a new project on Coverity for myself as the existing project could not handle both branches with so much different structure in directory, modules and build tools. Last week I added some of the active developers of the DFG team to this new project. In near future it could be possible that this new Coverity project will be opened or the current open project will re-adjusted to the master branch.

Kitodo-Docket/src/main/java/org/kitodo/docket/Docket.java
stweil: It would be even more elegant if you simply omitted the catch part. It is not needed to achieve the desired behaviour. See https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html. I suggest to fix that for all code locations.

PR #968, opened 2017-09-11
Mapping

Beacze:
- add JSON files
- adjust Data Management module
- use in testing
- initial adjustment in frontend


Kitodo/src/main/java/de/sub/goobi/forms/IndexingForm.java
```diff
@@ -843,22 +854,13 @@ private void resetGlobalProgress() {
...
-            indexRestClient.initiateClient();
-            indexRestClient.setIndex(ConfigMain.getParameter("elasticsearch.index", "kitodo"));
-            indexRestClient.enableSortingByTextField(type, field);
-        } catch (CustomResponseException | IOException e) {
-            logger.error(e);
-        }
+    private static String readMapping() throws IOException, ParseException {
+        JSONParser parser = new JSONParser();
+        ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+        InputStream inputStream = classloader.getResourceAsStream("mapping.json");
```
henning-gerhardt: You are open a stream object without closing it never. Maybe you can use the try-with-resources statement here and avoid / handle the unhandled IOException here.

Kitodo/src/main/resources/mapping.json
```diff
@@ -0,0 +1,648 @@
+{
```
henning-gerhardt: Indent looks strange: between 1, 2 and 4 space indent. Could this be uniformed to one identical indent (Java code use 4 space indent).

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -23,6 +23,7 @@ Bibliothekarische\ Aufnahme\ f\u00FCr\ den\ lokalen\ OPAC=Bibliothekarische Aufn
 CatalogueChangeDocType=Dokumententyp ge\u00E4ndert zu
 CatalogueUnKnownType=Unbekannte Gattung
 CatalogueUsage=Verwendeter Katalog
+createMapping=ES Mapping erzeugen
```
henning-gerhardt: Please use full name for Elastic Search instead of the short name only understable by programmers.

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -59,7 +60,7 @@ Korrektur\ notwendig=Korrektur notwendig
 KorrekturFuer=Korrektur f\u00FCr Schritt
 KorrekturloesungFuer=Korrekturl\u00F6sung f\u00FCr Schritt
 Lizenzeinwerbung=Lizenzeinwerbung
-mappingUpdate=ElasticSearch Mapping aktualisieren
+mapping=ES Mapping
```
henning-gerhardt: Same here for short name

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -24,6 +24,7 @@ CatalogueChangeDocType=document type changed to
 CatalogueUnKnownType=unknown category
 CatalogueUsage=used Catalogue
 Corrections/Errors=Corrections / Errors
+createMapping=Create ES mapping
```
henning-gerhardt: Use full name instead of shorted one.

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -60,7 +61,7 @@ Korrektur\ notwendig=correction required
 KorrekturFuer=Correction for step
 KorrekturloesungFuer=Correction done for step
 Lizenzeinwerbung=Application for a licence
-mappingUpdate=Update ElasticSearch mapping
+mapping=ES mapping
```
henning-gerhardt: Use full name instead of shorted one.

Kitodo/src/test/java/org/kitodo/MockDatabase.java
```diff
@@ -114,6 +122,22 @@ private static IndexRestClient initializeIndexRestClient() {
         return restClient;
     }
 
+    private static String readMapping() {
+        JSONParser parser = new JSONParser();
+        JSONObject jsonObject = new JSONObject();
+        ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+        InputStream inputStream = classloader.getResourceAsStream("mapping.json");
```
henning-gerhardt: Opening a stream without closing it - this should be avoided and be replaced by try-with-resources statements.

Kitodo/src/test/resources/mapping.json
```diff
@@ -0,0 +1,648 @@
+{
```
henning-gerhardt: Same indent mix and did we need really to JSON files? I think one should be used for real and test usage as both will differ in future and tests did not check the real execution.

Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/KitodoRestClient.java
```diff
@@ -132,4 +150,14 @@ public String getType() {
     public void setType(String type) {
         this.type = type;
     }
+
+    protected int processStatusCode(StatusLine statusLine) throws CustomResponseException {
+        int statusCode = statusLine.getStatusCode();
+        if (statusCode >= 400 && statusCode < 452) {
```
henning-gerhardt: The whole 4xx range (400 until 499) is defined for client errors.

Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/KitodoRestClient.java
```diff
@@ -132,4 +150,14 @@ public String getType() {
...
+            throw new CustomResponseException("Client error: " + statusLine.toString());
+        } else if (statusCode >= 500 && statusCode < 512) {
```
henning-gerhardt: The whole 5xx range (500 until 599) is defined for server errors.

Kitodo/src/main/resources/messages/messages_de.properties
Beacze: But this thing must be changed in future task when "real" button will be made.

Kitodo/src/test/resources/mapping.json
Beacze: and how to enforce that test execution uses mapping from outside?

Kitodo/src/main/resources/messages/messages_de.properties
henning-gerhardt: Is there a problem if we now use a good and readable name and it is switch to an other kind of visibleness in future?

Kitodo/src/test/resources/mapping.json
henning-gerhardt: So far as I could discover it is not so easy to solve this or I did not found a good source for knowledge. We must stay with both files until we found a good and practicable way and changing of both files must be done.

PR #975, opened 2017-09-12
Fix user administration

solth:
Replace Jenia Popup windows with Primefaces Dialog in user administration to re-enable adding projects and user groups.


henning-gerhardt:
Please fix the merge conflict in `Kitodo/src/main/java/de/sub/goobi/forms/IndexingForm.java`.

solth:
Resolved conflict from rebase.

PR #977, opened 2017-09-13
Master-Template for new frontend design

solth:
Master-Template `base.xhtml` incorporates the new frontend design for the new version of Kitodo.Production and makes use of new "[Facelets Templating](http://docs.oracle.com/javaee/7/javaserver-faces-2-2/vdldocs-facelets/ui/tld-summary.html)" technique of JSF 2:
- controll the basic layout "frame" of all pages in a single master-template (`base.xhtml`)
- master-template defines mulitple template parts like `head`, `header`, `content` etc. using the `ui:insert` tag
- default values for those template parts are included in the master-template using the `ui:include` tag to include separate files like `header.xhtml`, `content.xhtml` etc.
- template-clients (templates for specific pages within Kitodo.Production) make use of the master-template and override only those defined template parts (mainly probably the `content` part) that have contents that are specific to that particular page; the other defined template parts can retain their default values and don't need to be overwritten
- an example for a page specific template-client is `webapp/pages/start.xhtml`
- this way the same header, footer, navigation etc. files don't have to be re-included in every single page template as is the case right now in Kitodo.Production
- `base.xhtml` and the template-parts `head.xhtml`, `header.xhtml`, `content.xhtml` and `footer.xhtml` are located in the `webapp/WEB-INF/templates` folder to restrict direct access to them to the system (they don't need to be exposed to the clients)
- in the same way, resources (stylesheets, Javascript files, images) are located in `webapp/WEB-INF/resources` (possible since JSF 2.2)


Kitodo/src/main/webapp/WEB-INF/templates/includes/head.xhtml
```diff
@@ -0,0 +1,26 @@
...
+
+<ui:composition xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
+                xmlns:h="http://xmlns.jcp.org/jsf/html">
+    <title><ui:insert name="title">Kitodo.Production</ui:insert></title>
+
+    <meta name="description" content="KITODO.PRODUCTION Base-Theme"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, shrink-to-fit=no"/>
+
+    <h:outputStylesheet name="css/kitodo.css"/>
+    <link href='http://fonts.googleapis.com/css?family=Raleway:300,400,500'/>
```
henning-gerhardt: I would suggest to switch to a protocol-less URI for loading this font as a secure connection would not allow to load a non-secure accessible data.

henning-gerhardt: Other question regarding to this font `Raleway`: it is allowed to use and distribute this font in our project?

Kitodo/src/main/webapp/WEB-INF/templates/includes/navigation.xhtml
```diff
@@ -0,0 +1,3 @@
+<ui:composition xmlns:ui="http://xmlns.jcp.org/jsf/facelets" xmlns:h="http://java.sun.com/jsf/html">
```
henning-gerhardt: Missing license header for this xhtml file.

Kitodo/src/main/webapp/WEB-INF/templates/includes/head.xhtml
solth: > Other question regarding to this font Raleway: it is allowed to use and distribute this font in our project?

According to https://www.fontsquirrel.com/license/raleway this font uses the [SIL Open Font License](https://en.wikipedia.org/wiki/SIL_Open_Font_License), which - according to Wikipedia - "_permits the fonts to be used, modified, and distributed freely (so long as the resulting fonts remain under the Open Font License). However, the copyright holder may declare the font's name as being a "Reserved Font Name", which modified versions then cannot bear. The License permits covered fonts to be freely embedded in documents under any terms, but it requires that fonts be packaged with software if they are sold._"

henning-gerhardt: Thank you for this explanation for the license of the font. Can you add a comment with a link to the license? Thank you.

PR #982, opened 2017-09-14
Partial indexing

Beacze:


henning-gerhardt:
Please rebase against master as there is at least one merge conflict in `Kitodo/src/main/java/org/kitodo/services/data/RulesetService.java`.

PR #986, opened 2017-09-15
Fix task administration

Beacze:
- fix NPE exceptions
- fix popup windows for insert / edit tasks (ProzessverwaltungForm)

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/BenutzerverwaltungForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/BenutzerverwaltungForm.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 import java.util.regex.Matcher;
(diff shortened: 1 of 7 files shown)


Kitodo/src/main/java/de/sub/goobi/forms/BenutzerverwaltungForm.java
```diff
@@ -131,6 +132,20 @@ public String filterAlleStart() {
     }
 
     /**
+     * Return list of users.
+     *
+     * @return list of user groups
+     */
+    public List<UserDTO> getUsersForTasks() {
```
Kathrin-Huber: Where is the correlation the the task here? You just get allActiveUsers

Beacze: you need this list for popup window with users

Kathrin-Huber: but why not call it "getActiveUsers"?

Beacze: Because in the future getActiveUsers will contain active users with project and user group titles, and this one will contain only list of users without any additional data - so slim version used only for tasks' popup.

PR #989, opened 2017-09-19
Hibernate 5 migration

Beacze:
- change version to 5.2.10
- introduce proper id generation (no rely on default)
- introduce transactions in DAO classes
- reverse ownership between Project and User (Project is added to User, User performs save)
- clear separation for HibernateHelper
- use only one HibernateUtil (remove other)
- refresh result lists after save/edit/remove
- run IT independently (clean MockDatabase, separate ES and Hibernate)

Depends on #986 and #994


Kitodo/src/main/java/org/kitodo/services/data/base/SearchService.java
```diff
@@ -233,6 +233,11 @@ public void save(T baseIndexedBean) throws DataException {
             saveToDatabase(baseIndexedBean);
             saveToIndex(baseIndexedBean);
             manageDependenciesForIndex(baseIndexedBean);
+            try {
```
henning-gerhardt: This and the similar changes in `remove` method looks like a workaround / hack for a problem. You should add at least a `TODO` for this as it could not fixed yet.

Kitodo/src/main/java/org/kitodo/services/data/base/SearchService.java
```diff
@@ -290,6 +295,11 @@ public void remove(T baseIndexedBean) throws DataException {
             saveToDatabase(baseIndexedBean);
             removeFromIndex(baseIndexedBean);
             manageDependenciesForIndex(baseIndexedBean);
+            try {
```
henning-gerhardt: See my comment in the `save' method.

Kitodo/src/main/java/org/kitodo/services/data/base/SearchService.java
Beacze: It is not a hack or workaround. It is necessary to "give" time to ES to index all changes.

Kitodo/src/main/java/org/kitodo/services/data/base/SearchService.java
Beacze: Like above.

Pf course if someone knows better way to handle it I'm open to discuss it :)

PR #991, opened 2017-09-20
IT tests

Beacze:
- decrease sleep time in tests
- insert only dockets in DocketServiceIT
- insert only ldap groups and disable usage of ES node in LdapGroupServiceIT
- insert only rulesets in RulesetServiceIT
- insert only users and users' groups in UserGroupServiceIT
- re-enable SQL logging during tests to console

Depends on #989 and #990


henning-gerhardt:
There are merge conflicts, please resolve them.

PR #992, opened 2017-09-20
Create mapping and delete index from indexing page

solth:


Kitodo/src/main/java/de/sub/goobi/forms/IndexingForm.java
```diff
@@ -110,6 +106,10 @@ public void run() {
 
     private boolean indexingAll = false;
 
+    // -2 = deletion failed; -1 = mapping failed
+    // 1 = mapping successful; 2 = deletion successfull
+    private int indexState = 0;
```
henning-gerhardt: Why you are not using a enum for possible values? You describe them already, why not using them? It would even made code more readable as some magic numbers.

Kitodo/src/main/webapp/pages/Indexing.xhtml
```diff
@@ -32,10 +44,40 @@
                             PF('progressPoll').stop();
                             $('.refreshTable').click();
                         }
+                        else if (message.includes("mapping")) {
```
henning-gerhardt: I would prefer a similar formatting style for JavaScript and Java. This would mean, that `else` blocks should begin in them line as normal `if` block ends.
I'm open for other opinions as I did not wrote many JavaScript based code.

PR #998, opened 2017-09-21
Improve JSON objects conversion in services' classes

Beacze:


Kitodo/src/main/java/org/kitodo/services/data/BatchService.java
```diff
@@ -281,10 +281,12 @@ public void addAllObjectsToIndex() throws CustomResponseException, InterruptedEx
...
         BatchDTO batchDTO = new BatchDTO();
-        batchDTO.setTitle(getStringPropertyForDTO(jsonObject, "title"));
-        batchDTO.setType(getStringPropertyForDTO(jsonObject, "type"));
+        batchDTO.setId(getIdFromJSONObject(jsonObject));
+        JSONObject batchObject = getSource(jsonObject);
+        batchDTO.setTitle(getStringPropertyForDTO(batchObject, "title"));
+        batchDTO.setType(getStringPropertyForDTO(batchObject, "type"));
         if (!related) {
-            batchDTO = convertRelatedJSONObjects(jsonObject, batchDTO);
+            batchDTO = convertRelatedJSONObjects(batchObject, batchDTO);
```
Kathrin-Huber: batchObject sounds too much like it's a bean.
I would prefere batchJSONObject

Kitodo/src/main/java/org/kitodo/services/data/BatchService.java
```diff
@@ -281,10 +281,12 @@ public void addAllObjectsToIndex() throws CustomResponseException, InterruptedEx
     @Override
     public BatchDTO convertJSONObjectToDTO(JSONObject jsonObject, boolean related) throws DataException {
         BatchDTO batchDTO = new BatchDTO();
-        batchDTO.setTitle(getStringPropertyForDTO(jsonObject, "title"));
-        batchDTO.setType(getStringPropertyForDTO(jsonObject, "type"));
+        batchDTO.setId(getIdFromJSONObject(jsonObject));
+        JSONObject batchObject = getSource(jsonObject);
```
Kathrin-Huber: why did you move the "getSource" out of the methods?

Beacze: I wanted to avoid this one level querying _source for every property. Probably it doesn't give much increase of the speed for one object, but for 1000 or 10000 it can already make a difference.

Kitodo/src/main/java/org/kitodo/services/data/BatchService.java
Beacze: renamed

PR #999, opened 2017-09-21
Fix nullruleset_slubdd.xml in RulesetService test

Beacze:
- add test ruleset
- ad variable for ruleset directory to test config file
- test content of ruleset file

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/services/data/RulesetServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/services/data/RulesetServiceIT.java
@@ -29,6 +29,8 @@ import org.kitodo.MockDatabase;
 import org.kitodo.data.database.beans.Ruleset;
 import org.kitodo.data.database.exceptions.DAOException;
 
+import ugh.dl.DocStructType;
+
 /**
  * Tests for RulesetService class.
  */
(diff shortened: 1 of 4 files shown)


Kitodo/src/test/resources/ruleset_test.xml
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
```
henning-gerhardt: Please add default Kitodo license to this file.

Kitodo/src/test/java/org/kitodo/services/data/RulesetServiceIT.java
```diff
@@ -127,10 +129,10 @@ public void shouldFindByTitle() throws Exception {
     public void shouldFindByFile() throws Exception {
         RulesetService rulesetService = new RulesetService();
 
-        JSONObject ruleset = rulesetService.findByFile("ruleset_slubdd.xml");
+        JSONObject ruleset = rulesetService.findByFile("ruleset_slubhh.xml");
```
Kathrin-Huber: why rename it from slubdd to slubhh?
ruleset_test, like in the other place, is way better

Beacze: because slubdd was renamed to test, I didn't want to use real slubdd file in test

Kathrin-Huber: so rename it to something without slub OR hh in it.
ruleset_title_test... whatever...

Kitodo/src/test/resources/ruleset_test.xml
Beacze: done

Kitodo/src/test/java/org/kitodo/services/data/RulesetServiceIT.java
Beacze: done

PR #1006, opened 2017-09-25
Check process completeness

Kathrin-Huber:


Kitodo/src/test/java/org/kitodo/services/data/ProjectServiceTest.java
```diff
@@ -0,0 +1,50 @@
+package org.kitodo.services.data;
```
henning-gerhardt: Missing license header - please add missing header.

Kitodo/src/main/java/org/kitodo/services/data/ProjectService.java
```diff
@@ -361,4 +363,21 @@ private ProjectDTO convertRelatedJSONObjects(JSONObject jsonObject, ProjectDTO p
...
+     *
+     * @param project
+     *            The project to check
+     * @return true, if project is complete and can be used, false, if project
+     *         is incomplete
+     */
+    public boolean isProjectComplete(Project project) {
+        boolean projectsXmlExists = (new File(ConfigCore.getKitodoConfigDirectory() + "kitodo_projects.xml")).exists();
+        boolean digitalCollectionsXmlExists = (new File(
+                ConfigCore.getKitodoConfigDirectory() + "kitodo_digitalCollections.xml")).exists();
```
henning-gerhardt: For `kitodo_digitalCollections.xml` is in Kitodo/src/main/java/org/goobi/production/constants/FileNames.java a constant value defined. It should be considered to use this and add maybe an other for `kitodo_projects.xml`. Both should be used as much as possible instead of hard coding them everywhere.

PR #1008, opened 2017-09-25
Add resource bundle for message files

solth:


solth:
Defining the resource bundle in the `faces-config.xml` this way makes the corresponding messages available on a global scale in the whole application. Using `f:loadBundle` instead - as it is the case in Kitodo.Production right now - only makes the messages available on a local scale to individual views or pages. 

This can - and currently _is_ - mostly be mitigated by adding the `f:loadBundle` tag to the `head.xhtml` file - which in turn normally is included in every other template - but using the `resource-bundle` to _explicitely_ declare the resource bundle globally seems like the cleaner approach, IMHO.

In addition, the IntelliJ IDE recognizes globally defined resource bundles and allows development aids like auto completion for strings in the corresponding message files (for locally defined resource bundles, this only works for the files that explicitely include the `f:loadBundle` tag, not for those that get access to the resource bundle implicitly by including the standard `head.xhtml` template fragment).

See https://www.mkyong.com/jsf2/jsf-2-0-and-resource-bundles-example/ for addtional information.

henning-gerhardt:
Thank you for this explanation - without it was not possible for me to understand this change.

PR #1025, opened 2017-09-29
Removed "required"-property of project title on tab change, updated buttons to primefaces

tim-boerner:


henning-gerhardt:
@tim-boerner : Your pull request #1033 made this changes superfluous as this contains a special solution for one page and #1033 a more general solution.

PR #1030, opened 2017-09-29
Fix task save fix redirecting

tim-boerner:


Beacze:
Depends on #1030

Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
```diff
@@ -714,11 +714,10 @@ public String newTask() {
         } catch (DataException e) {
             logger.error(e);
         }
-        save();
+        //save();
```
henning-gerhardt: Do not comment out obsolete code - remove it. You should squash this and your later delete commit to one.

henning-gerhardt:
There are two merge conflicts in `Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java` - please resolve them.

PR #1031, opened 2017-10-04
Minor fixes task save

Beacze:
Depends on #1030


Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
```diff
@@ -741,7 +745,7 @@ public void saveTask() {
      * @return page
      */
     public String deleteTask() {
-        this.process = this.task.getProcess();
+        //this.process = this.task.getProcess();
```
henning-gerhardt: If this code is not needed, remove it or don't touch it, if this line get removed in your next commit.

PR #1029, opened 2017-09-29
Fix task add / edit / delete

Beacze:
Redirection after insert / delete of task is incorrect.

Depends on #1028

PR #1032, opened 2017-10-04
Fix change processing status

Beacze:
Depends on #1029

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/persistence/apache/FolderInformation.java
+++ b/Kitodo/src/main/java/de/sub/goobi/persistence/apache/FolderInformation.java
@@ -63,28 +63,28 @@ public class FolderInformation {
         /* nur die _tif-Ordner anzeigen, die nicht mir orig_ anfangen */
         FilenameFilter filterDirectory = new FileNameEndsAndDoesNotBeginWithFilter(DIRECTORY_PREFIX + "_",
                 "_" + DIRECTORY_SUFFIX);
-        URI tifOrdner = null;
+        URI tifDirectory = null;
         ArrayList<URI> directories = fileService.getSubUris(filterDirectory, dir);
         for (URI directory : directories) {
-            tifOrdner = directory;
+            tifDirectory = directory;
         }
 
-        if (tifOrdner == null && useFallBack) {
+        if (tifDirectory == null && useFallBack) {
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/java/de/sub/goobi/persistence/apache/FolderInformation.java
```diff
@@ -63,57 +63,46 @@ public URI getImagesTifDirectory(boolean useFallBack) {
...
 
-        return rueckgabe;
-    }
-
-    /**
-     * Get tif directory exists.
-     *
-     * @return true if the Tif-Image-Directory exists, false if not
-     */
-    public Boolean getTifDirectoryExists() {
```
henning-gerhardt: You can not remove method `getTifDirectoryExists()` as this method is still used in `Kitodo/src/main/webapp/pages/inc_Prozessverwaltung/Prozesse_Liste.xhtml` lines 468, 476, 483, 490, 496 and 503!

Beacze: item.tifDirectoryExists - item is process object, so this method should be in ProcessDTO:

`<t:dataTable id="auflistung" styleClass="standardTable" width="100%" cellspacing="1px" cellpadding="1px"
                 headerClass="standardTable_Header"
                 rowClasses="standardTable_Row1,standardTable_Row2,standardTable_Row1,standardTable_Row2" var="item"
                 value="#{ProzessverwaltungForm.page.listReload}">`

henning-gerhardt: Then this method must be available in `ProcessDTO` which is not the case yet.

Beacze: Actually it is:

public boolean isTifDirectoryExists() {
        return tifDirectoryExists;
    }

As it is boolean, it works with is....: https://stackoverflow.com/questions/14400222/boolean-properties-starting-with-is-does-not-work

henning-gerhardt: Ah, ok. Thank you for this explanation. Then this method could be removed.

henning-gerhardt:
There is a merge conflict in `Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java` - please resolve this conflict.

PR #1033, opened 2017-10-04
added encoding filter

tim-boerner:


Kitodo/src/main/java/de/sub/goobi/helper/servletfilter/RequestControlFilter.java
```diff
@@ -71,6 +73,11 @@
     @SuppressWarnings({"unchecked", "rawtypes" })
     public void init(FilterConfig config) throws ServletException {
 
+        encoding = config.getInitParameter("requestEncoding");
```
henning-gerhardt: Where must this parameter be defined to set the value? It would be nice if you added the proper value to this file.

tim-boerner:
done

PR #1040, opened 2017-10-05
Fix tomcat url for selenium tests

tim-boerner:


henning-gerhardt:
Merging this pull request as emergency fix with reason: new Tomcat version was released and old used version is only available on archive server on the Tomcat project.

PR #1047, opened 2017-10-06
Cleanup of C3P0 usage

henning-gerhardt:
- remove obsolete hibernate property
- remove obsolete and unused dependency


Kitodo/src/main/resources/hibernate.cfg.xml
```diff
@@ -38,11 +38,6 @@
         <property name="hibernate.c3p0.acquire_increment">1</property>
```
Beacze: Are not this two properties also related to it?

henning-gerhardt: No. This parameters are used and should not removed. Removed dependency contains only a debug class and a Oracle utils class which we did not need.

PR #1022, opened 2017-09-28
Remove msgs.get - method get doesn't exists in PropertyResourceBundle

Beacze:
Fix for:
javax.el.MethodNotFoundException - Method not found: class java.util.PropertyResourceBundle.get(java.lang.String)

PR #1049, opened 2017-10-06
Rename parameters to display details of tasks

Beacze:


Beacze:
We store already formatted date String in index. So index returns as String. 

Translations are not working, as get method is not supported (https://github.com/kitodo/kitodo-production/pull/1022).

henning-gerhardt:
Formatting of date including time should be done on displaying in context of used browser / user language and not while indexing. Indexing should be with a neutral date / time value. This was the reason why this "AsFormattedString" methods was used - beside the wrong name. Translation of values could even be done on accessing this values by an own method.

Beacze:
Translations are added.

PR #1050, opened 2017-10-06
Add missing tests for TaskService and minor fix for DTO conversion

Beacze:
Fix NPE on processing user fetching.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
+++ b/Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
@@ -360,7 +360,9 @@ public class TaskService extends TitleSearchService<Task, TaskDTO, TaskDAO> {
         Integer process = getIntegerPropertyForDTO(jsonObject, "process");
         taskDTO.setProcess(serviceManager.getProcessService().findById(process, true));
         Integer processingUser = getIntegerPropertyForDTO(jsonObject, "processingUser");
-        taskDTO.setProcessingUser(serviceManager.getUserService().findById(processingUser, true));
+        if (processingUser != 0) {
+            taskDTO.setProcessingUser(serviceManager.getUserService().findById(processingUser, true));
+        }
         taskDTO.setUsers(convertRelatedJSONObjectToDTO(jsonObject, "users", serviceManager.getUserService()));
         taskDTO.setUserGroups(
                 convertRelatedJSONObjectToDTO(jsonObject, "userGroups", serviceManager.getUserGroupService()));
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -358,7 +358,9 @@ private TaskDTO convertRelatedJSONObjects(JSONObject jsonObject, TaskDTO taskDTO
         Integer process = getIntegerPropertyForDTO(jsonObject, "process");
         taskDTO.setProcess(serviceManager.getProcessService().findById(process, true));
         Integer processingUser = getIntegerPropertyForDTO(jsonObject, "processingUser");
-        taskDTO.setProcessingUser(serviceManager.getUserService().findById(processingUser, true));
+        if (processingUser != 0) {
```
henning-gerhardt: If no user is currently working on this task as he / she aborted this task, this should be reflected into the data.

Beacze: That's exactly what this line does. It checks if some user works on this, if yes, it requests information about this user, if not, it leaves this attribute null.

henning-gerhardt: Sorry, I did not see that you are resetting the user. Your code set only the code if the `processingUser` is not `0` and if I look into `getIntegerPropertyForDTO`  it is returning `0` if the value even contains value `null`. So it is never set to an other value even no user is working any more on this task.

Beacze: Could you explain more detailed what is a problem here?

henning-gerhardt: I would only ensure that a formally attached user to a task is really removed from indexed data if this user is not involved with this task any more.

Beacze: Method responsible for it is manageProcessingUserDependenciesForIndex(Task task), but I see that there must be actual check if ProcessingUser is not empty.

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -117,7 +120,18 @@ private void manageProcessingUserDependenciesForIndex(Task task) throws CustomRe
...
             User user = task.getProcessingUser();
-            serviceManager.getUserService().saveToIndex(user);
+            if (user != null) {
+                serviceManager.getUserService().saveToIndex(user);
+            }
+            reSaveUserAfterRemoveFromProcessing(task);
+        }
+    }
+
+    private void reSaveUserAfterRemoveFromProcessing(Task task)
```
henning-gerhardt: It is not more a re-index after removing as your method does apply changes only to index?

PR #1052, opened 2017-10-06
Fix triple save on process properties

Beacze:


henning-gerhardt:
Can you please provide some tests as current tests did not cover this wrong behaviour.

Beacze:
Actually it can be quiet hard to write tests for it as this.processPropertyList doesn't have set method and it uses directly data passed from the frontend.

PR #1053, opened 2017-10-06
Fix double property duplication

Beacze:
Depends on #1052


henning-gerhardt:
Can you please provide some tests as current tests did not cover this wrong behaviour.

Beacze:
The same situation like in the PR #1052.

PR #1056, opened 2017-10-13
Import lugh from 2.x rebased

matthias-ronge:
Add new module to prepare for substitution of UGH library.

All source code does not yet interact with the existing code, so this should not cause any inferences. Source code is covered with tests and documented in Javadoc.


matthias-ronge:
Review abgeschlossen, kann jetzt gemerged werden

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+    public MemoryResult get(String relation) {
+        return new MemoryResult(edges.get(relation));
+    }
+
+    /**
+     * {@inheritDoc}<!-- . -->
+     */
+    @Override
+    public MemoryResult get(String relation, IdentifiableNode identifierRelation, ObjectType identifierValue) {
+        return get(relation != null ? Arrays.asList(new String[] {relation }) : null,
```
henning-gerhardt: Return value `null` should be avoided. Is a `new MemoryResult()` not a better solution?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/RDF.java
```diff
@@ -0,0 +1,233 @@
...
+
+import java.math.BigInteger;
+import java.util.HashMap;
+import java.util.Map;
+import org.kitodo.dataaccess.storage.memory.MemoryNodeReference;
+
+/**
+ * The {@code http://www.w3.org/1999/02/22-rdf-syntax-ns} namespace.
+ */
+public class RDF {
```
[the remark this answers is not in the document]
Beacze: https://stackoverflow.com/questions/20874000/use-of-enums-in-java-with-different-datatypes here good answer how to implement enums for different datatypes

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/RDFS.java
```diff
@@ -0,0 +1,144 @@
...
+package org.kitodo.dataaccess;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.kitodo.dataaccess.storage.memory.MemoryNodeReference;
+
+/**
+ * The {@code http://www.w3.org/2000/01/rdf-schema} namespace.
+ */
+public class RDFS {
```
Beacze: like previous one

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+            if (value != null) {
+                if ((first == null) || (value < first)) {
+                    first = value;
+                }
+                if ((last == null) || (value > last)) {
+                    last = value;
+                }
+            }
+        }
+        return (first == null) || (last == null) ? null : new Indices(first, last);
```
henning-gerhardt: Returning a `null` value should be avoided. Is a `new Indeces()` not a better solution?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/SerializationFormat.java
```diff
@@ -0,0 +1,170 @@
...
+                    bos.close();
+                } catch (IOException e) {
+                    /* close silently */
+                }
+            }
+            if (fos != null) {
+                try {
+                    fos.close();
+                } catch (IOException e) {
+                    /* close silently */
```
Beacze: no empty catch blocks, at least logger

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/AmbiguousDataExceptionTest.java
```diff
@@ -0,0 +1,24 @@
...
+package org.kitodo.dataaccess;
+
+import org.junit.Test;
+import org.kitodo.dataaccess.AmbiguousDataException;
+
+public class AmbiguousDataExceptionTest {
+
+    @Test(expected = AmbiguousDataException.class)
+    public void testAmbiguousDataExceptionCanBeThrown() throws AmbiguousDataException {
+        throw new AmbiguousDataException();
```
henning-gerhardt: What this kind of test? You are testing the test framework!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/XMLSchema.java
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * The XML Schema datatypes.
+ *
+ * @see "https://www.w3.org/TR/xmlschema-2/"
+ */
+public class XMLSchema {
+    /**
+     * A reference to a named linked data node that represents an integer type.
+     */
+    public static class IntegerDatatype implements NodeReference {
```
Beacze: In Coding Guideline we have pointed out that every file can contain only one class.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/LinkedDataExceptionTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.Test;
+import org.kitodo.dataaccess.AmbiguousDataException;
+import org.kitodo.dataaccess.LinkedDataException;
+import org.kitodo.dataaccess.NoDataException;
+
+public class LinkedDataExceptionTest {
+
+    @Test(expected = LinkedDataException.class)
+    public void testAmbiguousDataExceptionCanBeThrownAsLinkedDataException() throws LinkedDataException {
+        throw new AmbiguousDataException();
```
henning-gerhardt: What this kind of test? You are testing the test framework!

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/LinkedDataExceptionTest.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Test(expected = LinkedDataException.class)
+    public void testNoDataExceptionCanBeThrownAsLinkedDataException() throws LinkedDataException {
+        throw new NoDataException();
```
henning-gerhardt: What this kind of test? You are testing the test framework!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Namespaces.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    public static String namespaceOf(String url) {
+        int numberSign = url.indexOf('#');
+        if (numberSign > -1) {
+            return url.substring(0, numberSign + 1);
+        }
+        return url.substring(0, url.lastIndexOf('/') + 1);
+    }
+
+    // /**
```
Beacze: Don't leave commented code.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/GraphPath.java
```diff
@@ -0,0 +1,313 @@
...
+ * Inspired by XPath.
+ */
+public class GraphPath extends MemoryNode {
+
+    /**
+     * Parser to recursively parse an object. The parsing method returns the
+     * number of code points consumed, the result is provided in the field
+     * {@code result}.
+     */
+    private static class ObjectParser {
```
Beacze: Again only one class in one file.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/GraphPath.java
```diff
@@ -0,0 +1,313 @@
...
+                predicates.add(predicate.getIdentifier());
+            }
+            Collection<ObjectType> objects = new LinkedList<>();
+            toSegment.get(RDF.OBJECT).forEach(objects::add);
+            for (Node node : nodesOnly) {
+                node.get(predicates, objects).forEach(thisResult::add);
+            }
+            apply(thisResult, toSegment).forEach(result::add);
+        }
+
```
Beacze: Is there are specific reason to leave here two blank lines before return and in next method none?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NoDataExceptionTest.java
```diff
@@ -0,0 +1,24 @@
...
+package org.kitodo.dataaccess;
+
+import org.junit.Test;
+import org.kitodo.dataaccess.NoDataException;
+
+public class NoDataExceptionTest {
+
+    @Test(expected = NoDataException.class)
+    public void testNoDataExceptionCanBeThrown() throws NoDataException {
+        throw new NoDataException();
```
henning-gerhardt: What this kind of test? You are testing the test framework!

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
```diff
@@ -0,0 +1,134 @@
...
+            NodeReference other = storage.createNodeReference("http://names.example/foo");
+
+            assertThat(one.hashCode() == other.hashCode(), is(true));
+        }
+    }
+
+    @Test
+    public void testNodeReferenceCanBeCreatedFromHttpUrl() {
+        for (Storage storage : TestConfig.STORAGES_TO_TEST_AGAINST) {
+            storage.createNodeReference("http://names.example/foo#bar");
```
henning-gerhardt: What you are testing here? Did not look like a test as no comparison or assertion is done.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    @Test
+    public void testNodeReferenceCanBeCreatedFromNonHttpUrl() {
+        for (Storage storage : TestConfig.STORAGES_TO_TEST_AGAINST) {
+            storage.createNodeReference("urn:example:kitodo-123456789-xyz");
```
henning-gerhardt: What you are testing here? Did not look like a test as no comparison or assertion is done.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/GraphPath.java
```diff
@@ -0,0 +1,313 @@
...
+    /**
+     * Creates an empty GraphPath. An empty GraphPath points to the node it is
+     * applied on.
+     *
+     * <p>
+     * This method is used in tests to build comparable objects
+     * programmatically.
+     */
+    protected GraphPath() {
+        super(GRAPH_PATH);
```
Beacze: Why make her something what is only specifically fot tests? Can't tests use public constructor?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/GraphPath.java
```diff
@@ -0,0 +1,313 @@
...
+            }
+
+            if ((index < length) && (string.codePointAt(index) == '[')) {
+                index++;
+
+                ObjectParser parseObjectRecursive = new ObjectParser(prefixes);
+                index += parseObjectRecursive.consume(string.substring(index));
+                index++;
+                graphPosition.put(RDF.OBJECT, parseObjectRecursive.result);
+
```
Beacze: Those blank lines are not needed here.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
```diff
@@ -0,0 +1,1134 @@
...
+            assertThat(one, is(equalTo(another)));
+        }
+    }
+
+    /**
+     * Tests {@code find(Node)}.
+     */
+    @Test
+    public void testFind() {
+        new GraphPathTest().testApplyingAGraphPathToANode();
```
henning-gerhardt: What did you test here? It did not look like a test.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryLangString.java
```diff
@@ -0,0 +1,186 @@
...
+                default:
+                    return false;
+            }
+            return super.matches(condition);
+        }
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}<!-- . -->
```
Beacze: Overridden methods don't need JavaDocs.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryLiteral.java
```diff
@@ -0,0 +1,332 @@
...
+        } else if (!value.equals(other.value)) {
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * {@inheritDoc}<!-- . -->
+     */
+    @Override
```
Beacze: Overridden methods don't need JavaDocs.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
```diff
@@ -0,0 +1,1134 @@
...
+        }
+    }
+
+    /**
+     * Tests {@code put(IdentifiableNode, ObjectType)}.
+     */
+    @Test
+    public void testPutNodeReferenceObjectType() {
+        for (Storage storage : TestConfig.STORAGES_TO_TEST_AGAINST) {
+            storage.createNode().put(Mods.NAME, storage.createLeaf("Wilhelm Busch", "de"));
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
```diff
@@ -0,0 +1,1134 @@
...
+        }
+    }
+
+    /**
+     * Tests {@code put(IdentifiableNode, String)}.
+     */
+    @Test
+    public void testPutNodeReferenceString() {
+        for (Storage storage : TestConfig.STORAGES_TO_TEST_AGAINST) {
+            storage.createNode().put(Mods.NAME, "Wilhelm Busch");
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
```diff
@@ -0,0 +1,1134 @@
...
+        }
+    }
+
+    /**
+     * Tests {@code put(String, ObjectType)}.
+     */
+    @Test
+    public void testPutStringObjectType() {
+        for (Storage storage : TestConfig.STORAGES_TO_TEST_AGAINST) {
+            storage.createNode().put("http://www.loc.gov/mods/v3#name", storage.createLeaf("Wilhelm Busch", "de"));
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryLiteral.java
```diff
@@ -0,0 +1,332 @@
...
+    public String getValue() {
+        return value;
+    }
+
+    /**
+     * Returns a hash code of the Literal.
```
Beacze: If this method somehow differs from predecessor, please write complete JavaDoc.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
```diff
@@ -0,0 +1,1134 @@
...
+        }
+    }
+
+    /**
+     * Tests {@code put(String, String)}.
+     */
+    @Test
+    public void testPutStringString() {
+        for (Storage storage : TestConfig.STORAGES_TO_TEST_AGAINST) {
+            storage.createNode().put("http://www.loc.gov/mods/v3#name", "Wilhelm Busch");
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+        private ChildNodeIterator() {
+            outer = edges.entrySet().iterator();
+        }
+
+        /**
+         * Looks for a next node.
+         *
+         * @return a next node, or null if there is none
+         */
+        private ObjectType findNextNode() {
```
Beacze: Return empty object instead of null.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+
+    /**
+     * The edges leading from this node to other nodes and leaves.
+     */
+    protected final HashMap<String, Collection<ObjectType>> edges = new HashMap<>();
+
+    /**
+     * Creates an empty node.
+     */
+    public MemoryNode() {
```
Beacze: Not needed declaration of the constructor.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+    }
+
+    /**
+     * The pair of the first and last index in use, as return type of the
+     * {@code range()} function.
+     */
+    private class Indices {
```
Beacze: Only one class for file.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+            edges.put(RDF.toURL(i), edges.remove(RDF.toURL(i - 1)));
+        }
+
+        // add the given element at the given position
+        edges.put(RDF.toURL(index), new HashSet<>(Arrays.asList(new ObjectType[] {element })));
+    }
+
+    /**
+     * {@inheritDoc}<!-- . -->
+     */
```
Beacze: No JavaDoc here.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+                        // Attach any literals as rdf:value
+
+                        result.put(RDF.VALUE.getIdentifier(), object);
+                    }
+                }
+
+            } else {
+
+                // Nodes that are not attached numerically or by value will just
+                // be
```
Beacze: Comment is bad formatted.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
```diff
@@ -0,0 +1,1096 @@
...
+                // copied.
+
+                result.putAll(relation, edge.getValue());
+            }
+        }
+
+        // If the result is one single rdf:value, just return the value
+
```
Beacze: Don't make blank line between one line comment and code

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/Mets.java
```diff
@@ -0,0 +1,149 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import org.kitodo.dataaccess.NodeReference;
+import org.kitodo.dataaccess.storage.memory.MemoryStorage;
+
+/**
+ * The {@code http://www.loc.gov/METS/} namespace.
+ */
+public class Mets {
```
Beacze: Enum

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/Mods.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import org.kitodo.dataaccess.NodeReference;
+import org.kitodo.dataaccess.storage.memory.MemoryStorage;
+
+/**
+ * The {@code http://www.loc.gov/mods/v3} namespace.
+ */
+public class Mods {
```
Beacze: Enum

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNamedNodeTest.java
```diff
@@ -0,0 +1,102 @@
...
+    public void testHashCodeIsInequalForADifferentURI() {
+        MemoryNamedNode one = new MemoryNamedNode("http://names.kitodo.org/test#Object1");
+        MemoryNamedNode another = new MemoryNamedNode("http://names.kitodo.org/test#Object2");
+
+        assertThat(one.hashCode() == another.hashCode(), is(false));
+    }
+
+    @Test
+    public void testMemoryNamedNodeCanBeCreatedWithAHttpUri() {
+        new MemoryNamedNode("http://names.kitodo.org/test#Object1");
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
```diff
@@ -0,0 +1,1134 @@
...
+            assertThat(node.getEnumerated().stream().mapToInt(r -> r.isAny() ? 1 : 0).sum(), is(3));
+            assertThat(node.remove(bob), is(true));
+            assertThat(node.getEnumerated().stream().mapToInt(r -> r.isAny() ? 1 : 0).sum(), is(2));
+            assertThat(node.remove(bob), is(false));
+            assertThat(node.getEnumerated().stream().mapToInt(r -> r.isAny() ? 1 : 0).sum(), is(2));
+        }
+    }
+
+    /**
+     * Tests {@code replace(String, Set<ObjectType>)}.
```
Beacze: No JavaDocs for tests classes.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNamedNodeTest.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Test
+    public void testMemoryNamedNodeCanBeCreatedWithANonHttpUri() {
+        new MemoryNamedNode("urn:example:kitodo:test-1234567-xyz");
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
```diff
@@ -0,0 +1,92 @@
...
+    public void testHashCodeIsEqualForTwoEqualInstances() {
+        MemoryNodeReference one = new MemoryNodeReference("http://names.example/foo");
+        MemoryNodeReference other = new MemoryNodeReference("http://names.example/foo");
+
+        assertThat(one.hashCode() == other.hashCode(), is(true));
+    }
+
+    @Test
+    public void testMemoryNodeReferenceCanBeCreatedFromHttpUrl() {
+        new MemoryNodeReference("http://names.example/foo#bar");
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    public void testMemoryNodeReferenceCanBeCreatedFromNonHttpUrl() {
+        new MemoryNodeReference("urn:example:kitodo-123456789-xyz");
```
henning-gerhardt: What did you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeTest.java
```diff
@@ -0,0 +1,947 @@
...
+
+        assertThat(node.matches(condition), is(false));
+    }
+
+    /**
+     * Tests {@code MemoryNode()} constructor.
+     */
+    @Test
+    public void testMemoryNodeCanBeCreatedEmpty() {
+        new MemoryNode();
```
henning-gerhardt: What are you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeTest.java
```diff
@@ -0,0 +1,947 @@
...
+        assertThat(node.get(Mods.NAME).countUntil(Long.MAX_VALUE), is((long) 3));
+    }
+
+    /**
+     * Tests {@code put(String, ObjectType)}.
+     */
+
+    @Test
+    public void testPutStringObjectType() {
+        new MemoryNode().put("http://www.loc.gov/mods/v3#name", MemoryLiteral.createLeaf("Wilhelm Busch", "de"));
```
henning-gerhardt: What are you testing here?

Beacze:
On mor general remark: 
commits like fix javadocs / fix tests should be squashed.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeTest.java
```diff
@@ -0,0 +1,947 @@
...
+    }
+
+    /**
+     * Tests {@code put(String, String)}.
+     */
+    @Test
+    public void testPutStringString() {
+        new MemoryNode().put("http://www.loc.gov/mods/v3#name", "Wilhelm Busch");
```
henning-gerhardt: What are you testing here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/LangStringTest.java
```diff
@@ -0,0 +1,161 @@
...
+            LangString other = storage.createLangString("Hoc est corpus meum.", "la");
+
+            assertThat(other.hashCode(), is(equalTo(one.hashCode())));
+        }
+    }
+
+    @Test
+    public void testLangStringCanBeCreatedWithBothTextAndLanguageGiven() {
+        for (Storage storage : TestConfig.STORAGES_TO_TEST_AGAINST) {
+            storage.createLangString("In vino veritas est.", "la");
```
henning-gerhardt: What this kind of test? You are not testing anything.

Kitodo-DataAccess/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+    <name>Kitodo - Data Access</name>
+    <url>https://github.com/kitodo/kitodo-production</url>
+    <organization>
+        <name>kitodo.org</name>
+        <url>http://kitodo.org</url>
+    </organization>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.jena</groupId>
+            <artifactId>apache-jena-libs</artifactId>
```
henning-gerhardt: This artefact depends on `slf4j` but our logging framework is `log4j2`. `slf4j` must be excluded and the log4j2 bridge `log4j-slf4j-impl` must be added beside a minimal logging configuration.

Kitodo-DataAccess/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
```
henning-gerhardt: This dependency is not needed if you switch to `log4j-slf4j-impl`

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/RDF.java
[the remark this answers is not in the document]
matthias-ronge: This was what I originally wanted. Thank you!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/RDFS.java
matthias-ronge: same here

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Namespaces.java
matthias-ronge: fixed

Kitodo-DataAccess/pom.xml
matthias-ronge: slf4j cannot be excluded. Apache Jena needs it. For tests, the non-operational bridge has been included (which should be ok for tests. If the tests fail, it is quite sure a problem in our code, not in Jena). For other uses, the module that will depend on this module has to include the appropriate bridge.

Kitodo-DataAccess/pom.xml
matthias-ronge: This is only to prevent a warning when running tests . Note `<scope>test</scope>`. It does not target productive use.

Kitodo-DataAccess/pom.xml
henning-gerhardt: It can as we did this already to remove a second logging framework. Jena needs only a implementation of slf4j for its own logging. This implementation could be provided by the slf4j package (which is delivered automatically as a Maven dependency) or by `log4j-slf4j-impl` package.

Kitodo-DataAccess/pom.xml
henning-gerhardt: But even in test scope we are using log4j2 as logging framework. Introducing second logging framework is not a good idea and complicate all logging access.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/GraphPath.java
matthias-ronge: no, because I need a reference object to test the object the constructor created. The reference object is created in the test class like what the constructor should deliver. For the reason that equals() is true, the comparison object must be of class GraphPath() as well.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/AmbiguousDataExceptionTest.java
matthias-ronge: How else then to test an exception? An exception’s only use is to be thrown, isn’t it?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/LangStringTest.java
matthias-ronge: No, I am testing that the method `createLangString()` does **not** throw an exception when both parameters are in their range. Testing also means to test the working situation, not only the error cases. Assume the the parameter check for the second parameter would be wrong, like `if(language == null || !language.isEmpty()) throw new IllegalArgumentException("Language must not be null or empty");` then this function would throw an exception and the test would fail. (Correctly, language.isEmpty() must not be negated here.)

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/LinkedDataExceptionTest.java
matthias-ronge: Tests that AmbiguousDataException is a subclass of LinkedDataException, which must be the case.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/LinkedDataExceptionTest.java
matthias-ronge: Tests that NoDataException is a subclass of LinkedDataException, which must be the case. See also the test method names which describe what the test should be testing.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NoDataExceptionTest.java
matthias-ronge: Tests that NoDataException can be thrown (i.e., does not throw another exception by itself).

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
matthias-ronge: No assertion necessary. Test is to prove that the method does not throw an exception, what it shall do only if the string in the constant would not be an URL. (Tests that URLs pass the URI validation correctly, not giving false positives.)

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
matthias-ronge: No assertion necessary. Test is to prove that the method does not throw an exception, what it shall do only if the string in the constant would not be an URI. (Tests that URNs pass the URI validation correctly, not giving false positives.)

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNamedNodeTest.java
matthias-ronge: Tests that HTTP URLs are accepted as an URI, no exception being thrown.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNamedNodeTest.java
matthias-ronge: Tests that URNs are accepted as an URI, no exception being thrown.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
matthias-ronge: Tests that URNs are accepted as an URI, no exception being thrown.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
matthias-ronge: Tests that HTTP URLs are accepted as an URI, no exception being thrown.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NoDataExceptionTest.java
Beacze: But it is like testing if Java language works...

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
Beacze: Tests without assertions are useless.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
Beacze: https://stackoverflow.com/questions/17731234/how-to-test-that-no-exception-is-thrown

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/AmbiguousDataExceptionTest.java
Beacze: Actually it is not necessary to test it. We don't need 100% code coverage. Our tests should at first be useful.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryNode.java
matthias-ronge: The constructor is needed or you cannot create an empty object. The default constructor is not created if the class has other constructors, which is the case.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NoDataExceptionTest.java
matthias-ronge: This is true for all tests. I think it should not be that an exception constructor throws an exception by itself, and it would be a bug that is hard to find because the exception ‘almost’ works (it still breaks the code flow, just a corresponding `catch` will not be invoked.)

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
matthias-ronge: No, not at all. They pass or they fail. If the URI validation misbehaves and throws an exception here (which it must not), the test will fail, this is what the test is for.

matthias-ronge: Or do you mean I should put `assert true;` there?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
matthias-ronge: The test in class NodeTest invokes the Test in class GraphPathTest which makes use of the find method that I want to test.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
matthias-ronge: Tests that put method works (i.e. does not throw an exception). This seems trivial for the current implementation, but the test is written on the interface and if extended to a future implementation of the interface, maybe it is not that trivial.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
matthias-ronge: Test that put method which takes a String as second parameter works. (no exception being thrown)

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
matthias-ronge: Test that put method which takes a String as the first argument works, not throwing an exception.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
matthias-ronge: @Beacze , see comment to question what you posted: “A JUnit test is judged to have failed if it throws any exception other than an expected exception. Usually no exceptions are expected.” So this is perfectly fine.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeTest.java
matthias-ronge: The constructor. It initializes an object and does not throw an exception.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeTest.java
matthias-ronge: Test that put method which takes two string arguments works. (no exception being thrown)

matthias-ronge:
Now where I changed logging framework tests fail because they want to create a log directory that does not exist on travis. What is the solution here?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/LangStringTest.java
Beacze: But this method can't throw any exception (except RuntimeException) as its declaration doesn't contain any.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
matthias-ronge: Tests that put method with two strings as arguments does not throw exception

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeTest.java
matthias-ronge: tests that put method with string and object type arguments works and do not throw an exception

Kitodo-DataAccess/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+            <version>1.4-atlassian-1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
```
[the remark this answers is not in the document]
matthias-ronge: This import looks exactly the same as in the other pom.xml files. What is wrong with it?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
Beacze: I'm a bit confused. Do you mean that this test once passes and once not?

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeTest.java
Beacze: Is it not violation of rule that tests should be independent of each other?

Kitodo-DataAccess/pom.xml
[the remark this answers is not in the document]
henning-gerhardt: In current master we are using a Maven property to set the version of used Junit. Your master / branch is 190 commits behind, so you did not see all the changes which are done in between. Maybe you should rebase your branch against current master.

henning-gerhardt:
In commit https://github.com/kitodo/kitodo-production/pull/1056/commits/8db55ac136b1841a9b450413ca756396c00db5f9 you added a log4j2 configuration file which try to write a log file which is failing as permission is denied. Other modules using a [log4j2 configuration file](https://github.com/kitodo/kitodo-production/blob/master/Kitodo-DataManagement/src/test/resources/log4j2.xml) which did log only to console.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
Beacze: Actually I was more meaning that we could use for .e.g assertNull and check if the exception is null, as test case which is here is very misleading.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/NodeReferenceTest.java
matthias-ronge: The test passes if it does not throw an exception. This test is part of testing the parameter scope of createNodeReference. It must accept a valid URI (and must not throw an exception).

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/storage/memory/MemoryNodeReferenceTest.java
matthias-ronge: What you mean would be like this:

    Throwable t = null;
    try {
        new MemoryNodeReference("urn:example:kitodo-123456789-xyz");
    } catch (Throwable t1) {
        t = t1;
    }
    assertThat(t, is(nullValue()));
But actually this is overkill. It is very OK to have tests without exceptions if the only condition of the test is that a method does not throw an exception. It is the default behaviour of the test framework.

matthias-ronge:
@henning-gerhardt Is there anything still missing from your perspective?

henning-gerhardt:
Your way to write tests is total different from my understanding to write tests and how tests are handled. This could not solved in a short time so I will accept your tests even as I don't like them but this kind of tests are better then no tests at all.

Last two issues with this pull requests:
- the amount of your commits: most of them could be squashed as they only fix stuff and did not show the way how the code is developed
- the created module is not installed inside the `Kitodo/module` directory as like the other non-core modules nor get installed in the path defined in the `kitodo_config.properties` file. For solving this you can look into the other modules pom file like `Kitodo-FileManagement` or `Kitodo-Validation`.

matthias-ronge:
_About the tests:_ Testing cannot prove the software is correct. However, it can show the absence of a wide class of errors. There is not the golden approach. There are several approaches, [black box and white box tests](https://de.wikipedia.org/wiki/Black-Box-Test) mainly. I decided for black box tests with a (hopefully) complete coverage of valid and invalid parameter sets. I decided for black box tests mainly because most of the code is interface, and (future) implementation is not yet known. If we add a `Storage` like MySQL or Fedora later, the tests are already there. It is not said that black box tests is the best approach, but white box testing isn’t necessarily, either (and must be further distinguished. See [Kontrollflussorientierte Testverfahren](https://de.wikipedia.org/wiki/Kontrollflussorientierte_Testverfahren) for more details).

_About the commits:_ I prefer not to squash them, because on each line of the code the last commit will be annotated. (Use the “Blame” button to see what I mean.) Now, if I add some feature, let’s say, metadata reading, and do a commit. Then I edit a different file to fix a typo in it. If I squash those, the file that I fixed a typo in will be labelled as “reading meta-data” which does not relate to it at all. This is why I do small commits. For Git, it’s meaningless. Other extreme would be to squash all commits in the merge, because from the view of the project history, it is one contribution.

I will review the history. Consecutive things that relate to the same “topic” (such as various fixes of checkstyle) can definitely be squashed. I don’t know for sure, but I believe these changes cannot be undone, this is why I am carfuel about them.

matthias-ronge:
Now all things should be fixed

PR #1061, opened 2017-10-17
Created new image visualization and pagination for metadata editor

tim-boerner:


Kitodo/src/main/java/de/sub/goobi/converter/URIConverter.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.faces.convert.ConverterException;
+import javax.faces.convert.FacesConverter;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+@FacesConverter("URIConverter")
+public class URIConverter implements Converter {
+    public static final String CONVERTER_ID = "URIConverter";
+    private static final Logger logger = LogManager.getLogger(DocketConverter.class);
```
henning-gerhardt: Wrong name for logging instance: please change it from `DocketConverter.class` to `URIConverter.class`

PR #1063, opened 2017-10-17
Prevent thread leakage

solth:
Reuse unique instances of SearchRestClient and IndexRestClient to solve the problem of leaking threads.


Kathrin-Huber:
What a beautiful singleton ;)

henning-gerhardt:
You are lowering the amount of check style violations (36 on my system) then you can even lower the value of maximum allowed violations (87) in the according pom file.

PR #1060, opened 2017-10-17
Frontend module

Kathrin-Huber:
Making this Pull Request in the name of Julien

PR #1065, opened 2017-10-19
Frontend module

solth:
Pull request incorporating change requests from https://github.com/kitodo/kitodo-production/pull/1060

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
+++ b/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
@@ -62,7 +62,7 @@ public class KitodoServiceLoader<T> {
      * @return A module with type T.
      */
     @SuppressWarnings("unchecked")
-    public T loadModule() throws IOException {
+    public T loadModule() {
 
         loadModulesIntoClasspath();
         loadFrontendFilesIntoCore();
@@ -77,7 +77,7 @@ public class KitodoServiceLoader<T> {
      * of the core module. Before copying, existing frontend files of the same module will be deleted from the
      * core module. Afterwards the created temporary folder will be deleted as well.
      */
-    private void loadFrontendFilesIntoCore() throws IOException {
+    private void loadFrontendFilesIntoCore() {
         Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
 
         try {
@@ -99,7 +99,7 @@ public class KitodoServiceLoader<T> {
             }
         }
         catch (Exception e) {
-            throw e;
+            logger.error("Classpath could not be accessed", e.getMessage());
         }
     }
 
--- a/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
+++ b/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
@@ -13,6 +13,8 @@ package org.kitodo.serviceloader;
 
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
@@ -23,12 +25,12 @@ import java.nio.file.DirectoryStream;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.Properties;
-import java.util.Random;
 import java.util.ServiceLoader;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
@@ -43,6 +45,14 @@ import org.apache.logging.log4j.Logger;
 public class KitodoServiceLoader<T> {
     private Class clazz;
     private String modulePath;
+
+    private static final String POM_PROPERTIES_FILE = "pom.properties";
+    private static final String ARTIFACT_ID_PROPERTY = "artifactId";
+    private static final String TEMP_DIR_PREFIX = "kitodo_";
+    private static final String META_INF_FOLDER = "META-INF";
+    private static final String RESOURCES_FOLDER = "resources";
+    private static final String PAGES_FOLDER = "pages";
+
     private static final Logger logger = LogManager.getLogger(KitodoServiceLoader.class);
 
     /**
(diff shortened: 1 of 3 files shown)


Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -52,20 +74,242 @@ public KitodoServiceLoader(Class clazz, String modulePath) {
...
 
     /**
+     * Loads bean classes and registers them to the FacesContext. Afterwards they can be used in all
+     * frontend files
+     */
+    private void loadBeans() {
+        try {
+
+            Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+            DirectoryStream<Path> stream = Files.newDirectoryStream(moduleFolder, "*.jar");
```
henning-gerhardt: Class `DirectoryStream` implements the `Closeable` / `Autocloseable` interface which means that we can use the try-with-resources-statement here. This stream object is not closed in current implementation which is an other good reason to change it.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -52,20 +74,242 @@ public KitodoServiceLoader(Class clazz, String modulePath) {
...
+    /**
+     * If the found jar files have frontend files, they will be extracted and copied into the frontend folder
+     * of the core module. Before copying, existing frontend files of the same module will be deleted from the
+     * core module. Afterwards the created temporary folder will be deleted as well.
+     */
+    private void loadFrontendFilesIntoCore() {
+        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+
+        try {
+            DirectoryStream<Path> stream = Files.newDirectoryStream(moduleFolder, "*.jar");
```
henning-gerhardt: Class `DirectoryStream` implements the `Closeable` / `Autocloseable` interface which means that we can use the try-with-resources-statement here. This stream object is not closed in current implementation which is an other good reason to change it.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -52,20 +74,242 @@ public KitodoServiceLoader(Class clazz, String modulePath) {
...
+     * frontend files
+     */
+    private void loadBeans() {
+        try {
+
+            Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+            DirectoryStream<Path> stream = Files.newDirectoryStream(moduleFolder, "*.jar");
+
+            for (Path f : stream) {
+                JarFile jarFile = new JarFile(f.toString());
```
henning-gerhardt: Class `JarFile` implements the `Closeable` / `Autocloseable` interface which means that we can use the try-with-resources-statement here. This stream object is not closed in current implementation which is an other good reason to change it.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -52,20 +74,242 @@ public KitodoServiceLoader(Class clazz, String modulePath) {
...
+    private void loadFrontendFilesIntoCore() {
+        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+
+        try {
+            DirectoryStream<Path> stream = Files.newDirectoryStream(moduleFolder, "*.jar");
+
+            for (Path f : stream) {
+
+                File loc = new File(f.toString());
+                JarFile jarFile = new JarFile(loc);
```
henning-gerhardt: Class `DirectoryStream` implements the `Closeable` / `Autocloseable` interface which means that we can use the try-with-resources-statement here. This stream object is not closed in current implementation which is an other good reason to change it.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -52,20 +74,242 @@ public KitodoServiceLoader(Class clazz, String modulePath) {
...
+                        String resourceFolder = String.join(File.separator,
+                                Arrays.asList(temporaryFolderName, META_INF_FOLDER, RESOURCES_FOLDER));
+                        copyFrontEndFiles(resourceFolder, filePath);
+                        FileUtils.deleteDirectory(tempFolderFile);
+                    } else {
+                        logger.info("No module found in JarFile '" + jarFile.getName() + "'.");
+                    }
+                }
+            }
+        }
```
henning-gerhardt: Formatting issue: next line should be in this line.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -52,20 +74,242 @@ public KitodoServiceLoader(Class clazz, String modulePath) {
...
+     *            destination path, where the frontend files will be extracted to
+     *
+     */
+    private void extractFrontEndFiles(String jarPath, String destinationPath) throws IOException {
+        File destinationFolder = new File(destinationPath);
+        if (!destinationFolder.exists()) {
+            destinationFolder.mkdir();
+        }
+
+        JarFile jar = new JarFile(jarPath);
```
henning-gerhardt: Class `DirectoryStream` implements the `Closeable` / `Autocloseable` interface which means that we can use the try-with-resources-statement here and the later jar.close() could be removed.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -44,28 +69,250 @@
...
+    /**
+     * Extracts the module name of the current module by finding the pom.properties
+     * in the given temporary folder
+     *
+     * @param temporaryFolder
+     *            folder in which the pom.properties file will be searched for
+     *
+     * @return boolean
+     */
+    private String extractModuleName(File temporaryFolder) throws IOException {
```
henning-gerhardt: Return value of method is String but JavaDoc claims boolean. I think that JavaDoc needs an update.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -44,28 +69,250 @@
...
+
+    /**
+     * Tries to find a file by a given file name in a folder by folder name.
+     *
+     * @param name
+     *            file name that will be searched for
+     * @param folder
+     *            folder that will be searched
+     *
+     * @return File, null
```
henning-gerhardt: `null` is not correct as in case file is not found an exception is thrown.

PR #1071, opened 2017-10-23
Additional data in BatchType, ProjectType and ProcessType

Beacze:
Add additional data to store in index. Needed for decrease amount of queries to get exact objects.


Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java
```diff
@@ -406,7 +407,11 @@ public Double findSumForSortHelperImages(Integer projectId) throws DataException
     }
 
     private QueryBuilder getQueryProjectId(Integer id) {
-        return createSimpleQuery("project", id, true);
+        return createSimpleQuery("project.id", id, true);
+    }
+
+    public QueryBuilder getQueryProjectTitle(String title) {
```
henning-gerhardt: New public method without JavaDoc.

PR #1076, opened 2017-10-24
Lazy data tables

solth:
Replace the old Tomahawk datatables with new Primefaces datatables. The Primefaces tables use a lazy data model that only load the number of datasets that is displayed on the current page.

**Note**: in order to work effectively, the changes in https://github.com/kitodo/kitodo-production/pull/1063 are required.

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/BasisForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/BasisForm.java
@@ -13,6 +13,7 @@ package de.sub.goobi.forms;
 
 import de.sub.goobi.helper.Helper;
 import de.sub.goobi.helper.Page;
+import de.sub.goobi.model.LazyDTOModel;
 
 import java.io.Serializable;
 import java.util.List;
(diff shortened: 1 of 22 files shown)


Kitodo/src/main/java/de/sub/goobi/forms/BasisForm.java
```diff
@@ -32,6 +33,16 @@
     protected User user;
     protected String sortierung = "prozessAsc";
 
+    private LazyDTOModel lazyDTOModel = null;
+
+    public LazyDTOModel getLazyDTOModel() {
```
henning-gerhardt: New public method without JavaDoc.

Kitodo/src/main/java/de/sub/goobi/forms/BasisForm.java
```diff
@@ -32,6 +33,16 @@
...
+        return lazyDTOModel;
+    }
+
+    public void setLazyDTOModel(LazyDTOModel lazyDTOModel) {
```
henning-gerhardt: New public method without JavaDoc.

PR #1095, opened 2017-11-02
Remove Criteria from ProjectHelper

Beacze:
Criteria are deprecated in current used version of Hibernate.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
@@ -123,6 +123,24 @@ public abstract class BaseDAO<T extends BaseBean> implements Serializable {
         return (List<T>) session.createQuery(query).list();
     }
 
+    @SuppressWarnings("unchecked")
+    List<Double> getAverage(String field, String query) {
+        Session session = HibernateHelper.getHibernateSession();
+        return (List<Double>) session.createQuery("SELECT AVG(" + field + ") " + query).list();
+    }
+
+    @SuppressWarnings("unchecked")
+    List<Long> getCount(String field, String query) {
+        Session session = HibernateHelper.getHibernateSession();
+        return (List<Long>) session.createQuery("SELECT COUNT(" + field + ") " + query).list();
+    }
+
+    @SuppressWarnings("unchecked")
+    List<Long> getSum(String field, String query) {
+        Session session = HibernateHelper.getHibernateSession();
+        return (List<Long>) session.createQuery("SELECT SUM(" + field + ") " + query).list();
+    }
+
     /**
      * Count all rows in database.
      *
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/TaskDAO.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/TaskDAO.java
@@ -125,4 +125,94 @@ public class TaskDAO extends BaseDAO<Task> {
         return getByQuery("FROM Task WHERE process_id = " + processId + " AND ordering < " + ordering
                 + " ORDER BY ordering DESC");
     }
+
+    /**
+     * Get tasks for non template processes for given project id and ordered by
+     * ordering.
+     *
+     * @param projectId
+     *            as Integer
+     * @return list of tasks
+     */
+    public List<Task> getTasksForNotTemplateProcessesForProjectIdOrderByOrdering(Integer projectId) {
+        return getByQuery(
+                "SELECT t FROM Task AS t INNER JOIN t.process AS p WITH p.template = 0 INNER JOIN p.project AS pr "
+                        + "WITH pr.id = " + projectId + " GROUP BY t.title ORDER BY t.ordering");
+    }
+
+    /**
+     * Get size of tasks for non template processes for given project id and ordered
+     * by ordering.
+     * 
+     * @param projectId
+     *            as Integer
+     * @return list of tasks
+     */
+    public List<Long> getSizeOfTasksForNotTemplateProcessesForProjectIdOrderByOrdering(Integer projectId) {
+        return getCount("t.id",
+                "FROM Task AS t INNER JOIN t.process AS p WITH p.template = 0 INNER JOIN p.project AS pr "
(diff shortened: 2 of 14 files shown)


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
```diff
@@ -123,6 +123,24 @@ public void remove(T baseBean) throws DAOException {
...
     }
 
+    @SuppressWarnings("unchecked")
+    List<Double> getAverage(String field, String query) {
+        Session session = HibernateHelper.getHibernateSession();
+        return (List<Double>) session.createQuery("SELECT AVG(" + field + ") " + query).list();
+    }
+
+    @SuppressWarnings("unchecked")
+    List<Long> getCount(String field, String query) {
```
henning-gerhardt: How can a SQL count() method return more then one value?

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/TaskDAO.java
```diff
@@ -125,4 +125,94 @@ public Task load(int id) throws DAOException {
...
+
+    /**
+     * Get size of tasks for non template processes for given project id and ordered
+     * by ordering.
+     * 
+     * @param projectId
+     *            as Integer
+     * @return list of tasks
+     */
+    public List<Long> getSizeOfTasksForNotTemplateProcessesForProjectIdOrderByOrdering(Integer projectId) {
```
henning-gerhardt: Counting objects should never be influenced by ordering and I think that this method is superfluous in current usage. See other comments on its usage.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/TaskDAO.java
```diff
@@ -125,4 +125,94 @@ public Task load(int id) throws DAOException {
...
+
+    /**
+     * Get size of tasks for exact processing status for non template processes for
+     * given project id and ordered by ordering.
+     *
+     * @param projectId
+     *            as Integer
+     * @return list of tasks sizes
+     */
+    public List<Long> getSizeOfTasksWithProcessingStatusForNotTemplateProcessesForProjectIdOrderByOrdering(
```
henning-gerhardt: Counting objects should never be influenced by ordering and I think that this method is superfluous in current usage. See other comments on its usage.

Kitodo/src/main/java/de/sub/goobi/helper/ProjectHelper.java
```diff
@@ -39,157 +34,92 @@
...
+                .findAmountOfImagesForNotTemplatesAndProjectId(project.getId());
 
-        proList.add(Projections.groupProperty(("title")));
-        proList.add(Projections.count("id"));
-        proList.add(Projections.avg("ordering"));
-
-        critSteps.setProjection(proList);
+        List<Task> tasks = serviceManager.getTaskService()
+                .getTasksForProjectHelper(project.getId());
+        List<Long> tasksSize = serviceManager.getTaskService()
```
henning-gerhardt: The amount of `tasks` and the number in `taskSize` is only equal if there no data inserted between both calls. Value of `taskSize` should be gathered from  `tasks.size()` call instead of quering the database again.

Kitodo/src/main/java/de/sub/goobi/helper/ProjectHelper.java
```diff
@@ -39,157 +34,92 @@
...
-
-            Object[] row = (Object[]) obj;
-
-            title = (String) (row[FieldList.stepName.fieldLocation]);
-            numberOfSteps = (Long) (row[FieldList.stepCount.fieldLocation]);
-            numberOfImages = (Long) (row[FieldList.imageCount.fieldLocation]);
+        List<Task> doneTasks = serviceManager.getTaskService()
+                .getTasksWithProcessingStatusForProjectHelper(
+                        TaskStatus.DONE.getValue(), project.getId());
+        List<Long> doneTasksSize = serviceManager.getTaskService()
```
henning-gerhardt: The amount of `doneTasks` and the number in `doneTaskSize` is only equal if there no data inserted between both calls. Value of `doneTaskSize` should be gathered from  `doneTasks.size()` call instead of quering the database again.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
Beacze: There are possibly many values with usage of Group By clausule.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/TaskDAO.java
Beacze: Original logic which was transferred was using ordering.

Kitodo/src/main/java/de/sub/goobi/helper/ProjectHelper.java
Beacze: IT is not exactly the same value. It counts ids of tasks inside Group By clausule.

Kitodo/src/main/java/de/sub/goobi/helper/ProjectHelper.java
```diff
@@ -39,157 +34,92 @@
...
-            ;
-        }
-
-        proList = null;
-        list = null;
-
-        Criteria critSteps = session.createCriteria(Task.class);
-
-        critSteps.createCriteria("process", "proc");
-        critSteps.addOrder(Order.asc("ordering"));
```
Beacze: Here is from comes ordering.

henning-gerhardt:
I did not realised that this changes are needed for our good statistic part. When our statistic needs this in this way, then I will accept it.

henning-gerhardt:
There are merge conflicts  in `Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java` and `Kitodo/src/main/java/org/kitodo/services/data/ProjectService.java`. Please resolve them.

henning-gerhardt:
There is a merge conflict in file `Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java`. Please resolve this conflict.

PR #1121, opened 2017-11-10
Remove accidental quotes

matthias-ronge:
These quotes definitely don’t belong there.


henning-gerhardt:
Thank you for discovering this small issue. As I see this is even existing in 2.x branch. Can you open a different / separate pull request against 2.x too or should I do this?

PR #1126, opened 2017-11-13
Fix NPE on exportMETS

Beacze:
Depends on #1125

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/export/download/ExportMets.java
+++ b/Kitodo/src/main/java/de/sub/goobi/export/download/ExportMets.java
@@ -26,7 +26,6 @@ import java.net.URI;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/java/de/sub/goobi/export/download/ExportMets.java
```diff
@@ -150,8 +150,13 @@ protected boolean writeMetsFile(Process process, URI metaFile, Fileformat gdzfil
         MetsModsImportExport mm = new MetsModsImportExport(this.myPrefs);
         mm.setWriteLocal(writeLocalFileGroup);
         mm = serviceManager.getSchemaService().tempConvert(gdzfile, this, mm, this.myPrefs, process);
-        mm.write(metaFile.toString());
-        Helper.setMeldung(null, process.getTitle() + ": ", "ExportFinished");
-        return true;
+        User user = (User) Helper.getManagedBeanValue("#{LoginForm.myBenutzer}");
```
henning-gerhardt: This should go into `ProzessverwaltungForm` class and there into the methods `exportMets`, `exportPdf` and `exportDMS`.

PR #1133, opened 2017-11-16
Fixing processlist navigation

chrizzor:
I've replaced the links with commandlink, otherwise the bean method doesn't get called.

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
@@ -178,10 +178,11 @@ public class ProzessverwaltungForm extends BasisForm {
     /**
      * Create new process.
      */
-    private void newProcess() {
+    public String newProcess() {
         this.process = new Process();
         this.newProcessTitle = "";
         this.modusBearbeiten = "prozess";
+        return "/pages/ProzessverwaltungBearbeiten";
     }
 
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/webapp/pages/inc_Prozessverwaltung/Prozesse_Liste.xhtml
```diff
@@ -36,9 +36,11 @@
     </h4>
 
     <!-- new process -->
-    <h:link outcome="/pages/ProzessverwaltungBearbeiten"
-            value="#{msgs.einenNeuenProzessAnlegen}"
-            rendered="#{(LoginForm.maximaleBerechtigung == 1 || LoginForm.maximaleBerechtigung == 2) and ProzessverwaltungForm.modusAnzeige!='vorlagen'}"/>
+    <h:commandLink action="#{ProzessverwaltungForm.newProcess}" immediate="true"
```
Beacze: PR https://github.com/kitodo/kitodo-production/pull/1142 removes this link, as it was decided that this method of task insert is incorrect. So this part can be ignored.

henning-gerhardt: I would go with @Beacze comment: do not change it now as this command link will be removed in the next pull request (or Beatrycze must adjust their pull request after this one is merged).

PR #1136, opened 2017-11-23
Remove module's attributes from Task bean and module manager

Beacze:


Kitodo/src/main/java/de/sub/goobi/helper/GoobiScript.java
```diff
@@ -528,7 +528,6 @@ private void addModuleToStep(List<Process> inProzesse) {
             if (proz.getTasks() != null) {
                 for (Task task : proz.getTasks()) {
                     if (task.getTitle().equals(this.myParameters.get("steptitle"))) {
-                        task.setTypeModuleName(this.myParameters.get("module"));
```
henning-gerhardt: This whole method is useless without this line. You should remove this method and any calls to this method.

PR #1138, opened 2017-11-27
Remove double automatic attribute of Task bean

Beacze:
There are two indicators in the code if Task is automatic:

- typeAutomatic 
- editTaskType - AUTOMATIC

This PR merges both of those properties in one.

Depends on #1136


henning-gerhardt:
This two attributes could not be merged as one attribute is necessary how to handle the step inside the workflow engine and the other one is used how the step is handled by execution.

PR #1155, opened 2017-12-08
Tmp/error in application individual header

BFallert:


henning-gerhardt:
There is already a pull request #1154  which fix this error. I'm closing this pull request. But thank you for your contribution.

PR #1156, opened 2017-12-11
Transfer necessary configuration step

BFallert:
Drop foreign key in front of drop key.

### DIFF
--- a/Goobi/setup/migrate_index/remove_old_foreign_keys_1.11.2.sql
+++ b/Goobi/setup/migrate_index/remove_old_foreign_keys_1.11.2.sql
@@ -1,0 +1,59 @@
+--
+-- Remove old FOREIGN KEY of Goobi.Production CE 1.11.2
+--
+--
+ALTER TABLE `batchesprozesse`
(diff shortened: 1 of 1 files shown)


BFallert:
Without this step I get only error message when removing the keys

henning-gerhardt:
How did you create your database for Kitodo.Production? 

So far as I know foreign keys are only created if your database is created by hibernate instead of using the SQL files. So standard migration could not recognize this foreign keys as they are created "outside" and with non determinable names.

BFallert:
The Database was created july 2014 when we introduced Goobi. It was done by a external supporter (@frank-ulrich-weber).

stweil:
Maybe we need a more sophisticated upgrade script then if the table names are random, non deterministic?

henning-gerhardt:
Generated names for index or foreign keys are not random generated but depending on used Hibernate version and some other information. It was difficult to get possible values for the index key names but it was not possible for me for the foreign key names as we (and other I asked in this time) did not have a Kitodo.Production database with foreign key relations.

PR #1160, opened 2017-12-14
Fix dead code issue in Active MQ connector

matthias-ronge:
This fix will put code that has become dead code back to work. I found this little bug due to a dead code warning. Maybe the underlying library changed slightly during a version upgrade.


henning-gerhardt:
Is the 2.x branch affected too as in both branches ActiveMQ version is 5.5.1.

matthias-ronge:
True. But it is too minor to fix. It only deals with the error case ‘closing a step without passing an ID’.

henning-gerhardt:
Can you provide a fix against `2.x` branch or should I do this? It is even for `2.x` branch important to fix even so small errors as it is currently used.

PR #1169, opened 2018-01-02
Removed filterKein from BenutzergruppenForm

oliver-stoehr:
Since we're using PrimeFaces lazy datatables filterKein and the Page object are not used anymore.


oliver-stoehr:
The title was misleading: The page class itself was not removed, but the _the use of the page class_ was removed from BenutzergruppenForm.

henning-gerhardt:
Ok, now it should be fine.

solth:
Now it has the same name as pull request #1168 😆

PR #1170, opened 2018-01-02
Fix user and usergroup templates

solth:


henning-gerhardt:
Please avoid to add changes after a pull request is reviewed and accepted for merge. Thank you.

PR #1177, opened 2018-01-03
Remove not thrown exceptions

Beacze:


Kitodo/src/main/java/de/sub/goobi/helper/servletfilter/RequestControlFilter.java
```diff
@@ -73,7 +73,7 @@
      */
     @Override
     @SuppressWarnings({"unchecked", "rawtypes" })
-    public void init(FilterConfig config) throws ServletException {
+    public void init(FilterConfig config) {
```
henning-gerhardt: Is this correct removement as there is a suppressed warnings annotation?

Beacze: Those warnings are not suppressing exception. If this exception would be necessary there IDE would inform about it as an error, not warning.

henning-gerhardt: Ok, then is this change ok.

PR #1182, opened 2018-01-04
Multiple authorizations

solth:
Replaced simple pull down menu with PrimeFaces "PickList" in `usergroupEdit` template. This allows adding multiple authorization levels to individual user groups and - in the future - saving them to the user group. 

**Note**: Saving the new authorizations to user groups is not implemented in the backend, yet!


Kitodo/src/main/java/de/sub/goobi/forms/BenutzergruppenForm.java
```diff
@@ -136,6 +139,41 @@ public int getUserGroupId() {
...
+    /**
+     * Assign the target property of given DualListModel of authorizations to
+     * 'myBenutzergruppe' using a PrimeFaces PickList object.
+     *
+     * @param authorizations
+     *            list of authorizations assigned to 'myBenutzergruppe'
+     */
+    public void setAuthorizations(DualListModel<Authorization> authorizations) {
+        try {
+            List<Authorization> assignedAuthorizations = new ArrayList<>();
```
Beacze: This method probably can be simplified to sometgin like that:
`public void setAuthorizations(DualListModel<Authorization> authorizations) {
        List<Authorization> assignedAuthorizations = new ArrayList<>();
        for (Authorization object : authorizations.getTarget()) {
            assignedAuthorizations.add(object); 
        }
        this.myBenutzergruppe.setAuthorizations(assignedAuthorizations);
    }`

solth: No, unfortunately, I think that change is not possible. `getTarget` returns just the values of the list items in the PickLists "target" list. These values are just the id's of the authorization objects, not the whole authorization objects themselves.

Kitodo/src/main/java/de/sub/goobi/forms/BenutzergruppenForm.java
```diff
@@ -136,6 +139,41 @@ public int getUserGroupId() {
...
+            for (Object object : authorizations.getTarget()) {
+                assignedAuthorizations
+                        .add(serviceManager.getAuthorizationService().getById(Integer.parseInt((String) object)));
```
henning-gerhardt: This looks ugly to me as you converting a list entry from `Object` to `String` and then you hope that this could be used as an `Integer` value. Is there no other way for converting and retrieving the necessary data? As there are no test which are proving your assumption made me the decision not easy.

PR #1194, opened 2018-01-10
Add 'docket edit' page in new layout

solth:


Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -1133,6 +1134,7 @@ wrongLogin=Fehler bei Authentifizierung
 year=Jahr
 years=Jahre
 yes=Ja
+youWillBeRedirected=Sie werden nach erfolgreichem Speichern weitergeleitet.
```
henning-gerhardt: This would be the fourth redirection message. Other three are `ldapgroupYouWillBeRedirected`, `projectYouWillBeRedirected` and `userYouWillBeRedirected`. Maybe you should add an prefix like `docket` or this messages should be unique used.

Kitodo/src/main/java/de/sub/goobi/forms/DocketForm.java
```diff
@@ -144,4 +145,17 @@ public int getDocketId() {
...
     }
+
+    // TODO:
+    // replace calls to this function with "/pages/projects" once we have completely
+    // switched to the new frontend pages
+    private String redirectToList(String urlSuffix) {
+        String referrer = FacesContext.getCurrentInstance().getExternalContext().getRequestHeaderMap().get("referer");
+        String callerViewId = referrer.substring(referrer.lastIndexOf("/") + 1);
+        if (!callerViewId.isEmpty() && callerViewId.contains("editDocket.jsf")) {
+            return "/pages/projects" + urlSuffix;
```
henning-gerhardt: Why are you returning to `/pages/projects`? Maybe a wrong copy and past?

solth: No, this path is correct, since "dockets" are now located in an additional tab of the "projects" page instead of having their own, dedicated page.

Kitodo/src/main/resources/messages/messages_de.properties
solth: There will even be a fifths redirect with the identical text in the upcoming pull request for the ruleset editing page by @oliver-stoehr . Since he doesn't have the neutral "youWillBeRedirected" message _without_ object type that I added here, we agreed to replace all object specific messages with this general one and remove the obsolete object specific messages after both PRs were merged.

henning-gerhardt: Then should @oliver-stoehr not add a fifth redirect message and change it this neutral message key. Afterwards this messages should be unified.

PR #1196, opened 2018-01-11
Show background animation only on first visit

solth:
The "slide-in" animation of the new layouts background graphic has been deemed obtrusive because it is shown for every page the user opens.
This change restricts the animation to the first page in new layout opened by the user per session. All following pages will show the background image without animation.

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java
@@ -56,6 +56,7 @@ public class LoginForm implements Serializable {
     private SecurityPasswordEncoder passwordEncoder = new SecurityPasswordEncoder();
     private transient ServiceManager serviceManager = new ServiceManager();
     private static final Logger logger = LogManager.getLogger(LoginForm.class);
+    private boolean firstVisit = true;
 
     /**
      * Log out.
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java
```diff
@@ -390,4 +391,12 @@ public static URI getCurrentUserHomeDir() throws IOException {
         }
         return result;
     }
+
+    public boolean isFirstVisit() {
```
henning-gerhardt: Public method without JavaDoc. Please add JavaDoc!

Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
```diff
@@ -23,9 +23,12 @@ body {
     line-height: 18px;
     background: url("#{resource['/images/background-top.svg']}") no-repeat #f3f3f3 center top fixed;
     background-size: auto 300px;
+    }
+
+body.first-visit {
     animation-duration: 1s;
     animation-name: slidein;
-    }
+}
```
henning-gerhardt: Is this indent correct? In whole file the closing bracket is indent by 4 spaces.

PR #1202, opened 2018-01-12
KIT-648 (1/3)

matthias-ronge:
- Adds the UGH interface description as separate Maven module


henning-gerhardt:
Please use a proper pull request name and no internal ticket number.

PR #1205, opened 2018-01-12
Remove FolderInformation class and replace usage of it with FileService

Beacze:


Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java
```diff
@@ -2232,6 +2227,28 @@ protected boolean writeMetsFile(Process process, String targetFileName, Fileform
...
+     * @return String
+     */
+    public List<URI> getDataFiles(Process process) throws InvalidImagesException {
+        URI dir;
+        try {
+            dir = getImagesTifDirectory(true, process);
+        } catch (Exception e) {
+            throw new InvalidImagesException(e);
+        }
+        /* Verzeichnis einlesen */
```
henning-gerhardt: Please use English comment and use `//` for single line comments.

PR #1208, opened 2018-01-15
Encapsulate UGH in Interface (1/3): Add Interface

matthias-ronge:
Adds an interface for the UGH classes


henning-gerhardt:
You introduced / added to many check style errors. Please correct them.

PR #1219, opened 2018-01-17
Drop down buttons for creating new elements

IkramMaalej:


Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -1152,9 +1152,9 @@ zurueckZuArbeitsschritt=Zur\u00FCck zu Aufgabe
 zurueckZurStartseite=Zur\u00FCck zur Startseite
 zusaetzlicheDetails=Zus\u00E4tzliche Details
 zusammenfassung=Zusammenfassung
-Create=Anlegen
-neuenRegelsatz=Neuen Regelsatz
-neuenProzess=Neuen Vorgang
-neuenLaufzettel=Neuen Laufzettel
+neu=Neu
```
Beacze: It would be nice if you could use English name of properties.

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -1152,3 +1152,9 @@ zurueckZuArbeitsschritt=Zur\u00FCck zu Aufgabe
 zurueckZurStartseite=Zur\u00FCck zur Startseite
 zusaetzlicheDetails=Zus\u00E4tzliche Details
 zusammenfassung=Zusammenfassung
+neu=Neu
```
henning-gerhardt: Why you are using a German message key? New keys should be in English.

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -1159,3 +1159,9 @@ zurueckZuArbeitsschritt=Back to task
 zurueckZurStartseite=Back to start page
 zusaetzlicheDetails=Additional details
 zusammenfassung=Summary
+neu=New
```
henning-gerhardt: Why you are using a German message key? New keys should be in English.

Kitodo/src/main/resources/messages/messages_de.properties
IkramMaalej: Because using "new" as a key is not possible that's can be interpreted as the java new Object().
that's why also I have modified it to newElement.

PR #1224, opened 2018-01-18
Add XML reader and writer

matthias-ronge:


Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/AttrIterable.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    private final int length;
+
+    /**
+     * Creates a class to iterate over the attributes of a DOM element.
+     *
+     * @param element
+     *            element over whose attributes shall be iterated
+     */
+    AttrIterable(NamedNodeMap attributes) {
```
henning-gerhardt: JavaDoc and method signature did not match. Please update the JavaDoc!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Returns whether white space must be preserved to correctly represent this
+     * string in XML.
+     *
+     * @param s
+     *            string to examine
+     * @return if preservation of white space is necessary
+     */
+    public static boolean requiresPreservation(String s) {
```
henning-gerhardt: Short method parameter names should be avoided. Please use a longer and understandable name.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Creates the enumeration’s elements. Populates their private fields.
+     *
+     * @param enumeratedValue
+     *            String value as by specification
+     * @param toPreserve
+     *            whether the white-space is to preserve
+     */
+    private Space(String enumeratedValue) {
```
henning-gerhardt: JavaDoc and method signature did not match. Please update the JavaDoc!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,561 @@
...
+     * @param lang
+     *            language set on some parent node, if a RDF.LANG_STRING is to
+     *            be created but there is no local language tag
+     * @param space
+     *            whether or not to trim white space around the element
+     * @param documentNS
+     *            the namespace of the document
+     * @return the literal node
+     */
+    private static ObjectType parseLiteralElement(Element element, NodeReference type, String lang, Space space,
```
henning-gerhardt: JavaDoc and method signature did not match. Please update the JavaDoc!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,561 @@
...
+     * @param lang
+     *            inherited xml:lang attribute value
+     * @param space
+     *            whether to preserve white space (defaults to false)
+     * @param documentNS
+     *            the address of the document, as fall-back namespace for local
+     *            XML elements and attributes
+     * @return the populated node
+     */
+    private static Node parseNodeElement(Element element, String lang, Space space, String documentNS, boolean root,
```
henning-gerhardt: JavaDoc and method signature did not match. Please update the JavaDoc!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,561 @@
...
+     * @param lang
+     *            inherited xml:lang attribute value
+     * @param space
+     *            whether to preserve white space
+     * @param documentNS
+     *            the address of the document, as fall-back namespace for local
+     *            XML elements and attributes
+     * @return the populated node
+     */
+    private static Set<ObjectType> parseRelationElement(Element element, String lang, Space space, String documentNS,
```
henning-gerhardt: JavaDoc and method signature did not match. Please update the JavaDoc!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,561 @@
...
+     * Parses an XML DOM tree to a linked data tree.
+     *
+     * @param root
+     *            root node of the XML DOM tree
+     * @param documentNS
+     *            the address of the document, as fall-back namespace for local
+     *            XML elements and attributes
+     * @return the root node of a linked data tree
+     */
+    private static Node toNode(Element root, String documentNS, Storage storage) {
```
henning-gerhardt: JavaDoc and method signature did not match. Please update the JavaDoc!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,561 @@
...
+     */
+    public static Node toNode(String data, String documentNS, Storage storage) throws SAXException {
+        try {
+            return toNode(new ByteArrayInputStream(data.getBytes("UTF-16")), Optional.of("UTF-16"), documentNS,
+                storage);
+        } catch (UnsupportedEncodingException e) {
+            /* "UTF-16" is always supported by Java */
+            throw new IllegalStateException(e.getMessage(), e);
+        } catch (IOException e) {
+            /* there is no IOException to expect when reading from a String */
```
henning-gerhardt: Why you are catching an exception which should not be thrown as you mentioned this in your comment?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,561 @@
...
+     * Returns an URI for a local file, with the host name of the machine the
+     * program is running on.
+     *
+     * @param file
+     *            file to create an URI for
+     * @return URI for file
+     * @throws IOException
+     *             if it fails
+     */
+    public static String uriForFile(File file) throws IOException {
```
henning-gerhardt: When I read this method then I would expect and URI object is returned but you are returning a String object? I think a URI object is a better solution or adjust the JavaDoc and method name.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,340 @@
...
+     *            element to populate
+     * @param xml
+     *            the dom document, needed to create the children
+     * @param abbr
+     *            the namespace handler initialized at the root node
+     * @throws NoSuchElementException
+     *             if a relation links to an empty set of objects. (If so, the
+     *             object model is corrupt.)
+     */
+    private static void nodeToElement(Node node, Document xml, Namespaces abbr, Element out) {
```
henning-gerhardt: JavaDoc and method signature did not match. Please update the JavaDoc!

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,340 @@
...
+     *            number of spaces of each indent level. A negative number
+     *            disables wrapping
+     * @param namespaces
+     *            an external defined mapping of namespaces to abbreviations
+     * @throws IOException
+     *             if the writing fails
+     */
+    public static void toFile(Node node, File file, int indent, Map<String, String> namespaces) throws IOException {
+        OutputStream out = null;
+        try {
```
henning-gerhardt: Please use a `try-with-resources` statement here instead of closing streams by yourself.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/format/xml/XMLReaderTest.java
```diff
@@ -0,0 +1,35 @@
+package org.kitodo.dataaccess.format.xml;
```
henning-gerhardt: Missing project license header - please add the license header!

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/format/xml/XMLWriterTest.java
```diff
@@ -0,0 +1,112 @@
+package org.kitodo.dataaccess.format.xml;
```
henning-gerhardt: Missing project license header - please add the license header!

Kitodo-DataAccess/src/test/resources/xmlWriterTest-expectedResult.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<mets:mets xmlns:mets="http://www.loc.gov/METS/" xmlns:mods="http://www.loc.gov/mods/v3" xmlns:xlink="http://www.w3.org/1999/xlink">
```
henning-gerhardt: It is possible to add a license header in this file? If so, please add the license header.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Signals that white-space processing is acceptable for this element.
+     */
+    DEFAULT("default") {
+        @Override
+        /**
+         * Trimming white space the XML way.
+         */
+        public String trim(String s) {
```
henning-gerhardt: Short method parameter names should be avoided. Please use a longer and understandable name.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Signals the intention that in that element the white space should be
+     * preserved.
+     */
+    PRESERVE("preserve") {
+        @Override
+        /**
+         * Trimming is doing nothing.
+         */
+        public String trim(String s) {
```
henning-gerhardt: 1) Why is this method is needed but do nothing?
2) Short method parameter names should be avoided. Please use a longer and understandable name.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Trims a String the XML way, if Space.DEFAULT is chosen. Does nothing if
+     * Space.PRESERVE is chosen.
+     *
+     * @param s
+     *            string to process
+     * @return trimmed string
+     */
+    public abstract String trim(String s);
```
henning-gerhardt: Short method parameter names should be avoided. Please use a longer and understandable name.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,340 @@
...
+    /**
+     * String to pass to the transformer to switch on indenting.
+     */
+    private static final String TRANSFORMER_INDENT_TRUE = "yes";
+
+    /**
+     * String to pass to the transformer to set the indenting width in
+     * characters.
+     */
+    private static final String TRANSFORMER_INDENT_VALUE = "{http://xml.apache.org/xslt}indent-amount";
```
henning-gerhardt: Is a closing `}` missing or the beginning `{` wrong ?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,340 @@
...
+     * @return the DOM document
+     */
+    private static Document toDocument(Node node, Map<String, String> namespaces) {
+        Namespaces abbr = namespaces instanceof Namespaces ? (Namespaces) namespaces : new Namespaces(namespaces);
+        DocumentBuilder builder;
+        try {
+            builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+        } catch (ParserConfigurationException e) {
+            String message = e.getMessage();
+            throw new RuntimeException(message != null ? message : e.getClass().getSimpleName(), e);
```
henning-gerhardt: Is a `RuntimeException` the only one solution in this case? It is hard to catch them and not described in the JavaDoc that this kind of Exception is thrown.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,340 @@
...
+                transformer.setOutputProperty(OutputKeys.INDENT, TRANSFORMER_INDENT_TRUE);
+                transformer.setOutputProperty(TRANSFORMER_INDENT_VALUE, Integer.toString(indent));
+            }
+            out = new FileOutputStream(file);
+            transformer.transform(new DOMSource(toDocument(node, namespaces)), new StreamResult(out));
+            out.flush();
+
+        } catch (TransformerException e) {
+            String message = e.getMessage();
+            throw new RuntimeException(message != null ? message : e.getClass().getSimpleName(), e);
```
henning-gerhardt: Is a `RuntimeException` a good solution for this case?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,340 @@
...
+                transformer.setOutputProperty(OutputKeys.INDENT, TRANSFORMER_INDENT_TRUE);
+                transformer.setOutputProperty(TRANSFORMER_INDENT_VALUE, Integer.toString(indent));
+            }
+            transformer.setOutputProperty(OutputKeys.ENCODING, charset);
+            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+            transformer.transform(new DOMSource(toDocument(node, namespaces)), new StreamResult(buffer));
+            return new String(buffer.toByteArray(), charset);
+        } catch (TransformerException | UnsupportedEncodingException e) {
+            String message = e.getMessage();
+            throw new RuntimeException(message != null ? message : e.getClass().getSimpleName(), e);
```
henning-gerhardt: Is a `RuntimeException` a good solution in this case?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
matthias-ronge: The enum has got an interface, so each enum member implements that interface, and behaves differently. The one case does reduce white space (indents) as defined by XML specification, the other one does nothing (behaviour like in HTML `<pre>` tags)

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
matthias-ronge: Because otherwise, the static method `toNode(String, String, Storage)` would have to declare `throws IOException` which it never does. Which would mean any usage of the method would have to care for that exception that is never thrown.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: No. Apache just wants *that* string.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: This exception will never be thrown. This is just a catch-and-wrap so that the method does not need to declare `throws ParserConfigurationException` which it never does, and the programmer using this method does not have to care for an exception that is never thrown. The definition is that a ParserConfigurationException will be thrown *if a DocumentBuilder cannot be created which satisfies the configuration requested*. Since I fetch a new instance and do not put any configuration before calling `newDocumentBuilder()`, this will never happen.

Kitodo-DataAccess/src/test/resources/xmlWriterTest-expectedResult.xml
matthias-ronge: No, this file is against what the test compares. It cannot be changed, sorry.

matthias-ronge:
2. Tests are present, for both writing and reading.
3. Yes—I am a lazy enaugh not to write code that is unneccessary.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,337 @@
...
+     *            File to write to.
+     * @param indent
+     *            number of spaces of each indent level. A negative number
+     *            disables wrapping
+     * @param namespaces
+     *            an external defined mapping of namespaces to abbreviations
+     * @throws IOException
+     *             if the writing fails
+     */
+    public static void toFile(Node node, File file, int indent, Map<String, String> namespaces) throws IOException {
```
henning-gerhardt: Should this method to be used to store data on local filesystem in a process directory?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,337 @@
...
+        String charset = optionalCharset.orElse(StandardCharsets.UTF_8.toString());
+
+        try {
+            Transformer transformer = TransformerFactory.newInstance().newTransformer();
+            if (indent > -1) {
+                transformer.setOutputProperty(OutputKeys.INDENT, TRANSFORMER_INDENT_TRUE);
+                transformer.setOutputProperty(TRANSFORMER_INDENT_VALUE, Integer.toString(indent));
+            }
+            transformer.setOutputProperty(OutputKeys.ENCODING, charset);
+            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
```
henning-gerhardt: You are opening a stream without closing it properly.

Kitodo-DataAccess/src/test/resources/xmlWriterTest-expectedResult.xml
henning-gerhardt: If I understood your code correct then comments should be ignored.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/format/xml/XMLWriterTest.java
```diff
@@ -0,0 +1,123 @@
...
+                put("xlink", "http://www.w3.org/1999/xlink#");
+            }
+        };
+        File expectedResult = new File("src/test/resources/xmlWriterTest-expectedResult.xml");
+        File testfile = File.createTempFile("xmlWriterTest-", ".xml");
+        try {
+            SerializationFormat.XML.write(data, namespaces, testfile);
+            assertThat(FileUtils.contentEquals(testfile, expectedResult), is(true));
+        } finally {
+            if (testfile.exists()) {
```
henning-gerhardt: Maybe you should move part of your code to `@Before` and `@After` methods to initialise and clean up your file space usage.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
henning-gerhardt: Your method already throws a `SAXException` in which case the programmer must react. Throwing a `RuntimeException` instead of a declared `IOException` is not a good way. In most cases `RuntimeException` are not fetched and broke the whole application in a not user friendly way.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
henning-gerhardt: I oversee the closing `}` and that the Apache library need the string in that way is not understandable for me but if this library needs this that it should go in.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
henning-gerhardt: But if this is happen because without a reason this exception is not thrown or declared a mostly not fetched `RuntimeException` is thrown.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,573 @@
...
+     * a {@code String}. The URI is formed with the host name of the machine the
+     * program is running on.
+     *
+     * @param file
+     *            file to create an URI for
+     * @return unique resource identifier for the file
+     * @throws IOException
+     *             if it fails
+     */
+    public static String globallyUniqueIdentifierForFile(File file) throws IOException {
```
henning-gerhardt: In Kitodo project files not accessed directly (except configuration files). For accessing files for example in the process directory of a process the module `Kitodo-FileManagement` should be used. Reason behind is that files are not necessary saved on a local file system and maybe provided be a media repository over http based requests. So maybe your self written method is not needed if you use the API from `Kitodo-FileManagement` module and URI objects.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,573 @@
...
+     *            file to read
+     * @param storage
+     *            storage implementation to read the XML file into
+     * @return the linked data node
+     * @throws SAXException
+     *             if the file is not wellformed
+     * @throws IOException
+     *             if the reading fails
+     */
+    public static Node toNode(File path, Storage storage) throws SAXException, IOException {
```
henning-gerhardt: See my comment for accessing file in for example process directory.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
Beacze: If compiler informs that IOException can be thrown it should be declared in throws.

matthias-ronge: A `SAXException` *can* happen, because the parsing of the XML may fail (like for `"<foo>bar</a>"`, which is not well-formed). The `IOException` *cannot* happen, because reading from a string does not cause a device error (such as: no read permission, network not responding, defective hard disk, …). It is only declared by the interface because you can pass various readers to the interface (also readers for ‘normal’ files), in which case this exception could happen.

For why I handle this situation like this, see [these recommendations](http://www.odi.ch/prog/design/newbies.php#25).

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: This is based on different understanding of name spaces. The syntax is `{<name space>}<simple name>`.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: Should be correct. `NoSuchElementException` is an unchecked exception that does not need to be declared. It is only explained in the Javadoc because the reason is not obvious if it happens. It may be thrown by `objects.iterator().next()` if the data model in memory is corrupted.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: This is the recommended way to handle exceptions that are “never” thown, see [here](http://www.odi.ch/prog/design/newbies.php#25).

Kitodo-DataAccess/src/test/resources/xmlWriterTest-expectedResult.xml
matthias-ronge: No. XMLWriterTest uses Apache FileUtils to do a compare if the output of the XML writer is equal to this file: `assertThat(FileUtils.contentEquals(testfile, expectedResult), is(true));`

Comments are ignored in the XML Reader, this is what you might be thinking of, but that’s a different thing.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: This method stores data in a simple XML file. I do not know yet which data it will store where. (I would prefer to store the data in the process directory in another format, which is more straightforward. But technically, you can.)

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: [`ByteArrayOutputStream`](https://docs.oracle.com/javase/8/docs/api/java/io/ByteArrayOutputStream.html#close--) does not need to be closed:

> Closing a `ByteArrayOutputStream` has no effect. The methods in this class can be called after the stream has been closed without generating an `IOException`.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
Beacze: Maybe for developer it is useful to make such a constructs, but for production not. RuntimeException crashes the whole application and is not user friendly. 

Additionally our coding guideline says that exceptions should be thrown up to form classes where they can be presented to the user in the friendly way as a message, not as the crash of the whole application.

In my personal opinion it is just adding not necessary blur to the implementation.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/format/xml/XMLWriterTest.java
matthias-ronge: Why? The code is specific to *this* test. `@Before` and `@After` methods will be executed before and after *every* test. If ever another test will be added (imagine something, `xmlWriter.isProperlyConfigured()`) the test file would be created and deleted, which does not make sense.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
matthias-ronge: This is part of reading configuration XML files and rule sets, yes.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
matthias-ronge: This is part of reading configuration XML files and rule sets, too.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
matthias-ronge: The method will not crash, because there are no IOException when reading from a string. A method should not declare an exception that it never throws.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/format/xml/XMLReaderTest.java
matthias-ronge: Was added.

matthias-ronge:
Is this sufficient for you now?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
henning-gerhardt: On many methods you are throwing unchecked exceptions (most of them are RuntimeExceptions) without any explanation in JavaDoc and here you added it? I can not follow your minds in this kind of writing code and documentation.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
henning-gerhardt: As this is a public static method, who will prevent the usage for other use cases? I will hope that you are right and it is really only used for configuration files and not the meta data files itself.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
henning-gerhardt: I hope, you are right.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
henning-gerhardt: So, you are adding methods which are maybe used? Why not implement a method when this method is needed?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,573 @@
...
+     *            XML elements and attributes
+     * @param storage
+     *            storage implementation to read the XML file into
+     * @return the created Node
+     * @throws SAXException
+     *             if the XML is semantically wrong
+     */
+    public static Node toNode(String data, String documentNS, Storage storage) throws SAXException {
+        try {
+            return toNode(new ByteArrayInputStream(data.getBytes("UTF-16")), Optional.of("UTF-16"), documentNS,
```
henning-gerhardt: You are opening a stream but closing it inside a other method. Is this a good way? Would it not be proper way to open and close a stream inside the same method?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
matthias-ronge: In this case, it has value (may help in the future) to understand what is going on, and the comment is already there. I don’t see a benefit in removing sensible documentation. In the Oracle JDK it is widespread practice to document runtime exceptions if they are important to a method, and otherwise not. I try to follow this practice. (Of course, there is *always* more that could be documented—feel free to do so.)

The `RuntimeException` cases are most likely cases where I wrap a checked exception that is the result of an either bad, or unconventionally used design. If you want to [delve deeper into this matter, I can really recommend you this article](http://www.onjava.com/pub/a/onjava/2003/11/19/exceptions.html#article).

**TL;DR**
* Checked exceptions should be used in cases where the *user* has to fix the problem and the front-end should generate a sensible error message. Common example: disk full.
* Unchecked exceptions should be used if the exception is the result of a programming error elsewhere, typically `NullPointerException` or `IllegalArgumentException`. Here, a *programmer* needs to fix it. A stacktrace in the front-end will be sufficient in this case, as the user cannot help it anyway.
* In the early days of Java some programmers did not fully understand the concepts behind checked and unchecked exceptions, not even always at Sun. If a library was implemented badly and throws checked exceptions where they are inappropriate, or in a case where a checked exception can be considered to be never thrown (like: `string.getBytes("UTF-8")` → throws checked UnsupportedEncodingException, but will never, since UTF-8 is built-in. The exception makes sense if the enconding string is *entered* manually by the user, but not if it is a constant that is known to work) best practice is to `catch (WhateverCheckedException e) { throw new RuntimeException(e.getMessage(), e); }` so the desing is clean, but *if* the exception ever happens, it is not lost. If it will ever happen, this is the class ‘programming error’ from above, because it was not previewn to be happening.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
henning-gerhardt: Even a `close` call has no effect (regarding to the documentation) it would be a proper way to handle all stream objects in the same way and to use a `try-with-resources' statement should be possible to add.

Kitodo-DataAccess/src/test/java/org/kitodo/dataaccess/format/xml/XMLWriterTest.java
matthias-ronge: Was added—It seems that GitHub does not automatically update, I don’t know why.
→ We have to check checkstyle setup as it should refuse missing license headers.

Beacze: Checkstyle checks for missing license header in java classes, but... we don't run checkstyle on test resources.

matthias-ronge:
Your comments made me rethink some things and I changed the following:

- The method with the unclear name in fact creates a name space and was in the wrong place. I moved it to the corresponding class and gave it a better name. However, it should be public, because unique namespaces will be a requirement for the data model (to allow for multi-client capability).
- I didn’t have the FileManagementInterface in mind. So I added a method to the XMLWriter to accept an outputstream (from the FileManagementInterface). Write file method is still in place because it is *very* helpful during development and debugging the data model.
- XMLWriterTest fails after checking it out on Windows again because the corresponding file was originally with Windows’ line endings (test worked), then convertet to Linux line endings (test worked on Travis), but now after check out again I get a Linux file (which is what it should) but comparison fails. So I made the test unaware of line endings now.

I hope you agree with my changes and there is nothing more to fix now.

matthias-ronge:
Is there something still open?

Kitodo-DataAccess/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>log4j-slf4j-impl</artifactId>
             <version>2.9.1</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>4.1</version>
+        </dependency>
```
henning-gerhardt: Please remove the white spaces at the end of the line. This was only visible after run a `git diff master...<your_branch>` on console.

henning-gerhardt:
I can not follow your way of re-throwing exception as Runtime based exception. You added a lot of links which "recommend" it but I must read them in the next days if I have time for it.

`RuntimeException` should used on rare situations but you are using them every where and this is not a good sign.

matthias-ronge:
“Everywhere” is a little bit over the top, there were four usages. I have chaged them either into IllegalStateException (because it is an illegal state if something happens that cannot happen), and one I found out that it may be a wrapped IOException, so I changed that one back—seems to be self-evident. This is at least more self-explaining than a plein RuntimeException.

henning-gerhardt:
A specific exception is better then a general exception even this specific exceptions depends on a RuntimeException. Even after reading your links I'm not convinced about the expanded usage of RuntimeExceptions and their derivatives.

matthias-ronge:
Is this branch okay now or is there still something missing?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
```diff
@@ -0,0 +1,576 @@
...
+     *            context to show the resource in
+     * @return the created link
+     */
+    public static Node createSimpleLink(Storage storage, Optional<IdentifiableNode> relation, String to,
+            Collection<String> type, Collection<LangString> title, Optional<Actuate> actuate, Optional<Show> show) {
+        Node node = storage.createNode();
+        node.put(HREF, to);
+        type.forEach(entry -> node.put(ROLE, entry));
+        relation.ifPresent(value -> node.put(ARCROLE, value.getIdentifier()));
+        if (title.size() == 1) {
```
Beacze: Why every of those enums have repeated code? It is repeated in this file 3 times.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
```diff
@@ -0,0 +1,576 @@
...
+
+    /**
+     * Returns a version of this node reference which, in a debugger, will
+     * symbolically represent it.
+     */
+    @Override
+    public String toString() {
+        return '↗' + identifier;
+    }
+
```
Beacze: General remark to the file, it would be more readable if it would be sorted in some way, for e.g. first attributes, next inner enums, at the end methods.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XML.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private String identifier;
+
+    /**
+     * Enum constants constructor.
+     *
+     * @param identifier
+     *            referenced URL
+     */
+    private XML(String identifier) {
```
Beacze: Modifier 'private' is redundant for Enum constructor - needs to be removed.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLReader.java
```diff
@@ -0,0 +1,523 @@
...
+     * @param space
+     *            whether or not to trim white space around the element
+     * @param documentNS
+     *            the namespace of the document
+     * @param storage
+     *            storage to read the XML into
+     * @return the literal node
+     */
+    private static ObjectType parseLiteralElement(Element element, NodeReference type, String lang, Space space,
+            String documentNS, Storage storage) {
```
Beacze: Parameter documentNS is not used in this method.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,352 @@
...
+     *             if a relation links to an empty set of objects. (If so, the
+     *             object model is corrupt.)
+     */
+    private static void nodeToElement(Node node, Document xml, Namespaces abbr, Element out) {
+        TreeMap<Long, ObjectType> orderedElements = new TreeMap<>();
+        HashMap<String, Collection<ObjectType>> bagElements = new HashMap<>();
+        long first = Long.MAX_VALUE;
+        long last = 0;
+
+        if ((node instanceof Node) && (node instanceof IdentifiableNode)) {
```
Beacze: Condition check _node instanceof Node_ should be removed, and if any check is needed here, it is null check _(Objects.nonNull(node))_.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XMLWriter.java
```diff
@@ -0,0 +1,352 @@
...
+     * @return the DOM document
+     */
+    private static Document toDocument(Node node, Map<String, String> namespaces) {
+        try {
+            Namespaces abbr = namespaces instanceof Namespaces ? (Namespaces) namespaces : new Namespaces(namespaces);
+            String qualifiedName = abbr.abbreviateElement(getTypeOrDefault(node));
+            Document document = DocumentBuilderFactory.newInstance().newDocumentBuilder().getDOMImplementation()
+                    .createDocument(Namespaces.namespaceOfForXMLFile(getTypeOrDefault(node)), qualifiedName, null);
+            Element root = document.getDocumentElement();
+            if ((node instanceof Node) && (node instanceof IdentifiableNode)) {
```
Beacze: Condition check node instanceof Node should be removed, and if any check is needed here, it is null check (Objects.nonNull(node)).

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    private final String enumeratedValue;
+
+    /**
+     * Creates the enumeration’s elements. Populates their private fields.
+     *
+     * @param enumeratedValue
+     *            String value as by specification
+     */
+    private Space(String enumeratedValue) {
```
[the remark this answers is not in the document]
matthias-ronge: I find this confusing. The constructor of an enum class is *private*, not package-private. Why shall I confusingly declare it package-private?

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
```diff
@@ -0,0 +1,576 @@
...
+     * @param title
+     *            link title(s) as human readable text
+     * @param actuate
+     *            when to load the resource
+     * @param show
+     *            context to show the resource in
+     * @return the created link
+     */
+    public static Node createSimpleLink(Storage storage, Optional<IdentifiableNode> relation, IdentifiableNode to,
+            Collection<LangString> title, Optional<Actuate> actuate, Optional<Show> show) {
```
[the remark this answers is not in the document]
matthias-ronge: The [XLink specification](https://www.w3.org/TR/xlink/) defines them so.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
matthias-ronge: The keyword `enum` implies that the resulting class is an *immediate* subclass of `java.lang.Enum`. So, sadly, I cannot make an intermediate generic class which summarizes this code. I wish I could.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
matthias-ronge: It is sorted:
1. enum constants, in alphabetical order
2. inner enums, in alphabetical order
3. (normal) constants, in alphabetical order
4. static methods, in alphabetical order
5. constructor, in alphabetical order
6. non-static methods, in alphabetical order

This is the order Eclipse provides me with when I use the "save action" to “sort all members”.

matthias-ronge:
Should be all fixed, notwithstanding that I do not really understand the one.

Branche was rebased against current master.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/Space.java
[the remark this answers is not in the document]
Beacze: The constructor actually isn't package-private... it's implicitly private the way interface methods are implicitly public even if you don't add the keyword.

What does mean that private keyword is not needed here.

https://stackoverflow.com/questions/7747948/why-can-a-enum-have-a-package-private-constructor/7748220

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/SeveralElementsForSameIndexException.java
```diff
@@ -0,0 +1,66 @@
...
+        // nothing to do
+    }
+
+    /**
+     * Creates a SeveralElementsForSameIndexException with a message string.
+     *
+     * @param s
+     *            message
+     */
+    public SeveralElementsForSameIndexException(String s) {
```
Beacze: Variable name - should be message.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
```diff
@@ -0,0 +1,576 @@
...
+        public static Actuate getEnum(String name) {
+            if (!reversed.containsKey(name)) {
+                throw new IllegalArgumentException("Unknown String Value: " + name);
+            }
+            return reversed.get(name);
+        }
+
+        private String name;
+
+        private Actuate(String name) {
```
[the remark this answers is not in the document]
Beacze: Modifier 'private' is redundant for Enum constructor - needs to be removed.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
```diff
@@ -0,0 +1,577 @@
...
+        public static Show getEnum(String name) {
+            if (!reversed.containsKey(name)) {
+                throw new IllegalArgumentException("Unknown String Value: " + name);
+            }
+            return reversed.get(name);
+        }
+
+        private String name;
+
+        private Show(String name) {
```
Beacze: Modifier 'private' is redundant for Enum constructor - needs to be removed.

Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/format/xml/XLink.java
```diff
@@ -0,0 +1,576 @@
...
+        public static Type getEnum(String name) {
+            if (!reversed.containsKey(name)) {
+                throw new IllegalArgumentException("Unknown String Value: " + name);
+            }
+            return reversed.get(name);
+        }
+
+        private String name;
+
+        private Type(String name) {
```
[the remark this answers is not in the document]
Beacze: Modifier 'private' is redundant for Enum constructor - needs to be removed.

matthias-ronge:
I thought, I had changed all of them. Sorry. String was renamed, too.

PR #1235, opened 2018-01-24
Use LazyDTOModel in "Meine Aufgaben"

oliver-stoehr:


Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -852,6 +853,21 @@ public void setMySchritt(Task task) {
...
+     *
+     * @param id passed as int
+     * @return task
+     */
+    public Task getTaskById(int id) {
+        try {
+            return serviceManager.getTaskService().getById(id);
+        } catch (DAOException e) {
+            logger.error(e.getMessage());
+            return null;
```
henning-gerhardt: 1) Returning `null` should be avoided as it introduce unnecessary null checks and null pointer exceptions.
2) Should the user not be informed in this case? You are only logging it but did not inform the user.

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -95,6 +97,74 @@ public static TaskService getInstance() {
...
+    public List<TaskDTO> findAll(String sort, Integer offset, Integer size, Map filters) throws DataException {
+
+        LoginForm login = (LoginForm) Helper.getManagedBeanValue("#{LoginForm}");
+        if (login == null) {
+            return new ArrayList<>();
+        }
+
+        // TODO: find other way than retrieving the form bean to access "hideCorrectionTasks" and "showAutomaticTasks"
+        // e.g. which tasks should be returned!
+        BoolQueryBuilder subquery = new BoolQueryBuilder();
```
Beacze: The same query is written two times. It could be build in private method and called here by some get...Query method

Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -175,18 +178,14 @@ private String sortList() {
...
-        if (this.sortierung.equals("batchDesc")) {
-            order = Order.desc("proc.batchID");
-        }
-        if (this.sortierung.equals("prozessdateAsc")) {
-            order = Order.asc("proc.creationDate");
-        }
-        if (this.sortierung.equals("prozessdateDesc")) {
-            order = Order.desc("proc.creationDate");
-        }*/
+        /*
```
henning-gerhardt: Please remove this out commented code. It is the fifth or sixth time that this code get formatted.

Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -852,6 +852,25 @@ public void setMySchritt(Task task) {
...
+     * @param id
+     *            passed as int
+     * @return task
+     */
+    public Task getTaskById(int id) {
+        try {
+            return serviceManager.getTaskService().getById(id);
+        } catch (DAOException e) {
+            logger.error(e.getMessage());
+            FacesContext facesContext = FacesContext.getCurrentInstance();
```
henning-gerhardt: Why you are introducing a new way to report error messages to the user? Even this message is not translatable.

henning-gerhardt:
This pull request needs a rebase as at least one merge conflict exists in `Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java`.

Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
solth: This is the Java bean alternative to rendering messages to templates that have `h:message` or `h:messages` tags. It should therefore be in line with the existing error reporting in templates like `editDocket.xhtml` and others.

solth:
Rebased.

PR #1237, opened 2018-01-25
Write protection for projects

IkramMaalej:


Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
```diff
@@ -292,6 +292,14 @@ public void setMyProjekt(Project inProjekt) {
         this.myProjekt = inProjekt;
     }
 
+    public boolean isLocked() {
```
henning-gerhardt: Public method without JavaDoc - please add a JavaDoc description.

Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
```diff
@@ -292,6 +292,14 @@ public void setMyProjekt(Project inProjekt) {
...
+        return locked;
+    }
+
+    public void setLocked(boolean locked) {
```
henning-gerhardt: Public method without JavaDoc - please add a JavaDoc description.

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -1174,3 +1174,5 @@ newBatch=New batch
 newUser=New user
 newUserGroup=New user group
 newLdapGroup=New LDAP group
+locked=Locked
```
Beacze: It would be useful to keep alphabetical order of properties in messages files. It would help to avoid double messages for the same purpose.

IkramMaalej:
For the second and third remarks : 
In the default case, the input fields of the "projectEdit" page should be deactivated in order to avoid unwanted changes. This "write protection" should be switched on and off via a button with a symbol "lock" in the project edit form. So every user have to unlock the form to make an edit.

PR #1238, opened 2018-01-26
Adds SQL to remove foreign key constraints for versions before 1.11.2

lutzhelm:
Hibernate auto generated schemas might contain foreign key constraints that result in errors when executing remove_old_index_prior_1.11.2.sql. Same problem as in pull request #1156 but for versions before 1.11.2.
Also adds some lines about it to the readme.


henning-gerhardt:
Thank you for contributing this migration file. I can not test it because I have no database with this kind of schema. Your changes will be visible for all with the next release.

PR #1241, opened 2018-01-29
Duplicate project

solth:
Lets the user duplicate existing projects from the project list view using a "duplicate project" button. The user is redirected to the project edit view of the new, duplicated project, with all fields apart from the title pre-populated with the values retrieved from the original project.

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
@@ -169,45 +169,10 @@ public class ProjekteForm extends BasisForm {
      * @return page address
      */
     public String duplicateProject(Integer itemId) {
-        this.myProjekt = new Project();
         this.itemId = 0;
         try {
-            Project baseProject = serviceManager.getProjectService().getById(itemId);
-
-            // Project _title_ should explicitly _not_ be duplicated!
-            this.myProjekt.setStartDate(baseProject.getStartDate());
-            this.myProjekt.setEndDate(baseProject.getEndDate());
-            this.myProjekt.setNumberOfPages(baseProject.getNumberOfPages());
-            this.myProjekt.setNumberOfVolumes(baseProject.getNumberOfVolumes());
-
-            this.myProjekt.setFileFormatInternal(baseProject.getFileFormatInternal());
-            this.myProjekt.setFileFormatDmsExport(baseProject.getFileFormatDmsExport());
-            this.myProjekt.setDmsImportErrorPath(baseProject.getDmsImportErrorPath());
-            this.myProjekt.setDmsImportSuccessPath(baseProject.getDmsImportSuccessPath());
-
-            this.myProjekt.setDmsImportTimeOut(baseProject.getDmsImportTimeOut());
-            this.myProjekt.setUseDmsImport(baseProject.isUseDmsImport());
-            this.myProjekt.setDmsImportCreateProcessFolder(baseProject.isDmsImportCreateProcessFolder());
-
-            this.myProjekt.setMetsRightsOwner(baseProject.getMetsRightsOwner());
-            this.myProjekt.setMetsRightsOwnerLogo(baseProject.getMetsRightsOwnerLogo());
-            this.myProjekt.setMetsRightsOwnerSite(baseProject.getMetsRightsOwnerSite());
-            this.myProjekt.setMetsRightsOwnerMail(baseProject.getMetsRightsOwnerMail());
-
-            this.myProjekt.setMetsDigiprovPresentation(baseProject.getMetsDigiprovPresentation());
-            this.myProjekt.setMetsDigiprovPresentationAnchor(baseProject.getMetsDigiprovPresentationAnchor());
-            this.myProjekt.setMetsDigiprovReference(baseProject.getMetsDigiprovReference());
-            this.myProjekt.setMetsDigiprovReferenceAnchor(baseProject.getMetsDigiprovReferenceAnchor());
-
-            this.myProjekt.setMetsPointerPath(baseProject.getMetsPointerPath());
-            this.myProjekt.setMetsPointerPathAnchor(baseProject.getMetsPointerPathAnchor());
-            this.myProjekt.setMetsPurl(baseProject.getMetsPurl());
-            this.myProjekt.setMetsContentIDs(baseProject.getMetsContentIDs());
-
-            this.myProjekt.setProcesses(new ArrayList<>(baseProject.getProcesses()));
-
+            this.myProjekt = serviceManager.getProjectService().duplicateProject(itemId);
             return redirectToEdit("?faces-redirect=true");
-
         } catch (DAOException e) {
             logger.error(e.getMessage());
             FacesContext facesContext = FacesContext.getCurrentInstance();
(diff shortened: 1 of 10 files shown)


Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
```diff
@@ -164,6 +163,29 @@ public String newProject() {
...
+     */
+    public String duplicateProject(Integer itemId) {
+        this.itemId = 0;
+        try {
+            this.myProjekt = serviceManager.getProjectService().duplicateProject(itemId);
+            return redirectToEdit("?faces-redirect=true");
+        } catch (DAOException e) {
+            logger.error(e.getMessage());
+            FacesContext facesContext = FacesContext.getCurrentInstance();
+            FacesMessage facesMessage = new FacesMessage("ERROR: Unable to duplicate selected project!");
```
henning-gerhardt: Output message is not translatable.

Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
```diff
@@ -792,6 +814,26 @@ public int getItemId() {
...
+     * @return String containing the templates titles of the project with the given
+     *         ID
+     */
+    public String getProjectTemplateTitles(int id) {
+        try {
+            return serviceManager.getProjectService().getProjectTemplatesTitlesAsString(id);
+        } catch (DAOException e) {
+            logger.error(e.getMessage());
+            FacesContext facesContext = FacesContext.getCurrentInstance();
+            FacesMessage facesMessage = new FacesMessage("ERROR: Unable to retrieve templates for project!");
```
henning-gerhardt: Output message is not translatable.

Kitodo/src/main/java/org/kitodo/services/data/ProjectService.java
```diff
@@ -282,4 +283,62 @@ public boolean isProjectComplete(Project project) {
...
+
+        duplicatedProject.setMetsPointerPath(baseProject.getMetsPointerPath());
+        duplicatedProject.setMetsPointerPathAnchor(baseProject.getMetsPointerPathAnchor());
+        duplicatedProject.setMetsPurl(baseProject.getMetsPurl());
+        duplicatedProject.setMetsContentIDs(baseProject.getMetsContentIDs());
+
+        // only duplicate templates, not all processes, of given base project
+        duplicatedProject.setProcesses(new ArrayList<>(
+                baseProject.getProcesses().stream().filter(Process::isTemplate).collect(Collectors.toList())));
+        return duplicatedProject;
```
henning-gerhardt: You are missing the mets file groups to duplicate.

Kitodo/src/test/java/de/sub/goobi/forms/ProjekteFormIT.java
```diff
@@ -74,4 +74,26 @@ public void shouldGetStatisticsManager() throws Exception {
         //assertEquals("Statistics Manager was not counted correctly!", Integer.valueOf(50), statisticsManager.getNumberOfPages());
         //assertEquals("Statistics Manager was not counted correctly!", Integer.valueOf(3), statisticsManager.getNumberOfVolumes());
     }
+
+    @Test
+    public void shouldDuplicateProject() throws Exception {
```
henning-gerhardt: This test belongs to `ProjectService` or `ProjectServiceIT` class as you test against the business logic and not the output of the form method. You can test both if you want as more tests are better then less tests.

Kitodo/src/main/java/org/kitodo/services/data/ProjectService.java
```diff
@@ -282,4 +283,62 @@ public boolean isProjectComplete(Project project) {
...
+     * Duplicate the project with the given ID 'itemId'.
+     *
+     * @return the duplicated Project
+     */
+    public Project duplicateProject(Integer itemId) throws DAOException {
+        Project duplicatedProject = new Project();
+
+        Project baseProject = getById(itemId);
+
+        // Project _title_ should explicitly _not_ be duplicated!
```
henning-gerhardt: How did you find the project after saving again if it has no title? Should the user not insert a new project title on starting duplicating a project and this inserted title is set here?

solth: The `title` field is empty when the user duplicates a project. Since the `title` field is mandatory, the user cannot save the duplicated project without entering a new title.

henning-gerhardt: OK if the user get the duplicated project displayed and must enter the project title, then it should work.

Kitodo/src/main/webapp/WEB-INF/templates/includes/header.xhtml
```diff
@@ -61,7 +61,7 @@
                             <li><h3>#{msgs.dashboard}<h:outputText rendered="#{LoginForm.maximaleBerechtigung == 1}"> <i>(Admin)</i></h:outputText></h3></li>
                             <li class="nav-pic-text"><h:link value="#{msgs.schritte}" outcome="tasks"><i class="fa fa-bell fa-lg"/></h:link></li>
                             <li class="nav-pic-text"><h:link value="#{msgs.prozesse}" outcome="processes"><i class="fa fa-hourglass-3 fa-lg"/></h:link></li>
-                            <li class="nav-pic-text"><h:link value="#{msgs.projekte}" outcome="projects" onclick="#{ProjekteForm.setLocked(true)}"><i class="fa fa-folder-open fa-lg"/></h:link></li>
+                            <li class="nav-pic-text"><h:link value="#{msgs.projekte}" outcome="projects"><i class="fa fa-folder-open fa-lg"/></h:link></li>
```
henning-gerhardt: Is this change really wanted? It is introduced by @Beacze .

Beacze: Without it I couldn't save copied project. I don't know for what is used this parameter.

solth: This onclick action has to remain in the code, probably. It was introduced in the "accidental change prevention" that @IkramMaalej introduced in #1237. Instead, the lock should be set to `false` when someone clicks the `duplicateProject` button. I will add a corresponding commit.

henning-gerhardt:
There are merge conflicts  in `Kitodo/src/main/resources/messages/messages_de.properties` and `Kitodo/src/main/resources/messages/messages_en.properties`. Please solve them.

solth:
Conflicts resolved.

Kitodo/src/main/webapp/WEB-INF/templates/includes/header.xhtml
henning-gerhardt: Is this now solved?

solth: Yes, this is resolved (after discussing the issue with @IkramMaalej, we decided that locking the project shouldn't be done in the header but instead only when someone clicks on the "project edit" button in the list view, which has been added in commit 9d73a74d7047899b706f9460ca5a066903c1d1d7)

PR #1245, opened 2018-01-31
Stylesheet update

solth:
Second stylesheet update pass.


henning-gerhardt:
There is at least one merge conflict in `Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css`. Please solve this conflict.

Kitodo/src/main/webapp/pages/kitodo-frontend/firstpage.xhtml
```diff
@@ -0,0 +1,40 @@
+<!--
```
henning-gerhardt: What is the reason to add this and the other file `secondpage.xhtml`?

Kitodo/src/main/webapp/pages/kitodo-frontend/test/secondpage.xhtml
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
```
henning-gerhardt: What is the reason to add this and the other file `firstpage.xhtml`?

solth:
> See my comment to `firstpage.xhtml` and `secondpage.xhtml`. I do not think that this files should be added or?

You are right, these were not supposed to be added with this commit. Therefore I removed them again. Thanks for pointing it out!

solth:
> There is at least one merge conflict in Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css. Please solve this conflict.

Rebased and conflict resolved.

PR #1248, opened 2018-02-01
Remove for loop from BatchForm which did not loop

Beacze:


Kitodo/src/main/java/de/sub/goobi/forms/BatchForm.java
```diff
@@ -481,34 +481,31 @@ public String exportBatch() {
             Helper.setFehlerMeldung("noBatchSelected");
             return null;
         }
-        for (Integer batchID : selectedBatches) {
```
henning-gerhardt: With this change it is not possible any more to export more then one batch even if you select more then one. Is this really wanted?

Beacze: It was not possible also earlier as this method was returning after first execution, so in fact it was loop which was not looping.

Beacze:
Problem needs to be discussed.

PR #1253, opened 2018-02-01
Optimize way of getting current user

Beacze:


Kitodo/src/main/java/de/sub/goobi/export/dms/ExportDms.java
```diff
@@ -494,9 +494,13 @@ public void imageDownload(Process process, URI userHome, String atsPpnBand, fina
...
                  * Benutzerberechtigung neu anlegen
                  */
-                User myUser = (User) Helper.getManagedBeanValue("#{LoginForm.myBenutzer}");
+                User user = Helper.getCurrentUser();
                 try {
-                    fileService.createDirectoryForUser(zielTif, myUser.getLogin());
+                    if (user != null) {
+                        fileService.createDirectoryForUser(zielTif, user.getLogin());
+                    } else {
+                        throw new IOException("noLoggedUser");
```
henning-gerhardt: An exception messages must not be translatable and they are not translated until now.

Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java
```diff
@@ -2156,17 +2156,20 @@ public void imageDownload(Process process, URI userHome, String atsPpnBand, fina
...
                  * wenn kein Agora-Import, dann den Ordner mit Benutzerberechtigung neu anlegen
                  */
-                User myUser = (User) Helper.getManagedBeanValue("#{LoginForm.myBenutzer}");
+                User user = Helper.getCurrentUser();
                 try {
-                    fileService.createDirectoryForUser(zielTif, myUser.getLogin());
+                    if (user != null) {
+                        fileService.createDirectoryForUser(zielTif, user.getLogin());
+                    } else {
+                        throw new IOException("noLoggedUser");
```
henning-gerhardt: An exception messages must not be translatable and they are not translated until now.

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -785,6 +785,7 @@ noBatchSelected=Es ist kein Batch ausgew\u00E4hlt.
 noFileSelected=Keine Datei ausgew\u00E4hlt.
 noImage=Keine Images vorhanden in
 noImageRendered=Image wurde nicht erstellt
+noLoggedUser=There is no logged user
```
henning-gerhardt: Use something like `Es ist kein Benutzer angemeldet.`

PR #1254, opened 2018-02-01
Add method for get of metadata language for current logged user

Beacze:
Depends on #1253


Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
```diff
@@ -411,6 +411,10 @@ public static User getCurrentUser() {
         return login != null ? login.getMyBenutzer() : null;
     }
 
+    public static String getMetadataLanguageForCurrentUser() {
```
henning-gerhardt: New public method without JavaDoc. Please add missing JavaDoc.

PR #1275, opened 2018-02-07
Add missing set for index in LazyDTOModel

Beacze:


solth:
Is this what causes the list pages (projects, processes etc.) to not display any entries until the indexingPage has been opened at least once?

Beacze:
I think so.

PR #1281, opened 2018-02-07
Animate checkbox toggle

oliver-stoehr:
The checkbox icons are now styled with CSS instead of an text icon to support animations.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
@@ -193,8 +193,6 @@ button .ui-icon {
 }
 
 .ui-chkbox .ui-chkbox-icon {
-  /*outline: 0;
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
```diff
@@ -184,6 +184,53 @@ button .ui-icon {
...
+.ui-chkbox .ui-chkbox-box, .ui-chkbox .ui-chkbox-box.ui-state-hover {
+    width: 40px;
+    height: 40px;
+    border-radius: 3px;
+    background: #999;
+    cursor: pointer;
+}
+
+.ui-chkbox .ui-chkbox-icon {
+  /*outline: 0;
```
henning-gerhardt: Why are you adding out commented style sheet information? If this is not needed, remove it please.

PR #1285, opened 2018-02-09
Decrease logic in ProzesskopieForm

Beacze:

PR #1287, opened 2018-02-12
Decrease logic prozesskopieform

tim-boerner:
Fix for PR #1285, so this PR makes PR #1285 obsolete


henning-gerhardt:
Further review not possible as it is currently not possible to create new processes from a template to get data from a catalogue OPAC.

henning-gerhardt:
Closed as super seeded by #1293

PR #1294, opened 2018-02-21
Remove not used process properties box from task frontend

Beacze:


solth:
Actually, the file Kitodo/src/main/webapp/WEB-INF/templates/includes/tasks/taskBoxProperties.xhtml is still used in the new frontend. Why is it removed, here?

PR #1296, opened 2018-02-22
Hotfix for null check

Beacze:
It seems that when null is passed it is better than with null check.


henning-gerhardt:
Merging before 24 hour delay as this introduced and now removed check broke big parts of the application.

PR #1297, opened 2018-02-22
Fix issues found by SonarCloud

Beacze:
https://www.ibm.com/developerworks/library/j-jtp05236/

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/IndexingForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/IndexingForm.java
@@ -85,6 +85,7 @@ public class IndexingForm {
                 sleep(pause);
             } catch (InterruptedException e) {
                 logger.error("Thread interrupted: " + e.getMessage());
+                Thread.currentThread().interrupt();
             }
 
             currentIndexState = ObjectType.NONE;
@@ -258,6 +259,7 @@ public class IndexingForm {
                     }
                 } catch (InterruptedException e) {
                     logger.error(e.getMessage());
+                    Thread.currentThread().interrupt();
                 }
             }
         }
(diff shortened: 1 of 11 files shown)


Kitodo/src/main/java/de/sub/goobi/forms/IndexingForm.java
```diff
@@ -84,6 +84,7 @@ public void run() {
             try {
                 sleep(pause);
             } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
```
henning-gerhardt: Move this line after logging the error, so in any case the error get logged.

Kitodo/src/main/java/de/sub/goobi/forms/MassImportForm.java
```diff
@@ -344,10 +341,10 @@ public void uploadFile() throws IOException {
             basename = basename.substring(1);
         }
         if (basename.contains("/")) {
-            basename = basename.substring(basename.lastIndexOf("/") + 1);
+            basename = basename.substring(basename.lastIndexOf('/') + 1);
         }
         if (basename.contains("\\")) {
-            basename = basename.substring(basename.lastIndexOf("\\") + 1);
+            basename = basename.substring(basename.lastIndexOf('\\') + 1);
```
henning-gerhardt: Are you sure about this change? `lastIndexOf` can be called with a string type and int type value and in this case the string value must be used as we are checking for double backslash characters. Or did I missed something?

Beacze: It is escape character, so this change works correctly.

henning-gerhardt: Thank you for this explanation - escaping characters confuse me sometimes.

PR #1305, opened 2018-02-26
fixing Bug when calling metadataEditor page

IkramMaalej:


Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/processList.xhtml
```diff
@@ -50,12 +50,12 @@
                     <i class="fa fa-pencil-square-o fa-lg"/>
                     <f:param name="id" value="#{item.id}" />
                 </h:link>
-                <h:commandLink id="readXML" action="#{Metadaten.readXml}" title="#{msgs.metadatenBearbeiten}"
-                               rendered="#{ProzessverwaltungForm.displayMode.mode != 'template'}">
+                <h:commandLink id="readXML" action="#{Metadaten.readXml}" title="#{msgs.metadatenBearbeiten}">
                     <h:outputText><i class="fa fa-file-o"/></h:outputText>
+                    <f:param name="nurLesen" value="true"/>
```
henning-gerhardt: Why are you adding this parameter which prevent administrators and process maintainer to get a write access to the meta data?

IkramMaalej: 1. I have found this parameter in the same button in the old design page (Prozesse_Liste.xhtml line 208).
2. The "metadataEditor"  has the following error when trying to access to the page:
/ UI Layout Initialization Error
The center-pane element does not exist.
The center-pane is a required element.

henning-gerhardt: While moving pages / content from 2.x to master there a some information got lost ([see old implementation](https://github.com/kitodo/kitodo-production/blob/2.x/Goobi/newpages/inc_Prozessverwaltung/Prozesse_Liste.jsp#L361-L380)). This value was only set in the past if the current user as not the right to get a write access to the metadata of this process. But in the current implementation is no user access check and with your change you will deny for every one the write access to meta data over the process list. I don't think that adding this parameter fix the UI issues or did it?

IkramMaalej: No,I can remove this parameter without any other problem.

PR #1309, opened 2018-02-28
Display errors

solth:
Most of the new Facelets now use one of two new base templates `baseEditView` and `baseListView` which both contain a global `messages` tag to ensure that errors can be rendered on every page.


Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
```diff
@@ -991,11 +977,11 @@ public void exportDMSHits() {
             try {
                 export.startExport(proz);
             } catch (Exception e) {
-                Helper.setFehlerMeldung("ExportError", e.getMessage());
-                logger.error(e);
+                Helper.setErrorMessage("ExportError", logger, e);
             }
         }
         Helper.setMeldung(null, "ExportFinished", "");
+        logger.info(Helper.getTranslation("ExportFinished"));
```
henning-gerhardt: If you add a log message it should be logged before the user get the information. Please move this log call in front of the `Helper.setMeldung()` call.

Kitodo/src/main/java/de/sub/goobi/forms/SearchForm.java
```diff
@@ -169,7 +169,8 @@ protected void initUserList() {
         try {
             this.user.addAll(serviceManager.getUserService().getAllActiveUsersSortedByNameAndSurname());
         } catch (RuntimeException rte) {
-            logger.warn("Catched RuntimeException. List of users could be empty!");
+            Helper.setErrorMessage("errorLoadingMany", new String[] {Helper.getTranslation("aktiveBenutzer") });
+            logger.warn("RuntimeException caught. List of users could be empty!");
```
henning-gerhardt: Same here. Please first log the error message / reason and then display it to the user.

solth:
> Why are you sometimes using new Object[] {...} and on other places new String[] {...} for creating the second parameter? Should this not be unique?

I used `Object[]` when the array contains elements with various types (a variable of type `String` with the translation of a term and an `int` with the ID of an element). 
I used `String[]` in cases where I do not include ID parameters in the message, to ensure only Strings are passed as parameters for String substitution.

henning-gerhardt:
Thank you for the explanation but there are many places where are you using `Object[]` for a translated message like on places with `String[]` for getting even a translated message.

henning-gerhardt:
There are merge conflicts in
- Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
- Kitodo/src/main/java/de/sub/goobi/forms/ProzesskopieForm.java 
Please solve them.

solth:
Rebased and resolved conflicts.

I now use `Object[]` instead of `String[]` as parameter type in all calls to `setErrorMethod` since the underlying function `Messages.format` that is used for string substitution in `setErrorMethod` explicitely expects an array of type `Object` instead of `String` and otherwise produces a warning.

PR #1321, opened 2018-03-05
Add tests for VariableReplacer

Beacze:


Kitodo/src/test/java/de/sub/goobi/helper/VariableReplacerTest.java
```diff
@@ -0,0 +1,55 @@
+package de.sub.goobi.helper;
```
henning-gerhardt: Missing license header - please add the license header.

PR #1333, opened 2018-03-09
Fixes for Metadaten

Beacze:


Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -1683,48 +1647,53 @@ public void identifyImage(int pageNumber) {
...
-                        try {
-                            if (!fileService.delete(temporaryTifFile.toURI())) {
-                                logger.error(
-                                    "Error while deleting temporary tif file: " + temporaryTifFile.getAbsolutePath());
-                            }
-                            // not working
-                        } catch (IOException e) {
-                            logger.error("Error while deleting temporary tif file: " + e.getMessage());
+                // Copy tif-file to temporary folder
+                InputStream tifFileInputStream = fileService.read(tifFile);
```
henning-gerhardt: Where is the stream closed? Maybe you should use a try with resources statement. You are only moved the code but we should fix it when we touch the code.

PR #1339, opened 2018-03-09
Implemented Security Access Controller and first Authorities

tim-boerner:
- Created security access controller and service for checking authorities of current user

- Implemented usage of security access controller at dashboard menu, some parts of user page and fully on clientspage

- Inserted some first authorities with flyway script


tim-boerner:
Changed V2_26__Add_first_authorities.sql, so that inserted docket authorities can be assigned only globally.

PR #1343, opened 2018-03-12
Make sure the order of the issues is correct

### LINKED ISSUE #1323: Maintain the order of issues
Production shall be extended to preserve the order of several issues on the same day, also when they are maintained as individual processes.

When generating the METS export, the issue name shall be encoded as the METS `LABEL` attribute of the corresponding issue `div` to allow to distinguish them in Presentation’s calendar view.

**Approach:** When creating tasks at the granularity level “issues” and there are several issues per day, an additional value `CurrentNoSorting` will be added in the issues as metadata, which describes the sort order of the issues on this day.

When creating the XML export from the metadata editor, output will be created in the specified order. When re-importing the XML, the order of the issues will be taken into account. If necessary, one issue must be reconstructed multiple times. When exporting newspapers, the order of the issues will be evaluated and the aggregation will be carried out in this order.

The development of this feature is funded by SLUB Dresden.

matthias-ronge:
* Add sorting numbers to issues if there are several issues on one day
* Sort issues by sorting number on export

Fixes #1323


henning-gerhardt:
There are merge conflicts in

- Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
- Goobi/src/org/goobi/production/model/bibliography/course/Course.java 

Please solve / fix them.

matthias-ronge:
Somehow I was afraid that would happen. That was not trivial, but now it should work.

PR #1344, opened 2018-03-12
Add tests for ConfigCore

Beacze:


Kitodo/src/test/java/de/sub/goobi/config/ConfigCoreTest.java
```diff
@@ -0,0 +1,60 @@
...
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package de.sub.goobi.config;
+
+import java.net.URI;
+import org.joda.time.Duration;
```
henning-gerhardt: Ordering of import statements is not correct.

Kitodo/src/test/java/de/sub/goobi/config/ConfigCoreTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class ConfigCoreTest {
+
+    @Test
+    public void shouldSetTempImagesPathAsCompleteDirectory() {
+        URI path = URI.create("/pages/newImagesTemp/");
+        ConfigCore.setImagesPath(URI.create("/pages/newImagesTemp/"));
+        System.out.println(ConfigCore.getTempImagesPathAsCompleteDirectory());
```
henning-gerhardt: Why did you add this `System.out.println()` call? Maybe a logger is a better solution.

Beacze: This needs to be removed.

PR #1350, opened 2018-03-13
Fill the mysql database with default data on travis to test the migration

tim-boerner:
added default.sql script when mysql database is created on travis


tim-boerner:
Depends on PR [1345](https://github.com/kitodo/kitodo-production/pull/1345)

PR #1351, opened 2018-03-13
Active projects

solth:
"List" and "edit" pages now consistently show whether a project is active (instead of displaying it's "active" status on the "list" page and it's "inactive" status on the "edit" page).
The database and corresponding data structures in the Java classes have been adapted accordingly.


Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
```diff
@@ -1986,12 +1986,12 @@ public void setShowClosedProcesses(boolean showClosedProcesses) {
         this.showClosedProcesses = showClosedProcesses;
     }
 
-    public void setShowArchivedProjects(boolean showArchivedProjects) {
-        this.showArchivedProjects = showArchivedProjects;
+    public void setShowInactiveProjects(boolean showInactiveProjects) {
```
henning-gerhardt: As you changing the method name and signature, please add JavaDoc to this public method.

Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
```diff
@@ -1986,12 +1986,12 @@ public void setShowClosedProcesses(boolean showClosedProcesses) {
...
+        this.showInactiveProjects = showInactiveProjects;
     }
 
-    public boolean isShowArchivedProjects() {
-        return this.showArchivedProjects;
+    public boolean isShowInactiveProjects() {
```
henning-gerhardt: As you changing the method name and signature, please add JavaDoc to this public method.

PR #1356, opened 2018-03-19
Migrate to library javax.json for JSON objects

Beacze:

PR #1361, opened 2018-03-20
Displaying only tasks that belong to a process and to a template

IkramMaalej:


Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -107,6 +107,14 @@ private BoolQueryBuilder createUserTaskQuery(User user) {
         query.must(createSimpleQuery("processingStatus", TaskStatus.LOCKED.getValue(), false));
         query.must(createSimpleQuery("processingStatus", TaskStatus.DONE.getValue(), false));
 
+        List<JSONObject> templateProcesses = serviceManager.getTemplateService().findByTemplate(true, null);
```
henning-gerhardt: Following code is already existing in method `getAmountOfCurrentTasks`. Instead of duplicating the code, the code should be moved in a separate method and called from both methods.

henning-gerhardt:
There is a merge conflict in

- Kitodo/src/main/java/org/kitodo/services/data/TaskService.java

Please solve / fix it. In #1356 is usage of Json based objects changed as a different Json library is now used!

PR #1367, opened 2018-03-22
Add link to create new process template

solth:
~~… processes without template~~

This pull request slightly changes the way how ~~processes and~~ process templates are created in the new frontend :

- the "New process template" sub-button under the "New" button on the `projects` page now leads to the `processEdit` page with corresponding "process template" labels instead of "process" labels in the header and breadcrumb menu (this link still led to a page in the old design until now).
- ~~the "New process" sub-button under the "New" button on the `processes` page and the "is template" switch on the "Create new process (template)" page have been removed until we decide whether this way to create new processes without templates is really required or not~~
- ~~instead, new processes now always have to be created using the "Create process from template" button in the template list of "Process templates" tab on the `projects` page.~~

**Note:** I restored the controversial part (creating processes without process templates) so that this PR only contains clear fixes.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/processEdit/details.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/processEdit/details.xhtml
@@ -61,12 +61,13 @@
                 </p:selectOneMenu>
             </div>
 
-            <div>
-                <p:outputLabel for="isTemplate" value="#{msgs.istTemplate}"/>
-                <p:selectBooleanCheckbox id="isTemplate" value="#{ProzessverwaltungForm.process.template}">
-                    <p:ajax event="change" oncomplete="jQuery('#editForm\\:saveButtonToggler').click()"/>
-                </p:selectBooleanCheckbox>
-            </div>
+            <!-- TODO: decide whether the following switch is required -->
+            <!--<div>-->
+                <!--<p:outputLabel for="isTemplate" value="#{msgs.istTemplate}"/>-->
+                <!--<p:selectBooleanCheckbox id="isTemplate" value="#{ProzessverwaltungForm.process.template}">-->
+                    <!--<p:ajax event="change" oncomplete="jQuery('#editForm\\:saveButtonToggler').click()"/>-->
+                <!--</p:selectBooleanCheckbox>-->
+            <!--</div>-->
         </p:row>
     </p:panelGrid>
 </ui:composition>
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/webapp/pages/processes.xhtml
```diff
@@ -22,7 +22,8 @@
             <h3>#{msgs.prozessverwaltung}</h3>
             <h:form id="addForm">
                 <p:menuButton value="#{msgs.newElement}" icon="fa fa-plus-circle fa-lg" iconPos="right" styleClass="button-filled-green">
-                    <p:menuitem value="#{msgs.newProcess}" outcome="/pages/processEdit" icon="ui-icon-plusthick" />
+                    <!-- TODO: decide whether creating processes from scratch - without template - is required (than we need the next line) -->
```
henning-gerhardt: In description of this pull request you are writing that it should not possible any more to create processes without a process template. I read this as this lines can be removed.

Kitodo/src/main/webapp/WEB-INF/templates/includes/processEdit/details.xhtml
```diff
@@ -61,12 +61,13 @@
                 </p:selectOneMenu>
             </div>
 
-            <div>
-                <p:outputLabel for="isTemplate" value="#{msgs.istTemplate}"/>
-                <p:selectBooleanCheckbox id="isTemplate" value="#{ProzessverwaltungForm.process.template}">
-                    <p:ajax event="change" oncomplete="jQuery('#editForm\\:saveButtonToggler').click()"/>
-                </p:selectBooleanCheckbox>
-            </div>
+            <!-- TODO: decide whether the following switch is required -->
```
henning-gerhardt: Maybe template handling should be independent of process handling. They are similar but not equal in every aspect.

Kitodo/src/main/webapp/pages/processes.xhtml
solth: No, the current implementation is not final:

> the "New process" sub-button under the "New" button on the processes page and the "is template" switch on the "Create new process (template)" page have been removed **until we decide whether this way to create new processes without templates is really required or not**

Kitodo/src/main/webapp/WEB-INF/templates/includes/processEdit/details.xhtml
solth: Yes, that is the reasoning why the "template" switch has been removed from the `processEdit` page for now. I outcommented the code instead of removing it entirely, since the decision whether this functionality will still be required in the future or not has not been made, yet.

henning-gerhardt:
Until final decision is made I will ignore this pull request.

solth:
I restored the `template` switch on the `processEdit` page and the link to create a new process without template on the `processes` page, so that this Pull Request now is just about fixing the link to create a new process template from the `projects` page (which led to an old page until now).

PR #1384, opened 2018-03-28
Bring client page layouts into line with other list and edit page layouts

solth:


henning-gerhardt:
Selenium tests are failing every time. You must maybe adjust the Selenium tests for this client page. Maybe you should as @tim-boerner for some more information?

PR #1388, opened 2018-04-04
Purge old pages

solth:
This pull request removes the templates of the old Kitodo.Production layout and the `redirectTo` methods in the form classes (these were only intended for the transition period in which new and old templates co-existed and the form classes needed to distinguish between calls from old pages and new pages in order to return navigation targets accordingly). The `redirectTo` methods are replaced by static strings leading to specific new templates.


Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -613,7 +610,7 @@ private void schrittPerParameterLaden() throws DAOException {
              * nachholen, damit die Liste vollstÃ¤ndig ist
              */
             if (this.page == null &&  getUser() != null) {
-                redirectToList();
+                // TODO: check if we still need this!
```
henning-gerhardt: - If you removing this method call then no different code is executed. Maybe you should first investigate if this method call is still needed and if not then remove it.
- In previous versions of this code the list of tasks / steps was refreshed (old and already removed `filterAll()` method) and a pagination object got new data. I don't know which error or not-refreshment was solved in the past.

Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -599,7 +596,7 @@ private void setAttributesForProcess() {
     }
 
     /**
-     * prüfen, ob per Parameter vielleicht zunÃ¤chst ein anderer geladen werden
+     * prüfen, ob per Parameter vielleicht zunächst ein anderer geladen werden
```
henning-gerhardt: On fixing the German umlaut issue maybe you can translate this JavaDoc into English?

Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
solth: This method call is unnecessary because `redirectToList` did not perform any internal program logic but instead only returned a string, which here obviously wasn't used - because at this place we only had a `redirectToList()` instead of `return redirectToList()` in other places. This method call to `redirectToList` was probably never working to begin with.

henning-gerhardt:
There are merge conflicts in

- Kitodo/src/main/java/de/sub/goobi/forms/ProjekteForm.java
- Kitodo/src/main/webapp/pages/projectEditDetail.xhtml
- Kitodo/src/main/webapp/pages/projectEditMets.xhtml
- Kitodo/src/main/webapp/pages/projectEditSpecification.xhtml 

Please solve them.

solth:
Rebased branch and resolved conflicts.

Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -481,7 +478,7 @@ public void calcHomeImages() {
      */
     public Task getMySchritt() {
         try {
-            schrittPerParameterLaden();
+            loadStepByParameter();
```
Beacze: Small remark: we have decided to translate schritt to task.

henning-gerhardt:
Some comments for German speaking files which are left over after your removing:

`incMeta/BaumZumVerschieben.xhtml`:
- used in `incMeta/StrukturdatenOld.xhtml` line 427
- could be removed if `incMeta/StrukturdatenOld.xhtml` is removed

`incMeta/File.xhtml`:
- no usage in any xhtml files
- could be removed

`incMeta/NeuMeta.xhtml`:
- no usage in any xhtml files
- could be removed

`incMeta/NeuPerson.xhtml`:
- no usage in any xhtml files
- could be removed

`incMeta/PersonenUndMetadaten.xhtml`:
- no usage in any xhtml files
- could be removed

`incMeta/StrukturdatenOld.xhtml` referenced in 
- `Kitodo/src/main/resources/messages/messages_en.properties` 
- `Kitodo/src/main/resources/messages/messages_de.properties`
- no usage in any xhtml files
- could be removed

`inc_Main/box3.xhtml`:
- no usage in any xhtml files
- could be removed

`inc_Prozessverwaltung/Prozesse_Liste_Action.xhtml`:
- used in `ProzessverwaltungAlle.xhtml` line 71
- could be remoeved if `ProzessverwaltungAlle.xhtml` is removed

`ProjekteBearbeiten.xhtml`:
- no usage in any xhtml files
- could be removed

`ProzessverwaltungAlle.xhtml` referenced in
- `Kitodo/src/main/webapp/pages/incMeta/metadataEditorHeader.xhtml` line 28
- `Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java` line 264
- no usage in any xhtml files
- could be removed if references are fixed

PR #1402, opened 2018-04-17
Hibernate without ssl

henning-gerhardt:
Flyway on Travis throws ugly messages

```
Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.
```
on every run. For now I'm disabling SSL access for flyway and normal database access. Changing this values from `false` to `true` re-enable it on established secure connections.

### DIFF
--- a/Kitodo/src/main/resources/hibernate.cfg.xml
+++ b/Kitodo/src/main/resources/hibernate.cfg.xml
@@ -24,12 +24,17 @@
         <property name="hibernate.connection.driver_class">com.mysql.jdbc.Driver</property>
 
         <property name="hibernate.connection.url">
-            jdbc:mysql://localhost/kitodo
+            jdbc:mysql://localhost/kitodo?useSSL=false
         </property>
         <property name="hibernate.connection.username">kitodo</property>
         <property name="hibernate.connection.password">kitodo</property>
         <property name="hibernate.connection.autoReconnect">true</property>
         <property name="hibernate.connection.autoReconnectForPools">true</property>
+
+        <!-- SSL connection properties -->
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/resources/hibernate.cfg.xml
```diff
@@ -24,13 +24,18 @@
...
         </property>
         <property name="hibernate.connection.username">kitodo</property>
         <property name="hibernate.connection.password">kitodo</property>
         <property name="hibernate.connection.autoReconnect">true</property>
         <property name="hibernate.connection.autoReconnectForPools">true</property>
 
+        <!-- SSL connection properties -->
+        <property name="hibernate.connection.requireSSL">false</property>
+        <property name="hibernate.connection.verifyServerCertificate">false</property>
+        <property name="hibernate.connection.useSSL">false</property>
```
Beacze: Why this property is defined here and also in the URL?

henning-gerhardt: It must be defined on both as it used in different ways inside Hibernate and it is depending on used MySQL Server version.

henning-gerhardt: Explanation: the value of the property is used on other places where the URL is not interpreted and the value in the URL is not set as property which is used on other places. So it must be defined on two places.

PR #1423, opened 2018-04-27
Add workflow id to task table

Beacze:
Needed to track changes done to tasks.

Depends on #1420

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/workflow/model/beans/KitodoTask.java
+++ b/Kitodo/src/main/java/org/kitodo/workflow/model/beans/KitodoTask.java
@@ -17,6 +17,7 @@ import org.camunda.bpm.model.bpmn.instance.Task;
 
 public class KitodoTask {
 
+    private String workflowId;
     private String title;
     private Integer priority;
     private Integer ordering;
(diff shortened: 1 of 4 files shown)


Kitodo/src/main/java/org/kitodo/workflow/model/beans/KitodoTask.java
```diff
@@ -43,6 +44,7 @@
      *            * determined out of sequence flow
      */
     public KitodoTask(Task task, int ordering) {
+        this.workflowId = task.getId();
```
henning-gerhardt: You are sure that this assignment is correct? Should it not be `task.getWorkflowId()`?

Beacze: Yes, it is correct. Task object is a task from BPMN model (not our bean).

henning-gerhardt: This is a little bit confusing....

Beacze: I know but I can't change name of the class inside the bpmn model library.

PR #1429, opened 2018-05-02
Add copy constructor for Task and use it in BeanHelper

beatrycze-volk:


henning-gerhardt:
There are merge conflicts in

- Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Task.java
- Kitodo/src/main/java/de/sub/goobi/helper/BeanHelper.java

Please solve this conflicts.

PR #1434, opened 2018-05-03
Update template after diagram is updated

Beacze:
Depends on #1433


Beacze:
Work on this feature is right now paused so I close this PR.

PR #1440, opened 2018-05-07
Remove docket and ruleset from workflow

Beacze:
Docket and ruleset are assigned to template later. Workflow doesn't store this information.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/workflow/model/Updater.java
+++ b/Kitodo/src/main/java/org/kitodo/workflow/model/Updater.java
@@ -40,9 +40,6 @@ public class Updater {
         List<Process> processes = this.template.getProcesses();
 
         for (Process process : processes) {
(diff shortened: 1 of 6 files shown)


Kitodo/src/main/java/org/kitodo/workflow/model/Updater.java
```diff
@@ -40,8 +40,6 @@ public void updateProcessesAssignedToTemplate() throws DataException {
         List<Process> processes = this.template.getProcesses();
 
         for (Process process : processes) {
-            process.setDocket(this.template.getDocket());
-            process.setRuleset(this.template.getRuleset());
```
henning-gerhardt: You can remove following empty line too.

Beacze: Removed

PR #1442, opened 2018-05-07
Fix CERT bugs

Beacze:
Replace deprecated ValueBinding: https://thomaswabner.wordpress.com/2007/09/27/replacing-deprecated-valuebindung-stuff-from-jsf-with-elresolver/


Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
```diff
@@ -493,7 +495,21 @@ public static void setCurrentUser(User user) {
...
-        return (String) Helper.getManagedBeanValue("#{LoginForm.myBenutzer.metadataLanguage}");
+        return Helper.getCurrentUser().getMetadataLanguage();
+    }
+
+    /**
+     * Get AktuelleSchritteForm.
+     *
+     * @return AktuelleSchritteForm
+     */
+    public static AktuelleSchritteForm getAktuelleSchritteForm() {
```
henning-gerhardt: Please change method name as they should be in English. I know that the German name is used still as class name.

Kitodo/src/main/java/org/goobi/mq/processors/CreateNewProcessProcessor.java
```diff
@@ -211,22 +213,19 @@ private static Template getTemplateByTitle(String templateTitle) {
...
-     * @throws IllegalArgumentException
-     *             if a docType is not applicable to the template or the docType
-     *             isn’t valid
      */
     private static boolean docTypeIsPossible(ProzesskopieForm dialog, String docType) {
         Boolean fieldIsUsed = dialog.getStandardFields().get("doctype");
         if (fieldIsUsed == null || fieldIsUsed.equals(Boolean.FALSE)) {
-            throw new IllegalArgumentException(
-                    "Bad argument “docType”: Selected template doesn’t provide " + "the standard field “doctype”.");
+            logger.warn("Bad argument “docType”: Selected template doesn’t provide the standard field “doctype”.");
```
henning-gerhardt: - Did you checked if return value of this method is used in calling methods? 
- Did you checked if previous thrown exception is handled in calling methods?

Kitodo/src/main/java/org/goobi/mq/processors/CreateNewProcessProcessor.java
```diff
@@ -238,8 +237,9 @@ private static boolean docTypeIsPossible(ProzesskopieForm dialog, String docType
         if (valueIsValid) {
             return true;
         }
-        throw new IllegalArgumentException("Bad argument “docType”: Selected template doesn’t provide "
+        logger.warn("Bad argument “docType”: Selected template doesn’t provide "
```
henning-gerhardt: Same as above

Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
Beacze: Hmmm... What can be correct English name for it?

henning-gerhardt:
Selenium tests are failing as you changed unintentionally something on the user / user group page. Design looks different and it is not possible any more to do something on this page.

Kitodo/src/main/java/org/goobi/mq/processors/CreateNewProcessProcessor.java
Beacze: Yes, it is used:

```java
if (docType != null && docTypeIsPossible(newProcess, docType)) {
```

But now when I take a second look at this both methods I have still feeling that something is not right with them...

Beacze:
I have removed commit causing problems in selenium tests.

henning-gerhardt:

- Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java

Please fix this merge conflict.

Kitodo/src/main/java/org/goobi/mq/processors/CreateNewProcessProcessor.java
```diff
@@ -27,6 +27,9 @@
 import org.apache.logging.log4j.Logger;
 import org.goobi.mq.ActiveMQProcessor;
 import org.goobi.mq.MapMessageObjectReader;
+import org.kitodo.api.ugh.exceptions.PreferencesException;
```
henning-gerhardt: Why did you add this and the following two exception classes which are not used?

PR #1453, opened 2018-05-14
Change handling for RuntimeException in getSessionFactory

Beacze:


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/HibernateUtil.java
```diff
@@ -67,10 +67,9 @@ private static SessionFactory getSessionFactory() {
                 Metadata metadata = sources.getMetadataBuilder().build();
                 sessionFactory = metadata.getSessionFactoryBuilder().build();
             } catch (RuntimeException e) {
-                logger.error(e.getMessage(), e);
-                if (registry != null) {
-                    StandardServiceRegistryBuilder.destroy(registry);
-                }
+                logger.error(
+                    "Probably you have forgotten to adjust your Hibernate configuration! Error: " + e.getMessage(), e);
```
henning-gerhardt: As any `RuntimeException` based exception is fetched any kind of error raised this exception class. One error reason could be not adjusted Hibernate configuration. So the logged message could be misleading or more confusing.

PR #1454, opened 2018-05-14
Fix class attributes

Beacze:


Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
```diff
@@ -596,4 +596,12 @@ public static String getLastMessage() {
         return compoundMessage;
     }
 
+    /**
+     * Set activeMQReporting.
+     *
+     * @param activeMQReporting as java.util.Map<java.lang.String,java.lang.String>
+     */
+    public static void setActiveMQReporting(Map<String, String> activeMQReporting) {
+        Helper.activeMQReporting = activeMQReporting;
```
henning-gerhardt: Instead of `Helper.activeMQReporting = activeMQReporting;` it should be `this.activeMQReporting = activeMQReporting;` or?

Beacze: Helper is a static class so it shouldn't be accessed by instance (this).

henning-gerhardt:
There are merge conflicts in

- Kitodo/src/main/java/de/sub/goobi/export/dms/DmsImportThread.java
- Kitodo/src/main/java/de/sub/goobi/export/dms/ExportDms.java
- Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
- Kitodo/src/main/java/org/kitodo/services/data/ProcessService.java 

Please solve this conflicts.

PR #1455, opened 2018-05-14
Remove unnecessary null check - instanceof returns false for nulls

Beacze:


henning-gerhardt:

- Kitodo/src/main/java/org/goobi/mq/MapMessageObjectReader.java

Please solve this conflict.

PR #1470, opened 2018-05-17
Fix exception throw config

Beacze:
Depends on #1469


Kitodo/src/test/resources/kitodo_config.properties
```diff
@@ -27,6 +27,8 @@ diagramsFolder=src/test/resources/diagrams/
 # Script to create the directory for a new process
 script_createDirMeta=src/test/resources/scripts/script_createDirMeta.sh
 script_createDirUserHome=src/test/resources/scripts/script_createDirUserHome.sh
+script_createSymLink=src/test/resources/scripts/script_createSymLink.sh
```
henning-gerhardt: Does this and the following changed lines belong to this pull request?

Beacze: Yes, as without this parameter tests in WorkflowControllerServiceIT fail.

PR #1498, opened 2018-05-28
Remove not needed config file name and use kitodo opac config file name

Beacze:
Depends on #1497

### DIFF
--- a/Kitodo/src/main/java/org/goobi/production/constants/FileNames.java
+++ b/Kitodo/src/main/java/org/goobi/production/constants/FileNames.java
@@ -17,10 +17,6 @@ package org.goobi.production.constants;
  * @author Matthias Ronge &lt;matthias.ronge@zeutschel.de&gt;
  */
 public class FileNames {
-    /**
-     * Production main configuration file name.
-     */
-    public static final String CONFIG_FILE = "kitodo_config.properties";
 
     /**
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/java/org/goobi/production/constants/FileNames.java
```diff
@@ -17,10 +17,6 @@
  * @author Matthias Ronge &lt;matthias.ronge@zeutschel.de&gt;
  */
 public class FileNames {
-    /**
-     * Production main configuration file name.
-     */
-    public static final String CONFIG_FILE = "kitodo_config.properties";
```
henning-gerhardt: Instead of fixing usage of this constant value in `Kitodo-API/src/main/java/org/kitodo/config/Config.java` and `Kitodo-DataManagement/src/main/java/org/kitodo/config/ConfigMain.java`, you are removing it?

Beacze: Those two places don't see this attribute - both of those modules can't access Core module.

henning-gerhardt: So this file name must be maintained in many places. I don't like this solution but you must have good reasons to do it in this way.

PR #1503, opened 2018-05-28
Remove some of remaining filterAll methods

Beacze:


Kitodo/src/main/java/de/sub/goobi/forms/AktuelleSchritteForm.java
```diff
@@ -84,7 +83,7 @@
     private transient ServiceManager serviceManager = new ServiceManager();
     private static final String ERROR_LOADING = "errorLoadingOne";
     private static final String WORK_TASK = "arbeitsschritt";
-    private String taskListPath = MessageFormat.format(REDIRECT_PATH, "tasks");
+    public String taskListPath = MessageFormat.format(REDIRECT_PATH, "tasks");
```
henning-gerhardt: If this content of this variable is used on other places, use a getter Method to access it instead of making this variable public available.

PR #1506, opened 2018-05-29
Fix conditional statements and asserts

Beacze:


Kitodo/src/main/java/de/sub/goobi/helper/tasks/CreatePdfFromServletThread.java
```diff
@@ -150,7 +150,7 @@ public void run() {
                 logger.error("URI " + this.targetFolder + this.getProcess().getTitle() + ".PDF-ERROR.log is malformed",
                     e1);
             } catch (IOException e1) {
-                logger.error("Ressource " + uri + " could not be created", e);
+                logger.error("Resource could not be created", e);
```
henning-gerhardt: Why did you remove the name / content / value of not created URI? This could help on search for error reason.

Kitodo/src/main/java/org/goobi/production/model/bibliography/course/Block.java
```diff
@@ -263,7 +264,7 @@ public boolean isEmpty() {
      * @return whether the given string is equals to the assigned variant
      */
     public boolean isIdentifiedBy(String variant) {
-        return variant == null && this.variant == null || this.variant.equals(variant);
+        return variant == null && this.variant == null || Objects.nonNull(this.variant) && this.variant.equals(variant);
```
henning-gerhardt: Should comparing to null not replaced by `Objects.isNull` or `Objects.nonNull`?

Kitodo/src/main/java/de/sub/goobi/helper/tasks/CreatePdfFromServletThread.java
Beacze: uri is null here

henning-gerhardt: I see it now and removing is ok in this case.

PR #1525, opened 2018-06-04
Remove Util class

Beacze:


Beacze:
It was class with one method which had one line. I have ported it but question is if this set up of Map size is needed at all...

henning-gerhardt:
This code is written by @matthias-ronge. Maybe he can answer this question?

matthias-ronge:
By default, a map is initialized with a capacity of 16 and a load factor of 0.75. When adding more than 0.75*16=12 elements, it is resized and the contents re-mapped. I thought that was inefficient, if the size is known in advance, it can be calculated.

The `Util` can be removed because the same function is available from [MemoryResult, ll. 65-81](https://github.com/kitodo/kitodo-production/blob/fc399949bd4b088f2b152b5287da61d7503423e7/Kitodo-DataAccess/src/main/java/org/kitodo/dataaccess/storage/memory/MemoryResult.java#L65-L81).

henning-gerhardt:
Thank you for the explanation, Matthias. But we can not use it from a module which depends on Core module inside the core module. I think Beatrycze solution is good for now.

matthias-ronge:
Yes, it is just the answer for the question.

henning-gerhardt:

- Kitodo/src/main/java/de/sub/goobi/helper/Helper.java

Please solve this conflict.

PR #1526, opened 2018-06-04
Improvements File Management module

Beacze:


Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -446,6 +450,7 @@ public boolean createSymLink(URI homeUri, URI targetUri, boolean onlyRead, Strin
         File imagePath = new File(fileMapper.mapAccordingToMappingType(homeUri));
         File userHome = new File(getDecodedPath(targetUri));
         if (userHome.exists()) {
+            System.out.println("Got here!");
```
henning-gerhardt: Is this line is really needed here?

Beacze: Definitely not :D

PR #1547, opened 2018-06-07
Remove not used methods from LoginForm and rename remaining ones

Beacze:


henning-gerhardt:

- Kitodo/src/main/java/de/sub/goobi/forms/LoginForm.java

Please solve this conflict.

PR #1548, opened 2018-06-07
Add selenium test for delete process

Beacze:


Kitodo/src/test/java/org/kitodo/selenium/RemovingST.java
```diff
@@ -0,0 +1,52 @@
...
+    public void removeProcessTest() throws Exception {
+        if (SystemUtils.IS_OS_LINUX) {
+            File scriptCreateDirMeta = new File(ConfigCore.getParameter("script_createDirMeta"));
+            File scriptCreateDirUserHome = new File(ConfigCore.getParameter("script_createDirUserHome"));
+            ExecutionPermission.setExecutePermission(scriptCreateDirMeta);
+            ExecutionPermission.setExecutePermission(scriptCreateDirUserHome);
+        }
+
+        int numberOfProcessesDisplayed = Pages.getProcessesPage().countListedProcesses();
+        Assert.assertTrue("Process list is empty", numberOfProcessesDisplayed > 0);
```
tim-boerner: This should never fail, because even if there are no processes, we have at least one row in our table which is indicating that there are no processes. So the numberOfProcessesDisplayed will be 1.

Beacze: Can't you sleep? :grinning:

Good point. I will change condition.

PR #1553, opened 2018-06-08
Remove not used maven dependencies from Kitodo Core

Beacze:


Kitodo/pom.xml
```diff
@@ -289,15 +268,19 @@
                 </exclusion>
             </exclusions>
         </dependency>
-
         <!-- https://mvnrepository.com/artifact/org.apache.commons/commons-email -->
+        <dependency>
```
henning-gerhardt: Comment above did not fit to this dependency as this comment belong to the next dependency.

Kitodo/pom.xml
```diff
@@ -321,14 +304,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
-
-        <!-- https://mvnrepository.com/artifact/io.reactivex.rxjava2/rxjava -->
```
henning-gerhardt: Why did you remove this valid comment?

PR #1559, opened 2018-06-11
Remove a penetration on the innards of rt.jar

matthias-ronge:
Classes from the sun.* package hierarchy should not be used. Also, documents in the API that an exception can be expected here.

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/persistentidentifier/PersistentIdentifierGeneratorInterface.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/persistentidentifier/PersistentIdentifierGeneratorInterface.java
@@ -16,10 +16,12 @@ public interface PersistentIdentifierGeneratorInterface {
 
     /**
      * Registers the given identifier.
-     * 
+     *
      * @param urn
      *            the identifier to register.
      * @return true, if successful, false otherwise.
+     * @throws UnsupportedOperationException
+     *             if the register operation is not supported by this generator
      */
     boolean register(String urn);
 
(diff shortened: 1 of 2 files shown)


matthias-ronge:
@henning-gerhardt please request a review to someone bc I am not allowed to

This fixes an error report from Eclipse, there is no ticket for.

Kitodo-API/src/main/java/org/kitodo/api/persistentidentifier/PersistentIdentifierGeneratorInterface.java
```diff
@@ -16,10 +16,12 @@
 
     /**
      * Registers the given identifier.
-     * 
+     *
      * @param urn
      *            the identifier to register.
      * @return true, if successful, false otherwise.
+     * @throws UnsupportedOperationException
```
Beacze: It is interface, it doesn't need to have in JavaDoc UnsupportedOperationException. How can you be sure that there is not other implementation which implements this method?

matthias-ronge: There is at least one (the current) implementation that actually throws this exception. Documenting it means:
* to the users of the interface: that they must be aware of the fact that implementations may throw this (unchecked) exception (it is not an error) and handle the situation gracefully
* to implementers of this interface: that in the described case, they should throw *this* exception, and not some other.

See [Iterator.remove()](https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#remove--) for a corresponding example in the Java API.

PR #1560, opened 2018-06-11
Remove unknown plugins and GoobiHotfolder stuff

matthias-ronge:
Both "command" and "webapi" plugin types are not documented and not backed with open source plug-ins. Same is true for GoobiHotfolder stuff. This is to remove unused code.


matthias-ronge:
@henning-gerhardt please add a reviewer

Kitodo/src/main/java/org/goobi/production/flow/jobs/AbstractGoobiJob.java
```diff
@@ -1,91 +0,0 @@
...
- *
- * For the full copyright and license information, please read the
- * GPL3-License.txt file that was distributed with this source code.
- */
-
-package org.goobi.production.flow.jobs;
-
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.quartz.Job;
```
henning-gerhardt: As you remove last usage of `org.quartz` library the dependency in `Kitodo/pom.xml` and file `Kitodo/src/main/resources/quartz.properties` should be removed.

matthias-ronge: Good catch 👍

matthias-ronge:
Quartz removed: Less sand in the gears. ;-)

PR #1565, opened 2018-06-12
Unify all hibernate config files and change session context to thread

Beacze:
current_session_context_class: http://docs.jboss.org/hibernate/orm/4.3/manual/en-US/html/ch02.html#architecture-current-session

### DIFF
--- a/Kitodo/src/main/resources/hibernate.cfg.xml
+++ b/Kitodo/src/main/resources/hibernate.cfg.xml
@@ -52,7 +52,7 @@
         <property name="hibernate.cache.region.factory_class">org.hibernate.cache.ehcache.EhCacheRegionFactory</property>
 
         <!-- Enable Hibernate's automatic session context management -->
-        <property name="current_session_context_class">managed</property>
+        <property name="current_session_context_class">thread</property>
 
         <!-- Fix org.hibernate.LazyInitializationException: could not initialize proxy - no Session -->
         <property name="hibernate.enable_lazy_load_no_trans">true</property>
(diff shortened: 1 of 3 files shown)


Kitodo/src/test/java/org/kitodo/MockDatabase.java
```diff
@@ -824,10 +824,10 @@ private static void insertProjectFileGroups() throws DAOException, DataException
         fifthProjectFileGroup.setProject(project);
         serviceManager.getProjectFileGroupService().saveToDatabase(fifthProjectFileGroup);
 
-        project.getProjectFileGroups().add(fifthProjectFileGroup);
-        project.getProjectFileGroups().add(fifthProjectFileGroup);
-        project.getProjectFileGroups().add(fifthProjectFileGroup);
-        project.getProjectFileGroups().add(fifthProjectFileGroup);
+        project.getProjectFileGroups().add(firstProjectFileGroup);
```
henning-gerhardt: This changes did not fit in this commit nor pull request. Can you please move it to a separate commit or better pull request?

Kitodo/src/main/resources/hibernate.cfg.xml
```diff
@@ -52,7 +52,7 @@
         <property name="hibernate.cache.region.factory_class">org.hibernate.cache.ehcache.EhCacheRegionFactory</property>
 
         <!-- Enable Hibernate's automatic session context management -->
-        <property name="current_session_context_class">managed</property>
+        <property name="current_session_context_class">thread</property>
```
henning-gerhardt: What is the reason to switch from `managed` to `thread`? What kind of issue did you fix with this change?

Beacze: thread: current sessions are tracked by thread of execution.

managed: current sessions are tracked by thread of execution. However, you are responsible to bind and unbind a Session instance with static methods on this class: it does not open, flush, or close a Session.

I suspect that thread is more forgiving than managed...

henning-gerhardt: Thank you for the explanation.

PR #1570, opened 2018-06-13
Changes in hibernate session management

Beacze:
- use try with resources
- change sync object
- adjust tests

Edit:
- selenium test for remove was failing - cause delete of process needed to be changed

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
+++ b/Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
@@ -41,6 +41,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import javax.enterprise.context.SessionScoped;
 import javax.faces.context.FacesContext;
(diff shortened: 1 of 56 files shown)


Kitodo/src/test/java/org/kitodo/services/workflow/WorkflowControllerServiceIT.java
```diff
@@ -228,6 +228,7 @@ public void shouldSolveProblem() throws Exception {
         Process process = currentTask.getProcess();
         for (Task task : process.getTasks()) {
             if (currentTask.getOrdering() < task.getOrdering() && task.getOrdering() < correctionTask.getOrdering()) {
+                System.out.println(task.getTitle() + " " + task.getProcessingStatusAsString());
```
henning-gerhardt: I don't think, that this line is needed or?

Kitodo/src/main/java/de/sub/goobi/forms/ProzessverwaltungForm.java
```diff
@@ -206,14 +208,19 @@ public String saveAndRedirect() {
...
+            this.process.setTemplate(null);
+            List<Batch> batches = new CopyOnWriteArrayList<>(process.getBatches());
+            for (Batch batch : batches) {
                 batch.getProcesses().remove(this.process);
+                this.process.getBatches().remove(batch);
+                serviceManager.getBatchService().save(batch);
             }
             serviceManager.getProcessService().remove(this.process);
-        } catch (DataException e) {
+        } catch (DataException | RuntimeException e) {
```
henning-gerhardt: Catching a `RuntimeException` is really necessary here? Who is throwing this exception?

Beacze: Yes, it is necessary as if there happens some other problem e.g. NPE, application fails silently.

PR #1572, opened 2018-06-14
Implement ImageManagementInterface

matthias-ronge:
Creates new module ImageManagementModule

PR #1574, opened 2018-06-15
Update dependencies

Beacze:


Kitodo-DataManagement/pom.xml
```diff
@@ -142,9 +122,9 @@
             <version>${hibernate.version}</version>
         </dependency>
         <dependency>
-            <groupId>commons-lang</groupId>
-            <artifactId>commons-lang</artifactId>
-            <version>2.6</version>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.7</version>
```
henning-gerhardt: This library is used many times, please add a global version number property. This could be conflicting with needed changes on PR #1572 !

Kitodo/pom.xml
```diff
@@ -97,7 +97,7 @@
         <dependency>
             <groupId>commons-cli</groupId>
             <artifactId>commons-cli</artifactId>
-            <version>1.2</version>
+            <version>1.4</version>
```
henning-gerhardt: `commons-cli` in version 1.2 is still used in  `OpacPica-Plugin`. So if you did not change it there, this library version is still deployed.

Beacze: I don't see this one in pom of OpacPica-Plugin.

PR #1575, opened 2018-06-15
Replace anonymous inner classes with lambdas

Beacze:


Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
```diff
@@ -507,65 +500,58 @@ public static AktuelleSchritteForm getCurrentTaskForm() {
...
-            regexList.add(prefix + "\\.[jJ][pP][2]");
-            regexList.add(prefix + "\\.[pP][nN][gG]");
-            regexList.add(prefix + "\\.[gG][iI][fF]");
-
-            for (String regex : regexList) {
-                if (name.matches(regex)) {
-                    fileOk = true;
-                }
-            }
+    public static final FilenameFilter imageNameFilter = (dir, name) -> {
```
henning-gerhardt: Missing JavaDoc for public method. Or move this code into its own class to get ride out of this inner class usage.

Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
```diff
@@ -507,65 +500,58 @@ public static AktuelleSchritteForm getCurrentTaskForm() {
...
-            regexList.add(prefix + "\\.xls");
-            regexList.add(prefix + "\\.xlsx");
-            regexList.add(prefix + "\\.pptx");
-            regexList.add(prefix + "\\.ppt");
-
-            for (String regex : regexList) {
-                if (name.matches(regex)) {
-                    fileOk = true;
-                }
+    public static final FilenameFilter dataFilter = (dir, name) -> {
```
henning-gerhardt: Missing JavaDoc for public method. Or move this code into its own class to get ride out of this inner class usage.

Kitodo/src/main/java/de/sub/goobi/helper/tasks/EmptyTask.java
```diff
@@ -58,20 +58,10 @@
...
-         * @see java.lang.Thread.UncaughtExceptionHandler#uncaughtException(java.lang.Thread,
-         *      java.lang.Throwable)
-         */
-        @Override
-        public void uncaughtException(Thread origin, Throwable exception) {
-            if (origin instanceof EmptyTask) {
-                EmptyTask task = (EmptyTask) origin;
-                task.setException(exception);
-            }
+    public static final Thread.UncaughtExceptionHandler CATCH_ALL = (origin, exception) -> {
```
henning-gerhardt: JavaDoc did not matched or move this code into its own class to get ride out of this inner class usage.

Beacze: JavaDocs can't be made for lambdas - param attributes are not allowed there - and it is constant.

Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
Beacze: JavaDocs can't be made for lambdas - param attributes are not allowed there - and it is constant.

Kitodo/src/main/java/de/sub/goobi/helper/Helper.java
Beacze: JavaDocs can't be made for lambdas - param attributes are not allowed there - and it is constant.

henning-gerhardt: I would move this code to its own classes with proper JavaDoc and explanation and add tests for this class to prove its correct implementaion. Maybe I'm a old school programmer but this is more readable then this lambda expression.

Beacze: I have simplified this two lambdas.

PR #1585, opened 2018-06-18
Minor fixes

Beacze:
- remove unnecessary unboxing
- remove redundant substring calls
- fix arrays: singlet list and creation without rigid size list.toArray(new String[0]) (more efficient since java 6)
- replace Collections.sort with list.sort
- replace if with switch

### DIFF
--- a/Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
+++ b/Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
@@ -2853,13 +2853,13 @@ public class Metadaten {
         }
         List<String> newSelectionList = new ArrayList<>();
         for (Integer pageIndex : selectedPages) {
-            DocStructInterface firstpage = allPages.get(pageIndex - 1);
-            DocStructInterface secondpage = allPages.get(pageIndex);
-            switchFileNames(firstpage, secondpage);
+            DocStructInterface firstPage = allPages.get(pageIndex - 1);
+            DocStructInterface secondPage = allPages.get(pageIndex);
+            switchFileNames(firstPage, secondPage);
             newSelectionList.add(String.valueOf(pageIndex - 1));
         }
 
-        allPagesSelection = newSelectionList.toArray(new String[newSelectionList.size()]);
+        allPagesSelection = newSelectionList.toArray(new String[0]);
 
         retrieveAllImages();
         identifyImage(1);
@@ -2892,7 +2892,7 @@ public class Metadaten {
             newSelectionList.add(String.valueOf(pageIndex + 1));
         }
 
-        allPagesSelection = newSelectionList.toArray(new String[newSelectionList.size()]);
+        allPagesSelection = newSelectionList.toArray(new String[0]);
         retrieveAllImages();
         identifyImage(1);
     }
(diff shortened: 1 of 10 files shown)


Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -2859,13 +2859,13 @@ public void moveSeltectedPagesUp() {
...
-            DocStructInterface secondpage = allPages.get(pageIndex);
-            switchFileNames(firstpage, secondpage);
+            DocStructInterface firstPage = allPages.get(pageIndex - 1);
+            DocStructInterface secondPage = allPages.get(pageIndex);
+            switchFileNames(firstPage, secondPage);
             newSelectionList.add(String.valueOf(pageIndex - 1));
         }
 
-        allPagesSelection = newSelectionList.toArray(new String[newSelectionList.size()]);
+        allPagesSelection = newSelectionList.toArray(new String[0]);
```
henning-gerhardt: Why did you changes this? Amount of entries is well known.

Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -2898,7 +2898,7 @@ public void moveSelectedPagesDown() {
             newSelectionList.add(String.valueOf(pageIndex + 1));
         }
 
-        allPagesSelection = newSelectionList.toArray(new String[newSelectionList.size()]);
+        allPagesSelection = newSelectionList.toArray(new String[0]);
```
henning-gerhardt: Why did you changes this? Amount of entries is well known.

Kitodo/src/main/java/de/sub/goobi/metadaten/copier/MetadataPathSelector.java
```diff
@@ -252,7 +252,7 @@ protected String findIn(DocStructInterface supernode) {
      */
     public int getIndex() {
         if (index instanceof Integer) {
-            int a = ((Integer) index).intValue();
+            int a = (Integer) index;
```
henning-gerhardt: `Integer` class can contain `null` value but primitive not. You may replace `((Integer) index).intValue()` with `index.intValue()` to archive same behaviour?

Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
Beacze: Creating array without size is more efficient since java 6.

Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
Beacze: Creating array without size is more efficient since java 6.

Kitodo/src/main/java/de/sub/goobi/metadaten/copier/MetadataPathSelector.java
Beacze: Basically it is the same, doesn't matter if you use index or index.intValue:
https://stackoverflow.com/questions/13914317/assigning-a-null-value-to-an-int

For this case important part is also instance of - it returns false if index is null so assign will not be performed.

henning-gerhardt:

- Kitodo/src/main/java/de/sub/goobi/metadaten/MetadatenHelper.java

Please solve this conflict.

PR #1577, opened 2018-06-15
Added linking of physical and logical structMap at DataEditor module

tim-boerner:
Depends on PR1567

PR #1586, opened 2018-06-18
Change transformation from goobi person to metadata group of new data formate

tim-boerner:
- moved value of "name" attribute to value of "role" metadata child element.

depends on #1577


henning-gerhardt:
There are merge conflicts in

- Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/MetsKitodoWrapper.java
- Kitodo-DataEditor/src/test/java/org/kitodo/dataeditor/MetsKitodoWrapperTest.java
- Kitodo-DataFormat/pom.xml 

Please solve this conflicts.

PR #1587, opened 2018-06-18
Remove c3p0 library

Beacze:
It looks this one is not used anymore. Configuration in hibernate mapping uses hibernate c3p0.

### DIFF
--- a/Kitodo/src/main/resources/c3p0.properties
+++ b/Kitodo/src/main/resources/c3p0.properties
@@ -7,6 +7,14 @@
 #
 # For the full copyright and license information, please read the
 # GPL3-License.txt file that was distributed with this source code.
+
+# This configuration is not direct used inside the application but will
+# be used on usage. This both mentioned properties are for connection
+# testing and is used for unstable or bad programmed connections from
+# application to database.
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/resources/c3p0.properties
```diff
@@ -1,12 +0,0 @@
-#
```
henning-gerhardt: I don't think that removing this configuration file is necessary. Even, if we switch to `hibernate-c3p0`, which brings `c3p0` as a dependency back, the same file is used for configuration.

Beacze: That's true, hibernate imports it, but both properties were marked as not used. I suspect that now whole configuration happens inside the hibernate mapping.

henning-gerhardt: Sure, this configuration is not direct used inside the application but will be used on usage. This both mentioned properties are for [connection testing](https://www.mchange.com/projects/c3p0/#configuring_connection_testing) and was / is used for unstable or bad programmed connections from application to database.

PR #1588, opened 2018-06-19
Add ClassLoaderLeakPreventor

Beacze:
Depends on #1571 

https://stackoverflow.com/questions/7788280/memory-leak-when-redeploying-application-in-tomcat
http://java.jiderhamn.se/2012/03/04/classloader-leaks-vi-this-means-war-leak-prevention-library/

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -488,6 +488,11 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>se.jiderhamn.classloader-leak-prevention</groupId>
+            <artifactId>classloader-leak-prevention-servlet3</artifactId>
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/webapp/WEB-INF/web.xml
```diff
@@ -88,6 +88,12 @@
         <!--<param-value>none</param-value>-->
     <!--</context-param>-->
 
+    <listener>
```
henning-gerhardt: As the application build as a servlet 3.0+ application, then this is not needed any more, according to the [readme of the project](https://github.com/mjiderhamn/classloader-leak-prevention). Maybe you must adjust the pom entry too.

PR #1592, opened 2018-06-19
Add wait for response from ES index

Beacze:
Since version 5.x ES offers refresh property. Parameter wait_for forces to wait until object is indexed.


Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/index/IndexRestClient.java
```diff
@@ -68,41 +69,48 @@ public static IndexRestClient getInstance() {
      *            with document which is going to be indexed
      * @param id
      *            of document - equal to the id from table in database
+     * @param waitForRefresh
+     *            wait until index is refreshed - if boolean time of execution is
```
henning-gerhardt: Please add a better description of this parameter as `if boolean time of execution is longer` is not a good sentence. This is even more a copy and paste error on other places.

PR #1609, opened 2018-06-25
Remove not used methods from services

Beacze:


Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -519,16 +518,6 @@ public String getProcessingTimeAsFormattedString(Task task) {
         return Helper.getDateAsFormattedString(task.getProcessingTime());
     }
 
-    // a parameter is given here (even if not used) because jsf expects setter
-    // convention
-    public void setProcessingTimeNow(Task task) {
```
henning-gerhardt: You are sure to remove this getter and setter method as they referenced in `Kitodo/src/main/webapp/WEB-INF/templates/includes/tasks/taskBoxActivities.xhtml`?

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -699,16 +688,6 @@ private void abortTask(Task task) throws DataException {
         save(task);
     }
 
-    /**
-     * Returns whether this is a step of a process that is part of at least one
-     * batch as read-only property "batchSize". //TODO: is it ever used?
-     *
-     * @return whether this step’s process is in a batch
-     */
-    public boolean isBatchSize(Task task) {
```
henning-gerhardt: There are references in `Kitodo/src/main/webapp/pages/tasks.xhtml`.

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
Beacze: It uses getProcessingTimeNow() (it returns 1...) from Task bean. In general it would be better to remove also this occurrence from taskBoxActivities. Time now could be created just in background.

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
Beacze: Item with batchSize is for this case TaskDTO. So more likely it is needed to add this property to TaskDTO class... I will look into it closer.

PR #1632, opened 2018-07-03
Standard tables hover for rows

BFallert:
The active row in the standard tables is highlighted. Improves orientation in a table row. The current color is just a suggestion. This color is currently used in the University Library Mannheim.


henning-gerhardt:
Your change looks good and this is a good improvement if you handle or work with larger tables.

PR #1623, opened 2018-06-28
Make config string keys constants

matthias-ronge:
This is a follow-up pull request for #1573 which GitHub showed as “successfully merged” but as I found out actually it was never merged. I don’t understand it. Luckily, I was able to restore the commits from my local git repository and now rebased them onto the current master.

> This solves a long-aged TODO.
> (There is no ticket for this pull request.)

PR #1638, opened 2018-07-06
Create LongTimePreservationValidationModule

matthias-ronge:
This module implements the `LongTimePreservationValidationInterface` using the JHove library. It supports a various of file formats: images including Jpeg-2000, and also PDF.

JHove is included from Maven, the modul does not require manual installation.


Kitodo-LongTimePreservationValidationModule/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>kitodo-production</artifactId>
+        <groupId>org.kitodo</groupId>
+        <version>3.0-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+    <name>Kitodo − Long Time Preservation Validation Module</name>
```
henning-gerhardt: Please use a normal dash `-` as separator and not the long dash `−`. This is already happend in your Image Management Module.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    /**
+     * Method that is executed before the constructor of the parent class.
+     *
+     * @return {@code null}. The value is not used.
+     */
+    private static final String beforeParentConstructor() {
+        Logger.getLogger("edu.harvard.hul.ois.jhove.handler").setLevel(Level.OFF);
```
henning-gerhardt: Is switching of an other module logger as a hard coded program instruction a good solution? Why not disabling it through configuration so every user can switch it on if he need it.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        return null;
+    }
+
+    private final List<String> messages = new ArrayList<String>();
+
+    private ValidationResultState state;
+
+    protected KitodoOutputHandler() {
+        super(beforeParentConstructor(), null, new int[] {1970, 1, 1 }, null, null);
```
henning-gerhardt: What are this magic values `1970, 1, 1`? Please add at least a description to it which explains why this values are used.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    private void addMessage(Message message) {
+        StringBuilder line = new StringBuilder(127);
```
henning-gerhardt: How did you "calculate" this value of `127`?

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        line.append(message.getMessage());
+        String subMessage = message.getSubMessage();
+        if (subMessage != null) {
```
henning-gerhardt: In other modules an expression like `Objects.nonNull` is used for comparing to not `null`.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,160 @@
...
+     * @param e
+     *            exception to set
+     */
+    public void treatException(Exception e) {
+        state = ValidationResultState.ERROR;
+        if (e.getMessage() != null) {
+            messages.add(e.getMessage());
+        }
+        messages.add(e.getClass().getSimpleName());
+        StringWriter buffer = new StringWriter();
```
henning-gerhardt: Opening a stream without closing it should be avoided. Please use a try-with-resources statement here.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        e.printStackTrace(new PrintWriter(buffer));
```
henning-gerhardt: Same here: please use a try-with-resources statement.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/LongTimePreservationValidationModule.java
```diff
@@ -0,0 +1,85 @@
...
+            case JPEG_2000:
+                return "JPEG2000-hul";
+            case PDF:
+                return "PDF-hul";
+            case PNG:
+                return "PNG-gdm";
+            case TIFF:
+                return "TIFF-hul";
+            default:
+                throw new IllegalStateException("Complete switch");
```
henning-gerhardt: I think this is thrown exception is wrong as in this case a not supported module is called. At least get out a proper exception message as `complete switch` is not realy help- or meaningful.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/LongTimePreservationValidationModule.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class LongTimePreservationValidationModule implements LongTimePreservationValidationInterface {
+    /**
+     * Path to the {@code jhove.conf} file
+     *
+     * <p>
+     * Example:
+     * LongTimePreservationValidationModule.jhoveConf=/var/lib/tomcat8/webapps/kitodo-production/WEB-INF/classes/jhove.conf
+     */
+    private static final String PARAMETER_JHOVE_CONF = "LongTimePreservationValidationModule.jhoveConf";
```
henning-gerhardt: This must be adjusted as pull request #1623 is merged. Please add this to your todo list.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/LongTimePreservationValidationModule.java
```diff
@@ -0,0 +1,85 @@
...
+     *            file URI to validate
+     * @param fileType
+     *            file type to validate
+     */
+    @Override
+    public ValidationResult validate(URI fileUri, FileType fileType) {
+        KitodoOutputHandler result = new KitodoOutputHandler();
+        try {
+            JhoveBase jhoveBase = new JhoveBase();
+            jhoveBase.init(Config.getParameter(PARAMETER_JHOVE_CONF, "src/main/resources/jhove.conf"),
```
henning-gerhardt: Same here: this must be adjusted after #1623 is merged.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/ValidationResultState.java
```diff
@@ -0,0 +1,67 @@
...
+            case RepInfo.TRUE:
+                switch (valid) {
+                    case RepInfo.UNDETERMINED:
+                        return WELL_FORMED;
+                    case RepInfo.FALSE:
+                        return WELL_FORMED_BUT_NOT_VALID;
+                    case RepInfo.TRUE:
+                        return VALID;
+                    default:
+                        throw new IllegalStateException("Complete switch");
```
henning-gerhardt: Please adjust this exception as a not defined case is happend. At least a more meaningful hint / message should be thrown.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/ValidationResultState.java
```diff
@@ -0,0 +1,67 @@
...
+                }
+            default:
+                throw new IllegalStateException("Complete switch");
```
henning-gerhardt: Please adjust this exception as a not defined case is happend. At least a more meaningful hint / message should be thrown.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/ValidationResultState.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    private State state;
+
+    ValidationResultState(State state) {
+        this.state = state;
+    }
+
+    public State toState() {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo-LongTimePreservationValidationModule/src/main/resources/jhove.conf
```diff
@@ -0,0 +1,29 @@
...
+  <module>
+    <class>com.mcgath.jhove.module.PngModule</class>
+  </module>
+  <module>
+    <class>edu.harvard.hul.ois.jhove.module.TiffModule</class>
+  </module>
+  <outputHandler>
+    <class>edu.harvard.hul.ois.jhove.handler.TextHandler</class>
+  </outputHandler>
+</jhoveConfig>
```
henning-gerhardt: Missing new line at end of file.

Kitodo-LongTimePreservationValidationModule/src/main/resources/jhove.conf
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0"?>
+<jhoveConfig version="1.0"
```
henning-gerhardt: If this file should be part of a normal Kitodo.Production configuration environment then Kitodo license header is needed.

Kitodo-LongTimePreservationValidationModule/src/main/resources/jhove.conf
```diff
@@ -0,0 +1,29 @@
...
+ xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+ xmlns="http://hul.harvard.edu/ois/xml/ns/jhove/jhoveConfig"
+ xsi:schemaLocation="http://hul.harvard.edu/ois/xml/ns/jhove/jhoveConfig
+                     http://hul.harvard.edu/ois/xml/xsd/jhove/jhoveConfig.xsd">
+  <jhoveHome>src/test/resources</jhoveHome>
```
henning-gerhardt: This value `src/test/resources` is only valid for test execution and must be adjusted in normal environment or? If so why did you not add this kind of file to `/src/test/resources` and use it inside the tests?

Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -661,3 +661,13 @@ elasticsearch.password=kitodo
...
+
+
+# -----------------------------------
+# LongTimePreservationValidationModule
+# -----------------------------------
+
+# Jhove requires to read its configuration file from disk. You have to provide
+# the full path to the file:
+
+#LongTimePreservationValidationModule.jhoveConf=/full/path/to/jhove.conf
```
henning-gerhardt: Missing new line at end of file.

Kitodo-API/src/main/java/org/kitodo/api/validation/ValidationResult.java
```diff
@@ -28,7 +28,7 @@
 
     /**
      * Gets valid.
-     * 
+     *
```
henning-gerhardt: Change is good but why did your IDE fix this as you did not change anything at this file?

Kitodo-LongTimePreservationValidationModule/pom.xml
matthias-ronge: The usage of the "en dash" is [correct here. (Link in German)](https://typefacts.com/artikel/binde-und-gedankenstrich#8)

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
matthias-ronge: The program reads its configuration later, so this output is not affected. It only prints "Initializing", which is not helpful for any user.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
matthias-ronge: The numbers are used to initialize a calender object which is never used later, Hovever, the array must be in player or a `NullPointerException` will be thrown when JHove tries to access the array members to initialize the unused object. The Java epoch is defined in milliseconds since 1970-01-01, so these numbers practicsally mean *zero*.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
matthias-ronge: Not at all. But I just say that the message is longer than 16 characters, so it makes sense to initialize the StringBuilder a little bit larger. 127 is a good guess for that the buffer fits into a memory chunk.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
matthias-ronge: Javadoc of `Objects.nonNull()`:
> **API Note:**
> This method exists to be used as a [Predicate](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html), `filter(Objects::nonNull)`

Beacze: But it is not a problem, it can be safely used:
https://stackoverflow.com/questions/37972859/java-util-objects-isnull-vs-object-null

Kitodo-LongTimePreservationValidationModule/pom.xml
matthias-ronge: I will change it, but if there is an issue with it (unicode problems?) we should fix them.

Citing form [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s2.3.3-non-ascii-characters):
> **Tip:** Never make your code less readable simply out of fear that some programs might not handle non-ASCII characters properly. If that should happen, those programs are **broken** and they must be **fixed**.

Beacze: That's how looks current build - is there something not matching in this log?

[INFO]
[INFO] Kitodo Production .................................. SUCCESS [  0.278 s]
[INFO] OpacPica-Plugin .................................... SUCCESS [  2.863 s]
[INFO] Kitodo - API ....................................... SUCCESS [  2.438 s]
[INFO] Kitodo - Data Management ........................... SUCCESS [  4.114 s]
[INFO] Kitodo - Data Access ............................... SUCCESS [  8.282 s]
[INFO] Kitodo - Data Format ............................... SUCCESS [  3.208 s]
[INFO] Kitodo - Data Editor ............................... SUCCESS [  4.912 s]
[INFO] Kitodo ? Image Management Module ................... SUCCESS [  0.414 s]
[INFO] Kitodo - Persistent Identifier ..................... SUCCESS [  0.507 s]
[INFO] Kitodo - Docket .................................... SUCCESS [  2.520 s]
[INFO] Kitodo - Command ................................... SUCCESS [  0.976 s]
[INFO] Kitodo - File Management ........................... SUCCESS [  1.188 s]
[INFO] Kitodo - Schema Converter .......................... SUCCESS [  0.165 s]
[INFO] Kitodo - Validation ................................ SUCCESS [  0.163 s]
[INFO] Kitodo - Core ...................................... SUCCESS [07:21 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
matthias-ronge: Of course it is safe to use, where else would we go? But it is absolutely unnecessary.

The thing is:  
Code reviews should be about getting safe and understandable code. I consider it absolutely not expedient to constantly discuss any functionally identical little things. We win nothing, we lose: time to focus on the essentials. My and yours. So please ...

Beacze: True, but point is here that @henning-gerhardt is right with usage consistent code style in all places. It doesn't hurt to stick to it.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/LongTimePreservationValidationModule.java
matthias-ronge: The thing is that you *cannot* call any other than the above cases because the enum `FileType` does not have any more constants. Howeyer, the compiler will complain about a missing return statement if I do nothing.

Kitodo-API/src/main/java/org/kitodo/api/validation/ValidationResult.java
matthias-ronge: Probably because I had the file open

matthias-ronge:
It works without config file now.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
matthias-ronge: It does hurt because of the time it takes to

* switch the IDE to the corresponding branch
* wait for the IDE to compile
* figure out which module is affected, open it
* figure out which file is affected, navigate to it
* find the correct line
* Change it
* find unit tests
* run unit tests
* see that file was not yet saved
* Click save
* run unit tests again
* Find ‘Git Staging’ tab
* Craft a git commit
* Look up some english word in the dictionary that you need for the commit message, but are not sure any more how it is written
* Commit
* open Travis in web browser to see what is happening
* find out that branch was pushed to wrong name, because you renamed it recently when cleaning up, and forgot
* push branch again manually
* … etc …
And two days later:
* find the e-mail sent by GitHub about the comment and spend time to think about if that is something new or something I have already fixed
* your time to check if I changed it. If not: why? Are you on the correct branch? On the latest commit? Check back everything.

To put it simple, *every* remark during code review, and every change is taking time so we should spend it on more valuable development.

By the way, there are 2,130 usages of `someVariable != null` vs. 270 usages of `Objects.nonNull()` in my search, so I cannot see a consistent code style here yet.

Kitodo-LongTimePreservationValidationModule/pom.xml
matthias-ronge: I have changed Image Management Module as well.
It seems there is missing UTF-8 support somewhere

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,179 @@
...
+     *            exception to set
+     */
+    public void treatException(Exception e) {
+        state = ValidationResultState.ERROR;
+        if (e.getMessage() != null) {
+            messages.add(e.getMessage());
+        }
+        messages.add(e.getClass().getSimpleName());
+        try {
+            try (StringWriter buffer = new StringWriter()) {
```
henning-gerhardt: You can use two or more closeable classes in one statement which would reduce the number of courly brackets and made the code lines more readable.

Kitodo-LongTimePreservationValidationModule/src/main/java/org/kitodo/longtimepreservationvalidationmodule/KitodoOutputHandler.java
```diff
@@ -0,0 +1,179 @@
...
+
+    protected KitodoOutputHandler() {
+        super(beforeParentConstructor(), null, EPOCH_ZERO, null, null);
+    }
+
+    private void addMessage(Message message) {
+        StringBuilder line = new StringBuilder(127);
+        line.append(message.getMessage());
+        String subMessage = message.getSubMessage();
+        if (Objects.nonNull(subMessage = null)) {
```
henning-gerhardt: You are assigning a  value before checking its content with `null` - I do not think, that you want this. Maybe you mean `if (Objects.nonNull(subMessage)) {`.

henning-gerhardt:
@matthias-ronge : One more thing: a separate `log4j2.xml`file is needed in test resources directory as the main resources file did not apply for tests.

matthias-ronge:
Why is this necessary here but was not for ImageManagementModule? Was there any change I am not aware of? Aside of this, I don’t had any problem on my machine, it found the files while testing too. (And this is what I would aspect because the test does always depend on the productive code, an dthe resources is part of the code as well.) Perhaps something to configuer which resources to debloy for tests?

henning-gerhardt:
If this file is missing in ImageManagementModule then I missed that - if you want fix this then do it please in a separate pull request. Maybe it does not complain because there is no test which is triggering a log event and until this there is no need for a configuration file.

matthias-ronge:
No, I don’t log anything from tests. I can run them right from the IDE, also with debugger attached, so logging wouldn’t make sense here for me:

test


I will fix the other module in a separate branch.

PR #1646, opened 2018-07-10
Set a lower log level when catching exceptions

matthias-ronge:


Kitodo-API/src/main/java/org/kitodo/config/Config.java
```diff
@@ -132,7 +133,7 @@ public static String getParameter(String key, String defaultValue) {
         try {
             return getConfig().getString(key, defaultValue);
         } catch (ConversionException e) {
-            logger.catching(e);
+            logger.catching(Level.DEBUG, e);
             logger.warn("Configuration found in kitodo_config.properties for key {} is defined, but not a String!",
```
matthias-ronge: Is this case at all possible? A configuration value that is defined is always a String when using `PropertiesConfiguration`, isn’t it?

matthias-ronge:
👍 I changed this

PR #1648, opened 2018-07-10
Unsaved changes

solth:
Added a warning when the user leaves an editing page with unsaved changes.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/form.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/form.js
@@ -10,8 +10,10 @@
  */
 
 $(function() {
-    $('#editForm :input').on('change', function() {
-        setConfirmUnload(true);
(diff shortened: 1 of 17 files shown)


henning-gerhardt:
Selenium tests are failing on Travis CI run.

Kitodo/src/main/webapp/WEB-INF/resources/js/form.js
```diff
@@ -0,0 +1,22 @@
...
+$(function() {
+    $('#editForm').on('input', ':input', function() {
+        if (!$(this).parent().parent().hasClass('locked-button')) {
+            setConfirmUnload(true);
+        }
+    });
+});
+
+function setConfirmUnload(on) {
+    window.onbeforeunload = (on) ? function() { return true; } : undefined;
```
stweil, line 21: @solth, would it also be possible to replace `undefined` by `null`? Codacy complains: "Unexpected use of undefined."

PR #1652, opened 2018-07-16
Update footer

solth:


Kitodo/src/main/webapp/WEB-INF/templates/includes/footer.xhtml
```diff
@@ -14,6 +14,11 @@
 <ui:composition xmlns:ui="http://xmlns.jcp.org/jsf/facelets">
     <div class="wrapper">
         <section id="footer-content">
+            <ul id="footer-meta-menu">
+                <li><a href="">#{msgs.imprint}</a></li>
```
henning-gerhardt: Is this empty link reference intended? Same question for the next to lines.

solth: These links are empty because we do not have the corresponding pages, yet. Once we have those, the links will be updated.

henning-gerhardt: Ok, then you know that this pages are missing and will added later.

PR #1658, opened 2018-07-20
Fix migration for split processes and templates

Beacze:


Kitodo-DataManagement/src/main/resources/db/migration/V2_29__Split_processes_and_templates.sql
```diff
@@ -78,23 +85,23 @@ SET t.template_id = temp.id,
...
 
--- 9. Remove templates from process table
+-- 10. Remove templates from process table
 --
 
 DELETE FROM process
 WHERE template = 1;
 
--- 10. Switch on safe updates
+-- 12. Switch on safe updates
```
henning-gerhardt: This should be number `11` and not `12`.

henning-gerhardt:
@Kathrin-Huber : What did you think about this kind of changes?

Kathrin-Huber:
If the migrationscript is incorrect, it should be fixed. It wouldn't be the first time, we change a migration script afterwards. of course it is not desirable. Scripts for migration (at least bigger once, like that) should be testet against large data in the future, before they are commited.
After merging this pull request a flyway-repair is needed. Remember to do that on the test systems!

henning-gerhardt:
Thank you for the explanation @Kathrin-Huber. I tested this changes in the past but my data did not contain this kind of data so this error was not visible for me. One issue remain in this pull request and must be fixed by @Beacze : removed foreign key constraints must be readded after work is done.

PR #1659, opened 2018-07-20
Fix misspelling Dms - DMS

Beacze:


Erikmitk:
I checked out the branch of @Beacze and fired up a diff tool. That’s the change that was made in this single line.

image

`typeExportDMS` has to be written in upper-case otherwise the editor cannot make the connection between the checked property and the one that should be written in the .xml of the generated workflow. This is necessary to create export tasks.

Erikmitk:
Can we please expedite this Pull Request, @henning-gerhardt?

This small change is needed for testing a complete run through a typical workflow in our test system. This system updates over night so **we lose a whole day** for testing if we can't get this in quickly.

The change is clearly visible and very minor but it is vital for creating workflows that have an export task.

PR #1665, opened 2018-07-24
Template view

Beacze:


henning-gerhardt:

- Kitodo/src/main/java/org/kitodo/forms/DesktopForm.java

Please fix / solve this conflict.

Kitodo/src/test/resources/mapping.json
```diff
@@ -623,6 +623,28 @@
...
                         }
                     }
+                },
+                "workflow": {
+                    "properties": {
+                        "title": {
+                            "type": "text",
+                            "fields": {
+                                "keyword": {
+                                "type": "keyword",
```
henning-gerhardt: Missing indent, so both `mapping.json` files differ.

Kitodo/src/main/java/org/kitodo/services/data/TemplateService.java
```diff
@@ -260,6 +276,33 @@ private void convertRelatedJSONObjects(JsonObject jsonObject, TemplateDTO templa
...
+    /**
+     * Get diagram image for current template.
+     *
+     * @return diagram image file
+     */
+    public InputStream getTasksDiagram(String fileName) {
+        if (Objects.nonNull(fileName) && !fileName.equals("")) {
+            File tasksDiagram = new File(ConfigCore.getKitodoDiagramDirectory(), fileName + ".svg");
+            try {
+                return new FileInputStream(tasksDiagram);
```
henning-gerhardt: Instead of returning a possible open stream maybe returning a `byte[]` value of the read file did work too? An empty `byte[]` value should returned if given file is not found.

Beacze: I would stay at the moment with this as it is.

henning-gerhardt: Ok, but this should be evaluated in future.

Beacze: That's true. In general the way of displaying this image should be checked out by frontend developers.

PR #1669, opened 2018-07-26
Fix compilation errors after merge

Beacze:


henning-gerhardt:
Emergency merge as compile errors was not known and visible in front of merging.

PR #1687, opened 2018-08-02
Remove getProjectTemplateTitles - list of template titles is stored in index

Beacze:
It is now displayed from project index object.


henning-gerhardt:

- Kitodo/src/main/java/org/kitodo/services/data/ProjectService.java

Please fix / solve this conflict.

PR #1689, opened 2018-08-02
Remove unused module Kitodo - Data Access

matthias-ronge:
The module Kitodo - Data Access is not used and will likely not be used in the future, so it can be removed. It can still be restored from the version history if it is ever needed.


henning-gerhardt:
@Kathrin-Huber should decide about removing of this module.

PR #1697, opened 2018-08-07
 Add selector for generation source, thumbs and page view folder

matthias-ronge:
Folders can be selected from a drop-down list in the GUI.


Kitodo-DataManagement/src/main/resources/db/migration/V2_40__Change_projectfilegroup_to_folder.sql
```diff
@@ -69,6 +69,50 @@ ALTER TABLE projectFileGroup
 ALTER TABLE projectFileGroup RENAME TO folder;
```
Kathrin-Huber: never change an existing migration! (Unless it is breaking)
If you need to migrate stuf, than add a new migration file.

matthias-ronge: I did this here because it fills the path column below. If I do it later you have to adjust the position twice. That does not make sense. I had planned this from the beginning, I only split it into two pull requests so that the individual packages do not get too big to review.

matthias-ronge:
I have already seen that. Apparently Travis performs the merge locally and tests the result. In that case, this results in the migration file being there twice, and then it goes awry.

matthias-ronge:
Now it finished without error.

Kitodo-DataManagement/src/main/resources/db/migration/V2_40__Change_projectfilegroup_to_folder.sql
matthias-ronge: Do you think this is okay for once. The rest of the migration is in a separate file below as you see.

matthias-ronge:
Closed, because of bug. Tasks get lost.

matthias-ronge:
I mixed something up. This branch is ok.

Kathrin-Huber:
Well... it's still failing ;)

matthias-ronge:
How has this happened? Did you restart it? It was OK the last time I checked it. Well, I have restarted it and now it is OK again. You know that most of the time you just have to restart the build if the Selenium tests fail. Henning did do this by himself. I believe to recall him saying that it could randomly take up to seven builds for the tests to succeed. As I was told the cause for this behaviour is that there is a lot of Javascript on the GUI buttons.

Kitodo-DataManagement/src/main/resources/db/migration/V2_40__Change_projectfilegroup_to_folder.sql
Kathrin-Huber: But the first version is already in the master! This will break the test system because we need to execute flyway-repair if a migration script changes. Regarding the fact, that we are close to a release and elena is testing constantly I would rather not risk this. Please make a new script or you need to wait untill the release is done :(

matthias-ronge:
I have undone the database script changes, and the Travis check has passed.

PR #1701, opened 2018-08-15
Metadata editor

solth:
These changes add a basic new design for the metadata and structure editor. The design consists of three resizable columns containing the structure information of a document, it's metadata and an image viewer with additional thumbnail previews. The main advantage of the new design is that the user can work on the structure and metadata concurrently without having to switch back and forth between different views. The new design also adds a new image viewer based on OpenLayers.

**Known issues:**
- pagination is not implemented, yet
- filepaths of images in the physical struct map of the created mets structure contain one duplicated path segment
- new CSS rules of metadata editor result in some small design quirks on the desktop page
- the third column contains three buttons for different views of the gallery, of which the second is unused as of yet and may be removed entirely
- the error message `image path is invalid` is displayed after generating the png images for the currently selected folder and then opening the detail view in the gallery without selecting a page beforehand


Kitodo/src/main/java/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -81,27 +92,42 @@
...
  */
 public class Metadaten {
     private static final Logger logger = LogManager.getLogger(Metadaten.class);
+
+    /**
+     * Get imageHelper.
+     *
+     * @return value of imageHelper
+     */
+    public MetadatenImagesHelper getImageHelper() {
```
henning-gerhardt: Adding a method in list of method variable declaration should not go through any kind of review...

PR #1709, opened 2018-08-22
Creating dummy images in metadata editor

tim-boerner:


pom.xml
```diff
@@ -164,7 +164,7 @@
                                     <encoding>UTF-8</encoding>
                                     <failOnViolation>true</failOnViolation>
                                     <failsOnError>false</failsOnError>
-                                    <logViolationsToConsole>false</logViolationsToConsole>
+                                    <logViolationsToConsole>true</logViolationsToConsole>
```
Kathrin-Huber: Is this a change from local debugging? I think we decided to not log the checkstyle failures, because its crowding the log too much.

Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -163,6 +163,10 @@ ImagePrefix=\\d{8}
 #       image names, all characters are supported
 ImageSorting=number
 
+# Insert a path to an tif image which is taken as dummy images for a process
+# in order to start with pagination editing.
+dummyImagePath=C:/dev/local/images/dummyImage.tif
```
Kathrin-Huber: Please don't use windows-style paths in kitodo config. You can use them in your local config.

pom.xml
tim-boerner: Actually yes, but i thought as we have only around 60 violations left in core, the log crowding is acceptable for saving much debugging time by already seeing the voilations at travis build log in case the build is failing because of to much violations.

PR #1714, opened 2018-08-28
Fix unlocking processes

oliver-stoehr:
Unlocks a process when the user finishes the current task or cancels the assigned task and unlocks a process when using the manual link for unlocking the current process.


Kitodo/src/main/java/org/kitodo/services/workflow/WorkflowControllerService.java
```diff
@@ -49,6 +50,7 @@
 
 public class WorkflowControllerService {
 
+    private final MetadatenSperrung msp = new MetadatenSperrung();
```
Kathrin-Huber: Please use meaningful variable names. 
The class "metadatensperrung" should be translated.

PR #1715, opened 2018-08-28
Pagination without png generation, fix drag drop of structure elements

tim-boerner:
know issue: drag and drop of page elements results in added the draged page to the new structure element but does not remove it form the old element

reimplemented the old pagination edit dialog

drag and drop of structure elements is working again


oliver-stoehr:
It seems that there is still an issue with moving nodes via drag and drop: Sometimes nodes disappear or can be dropped on the root level. I couldn't figure out yet whether there is a pattern behind this behavior.

PR #1726, opened 2018-08-29
Prevent creation of docket and ruleset with the same name

Beacze:


Kitodo/src/main/java/de/sub/goobi/forms/RulesetForm.java
```diff
@@ -133,6 +137,23 @@ public String removeRuleset() {
...
     }
 
+    private boolean existsRulesetWithSameName() {
+        List<Ruleset> rulesets = serviceManager.getRulesetService().getByTitle(this.ruleset.getTitle());
+        if (rulesets.isEmpty()) {
+            return false;
+        } else {
+            if (Objects.nonNull(this.ruleset.getId())) {
+                if (rulesets.size() == 1) {
+                    return !rulesets.get(0).getId().equals(this.ruleset.getId());
```
Kathrin-Huber: You are checking if the id is the same. What about same name but different if? Shouldn't the return value always be true if a ruleset is found?

Beacze: If id is the same, it means that it is update of ruleset so they can have the same name (maybe file was only changed) - return false here. If ids are different, it means that it updates object and wants to give to update object name which is already in use by other ruleset - return true here.

PR #1741, opened 2018-08-31
Restore missing functionalities for process list - single process

Beacze:
[image](https://i.imgur.com/O6qtSXS.png)

PR #1745, opened 2018-09-03
Fix docket generation - missing classes and version conflicts

Beacze:
Depends on #1741 

Loaded module doesn't provide own dependencies so it is now included also in Core. Probably it would be good if in the future it is included by Docket module.


Beacze:
It would be nice if someone with correct data would tests it, as I can execute methods but at the end I get error:

`[ERROR] 2018-09-03 15:24:56.356 [http-apr-8080-exec-1] FOUserAgent - Image not available. URI: (instream-object). Reason: org.apache.xmlgraphics.image.loader.ImageException: The file format is not supported. No ImagePreloader found for null (No context info available)
org.apache.xmlgraphics.image.loader.ImageException: The file format is not supported. No ImagePreloader found for null`

I suspect that error is caused by incomplete data (missing images) in the process for which I call it.

PR #1755, opened 2018-09-04
Add script for loading workflow diagrams

Beacze:
Depends on #1753


Beacze:
Doesn't work for Firefox.

PR #1756, opened 2018-09-05
Simplify and unify properties displayed in editor and edit forms

Beacze:
- documentation, outputName removed
- edit of tasks contains the same fields as in editor (probably at that moment it is still better to keep it)
- id field can't be removed - it is obligatory for BPMN - without it validation of diagram files

properties

task-edit


felixlohmeier:
I suggest to remove the explainatory questions ("is it task metadata type" etc.)
btw: what is russian export?

Beacze:
Ok, I will remove questions.

Russian export - I suspect it is information if scanned book is using cyrylic alphabet. We have special export classes for this alphabet.

Beacze:
I would keep them as the back up if something goes wrong with editor.

Erikmitk:
Our goal was to reduce duplication. So they should go for now, right @felixlohmeier?

felixlohmeier:
Yes, from a user perspective I prefer "full deduplication".

felixlohmeier:
This change needs further testing and fixes.

If I create a workflow with some properties in tab task checked (e.g. metadata) in development version 2018-09-08 (master branch) and create a template ("Produktionsvorlage") based upon this workflow, the template ("Produktionsvorlage") shows properties (e.g. metadata) unchecked.

PR #1764, opened 2018-09-07
Fix method getTableSize in User bean which is used for table size and remove it from UserService

Beacze:


Kitodo/src/main/java/org/kitodo/services/data/UserService.java
```diff
@@ -726,18 +726,6 @@ public boolean isLoginValid(String login, String filePath) {
         return true;
     }
 
-    /**
-     * Table size.
-     *
-     * @return table size
-     */
-    public Integer getTableSize(User user) {
```
Kathrin-Huber: Why did you move it to the user?

Beacze: I didn't move it. It was already there as it takes data from database. All pages are using user.getTableSize for data tables, so this one was not used at all.

PR #1776, opened 2018-09-10
Fix current task edit and take / give away task

Beacze:
Depends on #1771


Kitodo/src/main/java/org/kitodo/forms/CurrentTaskForm.java
```diff
@@ -501,16 +499,9 @@ public void setCurrentTask(Task task) {
      *
```
Kathrin-Huber: I think here is something wrong ;)

PR #1784, opened 2018-09-11
Remove user style css designs

solth:
UI design being customizable by the user is a leftover from earlier versions, does not have any effect on the new design and can be removed.


Kitodo-DataManagement/src/main/resources/db/migration/V2_48__Remove_css_column.sql
```diff
@@ -0,0 +1,16 @@
+--
```
Kathrin-Huber: v_48 is already in use

PR #1789, opened 2018-09-13
Fix possible NPE on messages and wrong bundle name

Beacze:


Kitodo/src/main/java/org/kitodo/helper/Helper.java
```diff
@@ -341,11 +341,19 @@ public static String getString(Locale language, String key) {
...
-                return messages.getString(key);
+            if (commonMessages.containsKey(language)) {
+                ResourceBundle messages = commonMessages.get(language);
+                if (messages.containsKey(key)) {
+                    return messages.getString(key);
+                }
+            }
+            if (errorMessages.containsKey(language)) {
+                ResourceBundle messages = errorMessages.get(language);
+                if (messages.containsKey(key)) {
```
Kathrin-Huber: Can you put this out of the brackets to avoid duplicated code?

Beacze: I have moved it own method.

PR #1799, opened 2018-09-14
Use Batch object instead of batch id in select list on batch list page

Beacze:
- add BatchConverter
- use ProcessConverter in ProzesskopieForm (getProcesses() more general)


Kathrin-Huber:
You have 46 Checkstyle violations. The maximum number of allowed violations is 45. ;)

PR #1827, opened 2018-09-19
Remove class DefaultValues- replace it with more clean solution

Beacze:


Kitodo/src/main/java/de/sub/goobi/metadaten/MetadataLock.java
```diff
@@ -30,8 +29,7 @@
      * Zeit, innerhalb der der Benutzer handeln muss, um seine Sperrung zu
      * behalten (30 min)
      */
-    private static final long sperrzeit = ConfigCore.getLongParameter(ParameterCore.METS_EDITOR_LOCKING_TIME,
-        DefaultValues.METS_EDITOR_LOCKING_TIME);
+    private static final long sperrzeit = ConfigCore.getLongParameterOrDefaultValue(ParameterCore.METS_EDITOR_LOCKING_TIME);
```
Kathrin-Huber: Should be translated

PR #1830, opened 2018-09-20
Make conditions for create / edit object more exclusive

Beacze:
@solth Could you review it?

### DIFF
--- a/Kitodo/src/main/webapp/pages/templateEdit.xhtml
+++ b/Kitodo/src/main/webapp/pages/templateEdit.xhtml
@@ -28,9 +28,12 @@
             window.onload = function () {checkForm('editForm')};
         </script>
         <h3>
-            <h:outputText value="#{msgs.editTemplate} (#{TemplateForm.template.title})" rendered="#{not empty TemplateForm.template.title}"/>
-            <h:outputText value="#{msgs.newTemplate}" rendered="#{empty TemplateForm.template.title}"/>
-            <h:outputText value="#{msgs.duplicateTemplate}" rendered="#{empty TemplateForm.template.title and not empty TemplateForm.template.docket}"/>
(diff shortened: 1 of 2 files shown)


Beacze:
Should I use id everywhere?

solth:
> Should I use id everywhere?

If the goal of the PR is what its title states and if it still works, than yes.

Kitodo/src/main/webapp/pages/templateEdit.xhtml
```diff
@@ -28,9 +28,12 @@
             window.onload = function () {checkForm('editForm')};
         </script>
         <h3>
-            <h:outputText value="#{msgs.editTemplate} (#{TemplateForm.template.title})" rendered="#{not empty TemplateForm.template.title}"/>
-            <h:outputText value="#{msgs.newTemplate}" rendered="#{empty TemplateForm.template.title}"/>
-            <h:outputText value="#{msgs.duplicateTemplate}" rendered="#{empty TemplateForm.template.title and not empty TemplateForm.template.docket}"/>
+            <h:outputText value="#{msgs.editTemplate} (#{TemplateForm.template.title})"
+                          rendered="#{TemplateForm.template.id ne null and TemplateForm.template.title ne null and not empty TemplateForm.template.docket}"/>
+            <h:outputText value="#{msgs.newTemplate}"
```
solth: The conditions are now mutually exclusive, but neither "New template" nor "Duplicate template" is displayed (only "Edit template"), so apparently neither condition is met in the corresonding situation (before these changes, at least "New template" was properly displayed when creating a new template)

Beacze: So for now I will just leave it without full exclusions before it gets totally destroyed ;)

PR #1831, opened 2018-09-20
Fix duplication of workflows

Beacze:
- use random generator for names
- save duplicated files (missing removing in case user abort copying)
- remove overwrite of copied workflow in load method

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/helper/Helper.java
+++ b/Kitodo/src/main/java/org/kitodo/helper/Helper.java
@@ -21,6 +21,7 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.security.SecureRandom;
 import java.text.DateFormat;
 import java.text.MessageFormat;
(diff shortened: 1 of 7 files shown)


Kitodo/src/main/java/org/kitodo/helper/Helper.java
```diff
@@ -584,4 +585,22 @@ public static String getLastMessage() {
...
     }
+
+    /**
+     * Generate random string.
+     * 
+     * @param length
+     *            of random string to be created
+     * @return random string
+     */
+    public static String generateRandomString(int length) {
```
Kathrin-Huber: Didn't we already have random generation somewhere!?

Beacze: We had in MockDatabase. It is removed from there in this PR and moved to Helper.

Kathrin-Huber: I have seen this, I thought we had it somewhere for image files or something, but however...

PR #1834, opened 2018-09-21
Add additional mapping validation to prevent users from creating their own mapping

Beacze:
@Erikmitk Could you check if still is possible that you create your own mapping and next click in the app?

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/forms/IndexingForm.java
+++ b/Kitodo/src/main/java/org/kitodo/forms/IndexingForm.java
@@ -421,6 +421,11 @@ public class IndexingForm {
         }
     }
 
+    public boolean isMappingEmpty() {
+        String emptyMapping = "{\n\"mappings\": {\n\n    }\n}";
+        return isMappingValid(emptyMapping);
+    }
+
     private boolean isMappingValid(String mapping) {
         try {
             JsonObject mappingExpected = Json.createReader(new StringReader(mapping)).readObject();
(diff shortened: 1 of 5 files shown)


Erikmitk:
Doesn't work for me.

I create my index via `curl -X PUT "localhost:9200/kitodo"` and just start indexing. I don't touch any mapping at all. So I'm not sure if `createMapping` is actually called. The button for mapping creation is only active when there's no index at all. But I have an index. It's just empty (no mapping). Maybe it is sufficient to check the existence of the index (+ correct mapping) right before indexing itself.

We wouldn't need the button to create the mapping if it's implicitly checked (and fixed) when the indexing process starts.

Beacze:
Create mapping is only called if you click the button. So maybe additionally it would be necessary to check if mapping is correct before you can click start indexing, but... this would prevent possibility that someone wants to use index with additional types. 

Removing buttons for delete/create is not good idea, as we don't want force users to make it manually. We want to convince users to make it by application.

Maybe I could add one more check - if mapping is not empty, as this option we want exclude.

Beacze:
Installation guide adjusted: https://github.com/kitodo/kitodo-production/wiki/Installationsanleitung-f%C3%BCr-Kitodo.Production-3.x#a7-configure-elasticsearch

henning-gerhardt:
@Beacze : You should adjust https://github.com/kitodo/kitodo-production/blob/master/docs/gettingstarted/development-version.md like https://github.com/kitodo/kitodo-production/wiki/Installationsanleitung-f%C3%BCr-Kitodo.Production-3.x : Removing this curl call to create the index.

Kitodo/src/main/java/org/kitodo/forms/IndexingForm.java
```diff
@@ -415,6 +421,22 @@ private static String readMapping() {
         }
     }
 
+    public boolean isMappingEmpty() {
```
Kathrin-Huber: Javadoc missing

PR #1835, opened 2018-09-21
Unify naming for mets editor and image properties

Beacze:
+ remove not used properties (dummy image and show statistics box - this box doesn't exists anymore)


Kathrin-Huber:
Please resolve conflicts 

There is an adjustment of config.properties necessary , right?

Beacze:
Yes, renamed properties must be also renamed in local config file.

PR #1819, opened 2018-09-18
Automatic generation when completing the task

matthias-ronge:
Generating images starts automatically when the task is completed.

Note: This branch is stacked on [matthias-ronge/KIT-1010](https://github.com/matthias-ronge/kitodo-production/tree/KIT-1010) and should be merged after pull request #1791.

PR #1837, opened 2018-09-21
 Generating images can be started manually

matthias-ronge:
Generating images can be started manually. For this purpose, corresponding action links are displayed in the task. If you click on one of the links, the generation will be triggered. As asked for by Henning, there is not just one link, but three different functions that trigger the creation of all images, the creation of missing images, or the creation of damaged and missing images.

Screenshot showing the links for generating pictures

**Implementation:** To do that, additional action links have been added to the `TaskBoxActivities` page. The functions have been implemented in `CurrentTaskForm`. Each parameterizes an `ImageGeneratorTask`, which is passed to the `TaskManager` for execution, since the creation of the images can take a while for large processes. This allows an operator to manage the progress of the task in `TaskManager`. The class `ImageGeneratorTask` has been implemented. For this a function had to be developed, which can list the contents of a folder. This is a tricky business, as a special bell has been implemented for the SLUB, that the project folders have to deal with the situation that the contents of several conceptual folders are in the same digital folder on the hard disk. This is where a complicated filtering process is used to straighten that out. The file list is ordered by the common part of the file name. To solve the problem that you could only sort correctly either numerically, but the file names could not contain any characters that are not numbers, or you could sort alphanumerically, which leads to the wrong order of numbers for file names without leading zeros in the past, a comparator class was added, which compares strings lexicographically when sorting the map. It is already ensured that the umlauts are sorted correctly. The class fulfills a wide range of functions and can be used in other places in the future where something has to be sorted.

Various changes had to be made in order to be able to use the properly functioning modules in the core:

- There must be a text file in the JAR files that indicates the mapping from the interface to the implementation. This file was missing in both modules and has been added.
- It was found that the modules can not be loaded if the library JARs they use are not referenced in the core. Therefore, the necessary JARs have been added to the pom.xml of the core module.
- During the implementation, it was also discovered that naming the JAR files of the modules Image Management Module and Long Time Preservation Validation Module does not follow the standard. For this, the artifact id has been adapted to the standard.

Other changes that have occurred in the development:

- An exception was written for the widespread case of complete switch statements. This should later be used in all places in the code where the case occurs.
- The Javadoc has been improved.
- The `FolderGenerator` class has been renamed `GeneratorSelect` because the name better describes what the class does. (This is also part of the JSF form, the one that allows you to select the generator module, which is a similar situation to `GeneratorSwitch`.)
- Functionality that was included in database bean classes has been swapped out to a separate class `ContentFolder` in the core. This class is now called via introspection to avoid dependencies of the Data Management Module on the core.


Kitodo/src/test/java/org/kitodo/forms/GeneratorSwitchTest.java
```diff
@@ -8,7 +8,8 @@
  * For the full copyright and license information, please read the
  * GPL3-License.txt file that was distributed with this source code.
  */
-package org.kitodo.util;
+
+package org.kitodo.forms;
```
Beacze: Merge conflict?

matthias-ronge: No. As I wrote above, this class is part of a JSF form class. That's why I correctly moved them to the `forms` package. By the way, you have already reviewed this code. You do not need to review it again.

Beacze: Exactly I have reviewed this code that's why I know that this package assignment is wrong. Could you show me page to which responds this form? Because I don't see any.  Even more, I can say 100% that this whole stuff doesn't work. But as you don't want my review I will not elaborate what is wrong.

Kitodo/src/main/webapp/WEB-INF/templates/includes/currentTasksEdit/taskBoxActivities.xhtml
```diff
@@ -123,6 +123,30 @@
                         <f:param name="zurueck" value="AktuelleSchritteBearbeiten"/>
                     </h:commandLink>
 
+                    <!-- Re-generate all images button -->
+                    <h:commandLink id="action12"
+                                   rendered="#{not empty task.typeGenerate and process.blockedUser == null}"
+                                   action="#{AktuelleSchritteForm.regenerateAllImagesButtonClick}"
```
henning-gerhardt: Form class `AktuelleSchritteForm` is renamed to `CurrentTaskForm` form. You should test your code after you solved merge conflicts. It would be better if you add Selenium tests for your added code.

Kitodo/src/main/webapp/WEB-INF/templates/includes/currentTasksEdit/taskBoxActivities.xhtml
```diff
@@ -123,6 +123,30 @@
...
+                                   title="#{msgs.regenerateAllImages}">
+                        <h:outputText><i class="fa fa-cog"/> #{msgs.regenerateAllImages}</h:outputText>
+                    </h:commandLink>
+
+                    <!-- Re-generate missing and damaged images button -->
+                    <h:commandLink id="action16"
+                                   rendered="#{not empty task.typeGenerate and process.blockedUser == null}"
+                                   action="#{AktuelleSchritteForm.regenerateMissingImagesDamagedButtonClick}"
```
henning-gerhardt: See my comment above.

Kitodo/src/main/webapp/WEB-INF/templates/includes/currentTasksEdit/taskBoxActivities.xhtml
```diff
@@ -123,6 +123,30 @@
...
+                                   title="#{msgs.regenerateMissingAndDamagedImages}">
+                        <h:outputText><i class="fa fa-cog"/> #{msgs.regenerateMissingAndDamagedImages}</h:outputText>
+                    </h:commandLink>
+
+                    <!-- Re-generate missing images button -->
+                    <h:commandLink id="action17"
+                                   rendered="#{not empty task.typeGenerate and process.blockedUser == null}"
+                                   action="#{AktuelleSchritteForm.generateMissingImagesButtonClick}"
```
henning-gerhardt: See my comment above.

Kitodo/src/test/java/org/kitodo/forms/GeneratorSwitchTest.java
Erikmitk: But… like @Beacze says it has changed since https://github.com/kitodo/kitodo-production/pull/1819/files#diff-a287091c81593b5384e4e9f24aaccdd4 so it has to be reviewed and discussed again. I know the PRs need each other but switching a class to another package between PRs doesn't seem right.

Kitodo/src/main/webapp/WEB-INF/templates/includes/currentTasksEdit/taskBoxActivities.xhtml
matthias-ronge: Thanks for the hint, Eclipse can not track that. Presumably that did not even lead to a merge conflict. Then I have to test it again, in fact.

You know, it's not that easy to test that because I need a whole setup for it. I need a process with pictures in it, if possible a broken one, and then I can configure everything (source folder, destination folder, ...) and then you can test the functionality, of course. And the next time I have to reload the database because of a change to the schema, the whole setup is gone. I also do not know how to solve this with Selenium tests, to be honest. Apart from the fact that they are also not running on my PC, probably a conflict with (unrelated) Visual Studio drivers, but I'm still not sure. What I mean is, even the Selenium test would need an appropriate setup. A process in the database whose ID and process title is known so that images can be copied to the right place, etc. Programming all this is probably more work than the development itself. Of course that would be nice, but my time is somehow finite.

matthias-ronge:
Thankfully, Henning says the code does not work anymore. Because the error occurs in the JSF, this error unfortunately did not attract attention. Since this revision will take a long time, I now close the ticket.

PR #1843, opened 2018-09-25
Migrate checkstyle to 8.12

Beacze:


pom.xml
```diff
@@ -160,7 +160,14 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-checkstyle-plugin</artifactId>
-                        <version>2.17</version>
+                        <version>3.0.0</version>
+                        <dependencies>
+                            <dependency>
+                                <groupId>com.puppycrawl.tools</groupId>
```
Kathrin-Huber: Why did you need to add a dependency here, which we did not need before?

Beacze: Because:

_This version of the plugin uses Checkstyle 6.18 by default and requires Java 7. But you can upgrade the version used at runtime._
Source: https://maven.apache.org/plugins/maven-checkstyle-plugin/

PR #1866, opened 2018-10-04
Remove client select list from project edit view

Beacze:
Assign client for new project according to session client of authenticated user.


Kitodo/src/main/java/org/kitodo/forms/ProjectForm.java
```diff
@@ -146,6 +146,7 @@ public String cancel() {
      */
     public String newProject() {
         this.myProjekt = new Project();
```
Kathrin-Huber: please translate

PR #1868, opened 2018-10-04
Inserts action links to create images and implements the functionality

matthias-ronge:
Functioning:
To manually start image generation, corresponding action links are displayed in the task. If you click on one of the links, the generation will be triggered. There are tree different functions that trigger the creation of all images, the creation of missing images, or the creation of damaged and missing images.

Implementation:
The additional action links have been added to the taskBoxActivities.xml page. The functions have been implemented in CurrentTaskForm. Each of them parameterizes an ImageGenerator, which is passed to a TaskImageGeneratorThread for execution in the task manager, since the creation of the images can take a while for large processes. This allows an operator to manage the progress of the task in TaskManager. The class ImageGeneratorTask has been implemented. For this a function had to be developed, which can list the contents of a folder. This is a tricky business, as a special bell has been implemented, that the project folders have to deal with the situation that the contents of several conceptual folders are in the same digital folder on the hard disk. A complicated filtering process is necessary to straighten that out. To be able to use the properly functioning modules in the core, there must be a text file in the JAR files that indicates the mapping from the interface to the implementation. This file was missing in two modules and has been added. It was found that the modules can not be loaded if the library JARs they use are not referenced in the core. Therefore, the necessary JARs have been added to the pom.xml of the core module.


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Task.java
```diff
@@ -345,6 +350,24 @@ public void setProcess(Process process) {
...
+    public List<Project> getProjects() {
+        if (Objects.nonNull(process)) {
+            return Arrays.asList(process.getProject());
+        } else if (Objects.nonNull(template)) {
+            return template.getProjects();
+        } else {
+            return Collections.emptyList();
+        }
+    }
+
```
Kathrin-Huber: This is something which should be added to the TaskService. Please don't add functionality to beans.

Kitodo-ImageManagementModule/src/main/resources/META-INF/services/org.kitodo.api.imagemanagement.ImageManagementInterface
```diff
@@ -0,0 +1,12 @@
...
+#
+# This file is part of the Kitodo project.
+#
+# It is licensed under GNU General Public License version 3 or later.
+#
+# For the full copyright and license information, please read the
+# GPL3-License.txt file that was distributed with this source code.
+#
+
+org.kitodo.imagemanagementmodule.ImageManagementModule
```
Kathrin-Huber: I think, you renamed it, right?

Kitodo-LongTimePreservationValidationModule/src/main/resources/META-INF/services/org.kitodo.api.validation.longtimepreservation.LongTimePreservationValidationInterface
```diff
@@ -0,0 +1,12 @@
...
+#
+# This file is part of the Kitodo project.
+#
+# It is licensed under GNU General Public License version 3 or later.
+#
+# For the full copyright and license information, please read the
+# GPL3-License.txt file that was distributed with this source code.
+#
+
+org.kitodo.longtimepreservationvalidationmodule.LongTimePreservationValidationModule
```
Kathrin-Huber: Same here. Is it still called "...Module"?

Kitodo/src/main/java/org/kitodo/services/image/GenerationMode.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.services.image;
```
Kathrin-Huber: Please add enums to org.kitodo.enums

Kitodo/src/main/java/org/kitodo/forms/CurrentTaskForm.java
```diff
@@ -426,6 +431,47 @@ private void download() {
         }
     }
 
+    /**
+     * Regenerate all images.
```
Kathrin-Huber: please adjust javadoc to new methodname

Kitodo/src/main/java/org/kitodo/forms/CurrentTaskForm.java
```diff
@@ -426,6 +431,47 @@ private void download() {
...
+     */
+    public void generateAllImages() {
+        generateImages(GenerationMode.ALL, "regenerateAllImagesStarted");
+    }
+
+    /**
+     * Regenerate missing and damaged images.
```
Kathrin-Huber: Same here

Kitodo/src/main/java/org/kitodo/model/Subfolder.java
```diff
@@ -0,0 +1,301 @@
...
+    /**
+     * Convenience function to create a bunch of subfolders in one.
+     * 
+     * @param process
+     *            the process this subfolder belongs to
+     * @param folders
+     *            The general metrics of the kinds of subfolders to create
+     * @return subfolders
+     */
+    public static List<Subfolder> createAll(Process process, Collection<Folder> folders) {
```
Kathrin-Huber: Please don't create a bunch of entitys from the class itself. Should be moved to a service.

Kitodo/src/main/java/org/kitodo/model/Subfolder.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /**
+     * Returns a key mapper for a compiled regular expression.
+     *
+     * @param pattern
+     *            compiled regular expression to determine the key
+     * @return a mapping function for the key of the map
+     */
+    private static Function<URI, String> createKeyMapperForPattern(final Pattern pattern) {
```
Kathrin-Huber: Sorry, I don't get, what is happening here. What is the key mapper used for? Why is it a Function?

Kitodo/src/main/java/org/kitodo/model/Subfolder.java
```diff
@@ -0,0 +1,301 @@
...
+            throw new UndeclaredThrowableException(e);
+        }
+    }
+
+    /**
+     * Returns the settings of the subfolder configured in the project.
+     * 
+     * @return the folder
+     */
+    public Folder getFolder() {
```
Kathrin-Huber: Is it a Folder or is it settings like stated in the javadoc? Maybe the name of the bean is missleading?

Kitodo/src/main/java/org/kitodo/model/Subfolder.java
```diff
@@ -0,0 +1,301 @@
...
+     * @param pattern
+     *            compiled regular expression to determine the key
+     * @return a mapping function for the key of the map
+     */
+    private static Function<URI, String> createKeyMapperForPattern(final Pattern pattern) {
+        Function<URI, String> result = new Function<URI, String>() {
+            @Override
+            public String apply(URI uri) {
+                Matcher matcher = pattern.matcher(FilenameUtils.getName(uri.getPath()));
+                matcher.matches();
```
Kathrin-Huber: Result of matcher.matches() is ignored

Kitodo/src/main/java/org/kitodo/services/image/ImageGenerator.java
```diff
@@ -0,0 +1,423 @@
...
+        this.toBeGenerated = new LinkedList<>();
+    }
+
+    /**
+     * Appends the element to the list of elements to be generated.
+     *
+     * @param elementToAppend
+     *            element to be appended to the list
+     */
+    void addToToBeGenerated(Pair<Pair<String, URI>, List<Subfolder>> elementToAppend) {
```
Kathrin-Huber: Please create a class, instead of using Pair<Pair<String, URI>, List<Subfolder>>

Kitodo/src/main/java/org/kitodo/services/image/GenerationMode.java
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    /**
+     * Returns the corresponding filter for the generator variant.
+     *
+     * @param canonical
+     *            canonical part of the file name
+     * @return the filter for the generator variant
+     */
+    public abstract Predicate<? super Subfolder> getFilter(String canonical);
```
Kathrin-Huber: What do you need the wildcard for?

Kitodo/src/main/java/org/kitodo/services/image/ImageGenerator.java
```diff
@@ -0,0 +1,423 @@
...
+     * @param sourceImage
+     *            address of the source image from which the derivative is to be
+     *            calculated.
+     * @param imageProperties
+     *            configuration for the target image
+     * @return an image in memory
+     * @throws IOException
+     *             if an underlying disk operation fails
+     */
+    private static Image retrieveJavaImage(URI sourceImage, Folder imageProperties) throws IOException {
```
Kathrin-Huber: Why is it explicitly "JavaImage" istead of Image?

Kitodo/src/main/java/org/kitodo/services/image/ImageGenerator.java
```diff
@@ -0,0 +1,423 @@
...
+     * @param imageProperties
+     *            folder whose configuration specifies what should be created
+     * @param fileFormat
+     *            the file format to be created
+     * @param destinationImage
+     *            path to the target file to be generated
+     * @throws IOException
+     *             if filesystem I/O fails
+     */
+    private static void generateDerivative(URI sourceImage, Subfolder destinationImage, String canonical)
```
Kathrin-Huber: Javadoc params doesn't fit

Kitodo/src/main/java/org/kitodo/services/image/ImageGenerator.java
```diff
@@ -0,0 +1,423 @@
...
+     *            destination folder. The type of derivative to be generated is
+     *            defined in the properties of the destination folder.
+     */
+    void createDerivatives(Pair<Pair<String, URI>, List<Subfolder>> instruction) {
+        try {
+            Pair<String, URI> imageSource = instruction.getLeft();
+            for (Subfolder destinationFolder : instruction.getRight()) {
+                generateDerivative(imageSource.getValue(), destinationFolder, imageSource.getKey());
+            }
+        } catch (IOException e) {
```
Kathrin-Huber: Why do you replace the IOException with an UndeclaredThrowableException?

Kitodo/src/main/java/org/kitodo/services/image/MissingImagesFilterPredicate.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Check if there is a corresponding file in the folder. The name of the
+     * file results from the settings of the folder, and the canonical name part
+     * and the variables passed in the constructor.
+     * 
+     * @return true, if the picture needs to be generated
+     */
+    @Override
+    public boolean test(Subfolder destination) {
```
Kathrin-Huber: Please name the method more narrative. What is tested? why is it tested?

Kitodo/src/main/java/org/kitodo/model/Subfolder.java
matthias-ronge: Both, in a sense. One is a folder at the core level (and in this sense the parameters of the physical folder), the other is a physical folder on the disk. It's a bit like newspapers. On the one hand there is the newspaper as a conceptual unit (complete edition, ie all the newspapers in a newspaper that have ever appeared). This corresponds to the folder object. Then there's the newspaper, a single one that's just outside the door. This corresponds to the subfolder object.

Kitodo/src/main/java/org/kitodo/model/Subfolder.java
matthias-ronge: Right. I do not need it because it is always true. But without the method call, accessing `group()` below would not produce any result.

Kitodo/src/main/java/org/kitodo/services/image/ImageGenerator.java
matthias-ronge: Because these methods return a java image object in memory. In contrast, the image created by createDerivative is not a Java image (object), but only an image (written to disk). I wanted to better distinguish the two things with this choice of words.

Kitodo/src/main/java/org/kitodo/services/image/ImageGenerator.java
matthias-ronge: Because the method is called within a consumer, and the interface consumer does not allow a checked exception. I could also throw the exception here and catch it at the next higher level, but I would have to do it in one of the two places, and that method is shorter, so I did it here so that both methods stay clearer.

Kitodo/src/main/java/org/kitodo/services/image/MissingImagesFilterPredicate.java
matthias-ronge: I can not do that. The class implements the Predicate interface and the interface defines what the method should be called. It is tested whether there is a matching file in the given folder to the canonical part of the file name transferred during the creation of the object. If not, true is returned to enable the folder to be generated for this derivative.

Kitodo/src/main/java/org/kitodo/services/image/GenerationMode.java
matthias-ronge: I did, even though I find it awful, because I had to declare quite a few methods `public` that should not be. The enum is very closely linked to the image generator, is only used internally in it, and has no other purpose.

matthias-ronge:
Erstmal: Ich finde es schön dass du dir die Mühe machst den Code funktional zu verstehen.

Question about the file service: Is it obligatory to use the service? I do not quite understand the concept behind it. If I only use the module at this point I have no further dependencies. I think that's better. Also, the methods I need (which are available on the interface) are missing from the service. Is there a documentary?

Kitodo/src/main/java/org/kitodo/services/image/MissingImagesFilterPredicate.java
Beacze: Is it necessary to make this class as an implementation of Predicate?

matthias-ronge:
That should now be integrated, apart from the renaming of the modules, these as mentioned above in a separate branch, because it has to be touched a lot. (The branch is already done, and once this branch is merged, I'll put it online for review.) It would be nice if you could still merge it before jour holidays, even though it's after today's release, so I can go ahead.

Kitodo/src/main/java/org/kitodo/services/image/MissingImagesFilterPredicate.java
matthias-ronge: Yes, otherwise I would not have done it.

PR #1870, opened 2018-10-04
Fix deleting LDAP group and add the empty required fields.

IkramMaalej:

PR #1872, opened 2018-10-05
Restrict ldap group update to prevent overwriting existing data.

henning-gerhardt:
Prevent overwriting of existing data which was introduced by #1870 .

### DIFF
--- a/Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
+++ b/Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
@@ -29,4 +29,4 @@ sambaPasswordMustChange = '2147483647',
 sambaPasswordHistory = '00000000000000000000000000000000000000',
 sambaLogonHours = 'FFFFFFFFFFFFFFFFFFFF',
 sambaKickoffTime = '0'
-WHERE id = 2 ;
+WHERE id = 2 AND homeDirectory IS NULL AND gidNumber IS NULL AND sambaSID IS NULL;
--- a/Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
+++ b/Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
@@ -11,7 +11,7 @@
 
 -- 1. Update the dummy LDAP group
 
-UPDATE ldapGroup SET homeDirectory = '/home/test/users/{login}',
+UPDATE ldapGroup SET homeDirectory = '/usr/local/kitodo/users/{login}',
 gidNumber = '242',
 userDN = 'cn={login},ou=user,o=TestOrg,dc=kitodo,dc=org',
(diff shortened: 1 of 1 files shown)


Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
```diff
@@ -29,4 +29,4 @@ sambaPasswordMustChange = '2147483647',
 sambaPasswordHistory = '00000000000000000000000000000000000000',
 sambaLogonHours = 'FFFFFFFFFFFFFFFFFFFF',
 sambaKickoffTime = '0'
-WHERE id = 2 ;
+WHERE id = 2 AND homeDirectory IS NULL AND gidNumber IS NULL AND sambaSID IS NULL;
```
Kathrin-Huber: does the id needs to be 2?

henning-gerhardt: I did not add id with value 2 as this was in the previous and merged pull request. My goal is only not to overwrite existing data.

henning-gerhardt:
I noticed yet that there is maybe even an wrong data inserted for homeDirectory (see my comment in #1870 ). Should this be fixed in this pull request or in a new migration file?

Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
Beacze: I think default one has name test, so maybe we could use name.

Beacze:
This migration should be updated. Adding new one is not necessary.

Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
Kathrin-Huber: I think searching by name is the better idea here.

henning-gerhardt: Ok. I will change it.

henning-gerhardt:
Changing WHERE condition on migration from id to title did not work:

```
[ERROR] Message    : You are using safe update mode and you tried to update a table without a WHERE that uses a KEY column
[ERROR] Location   : Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql (/home/travis/build/henning-gerhardt/kitodo-production/Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql)
```
So I must stay with `id = 2`

Kitodo-DataManagement/src/main/resources/db/migration/V2_54__update_LDAP_Group_table.sql
henning-gerhardt: Did not work as flyway migration complain about missing KEY column in WHERE condition.

Beacze:
Switch off safe update: 
SET SQL_SAFE_UPDATES = 0;

PR #1889, opened 2018-10-08
comment dataFormat module due to connection problems with loc.gov

Kathrin-Huber:


pom.xml
```diff
@@ -120,7 +120,7 @@
         <module>Kitodo-DataManagement</module>
         <module>Kitodo-API</module>
         <module>Kitodo-DataEditor</module>
-        <module>Kitodo-DataFormat</module>
+        <!--<module>Kitodo-DataFormat</module>-->
```
solth: You should add a comment here to explain why this module is commented out.

PR #1890, opened 2018-10-08
Fix saving of workflows

Erikmitk:
This PR removes the remote command from the `workflowEditDetails` form and writes the content of the saved diagram (xml + svg) inside an input field of the form. This fixes the duplicate POST request issue and we get the data in the Java Backend on the first try.


Kitodo/src/main/webapp/WEB-INF/resources/js/modeler-custom.js
```diff
@@ -27,4 +27,4 @@ $(window).on("load", function () {
 			availableUserRoles.push(role);
 		});
 	});
-});
+});
```
Kathrin-Huber: last line missing

Kitodo/src/main/webapp/WEB-INF/resources/js/modeler-custom.js
```diff
@@ -9,7 +9,7 @@
  * GPL3-License.txt file that was distributed with this source code.
  */
 
- var availableUserRoles = [];
```
Kathrin-Huber: Change in format. I think this file isn't even touched!?

Erikmitk: I removed an empty space in front of the declaration. Now it's correctly formatted.

PR #1893, opened 2018-10-09
Add enum File for config files

Beacze:
Replace string collection FileNames.


Kitodo/src/main/java/org/kitodo/config/ConfigProjects.java
```diff
@@ -43,12 +42,12 @@ public ConfigProjects(String projectTitle) throws IOException {
      * @param configPath
      *            config path
      */
-    public ConfigProjects(String projectTitle, String configPath) throws IOException {
-        if (!(new File(configPath)).exists()) {
-            throw new IOException("File not found: " + configPath);
+    private ConfigProjects(String projectTitle, File configPath) throws IOException {
```
henning-gerhardt: Method signature changed but not the JavaDoc.

Kitodo/src/main/java/org/kitodo/config/enums/File.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Get file.
+     *
+     * @return value of file
+     */
+    public java.io.File getFile() {
+        return this.ioFile;
+    }
+
+    public String getAbsolutePath() {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo/src/main/java/org/kitodo/config/enums/File.java
```diff
@@ -0,0 +1,106 @@
...
+    public static File getByName(String name) throws ConfigurationException {
+        for (File file : File.values()) {
+            if (file.getName().equals(name)) {
+                return file;
+            }
+        }
+        throw new ConfigurationException("Configuration file '" + name + "' doesn't exists!");
+    }
+
+    public boolean exists() {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo/src/main/java/org/kitodo/config/enums/File.java
```diff
@@ -0,0 +1,106 @@
...
+
+package org.kitodo.config.enums;
+
+import org.apache.commons.configuration.ConfigurationException;
+import org.kitodo.config.ConfigCore;
+
+/**
+ * This class collects config files and their names used throughout the code.
+ */
+public enum File {
```
henning-gerhardt: I would suggest an other name for this enum class as `File` as this could be misleading or require full package names. What about `KitodoFile`?

Erikmitk: Why`KitodoFile` and not `ConfigFile` as it holds config files? ¯\_(ツ)_/¯

Erikmitk:
The enum of the enum cannot be `File` or even`KitodoFile`. (Adding Kitodo does not improve the naming and does not make any sense. "What kind of a file? Kitodo File!" ¯\\\_(ツ)\_\/¯ )

As the comment states `This class collects config files and their names`. Therefore, the name should reflect that as well. My suggestions is `ConfigFile` to make that purpose clear throughout its use.

Beacze:
Class is now called Kitodo onfigFile.

PR #1895, opened 2018-10-10
Enables you to enable validation for folders

matthias-ronge:
How it works:
To turn on validation for the contents of a folder, switches appear on the task property page. By clicking on a switch, the function can be switched on or off.

Implementation:
A new table `validationFolders` (`task` × `folder`) is created. This saves the variable `validationFolders` in the `task`. The `taskEditDetails.xhtml` page has been enhanced with switches to enable and disable the validation in a step. The functionality for this has been implemented in the class `TemplateForm`. The same functionality for generating images has also been moved here, and duplicate code has been summarized. The `GeneratorSwitch` class has been renamed to `FolderProcessingSwitch` because the switches are used equally for both types of processing. The tests were adapted or supplemented.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/forms/TemplateForm.java
+++ b/Kitodo/src/main/java/org/kitodo/forms/TemplateForm.java
@@ -14,8 +14,11 @@ package org.kitodo.forms;
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.MessageFormat;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.SessionScoped;
@@ -26,6 +29,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.kitodo.config.ConfigCore;
 import org.kitodo.config.enums.ParameterCore;
+import org.kitodo.data.database.beans.Folder;
+import org.kitodo.data.database.beans.Project;
 import org.kitodo.data.database.beans.Task;
 import org.kitodo.data.database.beans.Template;
 import org.kitodo.data.database.exceptions.DAOException;
@@ -34,7 +39,7 @@ import org.kitodo.enums.ObjectType;
 import org.kitodo.helper.Helper;
 import org.kitodo.helper.SelectItemList;
 import org.kitodo.model.LazyDTOModel;
-import org.kitodo.util.GeneratorSwitch;
+import org.kitodo.services.data.TaskService;
 import org.kitodo.workflow.model.Reader;
 
 @Named("TemplateForm")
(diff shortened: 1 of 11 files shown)


Kitodo/src/main/java/org/kitodo/forms/TemplateForm.java
```diff
@@ -395,6 +401,94 @@ public void setTask(Task task) {
         this.task = task;
     }
 
+    /**
```
Beacze: Is it needed to generate images for templates (Produktionsvorlagen)?

Kitodo/src/main/webapp/WEB-INF/templates/includes/taskTemplateEdit/taskEditDetails.xhtml
```diff
@@ -18,7 +18,7 @@
         xmlns:ui="http://xmlns.jcp.org/jsf/facelets">
     <p:panelGrid columns="2" layout="grid">
         <p:row>
-            <p:repeat var="generator" value="#{TemplateForm.task.generators}">
```
Beacze: According to design this page is going to be removed in very near future. Probably it is not the page to which you want to add it? What about taskEdit/taskEditDetails?

Kitodo/src/main/java/org/kitodo/forms/TemplateForm.java
matthias-ronge: No, the user only chooses in the template what to generate and what to validate. It is then generated or validated in the task. Look again at the page “Generieren von Bildern”, in particular the points 3 and 5. For the validating it is the same as the generation, only that the pictures are not generated, but validated. Or, if the user wants that, of course, both, so that they are first generated and then validated, whether they are OK.

Kitodo/src/main/webapp/WEB-INF/templates/includes/taskTemplateEdit/taskEditDetails.xhtml
matthias-ronge: According to what I have agreed with Elena, the additional switches should be added to the existing switches at tasks template. I had followed it on the edge that you have removed the other functions here. This has unfortunately overlapped. These changes should have been done before, if the branch would have been merged in mid-August, since then it's around. So basically these switches have to go to the others too. But that's what I would suggest in a separate ticket, otherwise it gets too confusing.

Beacze: But this is wrong page... It should be tasks of process. I will remove this page in next week.

matthias-ronge: No, that's not it. It can not be the task of the process, otherwise you would have to configure it individually for each process. It just needs to be configured just once, such as metadata step, script step, etc. The switches you have moved away. These switches must also be moved away if you want to remove the page. But one step at a time. I also find it awkward that this has overlapped in time now. But we have to live with that. In any case, it will need its own ticket, because there really has to be a lot of change. If I have seen it correctly, the configuration for metadata step, script step etc. are now in the workflow. If the configuration for the content to be generated and validated must be there, this means that the folder configuration must also be moved to the workflow. Basically, when I think about it, I'm absolutely in favor of it. The workflow also has a lot to do with the folders with content, they are finally processed in the workflow. Basically, they have always been an alien in the project because they have nothing to do with the project. But as I said, that's going to be a big change. Let's make our own ticket for it.

matthias-ronge:
I rebased the branch, but that was a big change. Wait until Travis runs through before you look at it again. If not I have to look again tomorrow. But I wanted to be home an hour ago ...

matthias-ronge:
Okay, Travis is green, you can review it if you have time.

PR #1905, opened 2018-10-11
Automatically generate images in a task

matthias-ronge:
Functioning:
When a task that has content folder generation enabled is completed, or when an automatic task runs, the requested subfolders are generated.

Implementation:
The task script thread, which always runs when completing a task or during automatic tasks, has been expanded to cause the images to be generated as well. The corresponding method was installed in the `TaskService`.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
+++ b/Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
@@ -58,11 +58,16 @@ import org.kitodo.data.elasticsearch.search.Searcher;
 import org.kitodo.data.exceptions.DataException;
 import org.kitodo.dto.TaskDTO;
 import org.kitodo.dto.UserDTO;
+import org.kitodo.enums.GenerationMode;
 import org.kitodo.helper.Helper;
 import org.kitodo.helper.VariableReplacer;
+import org.kitodo.helper.tasks.EmptyTask;
+import org.kitodo.model.Subfolder;
 import org.kitodo.services.ServiceManager;
 import org.kitodo.services.command.CommandService;
 import org.kitodo.services.data.base.TitleSearchService;
+import org.kitodo.services.file.SubfolderFactoryService;
+import org.kitodo.services.image.ImageGenerator;
 
 /**
  * The class provides a service for tasks. The service can be used to perform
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -700,6 +705,35 @@ private void abortTask(Task task) throws DataException {
...
+     *             If the task cannot be saved
+     */
+    public void generateImages(EmptyTask executingThread, Task task, boolean automatic) throws DataException {
+        Process process = task.getProcess();
+        ImageGenerator generator = new ImageGenerator(new Subfolder(process, process.getProject().getGeneratorSource()),
+                GenerationMode.ALL, SubfolderFactoryService.createAll(process, task.getContentFolders()));
+        generator.setSupervisor(executingThread);
+        if (automatic) {
+            if (executingThread.getException() == null) {
+                task.setEditType(TaskEditType.AUTOMATIC.getValue());
```
Beacze: In both cases the same TaskEditType is set up. In that case it should be set before if.

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
```diff
@@ -700,6 +705,35 @@ private void abortTask(Task task) throws DataException {
...
+                task.setProcessingStatus(TaskStatus.DONE.getValue());
+            } else {
+                task.setEditType(TaskEditType.AUTOMATIC.getValue());
+                task.setProcessingStatus(TaskStatus.OPEN.getValue());
+                save(task);
```
Beacze: Why task is saved only in case of exception? If generation was successful there is no need to save task with processing status set up to done?

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
matthias-ronge: You're right, that could have been better cleaned up, and I did that now. The function is now implemented only once and runs in script tasks and generate images equally.

Kitodo/src/main/java/org/kitodo/services/data/TaskService.java
matthias-ronge: Exciting question! I do not know either. But has always been like that and seems to work. Maybe it will then be saved elsewhere? Maybe it's also a bug. You can find out if you want it.

Beacze: You are re implementing it so you find out what is the correct way to do it.

henning-gerhardt:
Merge conflict in

- Kitodo/src/main/java/org/kitodo/services/data/TaskService.java

Please solve this conflict.

PR #1914, opened 2018-10-15
Remove not used find methods

Beacze:


henning-gerhardt:

- Kitodo/src/main/java/org/kitodo/services/data/BatchService.java

Please solve this conflict.

PR #1918, opened 2018-10-16
Remove not used methods in Project and Process forms

Beacze:
- processDTOS list is not used
- apply method is not used (code repeats save method)


henning-gerhardt:

- Kitodo/src/main/java/org/kitodo/forms/ProcessForm.java

Please solve this conflict.

PR #1919, opened 2018-10-16
Unify save of properties for process

Beacze:
Make insert, edit and delete permanent after save process button is clicked.


henning-gerhardt:

- Kitodo/src/main/java/org/kitodo/forms/ProcessForm.java

Please solve this conflict.

PR #1926, opened 2018-10-17
Extend indexing with partial indexing

Beacze:
Depends on #1925


Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -699,6 +699,8 @@ search=Suche
 start=Start
 startDate=Startdatum
 startIndexing=Indexierung starten
+startIndexingAll=Indexierung starten f\u00FCr allen
```
henning-gerhardt: Maybe better translation would be `Indexierung \u00FCber alles starten`

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -699,6 +699,8 @@ search=Suche
...
+startIndexingRemaining=Indexierung starten f\u00FCr nicht indexiert
```
henning-gerhardt: A maybe better translation would be `Indexierung f\u00FCr nicht Indexierte starten`

Kitodo/src/main/resources/messages/messages_de.properties
Erikmitk: What about: "Komplette Indexierung starten"

Kitodo/src/main/resources/messages/messages_de.properties
Erikmitk: Hmm… tough one… maybe `Indexierung vervollständigen`?

Beacze:
I have changed translations and rebased on master.

PR #1934, opened 2018-10-19
Implements a lock system for file access

matthias-ronge:
How it works:
There are four different types of locks that are enforced. The complete explanation can be found in the package-info.java.

Implementation:
The locks are implemented in the file management interface. There are four varieties of it to deal with all sorts of cases. The management is divided into three parts: the lock management, the stream management and the immutable read file management. The lock management manages the granted locks. The stream management manages the open streams through the locks. The immutable read file management manages the temporary files needed for the immutable read lock.


Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/ImmutableReadFileManagement.java
```diff
@@ -0,0 +1,240 @@
...
+     * @param originUri
+     *            URI for which the immutable read copy is to be returned
+     * @return the immutable read copy
+     * @throws IOException
+     *             if the file does not exist or if an error occurs in disk
+     *             access, e.g. because the write permission for the directory
+     *             is missing
+     */
+    private URI getUpToDateCopy(URI originUri) throws IOException {
+        URI upToDateCopy = upToDateCopies.get(originUri);
```
Beacze: Map.computeIfAbsent method could be used to replace the code like this:
      ```List<String> list = map.get(key);
            if (list == null) {
                list = new ArrayList<>();
                map.put(key, list);
             }```

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,511 @@
...
+                if (instance == null) {
+                    instance = new LockManagement();
+                }
+            }
+        }
+        return instance;
+    }
+
+    private static GrantedAccess tryCast(LockingResult lockingResult) throws AccessDeniedException {
+        if (lockingResult == null || !(lockingResult instanceof GrantedAccess)) {
```
Beacze: Unnecessary null check before instanceof - instanceof returns false in case of null.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,511 @@
...
+     * Checks for a group of locks if they can be returned.
+     *
+     * @param uris
+     *            the URIs on the locks to be returned
+     * @throws IllegalStateException
+     *             if there is still a stream flowing over the lock
+     */
+    private void canClose(Collection<URI> uris) {
+        String exceptionMessagePrefix = "The lock cannot be released because there are still streams flowing for the following URIs: ";
+        String exceptionMessage = uris.parallelStream().filter(uri -> streamManagement.isKnowingAnOpenStreamTo(uri))
```
Beacze: Lambda can be replaced with method reference.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,511 @@
...
+     * @param uri
+     *            URI for which the lock is desired
+     * @param requestedLock
+     *            which lock is desired
+     * @return {@code null} on success, otherwise a pair of the URI and all
+     *         users who have locks that conflict with the requested lock
+     */
+    private Pair<URI, Collection<String>> checkLockability(String user, URI uri, LockingMode requestedLock) {
+        if (grantedPermissions.get(uri) == null) {
+            return null;
```
Beacze: Null return should be avoided. Pair also should be avoided. For such a cases classes are better. In that case you could return class with empty collection of strings.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,511 @@
...
+     *
+     * @param locks
+     *            the locks that are returned
+     * @throws IllegalStateException
+     *             if there is still a stream flowing over the lock
+     */
+    void close(String user, Map<URI, AbstractLock> locks) {
+        synchronized (this) {
+            canClose(locks.keySet());
+            locks.entrySet().parallelStream().map(entry -> entry.getKey())
```
Beacze: Lambda can be replaced with method reference.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,511 @@
...
+     * @param <K>
+     *            type of map key
+     * @param <V>
+     *            type of map value
+     * @param pairStream
+     *            stream of pairs (future map entries)
+     * @return a map containing the pairs
+     */
+    private <K, V> Map<K, V> collectPairsToMap(Stream<Pair<K, V>> pairStream) {
+        return pairStream.collect(Collectors.toMap(pair -> pair.getKey(), pair -> pair.getValue()));
```
Beacze: Lambda can be replaced with method reference.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,511 @@
...
+     *
+     * @param <K>
+     *            type of map key
+     * @param <V>
+     *            type of map value
+     * @param pairStream
+     *            stream of pairs (future map entries)
+     * @return a map containing the pairs
+     */
+    private <K, V> Map<K, V> collectPairsToMap(Stream<Pair<K, V>> pairStream) {
```
Beacze: Class usage would save you such an operation.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,511 @@
...
+     *             if the file does not exist or if an error occurs in disk
+     *             access, e.g. because the write permission for the directory
+     *             is missing
+     */
+    private GrantedAccess grantLocks(LockRequests requests, GrantedAccess rights) {
+        Map<URI, AbstractLock> locks = createStreamOfLockRequests(requests)
+                .map(requestedLock -> Pair.of(requestedLock.getKey(),
+                    getInitializedGrantedPermissions(requestedLock.getKey()).addAGrantedLock(requests.getUser(),
+                        requestedLock.getValue(), streamManagement, immutableReadFileManagement, rights)))
+                .collect(Collectors.toMap(entry -> entry.getKey(), entry -> entry.getValue()));
```
Beacze: Lambda can be replaced with method reference.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockRequests.java
```diff
@@ -0,0 +1,130 @@
...
+        LockRequests other = (LockRequests) obj;
+        if (locks == null) {
+            if (other.locks != null) {
+                return false;
+            }
+        } else if (!locks.equals(other.locks)) {
+            return false;
+        }
+        if (user == null) {
+            if (other.user != null) {
```
Beacze: IF can be simplified to one statement.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/UpgradeableReadLock.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param uri
+     *            locked URI
+     * @param streamManagement
+     *            the stream management
+     * @param immutableReadFileManagement
+     *            the immutable read file management
+     * @param write
+     *            whether the lock is currently in write mode
+     */
+    UpgradeableReadLock(String owner, URI uri, StreamManagement streamManagement,
```
Beacze: StreamManagement streamManagement and ImmutableReadFileManagement immutableReadFileManagement are not used.

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
```diff
@@ -38,24 +40,89 @@
...
      *
      * @param uri
      *            the uri to write to
      * @return an writable OutputStream
+     * @deprecated This write function creates an exclusive lock with the
+     *             meaningless user name “System”. Therefore, this writing
+     *             function should not be used anymore. Use
+     *             {@link #write(URI, LockingResult)} instead.
      */
+    @Deprecated
```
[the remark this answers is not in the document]
matthias-ronge: The thing is, since the interface does not get passed a user, only one standard user “System” is ever used for the locks, which is stupid. And that every read requires an exclusive lock, I do not think so either. We can discuss that, but with whom?

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
```diff
@@ -38,24 +40,89 @@
...
+     *            opening of the stream
+     * @return an writable OutputStream
+     * @throws ProtocolException
+     *             if the file had to be first read in again, but this step was
+     *             skipped on the protocol. This error can occur with the
+     *             UPGRADE_WRITE_ONCE lock because its protocol form requires
+     *             that the file must first be read in again and the input
+     *             stream must be closed after the lock has been upgraded.
+     */
+    OutputStream write(URI uri, LockingResult permission) throws IOException;
```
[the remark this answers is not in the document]
matthias-ronge: Yes, the method throws an I/O excepiton. That's not explained, because that's obvious when opening an output stream. In the existing methods, the I/O exceptions are nowhere explained. I agree with that. What I have explained are the exceptions that have a special meaning. Protocol exception is a subclass of I/O exception and therefore does not need to be declared explicitly.

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
```diff
@@ -38,24 +40,89 @@
...
      *
      * @param uri
      *            the uri to write from
      * @return a readable InputStream
+     * @deprecated This read function creates an exclusive lock with the
+     *             meaningless user name “System”. Therefore, this reading
+     *             function should not be used anymore. Use
+     *             {@link #read(URI, LockingResult)} instead.
      */
+    @Deprecated
```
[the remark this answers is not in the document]
matthias-ronge: The thing is, since the interface does not get passed a user, only one standard user “System” is ever used for the locks. (see above)

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -25,25 +25,32 @@
...
+import org.kitodo.filemanagement.locking.GrantedAccess;
+import org.kitodo.filemanagement.locking.LockManagement;
 
 public class FileManagement implements FileManagementInterface {
 
     private static final Logger logger = LogManager.getLogger(FileManagement.class);
     private static final FileMapper fileMapper = new FileMapper();
 
     private static final String IMAGES_DIRECTORY_NAME = "images";
+    private static final String SYSTEM_USER = "System";
```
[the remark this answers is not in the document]
matthias-ronge: Not at all. That's why I wanted to change the methods at first; and now, after you asked me to stop that and just add new methods, mark them as deprecated, so they will not be used in the future, and hopefully they will be removed in the old code in the long term.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/DeniedAccess.java
```diff
@@ -0,0 +1,76 @@
...
+    private final Map<URI, Collection<String>> conflicts;
+
+    /**
+     * Generates a new result object in case locks could not be granted.
+     *
+     * @param conflicts
+     *            why did not work
+     */
+    public DeniedAccess(Map<URI, Collection<String>> conflicts) {
+        assert !conflicts.isEmpty() : "the conflict map should not be empty";
```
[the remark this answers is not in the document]
matthias-ronge: I do not understand what you have against assertions, but I changed it.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/ImmutableReadFileManagement.java
```diff
@@ -0,0 +1,240 @@
...
+
+    /**
+     * Map of the up-to-date copies.
+     */
+    private final Map<URI, URI> upToDateCopies = new HashMap<>();
+
+    /**
+     * Which copy of a URI was passed to which user.
+     */
+    private final Map<URI, Map<String, Pair<URI, AtomicInteger>>> urisGivenToUsers = new HashMap<>();
```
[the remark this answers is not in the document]
matthias-ronge: More classes? This pull request is already so big anyway. Well, if you want …

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/ImmutableReadFileManagement.java
matthias-ronge: Changed 👍

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
matthias-ronge: I did not know that yet! Great! 👍

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
matthias-ronge: It seems to me, own class for this is a bit of overkill, but if you want ...

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockRequests.java
matthias-ronge: No, this only works at first glance, but otherwise the else if block would also be executed if both objects are null (and would result in a NullPointerException). By the way: This code is generated automatically by Eclipse. I assume the underlying method was reviewed by a bunch of computer science professors worldwide. In short, that works, we do not need to review that.

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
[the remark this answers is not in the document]
matthias-ronge: The thing is, since the interface does not get passed a user, only one standard user “System” is ever used for the locks, which is stupid. And that every read requires an exclusive lock, I do not think so either. We can discuss that, but with whom?

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -72,16 +79,37 @@ private URI createResource(URI targetFolder, String fileName) throws IOException
     }
 
     @Override
+    @Deprecated
```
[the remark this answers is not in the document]
matthias-ronge: The thing is, since the interface does not get passed a user, only one standard user “System” is ever used for the locks. (see above)

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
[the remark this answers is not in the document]
matthias-ronge: Not at all. That's why I wanted to change the methods at first; and now, after you asked me to stop that and just add new methods, mark them as deprecated, so they will not be used in the future, and hopefully they will be removed in the old code in the long term.

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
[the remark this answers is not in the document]
matthias-ronge: The thing is, since the interface does not get passed a user, only one standard user “System” is ever used for the locks, which is stupid. And that every read requires an exclusive lock, I do not think so either. We can discuss that, but with whom?

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
```diff
@@ -38,24 +40,89 @@
...
+     * Attempts to get locks on one or more files. There are only two results:
+     * either, all locks can be granted, or none of the requested locks are
+     * granted. In the former case, the conflicts map in the locking result is
+     * empty, in the latter case it contains for each conflicting file the users
+     * who hold a conflicting lock on the file. If no locks have been granted,
+     * the call to {@code close()} on the locking result is meaningless, meaning
+     * that leaving the try-with-resources statement will not throw an
+     * exception. Just to mention that.
+     * 
+     * @param user
```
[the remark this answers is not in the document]
matthias-ronge: The core can determine the logged in user, see [here how this is done](https://github.com/kitodo/kitodo-production/pull/1934/files/89bbf3f482b6a4436ff3b1f1c17c39fd6a20682a#diff-ea3503a51ddbd81a2ac5534ab5840b77R202). Yes, exactly, before any reading or writing process, a request must first be made, in which a (lowest possible) appropriate lock is requested, and only if that succeeds the file can be read or written. That's the principle. However, if a file is to be written again later (i.e. in the meta-data editor), the lock should be fetched at the beginning and kept until the meta-data editor is closed again. Only then, the lock management can ensure that several conflicting requests from different users at the same time hinder each other as little as possible.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
matthias-ronge: It seems to me, own class for this is a bit of overkill, but if you want ...

henning-gerhardt:

Beacze:

matthias-ronge:
Hello @Kathrin-Huber, this pull request was reviewed by @Beacze but she would like you to look at it as well. As you can see, the code is well covered with tests that show it works. Let me know if I still have to do something here.

matthias-ronge:
Can this branch be merged?

Kitodo-FileManagement/pom.xml
```diff
@@ -44,12 +44,23 @@
...
+            <artifactId>commons-lang3</artifactId>
+            <version>${commons-lang3.version}</version>
+        </dependency>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
Kathrin-Huber: please use junit, we do not need several testing libraries. Or which aspect do you need, junit doesn't offer?

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/LockingResult.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Manages allocated locks or provides information about conflict originators in
+ * case of error.
+ */
+public interface LockingResult extends AutoCloseable {
```
Kathrin-Huber: I think "LockingResult" is a missleading name. The Javadoc says:"manages" maybe "LockingManager" is more appropriate?

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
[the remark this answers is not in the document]
Kathrin-Huber: But we don't need locks for every file, right? So i think we just need both. Maybe rename them to "concurrentWrite" and "lockedWrite" (or something like that, don't call them excatly like that)

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
```diff
@@ -0,0 +1,494 @@
...
+     * be used as singleton, meaning that there is only one lock management all
+     * over the class loader. For the usage, only the singleton should be used
+     * and no more instances of the class should exist. Everything else would
+     * not make sense or would even be dangerous.
+     *
+     * @return the singleton instance of this class
+     *
+     * @see "https://en.wikipedia.org/wiki/Double-checked_locking"
+     */
+    public static LockManagement singleton() {
```
Kathrin-Huber: Can you please use the standard "getInstance" method for the singleton?
I think a private constructore is missing here too

Kathrin-Huber:
There is a Java Library, which implements locking on files. Its using FileChannels.
See here: https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileLock.html

Kitodo-FileManagement/pom.xml
matthias-ronge: There is a long explanation from me in the corresponding ticket in the Jira, including examples. Abstract: It provides much better error descriptions when a test fails. Many tests can be written shorter and more readable.
Either way, that's a separate topic. We should discuss it there and decide how we handle it.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
matthias-ronge: I renamed the getter. A private constructor is not possible because otherwise the tests will not be able to get their own instance of the class. In operation, the class should function as a singleton, but in the tests each test must be able to get a newly initialized lock management.

matthias-ronge:
Yes, but it does not provide the flexibility needed and depends on the capabilities of the underlying operating system and file system. The javadoc explains that if the operating system does not support it, it simply generates exclusive locks. We can not use that here.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
Erikmitk: But that's not an acceptable reason. If each tests needs a newly initialized lock management that you have to write your tests or class accordingly. That is definitely possible.

You could implement a `clear`-method so you're able to wipe the instance if necessary. Then you can set up a new one for each test. Singleton doesn't mean that it has to stay around forever but that there can't be multiple ones at the same time.

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/LockingResult.java
Erikmitk: Maybe something like `LockProtocol` since it's defining the commands you can pass on.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
matthias-ronge: If the lock management has a clear method that is not better, you can also use it wrong. If possible, I would also like to avoid having methods in the productive code that are only used by tests. But if you care so much I can change it.

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/LockingResult.java
matthias-ronge: I would not say so. OK, you could rename it to `LockProtocol`. But then you would have to rename more, because then the whole metaphor does not match any more, isn’t it? At the moment, you call `tryLock()`. As a result, you get a `LockResult`, and that offers, for example, the method `isSuccesful()`. The result of trying to lock is successful, or maybe not. If not, I can `getConflicts()`, the conflict is a possible _result_ of the attempt to lock. If I rename that to _protocol_ now—on my request, trying to lock a file, will I get back a _protocol_ that I can then ask if it succeeded? What is a successful protocol? And then I can ask the _protocol_ if it has a conflict? I find this inconsistent, there was a conflict while trying to lock, and not a conflict in the protocol, then would rather thrown an `Exception`, isn’t it? A conflict in the protocol would be more likely for me if I request a copy of a file but the file can not be copied, not because of a locking conflict with another user, but, e.g. because the disk is full...

But if you find that much easier to understand, I rename it.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/locking/LockManagement.java
Erikmitk: > I would also like to avoid having methods in the productive code that are only used by tests

The only reason for violating the private constructor idea is that your tests should work. You're doing what you don't want to do as well.

I'd prefer it if the tests were independent and wouldn't need a new instance every time. But as this is not happening with the Singleton pattern a `clear`-method should be preferred. That's a clean interface that's not dictated by the tests.

> you can also use it wrong

But that's what exactly what code review is for. So that you don't use it wrong.😛

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/LockingResult.java
matthias-ronge: Okay, let's do this…!

matthias-ronge:
Is there anything else I need to do, or can it be merged that way?

PR #1954, opened 2018-10-30
Adds an interface for accessing METS

matthias-ronge:


Kitodo-API/src/main/java/org/kitodo/api/dataformat/service/AreaServiceInterface.java
```diff
@@ -0,0 +1,49 @@
...
+ * that describe their positioning.
+ */
+public interface AreaServiceInterface {
+    /**
+     * Returns a service to access the media unit ({@code <mets:file>} element)
+     * on which the view is based.
+     *
+     * @return a service to access the file
+     */
+    FileServiceInterface getFile();
```
Beacze: It is confusing - File or FileService? Method name says one, parameter says second.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/service/DivServiceInterface.java
```diff
@@ -0,0 +1,107 @@
...
+    Collection<MetaDataServiceInterface> getMetaData();
+
+    /**
+     * Returns a variant of the human-readable name of this outline element,
+     * which has been normalized so that it can be properly sorted by primitive
+     * sorting algorithms.
+     *
+     * @return the orderlabel
+     */
+    String getOrderlabel();
```
Beacze: OrderLabel?

Kitodo-API/src/main/java/org/kitodo/api/dataformat/service/DivServiceInterface.java
```diff
@@ -0,0 +1,107 @@
...
+     * @return the label
+     */
+    String getLabel();
+
+    /**
+     * Returns all meta-data that describe this outline level.
+     *
+     * @return the meta-data
+     */
+    Collection<MetaDataServiceInterface> getMetaData();
```
Beacze: MetaData or MetaDataService? I think metadata is an one word: https://www.wikiwand.com/en/Metadata

matthias-ronge: After all that I know, the only real corret spelling is “meta-data”, with dash. Therefore the camel case.

See [this excellent blog post](http://tdan.com/the-meta-data-fiasco/18502), starting from about one screen down where it says: “A second misperception with meta-data is the spelling.”

Kitodo-API/src/main/java/org/kitodo/api/dataformat/service/DivServiceInterface.java
matthias-ronge: The METS attribute is uppercase ORDERLABEL. So I concluded to call the getter without camel case.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/service/AreaServiceInterface.java
matthias-ronge: You will see that all methods return services (and of course strings, etc.), and in the names of the methods the "service" does not appear anywhere. This is because the interface should be named according to METS so that, knowing METS, it is intuitively understood. If I could decide that for myself, I would call the interfaces and methods quite differently, not even "File", because it's not a file at all. If you are not at home in METS, this seems strange at first glance, but it is primarily aimed at people who deal with METS on a daily basis. I have tried to describe in the javadocs what the objects mean exactly. You'll see that it's something completely different from what you would expect if you did not think in METS. But I've decided to name the classes that implement the interface differently, to better understand what they mean. I hope that's enough for both sides.

matthias-ronge:
Does it suit?

henning-gerhardt:

Kitodo-API/src/main/java/org/kitodo/api/dataformat/service/DivServiceInterface.java
Erikmitk: I agree with @Beacze that we should adopt the spelling as `metadata`. We can't go by a blogpost or the whole history of the word but how it is used today. (I know that might be hard for you @matthias-ronge since your background is in linguistics.) Even the blog posts says `Meta must be combined with a root word, either with a hyphen or by concatenation` So that's what we're gonna do.

 Every dictionary I checked uses the single word spelling and therefore we use it that way.

Beacze:

Kitodo-API/src/main/java/org/kitodo/api/dataformat/service/DivServiceInterface.java
matthias-ronge: No, it's not that bad for me. I just wanted to point out that there is another view here as well. And I was once told by an American that Production would be so difficult to understand because the English would be used so wrong in part. That's why I try a little bit to pay attention. But now it's clearly 2 to 1, so I've changed it now.

matthias-ronge:
@Kathrin-Huber, do you contact me for an appointment so that we can review this?

PR #1955, opened 2018-10-30
Adds an interface for accessing the ruleset

matthias-ronge:


henning-gerhardt:

Beacze:

matthias-ronge:
OK. But would be nice if this would go through the review in a timely manner, so I can continue there. The other two are not so urgent.

matthias-ronge:
I renamed the classes and methods as discussed. The `Map` we talked about at `getSortedVisibleMetadatas()` I replaced with a collection and added setter for access functions to the `RulesetManagementInterface`. Oh, and a function `load()`, that was missing completely what wasn’t noticed. I hope that’s what you think.

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/ComplexMetadataViewInterface.java
```diff
@@ -0,0 +1,69 @@
...
+     * @param <T>
+     *            the type of meta-data objects
+     * @param entered
+     *            meta-data objects that have already been entered, along with
+     *            their key
+     * @param additionallySelected
+     *            meta-data keys that the user has already selected
+     * @return the meta-data keys that the user can add
+     */
+    <T> Collection<MetadataViewInterface> getAddableMetadatas(Map<T, String> entered,
```
Kathrin-Huber: There is no plural "metadatas" , its just "metadata"

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/ComplexMetadataViewInterface.java
```diff
@@ -0,0 +1,69 @@
...
+     * @param entered
+     *            meta-data objects that have already been entered, along with
+     *            their key
+     * @param additionallySelected
+     *            meta-data keys that the user has already selected
+     * @return The sorted list of keys with the value objects. If keys are
+     *         already filled with values, the values are returned here. If a
+     *         key is a multiple-selection, values are grouped below it.
+     */
+    <T> List<MetadataViewWithValuesInterface<T>> getSortedVisibleMetadatas(Map<T, String> entered,
```
Kathrin-Huber: Same here

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/MetadataViewWithValuesInterface.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public interface MetadataViewWithValuesInterface<T> {
+    /**
+     * Returns the key to be displayed at this point. The interface can return
+     * entries where the key is {@code null} with values if there were values
+     * for excluded keys.
+     *
+     * @return the key
+     */
+    Optional<MetadataViewInterface> getMetadata();
```
Kathrin-Huber: Why do you use optional here?

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+public interface SimpleMetadataViewInterface extends MetadataViewInterface {
+    /**
+     * Maps a boolean value to a meta-data value.
+     *
+     * @param value
+     *            boolean input value
+     * @return value to save as a meta-data entry. If absent, delete the
+     *         meta-data entry.
+     */
```
Kathrin-Huber: The method returns the first key when true. What us rhe "mapping" part here?

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
```diff
@@ -0,0 +1,106 @@
...
+    default Optional<String> convertBoolean(boolean value) {
+        return value ? Optional.of(getSelectItems().entrySet().iterator().next().getKey()) : Optional.empty();
+    }
+
+    /**
+     * Returns the default value for the input type boolean.
+     *
+     * @return the default value
+     */
+    default boolean getBooleanDefaultValue() {
```
Kathrin-Huber: A default value, which depends on a condition?

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
```diff
@@ -0,0 +1,106 @@
...
+     * @return the default value
+     */
+    Collection<String> getDefaultItems();
+
+    /**
+     * Returns the default value for all input types except boolean and multiple
+     * selection.
+     *
+     * @return the default value
+     */
```
Kathrin-Huber: Why is there a list, if you always return the first object? And how can a value be valid for several input types?

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/MetadataViewWithValuesInterface.java
matthias-ronge: Because only then a meta-dataₖₑᵧ is returned, if it should also be displayed. If there is a value whose meta-dataₖₑᵧ is not to be displayed, the value is returned without a meta-dataₖₑᵧ. The value is then returned anyway, because in the changed view (after the user edited the meta-dataₑₙₜᵣᵧ) it still has to be saved again, even though it is hidden.

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
matthias-ronge: The two boolean values true and false are mapped to the two states in the METS, true = the first value, false = no value.

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
matthias-ronge: Yes exactly. Either the default strings contain a value, then it's `true`, if not it's `false`. The value should then be the value that, if `true`, is also written to the meta-dataₜₒₜₐₗᵢₜᵧ.

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
matthias-ronge: The rule set contains the element `<preset>` for the default values. However, the element can only occur once except for multiselect lists, so this method returns the first (and only) value. Multiselect lists, on the other hand, use the `getDefaultItems()` getter, which then returns all elements, because a multi-select list has more than one selected element (multiple `<preset>` elements in the rule set).

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/ComplexMetadataViewInterface.java
matthias-ronge: I changed the two.

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/ComplexMetadataViewInterface.java
matthias-ronge: I changed the two.

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
matthias-ronge: _How can a value be valid for several input types?_ A value, not a specific value, but the fact that the number of default values is 1 may be correct for multiple input types. A simple selection list may have exactly one preselected value, a date may have exactly one value in advance, and so on.

matthias-ronge:
@Kathrin-Huber, I had removed the two plural -s and answered all the questions. Can this be merged?

PR #1968, opened 2018-11-05
Correct the sorting of the current edition

matthias-ronge:
Fixes the sorting of multiple issues on one day in the `year.xml` file for the current issue of the day, deriving the sort order from the meta-data entry `CurrentNoSorting`.

### DIFF
--- a/Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
+++ b/Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
@@ -772,6 +772,48 @@ public class ExportNewspaperBatchTask extends EmptyTask {
     }
 
     /**
+     * Returns the index of the issue to insert between its siblings depending
+     * on its rank. A return value of {@code null} will indicate that no
+     * position could be determined which will cause
+     * {@link DocStruct#addChild(Integer, DocStruct)} to simply append the new
+     * child at the end.
+     *
+     * @param siblings
+     *            brothers and sisters of the child to add
+     * @param rank
+     *            rank of the child to insert
+     * @return the index position to insert the child
+     */
+    private static Integer positionByRank(List<DocStruct> siblings, Integer rank) {
+        if (siblings != null && rank != null) {
+            SIBLINGS: for (int i = 0; i < siblings.size(); i++) {
+                DocStruct aforeborn = siblings.get(i);
+                List<Metadata> allMetadata = aforeborn.getAllMetadata();
+                if (allMetadata != null) {
+                    for (Metadata metadataElement : allMetadata) {
+                        if (metadataElement.getType().getName().equals(IndividualIssue.RULESET_ORDER_NAME)) {
+                            try {
+                                if (Integer.parseInt(metadataElement.getValue()) > rank) {
+                                    return i;
+                                } else {
+                                    continue SIBLINGS;
+                                }
+                            } catch (NumberFormatException e) {
+                                if (metadataElement.getValue().compareTo(rank.toString()) > 0) {
+                                    return i;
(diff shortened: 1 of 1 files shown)


Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
```diff
@@ -771,6 +771,48 @@ private static Integer positionByRank(List<DocStruct> siblings, String metadataT
...
+     * {@link DocStruct#addChild(Integer, DocStruct)} to simply append the new
+     * child at the end.
+     *
+     * @param siblings
+     *            brothers and sisters of the child to add
+     * @param rank
+     *            rank of the child to insert
+     * @return the index position to insert the child
+     */
+    private static Integer positionByRank(List<DocStruct> siblings, Integer rank) {
```
henning-gerhardt: Why did you add this method as a static one?

Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
```diff
@@ -771,6 +771,48 @@ private static Integer positionByRank(List<DocStruct> siblings, String metadataT
...
+        if (siblings != null || rank != null) {
```
henning-gerhardt: Is your condition `||` correct? If `siblings` is `null` and `rank` not `null` then it would create a null pointer exception in the next line.

Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
```diff
@@ -771,6 +771,48 @@ private static Integer positionByRank(List<DocStruct> siblings, String metadataT
...
+            SIBLINGS: for (int i = 0; i < siblings.size(); i++) {
+                DocStruct aforeborn = siblings.get(i);
```
henning-gerhardt: As I did not find any translation: what did `aforeborn` mean or is there some spelling error?

Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
matthias-ronge: It doesn’t need access to fields of the class instance. I basically declare any method static if it is. You do not have to, I know. But why should I leave the assumption of a dependency if I know that there is no such?

Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
matthias-ronge: children − _Kinder_; siblings − _Geschwister_; aforeborn − _Vorgeborene/r, älterer Bruder/ältere Schwester_
At least now I wonder: Why are the outgoing relations of a tree structure actually called children?

Goobi/src/de/sub/goobi/helper/tasks/ExportNewspaperBatchTask.java
matthias-ronge: Yes, that was wrong, of course. Thank you. Of course it should be called `&&`.

henning-gerhardt:
@matthias-ronge: We tested your change but the ordering on the same day is still wrong. Current issue will be first regardless of its real position.

henning-gerhardt:
Sorry, I deployed the wrong branch without your changes to our test system. We are testing now a version with your changes.

matthias-ronge:
@henning-gerhardt, can this be merged?

henning-gerhardt:
First test exports looks good, so it could be merged in couple of next days.

PR #1989, opened 2018-11-12
Implement parallel tasks in workflow - part 1

Beacze:
- read task sequence of execution from diagram
- separate gateways on the two types  - exclusive and parallel
- add tests for reading diagrams
- add database migration


Kitodo-DataManagement/src/main/resources/db/migration/V2_62__Add_sequential_columns_to_task_table.sql
```diff
@@ -0,0 +1,16 @@
...
+-- It is licensed under GNU General Public License version 3 or later.
+--
+-- For the full copyright and license information, please read the
+-- GPL3-License.txt file that was distributed with this source code.
+--
+
+-- Add columns for sequential tasks to task table.
+
+ALTER TABLE task ADD previousTasks VARCHAR (255);
+ALTER TABLE task ADD currentTasks VARCHAR (255);
```
Kathrin-Huber: It has to be "concurrent" not "current"

PR #1996, opened 2018-11-15
Removes the name component module from modules called module

matthias-ronge:


Kathrin-Huber:
There is an IT test failing

matthias-ronge:
Now it works. I found and replaced a few more _times_. Apparently there was something crooked there. Note: You have to **run the build completely**. I think Idea does it by itself, but Eclipse does not and then the new module is missing.

PR #2002, opened 2018-11-15
Rename the method name

matthias-ronge:


Kitodo/src/main/java/org/kitodo/services/image/ImageGenerator.java
```diff
@@ -311,12 +311,12 @@ public GenerationMode getMode() {
      * Lets the supervisor do something, if there is one. Otherwise nothing
      * happens.
      *
-     * @param action
+     * @param something
```
Kathrin-Huber: If it is an action, call it action, not "something"

Erikmitk: @matthias-ronge Have you seen this comment? I agree with @Kathrin-Huber that you should change the comment and not the parameter.

`Let's the supervisor perform an action, if there is one.` is more fitting for the method.

matthias-ronge: Okay, I misunderstood! I thought the method signature was meant. I had not looked at the Javadoc.

Kathrin-Huber:
Please resolve colflicts

PR #2012, opened 2018-11-26
Minor fixes in metadata classes

Beacze:
- remove not used methods: getWert and getObjectType
- remove duplicated method setWert (setValue calls this method and does nothing more)
- rename classes Metadatum to Metadata
- rename some of the metadata variables
- add JavaDocs to Metadata interface


Kathrin-Huber:
please resolve conflicts

PR #2013, opened 2018-11-26
Integrate workflow editor build from repo

Erikmitk:
I created a [distinct repo](https://github.com/Erikmitk/kitodo-workflow-editor) for our workflow editor customizations. I tested locally that it's possible to create, save and edit workflows. The underlying functionality should therefore be the same as before since I have made no additional changes yet.

This PR simply integrates the files build by this repo at the time of following commit https://github.com/Erikmitk/kitodo-workflow-editor/commit/73648f5e734ea60bcf761c9e88101a7050a77da3. Additionally, we can also remove a bunch of css files that were redundant anyway. So far I have used the license header that we added to our custom editor build.

I know this is not yet a perfect solution but at least we now have a versioned history of our changes over there. Making further customizations can now be done and verified more easily.

We should have a discussion on how to integrate the other repo with Kitodo.Production. Changes to the editor should only be made over there so we don't lose functionality accidentally. Maybe there's a good way to integrate this into the build process somehow.


Erikmitk:
I have fixed a hack in the editor build. See: https://github.com/Erikmitk/kitodo-workflow-editor/commit/b36ab0ccd8b77f8b3b7e56f640692d56f5cd8b5d

PR #2016, opened 2018-11-26
Implement the ruleset management interface

matthias-ronge:
Functionality:
The ruleset management interface is implemented. The interface provides all required functions of the rule set. In particular, the meta-data processing mask is formed. The implementation will be backed up with a series of tests to verify that it works correctly.

Implementation:
The ruleset management interface is implemented in the Kitodo-DataEditor module in the `org.kitodo.dataeditor.ruleset` package. The implementation is based on the model-view-controller-service paradigm. The parser used is JAXB.


Erikmitk:
> I'm sorry that it is so big, but I do not know how it would be smaller.

I think this could improve if you'd just commit a lot more in between. "Commit early, commit often." as the Git saying goes.

The commit https://github.com/kitodo/kitodo-production/pull/2016/commits/4d31ce2d50d7f4be3fd223097cf035408a0c4b44 (`Implement the ruleset interface`) could easily be split into multiple commits i.e. one per class perhaps. So as a reviewer it's easier to follow your plan of action while implementing it. Just show how you got to your final state.

Same with 
https://github.com/kitodo/kitodo-production/pull/2016/commits/ab165e04073905f6c71121866c79fb6e01925531 (`Add tests`). Why not one commit for adding resources and multiple commits for the tests themselves. They don't have to be committed as a whole. Why not split them up as well? They could be split up via the interface they are testing or per "large" test.

That has the additional side effect (not in this case but in general) that possible merge conflicts are a lot smaller since each change is incremental instead of one huge chunk that has to be dealt with at once. The overall lines of code changes will be the same but it's easier to digest from a code review perspective. :)

matthias-ronge:
That's how I work, but I'm not working as linearly as you imagine when I'm developing such a new feature. If I uploaded my "grown" commit history here (as I did earlier), it would look like [this](https://github.com/kitodo/kitodo-production/pull/1837#commits-pushed-1213ba9). After being criticized, I do not do that anymore. Program development is, at least for me, a highly creative process in which I constantly change a lot, until the code is both working and well organized.

Keeping the commit history also has the distinct disadvantage that a rebase is no longer possible when it comes to conflicts that have to be resolved manually, because then you have to solve the conflict not just once, but for each commit new. This is not only a lot of work, but I do not succeed in the end, because I then have to solve the conflict in the source code state of yesteryear, which I do not understand at the end of the development for myself. That means I can no longer make a rebase but must make a merge, which was also undesirable.

If I should build up such a commit history as you imagine, then at the end of the development I would have to copy out my working code and then insert it back step for step (i.e. copy it in, until it looks like it did before). Here I have already done that on a small scale. I see that browsing 8,000 lines of code on a single page is unwieldy. This has already taken me half an hour, if I had to make it more finely granular, it would claim correspondingly even more project time. I do not think that can be our goal.

Erikmitk:
> “Keeping the commit history also has the distinct disadvantage that a rebase is no longer possible when it comes to conflicts that have to be resolved manually, because then you have to solve the conflict not just once, but for each commit new.”

I think that's false and doesn't make sense. If you run into a merge conflict you have to solve this conflict at that point in time. Why would you have to go back and change single commits? That's not how this is supposed to work. You rebase and when a conflict occurs you fix it on the spot and continue your rebase (`git rebase --continue`). This is 100% possible because that's how this works. 

BUT the whole idea of keeping the commit history with incremental changes is that this does not become an issue in the first place.

In your referenced PR the issue is that you somehow try to keep multiple branches in sync instead of rebasing each one with `master`. If you have multiple branches that are dependend of each other you have to rebase the first one with master and the second one with the first one. You just have to do that in the correct order instead of mixing and matching.

matthias-ronge:
These are always the more complicated cases. So suppose I removed a method and implemented something else, and that's in a commit. Later, there is a conflict at the rebase for this commit because somebody has improved on this deleted method. How should I solve it then: I have to change my newly implemented replacement. But now I have commits later (and that's weeks ago) undone the approach and restored the original method, then at last I have to fix the same problem for a second time. (When merging, it would not even have caused a conflict, because the final code would have been like the original code.) And the first fix was completely unnecessary. That’s for the content of the files.

Also, I've noticed that at least the Git tool in Eclipse has problems for me when files are renamed / moved, which causes unnecessary conflicts, and it reports the conflicts on every commit, and I have to fumble a long time to make Git aware of that the modified file is actually deleted and I have changed something elsewhere for it. Somehow Git cannot track that. That's just my experience. Of course it is possible somehow, but instead of minutes I would need hours, and that cannot be our intention.

It's true that this link was a different situation at the time, but I just wanted to show how such a commit history looks like, so it looks pretty much the same. Not pretty anyway.

Erikmitk:
> So suppose I removed a method and implemented something else, and that's in a commit. Later, there is a conflict at the rebase for this commit because somebody has improved on this deleted method. How should I solve it then: I have to change my newly implemented replacement

That's a communication problem. Either the improved method works for you. Then you'd use that and delete your replacement at the conflict. Or you go follow your intention and delete the improved method and use your replacement. What's the big deal here?

> But now I have commits later (and that's weeks ago) undone the approach and restored the original method, then at last I have to fix the same problem for a second time. (When merging, it would not even have caused a conflict, because the final code would have been like the original code.)

I agree that creativity is an important tool for solving problems. But chaotically creating and removing things is not a viable approach for working in a team. If these code parts only deal with your current work and yours alone that's not even a problem at all. But why are you simultaneously changing something someone else is working on in the first place. There has to be communication involved if multiple people touch the same parts of the code at the same time with different intentions. You should never delete a method someone else is working on/ improving while you figure your stuff out .If you think a method needs to be replaced to fit your needs tell people about it and make a pull request just for this.

These situations are not created by rebasing or git in the first place but are a result of a communication failure we should address instead.

> I've noticed that at least the Git tool in Eclipse has problems

Then don't use the Git tool in Eclipse. Switch to another tool or use the command line. If it's making you do more work, don't use it.

Renaming/Moving and Deleting can be done via git commands as `git rm`, `git mv` so git can deal with that without a hassle.

Erikmitk:
Another approach would be to create Pull Requests early and often so there are PRs for specific parts of the interface implementation. When all is in place the integration can begin. That would speed up development and review as well because parts can be commented on independently and in time instead of one PR that contains the full implementation that has to be taken in all at once.

matthias-ronge:
> That's a communication problem.

No, it happens because I try to find a way to develop it in the hope that it will work, but sometimes it does not work. I do not have to communicate this to anyone because it only happens locally in my branch. It's not until I'm done with my changes that I have to explain why I did something like this, but not what mistakes I might have made before that. A problem arises only if now some strange change in a Rebase gets into my history of development and I would have to go through some of the steps that I later changed over again. That would consume extra hours, and I do not want to waste them on that.

> chaotically creating and removing things is not a viable approach for working in a team

No, that's why it happens only locally with me, and in the end I turn everything so that it not only works (which it must do first) but that it also cleaned up.

> There has to be communication involved if multiple people touch the same parts of the code at the same time with different intentions.

That's exactly what did not happen when there was a blatant merge conflict. I'm not talking about the one-liners in which order two imports have to stand. But if, for example, one has removed a function that the other has expanded. Luckily, it does not happen that often anymore, but now and then it just happens.

> Renaming/Moving and Deleting can be done via git commands as `git rm`, `git mv` so git can deal with that without a hassle.

Thanks for the hint! I did not know that there is functionality in Git for that. There was no support for that a long time ago.

> Another approach would be to create Pull Requests early and often

Yes and no. There are two reasons. For one thing, if the pull request lasts a long time, then in the meantime I can not continue with the program part based on it. And we've had that often enough. (If it's an independent part, that's not a problem, and we're already practicing it.)

The second reason is that I have to do such a development in one go. I'm really into it then, I even dream of it at night, or wake up at night and take notes, because I just had an idea how I can solve a thing. If I can only work on one thing for two days and then it is on pause for two weeks, I will not be able to keep up because then I forgot the path I had in mind. Maybe that's how I work for _me_, but that's the way my head works.

Erikmitk:
>  I do not have to communicate this to anyone because it only happens locally in my branch.

If it were only happening locally on your branch that would not create a problem. But your creative approach has to interact with the changes by the other developers at the end. The problem seems to be that you go wild on a single branch and all these experiments that didn't work out have possible repercussions later on because they **are still there** since the PR and the rebase are not happening locally but with global changes.

If a solution doesn't work out and you have to revert it, you have to clean up your own branch. As simple as that. Change the order of commits so these reverts do not spread out across commits and cancel each other out. Or create a different branch and cherry-pick the commits that actually lead to your solution. You yourself could habe a feature-master-branch and several branches where you test stuff out. You merge your successful experiments and end up with one clean branch that can be used for the PR. Feel free to use branches locally. :)

In the end the branch that you use for the PR should be clean and failed experiments should not be visible. Those experiments can happen on different branches or on a single branch where commits were squashed and cleaned up. But don't squash everything into a single big commit. Just use this method to eliminate commits were you backtracked.

> Luckily, it does not happen that often anymore, but now and then it just happens.

That case is always a sign that we as a team screwed up and didn't talk enough.

> The second reason is that I have to do such a development in one go.

That's totally fair. That was just an idea for another approach. If that's how you work best, that should be the way to go. But again, I think the best way forward are PRs with a verbose but clean commit history. Of course there'll be failed solutions, but these should not be visible. The commit history should be either cleaned up OR all necessary commits are cherry-picked into a fresh branch so there are only the commits that are important.

matthias-ronge:
Back to the topic: Is there anything else I need to do here, or can it be gemerged now as it is?

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/FormRow.java
```diff
@@ -0,0 +1,65 @@
...
+     * A possible view on the key. None, if it is hidden.
+     */
+    private Optional<MetadataViewInterface> optionalKeyView;
+
+    /**
+     * The values. This is always one at most, except for multiple selections.
+     */
+    private Collection<T> values;
+
+    public FormRow(Optional<MetadataViewInterface> optionalKeyView, Collection<T> values) {
```
henning-gerhardt: Public constructor without JavaDoc.

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/KeyView.java
```diff
@@ -0,0 +1,274 @@
...
+            return true;
+        }
+        return universal.getPattern().get().matcher(value).matches();
+    }
+
+    void setScheme(String scheme) {
+        this.scheme = scheme;
+    }
+
+    public void setYearBegin(MonthDay yearBegin) {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Division.java
```diff
@@ -0,0 +1,121 @@
...
+    private List<Label> labels = new LinkedList<>();
+
+    /**
+     * In this element, if there is, it is stored that the division of this
+     * division is done by divisions, which map a calendar date.
+     */
+    @XmlElement
+    private SubdivisionByDateElement subdivisionByDate;
+
+    public Optional<String> getDates() {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Key.java
```diff
@@ -0,0 +1,184 @@
...
+        if (codomain == null) {
+            return Type.STRING;
+        }
+        if (codomain.getType() != null) {
+            return codomain.getType();
+        }
+        return codomain.getNamespace() != null ? Type.ANY_URI : Type.STRING;
+    }
+
+    public void setOptions(Collection<Option> options) {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Setting.java
```diff
@@ -0,0 +1,245 @@
...
+     * no attribute is written.
+     * 
+     * @param multiline
+     *            “multi-line” value to set
+     */
+    public void setMultiline(Boolean multiline) {
+        this.multiline = multiline;
+    }
+
+    public void setSettings(List<Setting> settings) {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo-DataEditor/src/main/resources/ruleset.xsd
```diff
@@ -0,0 +1,168 @@
...
+    </xs:simpleType>
+
+    <xs:simpleType name="Unspecified">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="unrestricted"/>
+            <xs:enumeration value="forbidden"/>
+        </xs:restriction>
+    </xs:simpleType>
+
+</xs:schema>
```
henning-gerhardt: New line at end of file is missing.

Kitodo/src/main/java/org/kitodo/services/dataeditor/RulesetManagementService.java
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+        return instance;
+    }
+    
+    private RulesetManagementService() {
+        rulesetManagementLoader = new KitodoServiceLoader<>(RulesetManagementInterface.class);
+    }
+
+    public RulesetManagementInterface getRulesetManagement() {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo-DataEditor/pom.xml
```diff
@@ -66,6 +66,17 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>${commons-lang3.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+            <version>${hamcrest-library.version}</version>
```
henning-gerhardt: Maven property `hamcrest-library.version` is not defined so build is failing.

Kitodo/src/main/java/org/kitodo/services/dataeditor/RulesetManagementService.java
```diff
@@ -0,0 +1,46 @@
...
+                if (Objects.equals(instance, null)) {
+                    instance = new RulesetManagementService();
+                }
+            }
+        }
+        return instance;
+    }
+    
+    private RulesetManagementService() {
+        rulesetManagementLoader = new KitodoServiceLoader<>(RulesetManagementInterface.class);
```
henning-gerhardt: This looks a little bit strange. @Kathrin-Huber what did you think?

matthias-ronge:
@Kathrin-Huber: Basically, what you need to review out of this is the two commits above:
* [Add classes to read the XML](https://github.com/kitodo/kitodo-production/pull/2016/commits/173aa6dcd975b4314f1af5002a57e7468f56ab39)
* [Implement the ruleset interface](https://github.com/kitodo/kitodo-production/pull/2016/commits/4d31ce2d50d7f4be3fd223097cf035408a0c4b44)

That is still a lot, and I feel sorry for that, but it is way less then the **whole** pull request, because a lot of lines of code is test ruleset XML files, required for the tests. I hope this breaks it down a bit for you.

PR #2023, opened 2018-11-30
Raise the version of the Hamcrest library to 2.1-rc3

matthias-ronge:


Erikmitk:
@Kathrin-Huber Just as a background info for this PR. The version we previously used was moved from Maven Central to another repo and is not available anymore. So tests in Travis are failing now. We discussed this in the dev team and concluded that we simply update the version number and use that for now. If and how we are dealing with Hamcrest as a dependency in general is still up for discussion.

PR #2029, opened 2018-12-03
Follow the renaming requirements of Hamcrest

matthias-ronge:
They have simply uploaded a new JAR file over the existing version number, which apparently does not contain the classes anymore.

### DIFF
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -182,8 +182,8 @@
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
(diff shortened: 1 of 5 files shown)


Erikmitk:
See: https://github.com/hamcrest/JavaHamcrest/commit/0da90880b3aa7364afba05b7f1a853a25c190a44

Kathrin-Huber:
Hamcrest all is 299kb when hamcrast library is only 51kb so maybe check if it is really needed.

Kathrin-Huber:
Ok, as I understood it, we get the big file anyway...

PR #2034, opened 2018-12-04
Add possibility to assign template during project edition

Beacze:


Kathrin-Huber:
I think it's better to use the client from template.project.client then to store it redundantly in the database

PR #2035, opened 2018-12-04
Restructure user edit page

Beacze:
- separate list pages to own compositions
- use parameters in included compositions
- use unified file naming (some files had prefix userEdit, other not - prefix removed to shorten file names)


Kathrin-Huber:
Please resolve conflicts

PR #2040, opened 2018-12-05
Increase hibernate and selenium version

Beacze:
Increase of the hibernate demanded also increase of selenium version ,as they both use net.bytebuddy:byte-buddy. Hibernate needs at least 1.8.x, but previous selenium version was bringing in 1.7.x

### DIFF
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -145,7 +145,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>5.4.1.Final</version>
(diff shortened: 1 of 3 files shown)


Kathrin-Huber:
Maybe the incrementation of the hibernate version makes it possible to remove the lazy loading hack :P

Beacze:
You can try, I was making approach to it but no success. Tests are failing because session is closed and after that it can't initialize related objects. If we would be able remove lazy hack, application would work incredible fast. It is main reason why login takes so long. It fetches all objects associated with the user and it takes forever.

Here is what I was trying: https://github.com/Beacze/kitodo-production/tree/hibernate-initalize

PR #2046, opened 2018-12-10
Implement the METS interface (1)

matthias-ronge:
This branch only contains the changes to the API files.


Beacze:
I would squash Add Javadoc commits.

`Add Javadoc and remove an unnecessary parameter
087c556
 @matthias-ronge
Add Javadoc
3ac5c5a
 @matthias-ronge
Add Javadoc
dba2d42`

Probably the others add javadoc could be also squashed with those three.

matthias-ronge:
Done. Apparently, squashing results in that all meanwhile merged commits are now displayed here as well.

Kathrin-Huber:
Then you have probably done something wrong while squashing, because usually this doesn't happen. Now I cannot review it, because there are abour 400 files changed :(

PR #2049, opened 2018-12-10
Implement the METS interface (2)

matthias-ronge:
This branch contains the access classes. To make the content easier to understand, look at the graphic on Intranet page “Das neue Production-Internformat” under the heading “Interne Objekte” as well.


Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/IdentifierProvider.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.kitodo.dataformat.access;
+
+import java.util.Iterator;
+
+/**
+ * Provides unique identifiers for elements in XML files.
+ */
+class IdentifierProvider implements Iterator<String> {
```
Kathrin-Huber: Why are you not using UUID?
For whatever reason you don't, then why are you implementing Iterator<String>?
Everything you need is:

```
private static long idCounter = 0;

public static synchronized String createID()
{
    return String.valueOf(idCounter++);
} 
```

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /**
+     * Returns the list of available media variants with the corresponding media
+     * files.
+     * 
+     * @return available media variants with corresponding media files
+     */
+    @Override
+    public Set<Entry<MediaVariant, MediaFile>> getAllUsesWithFLocats() {
```
Kathrin-Huber: why are you not returning the map?

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * A human readable label for the order of this media unit. This need not be
+     * directly related to the order number. Examples of order labels could be
+     * “I, II, III, IV, V,  - , 1, 2, 3”, meanwhile the order would be “1, 2, 3,
+     * 4, 5, 6, 7, 8, 9”.
+     */
+    private String orderlabel;
+
+    /**
+     * Public constructor for a new media unit. This constructor can be used
```
Kathrin-Huber: I don't get this javadoc. "It can be used with the service loader..."
Is this the only usage, or why are you pointing this out? A constructor can be used in every place where I want a new instance, right?

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
```diff
@@ -0,0 +1,207 @@
...
+     * 
+     * @param mediaVariant
+     *            variant of the added file
+     * @param mediaFile
+     *            media file to add
+     */
+    @Override
+    public void putFLocatForUse(UseXmlAttributeAccessInterface mediaVariant,
+            FLocatXmlElementAccessInterface mediaFile) {
+        mediaFiles.put((MediaVariant) mediaVariant, (MediaFile) mediaFile);
```
Kathrin-Huber: Please avoid casts.
If this method only accepts MediaVariant and MediaFile please show it in the signature

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaVariant.java
```diff
@@ -0,0 +1,118 @@
...
+     * 
+     * @param fileGrp
+     *            METS {@code <fileGrp>} from which a media variant is to be
+     *            created
+     * @throws IllegalArgumentException
+     *             if the MIME type is mixed within the METS {@code <fileGrp>}
+     */
+    MediaVariant(FileGrp fileGrp) {
+        this.use = fileGrp.getUSE();
+        Set<String> mimeTypes = fileGrp.getFile().parallelStream().map(fileType -> fileType.getMIMETYPE())
```
Kathrin-Huber: why is getUSE and getMIMETYPE in capital letters?
Sorry, I don't really get what is happening here :(

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetadataEntry.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+package org.kitodo.dataformat.access;
+
+import org.kitodo.api.dataformat.mets.MdSec;
+import org.kitodo.api.dataformat.mets.MetadataXmlElementAccessInterface;
+import org.kitodo.dataformat.metskitodo.MetadataType;
+
+/**
+ * A meta-data entry. This is probably the most important Java class of all
```
Kathrin-Huber: Please get rid of the prose! Cut it down to what is important for development.
```
A meta-data entry consists of a key and a value. The key is stored in the superclass {@link Metadata}.
````

should be enough

File: Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MultiMapCollector.java [hunk 7]
```diff
@@ -0,0 +1,94 @@
...
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collector;
+
+/**
+ * Collects an object stream to a multi-map. A multi-map is a
+ * {@code Map<K, Set<V>>}, which can store several objects under one key.
+ */
+class MultiMapCollector<T, K, V> implements Collector<T, Map<K, Set<V>>, Map<K, Set<V>>> {
```
Kathrin-Huber: Why don't you use an own class for this?
this:
```
Collector<T, Map<K, Set<V>>, Map<K, Set<V>>>
```
seems to be a bit confusing.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/OrderAwareList.java
```diff
@@ -0,0 +1,165 @@
...
+     * occurrences of the same number (unlike a TreeMap). The order of list
+     * items is changed only when necessary. The order of list members with the
+     * same ordinal number is not affected.
+     * 
+     * <p>
+     * The mathematical complexity of this algorithm is linear (n) if the order
+     * is already correct, which is already given in the vast majority of
+     * invocations of the method.
+     */
+    private void ensureOrder() {
```
Kathrin-Huber: Why don't you use the Comparable interface? Writing your own sorting algorithm I usually not a good idea!

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/ProcessingNote.java
```diff
@@ -0,0 +1,213 @@
...
+ * A processing note that can be placed in the header of the XML file.
+ */
+public class ProcessingNote implements AgentXmlElementAccessInterface {
+    /**
+     * Known roles in METS. If the role takes one of these values, the
+     * {@code ROLE} attribute is set to this value. Otherwise, the role is set
+     * to {@code OTHER} and the {@code OTHERROLE} attribute is set to the value
+     * of the role.
+     */
+    private static final List<String> KNOWN_ROLES = Arrays.asList("CREATOR", "EDITOR", "ARCHIVIST", "PRESERVATION",
```
Kathrin-Huber: Seems like an Enum would be appropriate

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/Structure.java
```diff
@@ -0,0 +1,430 @@
...
+import org.kitodo.dataformat.metskitodo.Mets;
+
+/**
+ * The tree-like outline structure for digital representation. This structuring
+ * structure can be subdivided into arbitrary finely granular
+ * {@link #substructures}. It can be described by {@link #metadata}.
+ */
+public class Structure implements DivXmlElementAccessInterface {
+    /**
+     * The name of the Q of Kitodo. This parameter is needed to assemble the
```
Kathrin-Huber: What is Q?
Alt Text

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/Structure.java
```diff
@@ -0,0 +1,430 @@
...
+                .collect(Collectors.toCollection(LinkedList::new));
+        type = div.getTYPE();
+        Set<MediaUnit> mediaUnits = mediaUnitsMap.get(div.getID());
+        if (mediaUnits != null) {
+            mediaUnits.stream().map(View::new).forEach(views::add);
+        }
+    }
+
+    /**
+     * Determined from a METS data structure of which type is a meta-data
```
Kathrin-Huber: Too much information in javadoc

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/Workpiece.java
```diff
@@ -0,0 +1,446 @@
...
+        Map<String, Set<MediaUnit>> mediaUnitsMap = mets.getStructLink().getSmLinkOrSmLinkGrp().parallelStream()
+                .filter(SmLink.class::isInstance).map(SmLink.class::cast)
+                .collect(new MultiMapCollector<>(SmLink::getFrom, smLink -> divIDsToMediaUnits.get(smLink.getTo())));
+        structure = getStructMapsStreamByType(mets, "LOGICAL")
+                .map(structMap -> new Structure(structMap.getDiv(), mets, mediaUnitsMap)).collect(Collectors.toList())
+                .iterator().next();
+    }
+
+    /**
+     * Returns the media units of this workpiece. An ordered list of media units
```
Kathrin-Huber: Too much information in javadoc

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/IdentifierProvider.java
matthias-ronge: I chose integer, because they are pretty short. If you want to use UUID we can still replace it—provided it works. I do not know what a UUID must contain and I do not know if METS IDs are allowed to contain all characters. I have implemented iterator because I find it so clear. But we can also leave it out. Should I remove it?

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
matthias-ronge: We had already discussed that when we defined the interface: The answer was that otherwise you could manipulate the map arbitrarily. So you get the result of the function `entrySet()` over which you can iterate if you want to process the whole map. But you can not damage it.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
matthias-ronge: It can only be used via the service loader because the content of the module is not visible from the outside. You can not see that in the code, so if you're looking for references to the constructor, you will not find one. Nevertheless one should not delete it therefore as unused code. That's why.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
matthias-ronge: This is not possible because of the interface. The method gets passed classes that implement an interface of the API. Internally I need the classes. This is owed to the API and therefore is not possible otherwise. If you want to change it, these classes would have to be in the API module.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaVariant.java
matthias-ronge: These are the method names that JXC (the generator for JAXB) extracts from METS. In METS, the attributes are written in uppercase letters. I have no influence on that.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetadataEntry.java
matthias-ronge: Okay, that can be shortened. TO DO

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MultiMapCollector.java
matthias-ronge: The interface is so defined in Java, I have no influence on that.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/OrderAwareList.java
matthias-ronge: Comparable just compares and does not sort. The problem at this point is that there may be several entries that are equal in comparison. In a collection implementation, they would overwrite each other, which is not wanted. One would have to implement a separate class of a collection (HashSet, TreeSet, etc.). The Comparable is the smaller half of this problem. The algorithm is very simple, I know that from the time I programmed Basic. If you would like to change that, we would have to solve this in an own ticket, that would be a major programming task.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/ProcessingNote.java
matthias-ronge: To check if a string matches a particular string? Not really, right?

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/Structure.java
matthias-ronge: To be honest, I have no idea. This is something that builds JAXB in memory like that and I copied it from the debugger. It works as it should.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/Structure.java
matthias-ronge: I have to look at it. To do

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/Workpiece.java
matthias-ronge: I have to look at it. TO DO

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
matthias-ronge: Okay, if you say so. How should that look?

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/Structure.java
matthias-ronge: I finally found it: [QName](https://docs.oracle.com/javase/8/docs/api/javax/xml/namespace/QName.html)

matthias-ronge: It would have been in the javadoc of the type, I should have just looked at it!!! (headbang)

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MediaUnit.java
Beacze: It would demand a bit of the additional work, so I would keep it for future. I can tell you tomorrow how it could look, so you could consider this change :)

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/IdentifierProvider.java
matthias-ronge: `static synchronized` is also nonsense. I do not want to allocate identifiers across Production uniquely (and then reset the counter back to 1 at each server restart). The identifiers need only be unique within a METS file so that the XML parser knows what is referring to what:

id_referencing

PR #2067, opened 2018-12-11
Implement the METS interface (3)

matthias-ronge:
This branch contains a test for the access classes.


matthias-ronge:
Branch 3 is back. As for your comments: I could format the METS file, of course, but I deliberately added it to the test as it currently looks in 2.x, because that's what the code must be able to handle. If you want, I can reformat it, just let me know.

PR #2079, opened 2018-12-12
Remove duplicate MassImportForm and rename DigitalCollections

Beacze:


Kitodo/src/main/java/org/kitodo/config/DigitalCollection.java
```diff
@@ -57,17 +57,17 @@ private DigitalCollections() {
...
     /**
      * Prepare digital collections and possible digital collections for process.
      *
-     * @param process
+     * @param projectBean
      *            object
      */
     @SuppressWarnings("unchecked")
-    public static void possibleDigitalCollectionsForProcess(Process process) throws JDOMException, IOException {
+    public static void possibleDigitalCollectionsForProcess(Project projectBean) throws JDOMException, IOException {
```
Kathrin-Huber: is this still correct?

Beacze: Yes, as this thing only uses project title. So better is to call it with the project parameter instead of process.

PR #2110, opened 2018-12-20
Add metadataclasses to api

Kathrin-Huber:


Kitodo-API/src/main/java/org/kitodo/api/MetadataGroup.java
```diff
@@ -0,0 +1,36 @@
...
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.api;
+
+import java.util.List;
+
+public class MetadataGroup extends Metadata{
+
+    //The value of the metadata is a list of metadatas.
```
Beacze: a list of metadata objects

oliver-stoehr: It should be "metadata objects" instead of "metadata", correct?

Beacze: Yes.

oliver-stoehr: According to [https://en.oxforddictionaries.com/definition/data](url) "data" is not countable. So it should be a "a list of metadata objects"

Beacze:
Checkstyle has failed because java docs have no description finished with the dot.

PR #2113, opened 2018-12-20
General fixes in new inserted code

beatrycze-volk:


Kitodo/src/main/java/org/kitodo/services/ServiceManager.java
```diff
@@ -380,7 +380,7 @@ public static ImageService getImageService() {
      *
      * @return LdapServerService object
```
henning-gerhardt: Copy and paste error?

PR #2123, opened 2019-01-03
Correct the incorrect use of the Service Loader in the METS API

matthias-ronge:
Due to a misunderstanding, the METS API was programmed incorrectly. As discussed at the December meeting, the data classes are provided in the API module in the `org.kitodo.api` package. The wrong use of the service loader is removed.


Kathrin-Huber:
Please don't put the classes in org.kitodo.api, but in te specific api package, where you need them.
Classes like "Metadata" which are universally applicable are already introduced in PR 2110

Kathrin-Huber:
The universal metadata classes are added to the master, please use them. They are exactly as we discussed them in the hackathon. I don't know what you need those other classes for, but they should go to the module-package

Kathrin-Huber:
there are conflicts in your branch

matthias-ronge:
I changed it. Now it should be better with the service loader.

matthias-ronge:
I've created two additional pull requests that anticipate in part the changes made in this pull request:
* [Add the domain to the meta-data class](https://github.com/kitodo/kitodo-production/pull/2136)
* [Add data format API classes](https://github.com/kitodo/kitodo-production/pull/2137)

PR #2137, opened 2019-01-14
Add data format API classes

matthias-ronge:
As part of the refactoring process to extract the incorrectly implemented service loaders for the METS module into data classes for the API, the data classes that will be created during the extraction are subsequently taken from the development and transferred to the API beforehand, so that these can be used as data classes for the data formats (METS, and possibly more in the future).

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
@@ -1,0 +1,87 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.api.dataformat;
+
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MediaUnit {
+    /**
+     * Each media unit can be available in different variants, for each of which
+     * a media file is available. This is in this map.
+     */
+    private Map<MediaVariant, URI> mediaFiles = new HashMap<>();
+
+    /**
+     * Sequence number of the media unit. The playback order of the media units
+     * when referenced from a structure is determined by this attribute (not by
+     * the order of the references).
+     */
+    private int order;
+
+    /**
+     * A human readable label for the order of this media unit. This need not be
+     * directly related to the order number. Examples of order labels could be
+     * “I, II, III, IV, V,  - , 1, 2, 3”, meanwhile the order would be “1, 2, 3,
+     * 4, 5, 6, 7, 8, 9”.
+     */
+    private String orderlabel;
+
+    /**
+     * Returns the map of available media variants with the corresponding media
+     * file URIs.
+     * 
+     * @return available media variants with corresponding media file URIs
+     */
+    public Map<MediaVariant, URI> getMediaFiles() {
+        return mediaFiles;
+    }
+
+    /**
+     * Returns the order number for this media unit.
+     * 
+     * @return the order number
+     */
+    public int getOrder() {
+        return order;
+    }
+
+    /**
+     * Returns the order label for this media unit.
+     * 
+     * @return the order label
+     */
+    public String getOrderlabel() {
+        return orderlabel;
+    }
+
+    /**
+     * Sets the order number for this media unit.
+     * 
+     * @param order
+     *            order number to set
+     */
+    public void setOrder(int order) {
+        this.order = order;
+    }
+
+    /**
+     * Sets the order label for this media unit.
+     * 
+     * @param orderlabel
+     *            order label to set
+     */
+    public void setOrderlabel(String orderlabel) {
+        this.orderlabel = orderlabel;
+    }
+}
--- a/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
@@ -57,13 +57,13 @@ public class MediaUnit {
     }
 
     /**
-     * Sets the order label for this media unit.
+     * Sets the order number for this media unit.
      * 
-     * @param orderlabel
-     *            order label to set
+     * @param order
+     *            order number to set
      */
-    public void setOrderlabel(String orderlabel) {
-        this.orderlabel = orderlabel;
+    public void setOrder(int order) {
+        this.order = order;
     }
 
     /**
@@ -76,12 +76,12 @@ public class MediaUnit {
     }
 
     /**
-     * Sets the order number for this media unit.
+     * Sets the order label for this media unit.
      * 
-     * @param order
-     *            order number to set
+     * @param orderlabel
+     *            order label to set
      */
-    public void setOrder(int order) {
-        this.order = order;
+    public void setOrderlabel(String orderlabel) {
+        this.orderlabel = orderlabel;
     }
 }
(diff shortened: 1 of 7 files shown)


Kitodo-API/src/main/java/org/kitodo/api/dataformat/SortedList.java
```diff
@@ -0,0 +1,140 @@
...
+        this.orderGetter = orderGetter;
+    }
+
+    /**
+     * Sorts the list by the media unit’s order value and can handle duplicate
+     * occurrences of the same number (unlike a TreeMap). The order of list
+     * items is changed only when necessary. The order of list members with the
+     * same ordinal number is not affected.
+     */
+    private void ensureOrder() {
```
Kathrin-Huber: I think you need to override the compare method, then you don't need to use your own "ensureOrder" method.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/SortedList.java
```diff
@@ -0,0 +1,140 @@
...
+        Collections.sort(this, (one, another) -> orderGetter.apply(one).compareTo(orderGetter.apply(another)));
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        ensureOrder();
```
Kathrin-Huber: ....and then you don't need to call it on every other method

Kitodo-API/src/main/java/org/kitodo/api/dataformat/Workpiece.java
```diff
@@ -0,0 +1,176 @@
...
+     * Returns the creation date of the workpiece.
+     * 
+     * @return the creation date
+     */
+    public GregorianCalendar getCreationDate() {
+        return creationDate;
+    }
+
+    /**
+     * Returns the edit history. The head of each METS file has space for
```
Kathrin-Huber: please cut down javadoc to 
Returns edit history.

If you really want to add information on the field, add it to the javadoc of the field, but I would rather have it clean.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/SortedList.java
Kathrin-Huber: Please read about lists and sets.
Here is a pretty nice comment about it:
https://stackoverflow.com/questions/8725387/why-is-there-no-sortedlist-in-java/8725470#8725470

Kitodo-API/src/main/java/org/kitodo/api/dataformat/SortedList.java
Beacze: At first we should ask ourself if this list needs to be sorted all the time or only in specific moment of usage like iteration over it. I have a feeling that it is a bit overkill to sort elements before adding new element, sort elements before iterator and so on. I assume that only moment when order really matters is iterating over it. Good solution for it is override compare method and next sort list right before you want to usage it.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/SortedCollection.java
```diff
@@ -37,104 +37,71 @@
...
      *            possibility to get an order number for T
      */
-    public SortedList(Function<T, Integer> orderGetter) {
+    public SortedCollection(Function<T, Integer> orderGetter) {
         this.orderGetter = orderGetter;
     }
 
+    @Override
+    public T get(int index) {
+        throw new UnsupportedOperationException("blocked");
```
Beacze: Why you disallow usage of for loop?

Kitodo-API/src/main/java/org/kitodo/api/dataformat/SortedCollection.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Iterator;
+import java.util.ListIterator;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * A collection that is aware of and enforces the order of its members
+ * prescribed by the ordinal number of the media units.
+ */
+class SortedCollection<T> extends ArrayList<T> implements Collection<T> {
```
Kathrin-Huber: Why don't you use a set? If you overwrite the compare method of a set, the elements are inserted at the correct position, you don't need to block the methods then.
When someone uses a for loop for this sortedCollection, then it's blocked, and it is not obvious why.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -0,0 +1,87 @@
...
+        return orderlabel;
+    }
+
+    /**
+     * Sets the order number for this media unit.
+     * 
+     * @param order
+     *            order number to set
+     */
+    public void setOrder(int order) {
```
Kathrin-Huber: I think you missarranged the setters and getters here :P
getOrder
setOrderlabel
getOrderlabel
setOrder

seems to be wrong

Kitodo-API/src/main/java/org/kitodo/api/dataformat/Structure.java
```diff
@@ -0,0 +1,167 @@
...
+    private String label;
+
+    /**
+     * The meta-data for this structure. This structure level can be described
+     * with any meta-data.
+     */
+    private Collection<Metadata> metadata = new HashSet<>();
+
+    /**
+     * The order label of this structure. This is needed very rarely. It is not
```
Kathrin-Huber: too much javadoc

Kitodo-API/src/main/java/org/kitodo/api/dataformat/Structure.java
```diff
@@ -0,0 +1,167 @@
...
+     * data type of this variable is a string, it is recommended to use a
+     * controlled vocabulary. If the generated METS files are to be used with
+     * the DFG Viewer, the list of possible structure types is defined.
+     * 
+     * @see "https://dfg-viewer.de/en/structural-data-set/"
+     */
+    private String type;
+
+    /**
+     * The views on media units that this structure level comprises. Currently,
```
Kathrin-Huber: too  much javadoc

Kitodo-API/src/main/java/org/kitodo/api/dataformat/View.java
```diff
@@ -0,0 +1,45 @@
...
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.api.dataformat;
+
+/**
+ * A view on a media unit. The individual levels of the {@link Structure} refer
+ * to {@code View}s on {@link MediaUnit}s. At the moment, each {@code View}
+ * refers to exactly one {@code MediaUnit} as a whole. This concept level has
```
Kathrin-Huber: remove last sentence of javadoc

Kitodo-API/src/main/java/org/kitodo/api/dataformat/Workpiece.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
+import org.kitodo.api.Metadata;
+
+/**
+ * The administrative structure of the product of an element that passes through
+ * a Production workflow.
+ * 
+ * <p>
```
Kathrin-Huber: The first sentence of the javadoc should be enough

Kitodo-API/src/main/java/org/kitodo/api/dataformat/SortedCollection.java
matthias-ronge: I have read your “I assume that only moment when order really matters is iterating over it” comment to the `ensureOrder()` function above that I should sort in `iterator()` only. Which in turn means all other functions that return objects must be blocked, because they would return the wrong order. Maybe I got you wrong. Sorry.

Beacze: I meant: Call sort before the usage of the list, not inside the implementation.

matthias-ronge:
All changed. I do not write this down to every single one.

PR #2139, opened 2019-01-15
METS-Export 2/2, implement XSLT.

IkramMaalej:


Kitodo/src/main/resources/xslt/ruleset_slubdd.xsl
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Kathrin-Huber: The name of the file is a bit confusing.

Erikmitk: Well… there's a reason for this name. The export has to know about the corresponding ruleset. So for now we added the convention that the export XSLT has to have the same name as the ruleset it refers to. So you have rulseset_slubdd.xml and ruleset_slubdd.xsl because that's what they are called here. We could change them to slubdd.xml and slubdd.xsl though, @IkramMaalej?

Hope that makes it clear. We have to document the naming convention.

PR #2145, opened 2019-01-18
Remove UGH

matthias-ronge:


Kitodo/src/main/java/org/kitodo/helper/metadata/LegacyDocStructHelperInterface.java
```diff
@@ -0,0 +1,364 @@
...
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.helper.metadata;
```
Kathrin-Huber: Shouldn't this go to the legacy-package?

Kitodo/pom.xml
```diff
@@ -767,15 +767,6 @@
                 <artifactId>maven-war-plugin</artifactId>
                 <version>3.2.2</version>
                 <configuration>
-                    <webResources>
-                        <resource>
-                            <directory>${parent.basedir}/OpacPica-Plugin/target</directory>
```
Kathrin-Huber: Can this already be removed? Even when, I think this should be done in another pull Request.

Kathrin-Huber:
Please resolve conflicts

Kitodo/src/main/java/org/kitodo/helper/metadata/LegacyDocStructHelperInterface.java
matthias-ronge: Good catch! 👍

Kitodo/pom.xml
matthias-ronge: It no longer works without dependency on the UGH.

Doing thi in another pull request is going to be difficult, I'm afraid. I mean, the newly implemented validation (Pull Request #2151) is based on this branch. How can that still be merged then?

What exactly should I do?

matthias-ronge:
The pull request is again put in pieces.

PR #2151, opened 2019-01-23
Re-implement validation

matthias-ronge:
The new validation uses the new ruleset for validation.
Please note the explanations in the individual commits. I have tried to better explain why I made the respective changes.

This pull request is based on #2145


Kathrin-Huber:
please rebase for better review

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/AbstractKeyView.java
```diff
@@ -31,6 +31,11 @@
      */
     protected final U universal;
 
+    /**
+     * The universal rule.
```
Kathrin-Huber: For what?

Kitodo/src/main/java/org/kitodo/production/helper/batch/BatchTaskHelper.java
```diff
@@ -455,7 +455,6 @@ private boolean isTaskValid(Task task) throws IOException {
     private boolean invalidMetadataExists(Task task) {
         if (task.isTypeMetadata()
                 && ConfigCore.getBooleanParameterOrDefaultValue(ParameterCore.USE_META_DATA_VALIDATION)) {
-            ServiceManager.getMetadataValidationService().setAutoSave(true);
```
Kathrin-Huber: why do you remove it? Has it something to do with validation?

Kitodo/src/main/resources/messages/errors_en.properties
```diff
@@ -86,18 +86,18 @@ metadataDigitalDocumentError=DigitalDocument error at\: {0}
...
+metadataInvalidData=The input "{0}" in field {1} is invalid.
 metadataIsEmpty=contains no data
-metadataMandatoryElement=missing input for\:
-metadataMissingIdentifier=missing value for identifier
+metadataMandatoryElement=Missing input for {0}.
+metadataMediaError=A structure has media assigned which aren't declared.
+metadataMediaUnassigned=The media unit is not assigned to a structure\:
+metadataMissingIdentifier=Missing value for the identifier.
 metadataNotAllowed={0} {1} is not allowed in new element {1}.
-metadataNotEnoughElements=have to exist at least once
```
Kathrin-Huber: The following error messages seems to be incomplete!?

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/AbstractKeyView.java
matthias-ronge: To get the `minOccurs` and `maxOccurs` value.

Kitodo/src/main/java/org/kitodo/production/helper/batch/BatchTaskHelper.java
matthias-ronge: Here, a variable is set, which has no further meaning. It is not used anywhere.

Kitodo/src/main/resources/messages/errors_en.properties
matthias-ronge: Exactly. That's why I changed it.

matthias-ronge:
Is that possible? Then the validation module would have to load the ruleset module. As far as I know, that does not work, right?

Beacze:
Since when it doesn't work? I have used command module inside file management module in May 2018. I think since then until now it works.

Kitodo/src/main/java/org/kitodo/production/helper/batch/BatchTaskHelper.java
Kathrin-Huber: I'm just wondering, why you changed it in this PR

Kathrin-Huber:
You can of course load modules in other modules :)

matthias-ronge:
Well, I did not know that. I always thought the modules were independent. Of course that makes things very different.

matthias-ronge:
I need a lot of methods that are only in the core. Do I have to duplicate all of them into the module?

- `metsService.createMetsXmlElementAccess()`, `fileService.tryLock()`, `fileService.read()` − to read METS file
- `rulesetManagementService.getRulesetManagement()`, `userService.getAuthenticatedUser().getMetadataLanguage()`, `Helper.getRequestParameter()` − to access the rule set
- `ConfigCore.getParameterOrDefaultValue()`, `ParameterCore.VALIDATE_IDENTIFIER_REGEX` − to validate the identifier against configured regex
- `processService.findAll()` − to check the identifier is not used in other process
- `Helper.getTranslation()` − to translate validation messages

matthias-ronge:
Here is a commit which shows the problem. Also I need `ProcessDTO` from the core to tell the title of the conflicting process in an error message.

Kathrin-Huber:
please resolve conflicts

Kitodo/src/main/java/org/kitodo/production/services/validation/MetadataValidationService.java
```diff
@@ -12,32 +12,120 @@
...
+
+    /**
+     * Message key if there are too little entries of a type.
+     */
+    private static final String MESSAGE_VALUE_TOO_RARE = "metadataNotEnoughElements";
+
+    private final MetadataValidationInterface metadataValidation;
+
+    public MetadataValidationService() {
+        metadataValidation = getValidationModule();
```
Kathrin-Huber: Please name it metadatavalidationmodule

PR #2152, opened 2019-01-24
Fix a bug that the meta-data is not saved

matthias-ronge:
The use of meta-data in legacy code is upside down: First, an empty meta-data is created and stored, and then the value is set, and the value is then supposed to propagate itself to the right place where it is stored. This had worked in the past by simply adding a mutable meta-data to a list. Now the matter is more complicated, because the meta-data value can also be in a field of the container and not in an own object at all. Basically, the implementation is an observer pattern or something that’s pretty much the same.


Kathrin-Huber:
please resolve conflicts

PR #2172, opened 2019-02-06
Remove UGH (4)

matthias-ronge:
Removes all remaining references to UGH interface methods

[compare against part 3](https://github.com/matthias-ronge/kitodo-production/compare/KIT-1155.3...KIT-1155.4)


Kitodo/src/main/java/org/kitodo/production/forms/ProzesskopieForm.java
```diff
@@ -531,32 +524,7 @@ private DocStructInterface getDocStruct(AdditionalField field) throws Preference
...
     }
 
-    /**
-     * Get together authors' names.
-     *
-     * @param persons
-     *            list of persons
-     * @return authors' names as String
-     */
-    protected String getAuthors(List<PersonInterface> persons) {
```
Kathrin-Huber: why did you remove this?

matthias-ronge: Because there is no dedicated `Person` object any more in the future. Complex object descriptions, such as persons, can and have to be done using meta-data groups.

PR #2175, opened 2019-02-06
Remove UGH (7)

matthias-ronge:
Removes the dependency to `kitodo-ugh`. In its place references JDOM, which was previously referenced transitively by the UGH.

[compare against part 6](https://github.com/matthias-ronge/kitodo-production/compare/KIT-1155.6...KIT-1155.7)


Kitodo/pom.xml
```diff
@@ -49,9 +49,9 @@
             <version>3.0.0-beta.2-SNAPSHOT</version>
         </dependency>
         <dependency>
-            <groupId>org.kitodo</groupId>
-            <artifactId>kitodo-ugh</artifactId>
-            <version>3.0-SNAPSHOT</version>
+            <groupId>jdom</groupId>
```
Kathrin-Huber: why did you add jdom here?

matthias-ronge: See description above: It was previously transitively added through the dependencies of the UGH. Removing the UGH requires to add it here manually for the project to compile.

PR #2176, opened 2019-02-06
Remove UGH (8)

matthias-ronge:
Clean up. We would be up to date with the first pull request except for deleting the PICA plugin which dropped out here as discussed.

[compare against part 7](https://github.com/matthias-ronge/kitodo-production/compare/KIT-1155.7...KIT-1155.8)


Kitodo/src/main/java/org/kitodo/production/metadata/copier/MetadataPathSelector.java
```diff
@@ -217,7 +217,11 @@ protected void createOrOverwrite(CopierData data, LegacyDocStructHelperInterface
      * of the path aren’t available. This works recursively, by calling itself
      * on the subnode, if found, or returning null otherwise.
      *
+<<<<<<< HEAD
```
Kathrin-Huber: you inserted merge-conflict code

Kitodo/src/test/resources/log4j2.xml
```diff
@@ -18,7 +18,7 @@
         </Console>
     </Appenders>
     <Loggers>
-        <Root level="debug">
+        <Root level="error">
```
Kathrin-Huber: don't change the log level here

Kitodo/src/main/java/org/kitodo/production/metadata/copier/MetadataPathSelector.java
matthias-ronge: Seems to be gone with the latest rebase.

Kitodo/src/test/resources/log4j2.xml
matthias-ronge: Can’t fix. This is the setting which caused Travis to crash due to too long logs. Obviously the config file was not considered in the past and is considered now (how it shoud be!), don’t ask me why.

Beacze: Interesting that until now it was not crashing... It crashes only on this PR. Which stuff is logged here which overload the log? 

_Obviously the config file was not considered in the past and is considered now (how it shoud be!), don’t ask me why._

I don't get what you mean here. It is config file for test logging. Standard logging has own configuration. Even more you, as developer, have in config-local own config file which you can adjust to your needs.

Kitodo/src/main/java/org/kitodo/production/metadata/copier/MetadataPathSelector.java
Beacze: Seems not be gone wit the latest rebase - still visible in the file view: https://github.com/kitodo/kitodo-production/blob/ab049a98d126acd2eef0d8587a5bae27173edfb7/Kitodo/src/main/java/org/kitodo/production/metadata/copier/MetadataPathSelector.java

Kitodo/src/test/resources/log4j2.xml
Kathrin-Huber: I'm going to merge it. I made a Jira Ticket to check log-level

PR #2177, opened 2019-02-07
delete modules folder on maven clean

Kathrin-Huber:
use maven clean plugin to delete modules folder.
We have a problem with the service loader, if we have several versions of the same jar in the modules folder.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,6 +135,23 @@
         <module>Kitodo</module>
     </modules>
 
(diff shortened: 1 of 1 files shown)


Beacze:
I assume that it should be only executed with development profile.

Kathrin-Huber:
There was a problem on moment of execution. I moved it to the parent pom, now it works

PR #2178, opened 2019-02-07
Configure lists

solth:
This PR adds the new authority `configureColumns`. Users with a role with this authority can adjust list views for processes, tasks, users and projects by adding or removing displayed columns in the corresponding lists via a small button next to the `new` button. The column configuration is saved client wide as soon as it is changed for the client with which the current user is logged in. 

In addition to default columns like `title`, `project` etc. the process list can also display process properties as additional "custom columns". In order to appear in the list of available columns, the titles of the corresponding process properties need to be configured in `kitodo_config.properties` with the new property key `processPropertyColumns`. (See `/Kitodo/src/main/resources/kitodo_config.properties` as a template for `config-local/kitodo_config.properties`)

Example: `processPropertyColumns=testProperty1, testProperty2, testProperty3`

Additionally, the new mapping entry for the class `org.kitodo.data.database.beans.ListColumn` has to be added to the local configuration file `hibernate.cfg.xml`.


Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -541,6 +541,12 @@ showOcrButton=false
 # Base path to OCR, without parameters
 ocrUrl=
 
+# -----------------------------------
+# Display process properties
+# -----------------------------------
+
+# Process properties to be displayed in process list
+processProperties=
```
Kathrin-Huber: The name is not self-explanatory enough
processPropertiesForProcessList 
or something

Kitodo/src/main/java/org/kitodo/production/forms/ClientForm.java
```diff
@@ -124,6 +125,7 @@ public void setClientById(int clientID) {
      */
     public void delete() {
         try {
+            this.client.getListColumns().clear();
```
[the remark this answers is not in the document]
Beacze: Nope, it needs to be first removed here. In other case it will be resaved by its relationship.

Kitodo/src/main/resources/kitodo_config.properties
solth: Would `processPropertyColumns` be ok?

Kathrin-Huber:
Obviously there are some chestyle errors, which are causing travis to fail :(

Kitodo-DataManagement/src/main/resources/db/migration/V2_67__Add_table_for_list_columns.sql
```diff
@@ -0,0 +1,101 @@
...
+INSERT INTO listColumn (title) VALUES ('ldapgroup.ldapgroup');
+INSERT INTO listColumn (title) VALUES ('ldapgroup.home_directory');
+INSERT INTO listColumn (title) VALUES ('ldapgroup.gidNumber');
+
+-- 3. Create client_x_listcolumn table
+CREATE TABLE client_x_listColumn (
+  client_id INT(11) NOT NULL,
+  column_id INT(11) NOT NULL,
+  PRIMARY KEY ( client_id, column_id ),
+  KEY FK_client_id (client_id),
```
Kathrin-Huber: The key is not according to our schema, it should be
FK_client_x_listColumn_client_id

PR #2204, opened 2019-02-12
Simplify a complex function

matthias-ronge:
Removes a code smell type ‘brain-overload’ reported by SonarCloud.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProzesskopieForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProzesskopieForm.java
@@ -911,25 +911,12 @@ public class ProzesskopieForm implements Serializable {
 
     private void iterateOverHigherLevelMetadata(LegacyDocStructHelperInterface enricher,
             Map<String, Map<String, LegacyMetadataHelper>> higherLevelMetadata) {
-        for (Entry<String, Map<String, LegacyMetadataHelper>> availableHigherMetadata : higherLevelMetadata
+        HIGHER: for (Entry<String, Map<String, LegacyMetadataHelper>> availableHigherMetadata : higherLevelMetadata
                 .entrySet()) {
             String enrichable = availableHigherMetadata.getKey();
-            boolean addable = false;
-            List<LegacyMetadataTypeHelper> addableTypesNotNull = enricher.getAddableMetadataTypes();
-            if (Objects.isNull(addableTypesNotNull)) {
-                addableTypesNotNull = Collections.emptyList();
+            if (!isAddable(enricher, enrichable)) {
+                continue HIGHER;
             }
-            for (LegacyMetadataTypeHelper addableMetadata : addableTypesNotNull) {
-                if (addableMetadata.getName().equals(enrichable)) {
-                    addable = true;
-                    break;
-                }
-            }
-            if (!addable) {
-                continue;
-            }
-
-            boolean metadataEqualsParent = false;
 
             for (Entry<String, LegacyMetadataHelper> higherElement : availableHigherMetadata.getValue().entrySet()) {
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/ProzesskopieForm.java
```diff
@@ -911,26 +911,13 @@ private void iterateOverAllMetadata(Map<String, Map<String, LegacyMetadataHelper
 
     private void iterateOverHigherLevelMetadata(LegacyDocStructHelperInterface enricher,
             Map<String, Map<String, LegacyMetadataHelper>> higherLevelMetadata) {
-        for (Entry<String, Map<String, LegacyMetadataHelper>> availableHigherMetadata : higherLevelMetadata
+        HIGHER: for (Entry<String, Map<String, LegacyMetadataHelper>> availableHigherMetadata : higherLevelMetadata
```
Kathrin-Huber: Please don't introduce labels, as I removed them in anoher pull request because sonarcloud complains about them (and not just sonarcloud)

Kitodo/src/main/java/org/kitodo/production/forms/ProzesskopieForm.java
```diff
@@ -911,25 +911,31 @@ private void iterateOverAllMetadata(Map<String, Map<String, LegacyMetadataHelper
...
                 for (LegacyMetadataHelper existentMetadata : amNotNull) {
                     if (existentMetadata.getMetadataType().getName().equals(enrichable)
                             && existentMetadata.getValue().equals(higherElement.getKey())) {
-                        continue HIGHER;
+                        breakMiddle = true;
+                        break;
                     }
                 }
-                enricher.addMetadata(higherElement.getValue());
+                if (breakMiddle) {
```
Kathrin-Huber: I think your change does not cause the same functionality.
In the first version, every iteration of the first for:loop is executed, in the second it breaks at some point.
The correct code should be:
```
if(!breakMiddle){
enricher.addMetadata(higherElement.getValue());
}
```

without a break.

matthias-ronge: Isn’t this what the `else` block does? I would think so.

Kathrin-Huber: No, in your case, if breakMiddle is true, the outer for_loop is breaked.
In my case, it goes to the next iteration.
I'm not sure what is the expected behaviour, but that's what the code was doing before and thats also what the label would have done.
Please write a test case for it, to be sure.

matthias-ronge: No, the outer for loop is not `break`ed. You can run this program to see what happens:

    public static void main(String[] args) {
        for (int i = 1; i <= 2; i++) { // outside loop
            System.out.print("A, ");
            for (int j = 1; j <= 2; j++) { // middle loop
                System.out.print("B, ");
                boolean breakMiddle = false;
                for (int k = 1; k <= 2; k++) { // inner loop
                    System.out.print("C, ");
                    breakMiddle = true;
                    break;
                }
                System.out.print("D, ");
                if (breakMiddle) {
                    System.out.print("E, ");
                    break;
                } else {
                    System.out.print("F, ");
                }
            }
            System.out.print("G, ");
        }
        System.out.println("H.");
    }

It prints:

    A, B, C, D, E, G, A, B, C, D, E, G, H.
                      ^

At the `^`, the outer loop goes to the next iteration. If it would not, it would print `A, B, C, D, E, H.` So it’s all right here.

Kathrin-Huber: Ah.. there is a THIRD For:loop...
I would argue, that it is STILL too complex...

PR #2206, opened 2019-02-12
Fix some sonar cloud issues

matthias-ronge:


Kitodo/src/main/java/org/kitodo/production/helper/metadata/MetadataHelper.java
```diff
@@ -46,6 +47,22 @@
     private static final int PAGENUMBER_FIRST = 0;
     private static final int PAGENUMBER_LAST = 1;
     private LegacyPrefsHelper prefs;
+    private final Comparator<? super LegacyReferenceHelper> referencesSorter = (firstObject, secondObject) -> {
```
Kathrin-Huber: please do this in a method not as an assignment

PR #2212, opened 2019-02-12
Configure XML secure processing

matthias-ronge:
Fixes a vulnerability reported by SonarCloud. For details, see [here](https://sonarcloud.io/organizations/henning-gerhardt-github/rules?open=squid%3AS2755&rule_key=squid%3AS2755).

### DIFF
--- a/OpacPica-Plugin/src/main/java/org/kitodo/production/plugin/opac/pica/GetOpac.java
+++ b/OpacPica-Plugin/src/main/java/org/kitodo/production/plugin/opac/pica/GetOpac.java
@@ -17,6 +17,7 @@ import java.io.StringReader;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
+import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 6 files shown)


Beacze:
Our processing is not secure. You can't use it for working with kitodo.xml or meta.xml.

matthias-ronge:
The link explains that this change simply disallows loading XML entities from other files. We don’t use that in any of our XMLs, so there shouldn’t be problems for us.

Beacze:
Our files itself are treated as unsafe.

OpacPica-Plugin/src/main/java/org/kitodo/production/plugin/opac/pica/GetOpac.java
```diff
@@ -16,16 +16,17 @@
...
 
+import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParserFactory;
 
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.methods.GetMethod;
-import org.apache.logging.log4j.LogManager;
```
henning-gerhardt: Why did you switch from log4j version 2 logging instance to an other? This change breaks the compilation of your pull request.

matthias-ronge: Oh, that's why! The import organizes Eclipse on its own, and somehow it has a habit of changing that occasionally. Why, I have not found out yet.

matthias-ronge:
Yes, that's exactly what I want to change. Or is there something against it? We do not use any XML entities.

OpacPica-Plugin/src/main/java/org/kitodo/production/plugin/opac/pica/GetOpac.java
Kathrin-Huber: but reorganizing the imports shouldn't change the source, right?
If really eclipse is doing this, you should absolutely find the setting and disable it, beacuse this is not acceptable! Furthermore you should check your Pull Requests to identify such behaviour.

PR #2217, opened 2019-02-13
Remove workflow file name, use only workflow title

Beacze:
Backend usage of of workflow id and name is removed. Field file name is replaced with title.

Workflow id and name (look image) needs to be somehow hidden by js. They can't be removed totally because they need to be available for bpmn converter. Without them xml files can't be created. Values stored inside them are ignored.

workflow


Kathrin-Huber:
Make workflowtitle unique 
resolve conflicts

PR #2230, opened 2019-02-18
Move getNormalizedTitle method to the Helper class

Beacze:
- unify getNormalizedTitle methods
- bonus: use StandardCharsets instead of String


Kathrin-Huber:
Please resolve conflicts

PR #2235, opened 2019-02-20
Extend media unit to support children and meta-data

matthias-ronge:
API change necessary to represent the physical structure tree with partial commissions.

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
@@ -23,6 +23,14 @@ import org.kitodo.api.Metadata;
 
 public class MediaUnit {
     /**
+     * The subordinate media units of this media unit, which form the media unit
+     * tree. The order of the substructures is subordinate media units by the
+     * order of the {@code <div>} elements in the
+     * {@code <structMap TYPE="PHYSICAL">} in the METS file.
+     */
+    private List<MediaUnit> children = new LinkedList<>();
+
+    /**
      * Each media unit can be available in different variants, for each of which
      * a media file is available. This is in this map.
      */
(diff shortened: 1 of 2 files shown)


Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -37,6 +49,23 @@
...
+     * {@code <structMap TYPE="PHYSICAL">} in the METS file.
+     */
+    private List<MediaUnit> subordinateUnits = new LinkedList<>();
+
+    /**
+     * Returns the subordinate media units associated with this media unit.
+     * 
+     * @return the subordinate media units
+     */
+    public List<MediaUnit> getChildren() {
```
solth: Wouldn't it be more intuitive to either call this method `getSubordinateUnits` or the corresponding field `children`?

Kitodo-API/src/main/java/org/kitodo/api/dataformat/Workpiece.java
```diff
@@ -37,11 +36,10 @@
...
 
     /**
-     * The media units that belong to this workpiece. The order of this
-     * collection is meaningful, but only describes the order in which the media
-     * units are displayed on the workstation of the compiler.
+     * The media unit that belongs to this workpiece. The media unit can have
+     * children, such as a bound book that can have pages.
      */
-    private List<MediaUnit> mediaUnits = new LinkedList<>();
+    private MediaUnit mediaUnit = new MediaUnit();
```
solth: I am not sure I understand the purpose of this change. Doesn't this mean the workpiece / root element of the physical structure can always only have one child element / subordinate unit, which than in turn can have multiple children? If that's the case, I think you should keep the existing list of `MediaUnit` objects as a field for `Workpiece`, since we do want multiple direct children beneath the root element.

matthias-ronge: This allows access to the BoundBook. And e.g. its metadata.

PR #2242, opened 2019-02-21
Add new default rule sets

matthias-ronge:
The repository still contains the old rule sets of the UGH, which no longer work. These are deleted and replaced by two new, working example rule sets.

This pull request is based on PR #2240


Kitodo/rulesets/subhh.xml
```diff
@@ -0,0 +1,1207 @@
...
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ *
+-->
+<ruleset xmlns="http://names.kitodo.org/ruleset/v2"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://names.kitodo.org/ruleset/v2 ../../Kitodo-DataEditor/src/main/resources/ruleset.xsd">
```
solth: Rulesets are normally copied to a local configuration directory, so this relative path to a subfolder of `Kitodo-DataEditor` won't normally work, I think.

Kathrin-Huber:
Please resolve conflicts

Kitodo/rulesets/subhh.xml
Kathrin-Huber: So what about this?

matthias-ronge: I would concider it the best to address the XSD file through an URL. This could be the GitHub URL (`https://raw.githubusercontent.com/kitodo/kitodo-production/master/Kitodo-DataEditor/src/main/resources/ruleset.xsd`), but I would prefer it on the web server `goobi.org`. What do you think?

henning-gerhardt: Which webserver goobi.org? :-D I think, you are meaning kitodo.org.

solth: How about putting `ruleset.xsd` into the same directory (`Kitodo/rulesets/`) until it's available via some URL? Then it would (hopefully) be clearer that this ruleset has to be copied as well when the user puts some custom ruleset file into his local ruleset configuration directory? And it could be referenced from the custom ruleset files without an awkward relative path.

matthias-ronge: ^^ You know that yesterday was Rose Monday, right? :smirk:
So it will go into the `rulesets/` directory then.

Kitodo-DataEditor/src/test/resources/testAnExtensiveRulesetCanBeLoaded.xml
```diff
@@ -13,7 +13,7 @@
 -->
 <ruleset xmlns="http://names.kitodo.org/ruleset/v2"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://names.kitodo.org/ruleset/v2 ../../main/resources/ruleset.xsd">
+         xsi:schemaLocation="http://names.kitodo.org/ruleset/v2 ../../../../Kitodo/rulesets/ruleset.xsd">
```
Kathrin-Huber: I thought, if it is in the same folder, there is no path required?

matthias-ronge: We have rule sets in **two** different folders: One to use and one for the tests. In one of both folders you logically need the path to the other folder. Alternative would be, of course, to copy the file into each folder. Personally, I am against it, because that brings with it all the problems of duplicate code. Say, if someone changes one of the two files you do not know in the end which is right now.

Beacze: Tests should be independent from real directories. So in this case copy would be suitable solution.

Beacze:
Travis will not pass it until it is not rebased.

PR #2246, opened 2019-02-25
Use select list for metadata languages

Beacze:


Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
```diff
@@ -386,6 +388,18 @@ public void load(int id) {
...
     }
 
+    /**
+     * Get map of supported metadata languages.
+     * 
+     * @return map of supported metadata languages
+     */
+    public Map<String, String> getMetadataLanguages() {
+        Map<String, String> metadataLanguages = new HashMap<>();
+        metadataLanguages.put("English", "en");
```
Kathrin-Huber: This seems to be pretty hardcoded, is there any way to make it configurable?

Beacze: I don't think so. I don't know if there is some list of available languages which could be used here.

henning-gerhardt: @Kathrin-Huber , @matthias-ronge, @Erikmitk :
Maybe it is possible to move the translation of the meta data and structure fields out of the ruleset files in translation files. The internal field names are used then as translation keys and translation is easily maintained out of ruleset files. Adding new language translations require then only an added translation bundle file and no change to the ruleset file.

Kathrin-Huber: > 
> 
> I don't think so. I don't know if there is some list of available languages which could be used here.

Can you add it to the config file? "metadataLanguages" or something

Beacze: Added.

PR #2248, opened 2019-02-26
Initialize LOCAL file group

matthias-ronge:
The local filegroup was not initialized. I have no idea where that was lost because that was once there. I have now made a true constant of it, which is now possible.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/metadata/legacytypeimplementations/LegacyInnerPhysicalDocStructHelper.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/metadata/legacytypeimplementations/LegacyInnerPhysicalDocStructHelper.java
@@ -37,11 +37,15 @@ public class LegacyInnerPhysicalDocStructHelper implements LegacyDocStructHelper
     private static final MetsService metsService = ServiceManager.getMetsService();
 
     /**
-     * A constant for local use. We cannot make this constant constant because
-     * the service loader is an instance variable.
+     * A constant for local use.
      */
     @Deprecated
-    public final MediaVariant local = new MediaVariant();
+    public static final MediaVariant LOCAL = new MediaVariant();
+
+    static {
+        LOCAL.setUse("LOCAL");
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/metadata/legacytypeimplementations/LegacyInnerPhysicalDocStructHelper.java
```diff
@@ -37,11 +37,15 @@
...
-     * A constant for local use. We cannot make this constant constant because
-     * the service loader is an instance variable.
+     * A constant for local use.
      */
     @Deprecated
-    public final MediaVariant local = new MediaVariant();
+    public static final MediaVariant LOCAL = new MediaVariant() {
+        {
+            setUse("LOCAL");
+            setMimeType("image/tiff");
```
henning-gerhardt: Should setting the mime type not depend of which kind of file is really used as local file? So far as I know institutes are using jpeg files for local usage and their tiff files are not used.

matthias-ronge: This is only a temporary solution as long as the helper classes for UGH replacement still exist. The value has always been written that way. It has never been used and read, it just needs to be in the field so there is no NullPointerException. Everything is configurable in the final solution. You do not even have to call the group LOCAL anymore.

Kathrin-Huber:
> 
> 
> I have tests changes. It creates for me now mimetypes but when I want to open in editor changed process, editor crashes with unknown error. Debugging didn't show any error, but editor is not responsive anymore.
> 
> error

@matthias-ronge can you reproduce this?

matthias-ronge:
No, it works for me. May you ask to @Beacze? We debugged this last week and it was some unrelated problem, but I don’t remember at the moment what it was.

matthias-ronge:
Is there still something missing here? Or can you merge the branch now?

PR #2249, opened 2019-02-26
Replace active and ready with status in workflow

Beacze:
workflow
workflow2


Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -27,7 +27,7 @@ active=Aktiv
 activeUsers=Aktive Benutzer
 addDemoTask=Beispielaufgabe hinzuf\u00FCgen
 addFilter=Filter hinzuf\u00FCgen
-addMessage=Nachricht hinzuf\u00FCgen
+addMessage=Nachricht hinzuf\u00FCgens
```
Kathrin-Huber: Why did you change it to "hinzufügens"?

Beacze: Mistake, now it is removed.

PR #2255, opened 2019-02-28
Update colors

oliver-stoehr:
Use css variables and replace colors not matching the new GUI specification.


Kathrin-Huber:
Nice! Thanks for cleaning this up!

PR #2259, opened 2019-02-28
Update initial roles

Beacze:


Beacze:
1. resetWorkflow has unknown usage and is with question mark in specification so I have left it out
2. fixed :)

PR #2280, opened 2019-03-04
Skip 'null' pages in method to retrieve logical page number

solth:


Kitodo/src/main/java/org/kitodo/production/metadata/MetadataProcessor.java
```diff
@@ -2129,13 +2129,13 @@ public void addNewImagesAndPaginate() {
...
      * Gets the logical page number from a paginated docstruct.
-     * 
+     *
      * @param docStruct
      *            The DocStruct opject.
      * @return The logical page number.
      */
     public String getLogicalPageNumber(LegacyDocStructHelperInterface docStruct) {
-        for (String page : allPages) {
+        for (String page : Arrays.stream(allPages).filter(Objects::nonNull).collect(Collectors.toList())) {
```
Beacze: Wouldn't it be better to skip insert of nulls to this list instead of filtering for them here?

PR #2292, opened 2019-03-06
Provide access to the type of a media unit

matthias-ronge:
Required for partial jobs

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
@@ -61,8 +61,13 @@ public class MediaUnit {
     private String orderlabel;
 
     /**
+     * The type of the media unit.
+     */
+    private String type;
+
+    /**
      * Returns the subordinate media units associated with this media unit.
-     * 
+     *
      * @return the subordinate media units
      */
     public List<MediaUnit> getChildren() {
@@ -72,7 +77,7 @@ public class MediaUnit {
     /**
      * Returns the map of available media variants with the corresponding media
      * file URIs.
-     * 
+     *
      * @return available media variants with corresponding media file URIs
      */
     public Map<MediaVariant, URI> getMediaFiles() {
@@ -81,7 +86,7 @@ public class MediaUnit {
 
     /**
      * Returns the meta-data on this structure.
-     * 
+     *
      * @return the meta-data
      */
     public Collection<Metadata> getMetadata() {
@@ -90,7 +95,7 @@ public class MediaUnit {
 
     /**
      * Returns the order number for this media unit.
-     * 
+     *
      * @return the order number
      */
     public int getOrder() {
@@ -99,7 +104,7 @@ public class MediaUnit {
 
     /**
      * Sets the order number for this media unit.
-     * 
+     *
      * @param order
      *            order number to set
      */
@@ -109,7 +114,7 @@ public class MediaUnit {
 
     /**
      * Returns the order label for this media unit.
-     * 
+     *
      * @return the order label
      */
     public String getOrderlabel() {
@@ -118,7 +123,7 @@ public class MediaUnit {
 
     /**
      * Sets the order label for this media unit.
-     * 
+     *
      * @param orderlabel
      *            order label to set
      */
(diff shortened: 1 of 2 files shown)


Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -118,20 +123,42 @@ public String getOrderlabel() {
...
+
     @Override
     public String toString() {
         String fileName = "No file (";
         if (!mediaFiles.isEmpty()) {
             URI uri = mediaFiles.entrySet().iterator().next().getValue();
-            fileName = FilenameUtils.getBaseName(uri.getPath()).concat(".* (");
+            fileName = FilenameUtils.getBaseName(uri.getPath()).concat(" (");
+        }
+        if (type != null) {
```
Beacze: We decided to use Objects.nonNull

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -144,6 +171,7 @@ public int hashCode() {
         result = prime * result + ((mediaFiles == null) ? 0 : mediaFiles.hashCode());
         result = prime * result + order;
         result = prime * result + ((orderlabel == null) ? 0 : orderlabel.hashCode());
+        result = prime * result + ((type == null) ? 0 : type.hashCode());
```
Beacze: We decided to use Objects.isNull

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -176,6 +204,13 @@ public boolean equals(Object obj) {
         } else if (!orderlabel.equals(other.orderlabel)) {
             return false;
         }
+        if (type == null) {
```
Beacze: We decided to use Objects.isNull

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -176,6 +204,13 @@ public boolean equals(Object obj) {
...
+            if (other.type != null) {
```
Beacze: We decided to use ObjectsnonNull

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -144,6 +171,7 @@ public int hashCode() {
         result = prime * result + ((mediaFiles == null) ? 0 : mediaFiles.hashCode());
```
Beacze: Is it really necessary to build up so complicated hashCode?

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
```diff
@@ -176,6 +204,13 @@ public boolean equals(Object obj) {
         } else if (!orderlabel.equals(other.orderlabel)) {
```
Beacze: This comment is about code not changed here but I find it important:

`if (obj == null) {
            return false;
        }
        if (!(obj instanceof MediaUnit)) {
            return false;
        }`

Do you know that _!(obj instanceof MediaUnit)_ checks also against null? You could skip first check.

matthias-ronge:
@Beacze this code is auto-generated by Eclipse

Beacze:
@matthias-ronge are we developers or Eclipse code generators?

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
matthias-ronge: Modified as discussed.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
matthias-ronge: Modified as discussed.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
matthias-ronge: Modified as discussed.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
matthias-ronge: Modified as discussed.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
matthias-ronge: Writing good hash functions is not easy. The hash function is not only intended to avoid hash collisions as much as possible, but also to distribute as well as possible in order to efficiently allocate hash maps. Because I'm not a professor of theoretical computer science, I prefer to rely on what Eclipse delivers, which is used and developed by computer scientists worldwide. They will do better than me.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaUnit.java
matthias-ronge: Above all, your comment is not on the affected line. I changed it.

PR #2294, opened 2019-03-07
Add security access for custom authorities for modules

Beacze:
- move SecurityAccess to the API module
- add page for insert / edit / delete authorities
- add initial state migration

Note: at the moment works for global authorities in outside modules. It is necessary to find way to store client id in way which will be accessible for outside modules / API. Maybe in session?

kitodo


Kitodo/src/main/java/org/kitodo/production/forms/AuthorityForm.java
```diff
@@ -0,0 +1,177 @@
...
+import org.kitodo.production.enums.ObjectType;
+import org.kitodo.production.helper.Helper;
+import org.kitodo.production.model.LazyDTOModel;
+import org.kitodo.production.services.ServiceManager;
+
+@Named("AuthorityForm")
+@SessionScoped
+public class AuthorityForm extends BaseForm {
+    private static final long serialVersionUID = 3541160917458068675L;
+    private static final Logger logger = LogManager.getLogger(RoleForm.class);
```
henning-gerhardt: Logger name did not match class name.

Kitodo/src/main/webapp/pages/authorityEdit.xhtml
```diff
@@ -0,0 +1,72 @@
+<ui:composition
```
henning-gerhardt: Missing license header.

PR #2295, opened 2019-03-07
Remove insert of resetWorkflow

Beacze:


henning-gerhardt:
Should this not be removed into a new migration file? This file is already in previous releases used.

Beacze:
There is no official release. Both releases were for developers. For them it doesn't matter if this one thing was inserted or not. Additionally there is frontend by which you can add / remove authority. Those all roles / authorities were inserted there only for testing purpose.

Regarding warnings from flyway: When message that mirgations don't match appears, you run repair and all is again good.

PR #2296, opened 2019-03-07
Add translations for authorities

Beacze:
Note: for missing translations still title from database is going displayed.

Depends on #2294 

auth1
auth2

PR #2297, opened 2019-03-08
Remove method getLocalizedTitle from Task bean - it doesn't translate

Beacze:
Translation of title can be achieved by adding translation keys to messages files. Next call method getTranslated from HelperForm (change is made in PR https://github.com/kitodo/kitodo-production/pull/2296).


Beacze:
Will be replaced after #2296 is merged.

PR #2310, opened 2019-03-12
Add DFG-Viewer mini ruleset

matthias-ronge:


matthias-ronge:
I had made this rule set for the workshop last week Tuesday. I do not understand why he's still lying around here.

PR #2325, opened 2019-03-13
Export images according to the folder configuration

matthias-ronge:


Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -102,6 +102,24 @@ URI createMetaDirectory(URI parentFolderUri, String directoryName) throws IOExce
         return URI.create(parentFolderUri.getPath() + '/' + directoryName);
     }
 
+    /**
+     * Creates a directory including any missing parent directories.
+     *
+     * @param pathToCreate
+     *            path to create
+     */
+    public void createDirectories(URI pathToCreate) throws IOException {
```
Kathrin-Huber: please write a test for this.
I think you get an indexOutOfBounds exception if you pass a path without a slash here, which can happen if you iterate over and over and get to the last directory. (I see you expect the process directories to be existent, but it is a public method an anyone can pass an arbitrary folder to be created)

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -481,14 +476,11 @@ public void deleteTiffHeaderFile(List<Process> processes) {
         Helper.setMessage(KITODO_SCRIPT_FIELD, "", "deleteTiffHeaderFile finished");
     }
 
-    private void exportDms(List<Process> processes, String exportImages, boolean exportFulltext) {
```
Kathrin-Huber: why is the fullText not exported anymore?

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -600,26 +501,40 @@ private void handleException(Exception e, String processTitle) {
...
-        String normalizedTitle = Helper.getNormalizedTitle(process.getTitle());
-
-        for (String processDir : processDirs) {
-            URI srcDir = ServiceManager.getProcessService().getProcessDataDirectory(process)
-                    .resolve(processDir.replace("(processtitle)", normalizedTitle));
-            URI dstDir = destination.resolve(processDir.replace("(processtitle)", normalizedTitle));
-
-            if (fileService.isDirectory(srcDir)) {
-                fileService.copyDirectory(srcDir, dstDir);
+    private void directoryDownload(Process process, URI destination) throws IOException, InterruptedException {
```
Kathrin-Huber: please add a test here too

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -205,9 +205,17 @@ private boolean prepareAndDownloadSaveLocation(Process process, URI destinationD
...
 
-        boolean downloadImages = downloadImages(process, userHome, destination);
-        if (!downloadImages) {
-            return false;
+        if (exportWithImages) {
+            try {
+                directoryDownload(process, destination);
+            } catch (IOException | InterruptedException | RuntimeException e) {
+                if (Objects.nonNull(exportDmsTask)) {
+                    exportDmsTask.setException(e);
```
Kathrin-Huber: there is no error message here

matthias-ronge: That's true. What kind of error message should there be?

Kathrin-Huber: How can I decide? You are the one to know, what is happening here.

matthias-ronge: Of course, I know what's going on here: If an exception has occurred, it will be set in the executing thread for it to appear in the UI. If a message is to be set, it must happen at the point where the exception is thrown. Here only an existing exception is forwarded to the user interface. Therefore, I do not know what I should add here.

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
matthias-ronge: The full text will be exported exactly if a folder is configured in the folder settings and if the switch for export is set to "on". This will delete the export of a hard-coded folder. If you want to continue exporting the folder, you must configure it.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
matthias-ronge: Test added

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
matthias-ronge: Test added

Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
```diff
@@ -56,4 +62,42 @@ public void testRenameFileWithLockedFile() throws IOException {
         }
 
     }
+
+    @Test
+    public void testCreateDirectories() throws IOException {
+        Path testBaseDirectoryPath = Paths.get("src/test/resources");
```
Kathrin-Huber: You can use it from the config file

matthias-ronge: Uset

matthias-ronge:
Well, that would finally be finished. Please merge!

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
Kathrin-Huber: You didn't change the code here and you are still getting an index out of bounds exception.
You can not assume, that any part of the directory is already there!

Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
```diff
@@ -56,4 +63,42 @@ public void testRenameFileWithLockedFile() throws IOException {
...
+        new FileService().createDirectories(URI.create(testBaseDirectoryURI + createDirectories));
+        assertTrue(Files.isDirectory(firstDirectoryOfTestPath.resolve("directories/can/be/created/at/once")));
+
+        // clean up
+        TreeDeleter.deltree(firstDirectoryOfTestPath);
+    }
+
+    @Test
+    public void testCreateDirectoriesWithTrailingSlash() throws IOException {
+        Path testBaseDirectoryPath = Paths.get(ConfigCore.getParameter(ParameterCore.DIR_XML_CONFIG));
```
Kathrin-Huber: The mapping to the kitodo directory is happening on the module

Kathrin-Huber: Thats why your test isn't failing with indexOutOfBounds.
In the core we don't know how the file system is managed, only the FileManagement knows.

Kitodo/src/test/java/org/kitodo/production/services/file/TreeDeleter.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.production.services.file;
```
Kathrin-Huber: Please move this file to "org.kitodo" like the other Util-files in test.
In the specific test-directorys are only testfiles for the corresponding packages.

Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
matthias-ronge: I did not understand it yet. What must the test look like to (currently) fail?

Beacze: She meant that all magic happens in File Management module, not here. That's why this test will not fail for whatever reason. Simply saying, there is not way to do it.

Kitodo/src/test/java/org/kitodo/export/ExportDmsIT.java
```diff
@@ -0,0 +1,133 @@
...
+            ocrAlto.list((directory, filename) -> filename.endsWith(".xml")).length);
+
+        File pdf = new File(downloadDir, "pdf");
+        assertTrue("Folder pdf should exist in destination location", pdf.isDirectory());
+        assertEquals("pdf should contain 184 PDFs", 184,
+            pdf.list((directory, filename) -> filename.endsWith(".pdf")).length);
+
+        float totalBytes = 184f * (358000 + 181554 + 1890 + 25157 + 3548885);
+        float mbps = totalBytes / tookNanos * 1953125 / 2048;
+        assertTrue("It should have been copied >50 MB/s (was: " + mbps + " MB/s)", mbps > 50);
```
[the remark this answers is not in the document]
matthias-ronge: Why not? Can you also test. The copying should be finished someday.

Kitodo/src/test/java/org/kitodo/production/services/file/TreeDeleter.java
matthias-ronge: Moved.

Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
matthias-ronge: The question is, if the method I am testing is allegedly flawed, then it must be possible to write a test that results in the error. If not, the method must be correct. If the error occurs in the file management module, then there is the same thing, then you would have to be able to write a test for it, which now produces an error first. The thing is: the method works for me and does what it should. But Kathrin said she had written a "tiny little" test that immediately threw an exception. If so, I want to know what the test looks like so that I can reproduce the error and then repair the method.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
matthias-ronge: I do not know what I would have to change, because the method throws no exception for me. I even tried `new FileService().createDirectories(new File("F:\\create\\several\\directories\\").toURI());` on an freshly formatted USB stick, so there was **definitely nothing** there before, and did not throw an exception. What does the code look like which you used to get an exception there?

matthias-ronge:
I had added the test of you and then also found the error and corrected the method. Everything should be right here, isn’t it?

Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
```diff
@@ -48,12 +61,61 @@ public void testRenameFileWithLockedFile() throws IOException {
...
 
         } finally {
             outputStream.close();
         }
 
     }
+
+    @Test
+    public void testCreateDirectories() throws IOException {
+        Path testBaseDirectoryPath = Paths.get(ConfigCore.getParameter(ParameterCore.DIR_XML_CONFIG));
```
Kathrin-Huber: You added the new test, but you didn't adjust the old ones.
this line needs to be removed here. Please do not give the directory here as the fileManagement is handling it.

Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
```diff
@@ -48,12 +61,61 @@ public void testRenameFileWithLockedFile() throws IOException {
...
+        // clean up
+        TreeDeleter.deltree(firstDirectoryOfTestPath);
+    }
+
+    @Test
+    public void testCreateDirectoriesWithTrailingSlash() throws IOException {
+        Path testBaseDirectoryPath = Paths.get(ConfigCore.getParameter(ParameterCore.DIR_XML_CONFIG));
+
+        // delete existing directories so that directories are actually created
+        Path firstDirectoryOfTestPath = testBaseDirectoryPath.resolve(SEVERAL);
```
Kathrin-Huber: I think the testenvironment should ensure, that there is no such directory. This part should be removed.

Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
```diff
@@ -48,12 +61,61 @@ public void testRenameFileWithLockedFile() throws IOException {
...
+        if (Files.isDirectory(firstDirectoryOfTestPath)) {
+            TreeDeleter.deltree(firstDirectoryOfTestPath);
+        }
+
+        String testBaseDirectoryURI = testBaseDirectoryPath.toUri().toString();
+        String createDirectories = "several/directories/can/be/created/with/trailing/";
+        new FileService().createDirectories(URI.create(testBaseDirectoryURI + createDirectories));
```
Kathrin-Huber: please use ServiceManager.getFileService() instead of new FileService()

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -112,17 +112,12 @@ private boolean executeScript(List<Process> processes) throws DataException {
                 setRuleset(processes);
                 break;
             case "exportDms":
-                exportDms(processes, this.parameters.get("exportImages"), true);
-                break;
             case "export":
-                exportDms(processes, this.parameters.get("exportImages"),
```
Kathrin-Huber: Why is this the same now?

PR #2338, opened 2019-03-15
Show progress cursor during AJAX request

matthias-ronge:
This improves the responsiveness a bit: If you click on a button that requires an AJAX request, a small load wheel or a small hourglass appears on the mouse (or the correspondingly configured mouse pointer). Unfortunately, it does not (yet) work on buttons that trigger a link to a new page. So: you can see it with buttons where you stay on the same page.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -95,6 +95,9 @@ sup {
     100% {opacity: 1;}
     }
 
+html.progressCursor * {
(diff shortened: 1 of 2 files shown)


matthias-ronge:
Apparently, Codacy does not like my suggestion. Unfortunately, I have too little idea about CSS to assess if this is “out of control”. I had found the example, implemented it and it worked. I have no idea how to proceed further.

Kathrin-Huber:
Maybe you ask Oliver. Is the !important really neccessary?

Kathrin-Huber:
So for me it is also working without the !important. can you test this please?

PR #2357, opened 2019-03-20
Unify prepare method for prepare process

Beacze:


Beacze:
It is included as part of #2358

PR #2387, opened 2019-03-26
Add improvements and tests for process fields

Beacze:


Beacze:
Value reassigning here is necessary to avoid null.

PR #2397, opened 2019-03-27
Update popups and add initial client

Kathrin-Huber:


Kitodo/src/test/java/org/kitodo/selenium/AddingST.java
```diff
@@ -187,6 +188,7 @@ public void addRulesetTest() throws Exception {
         assertTrue("Created Ruleset was not listed at rulesets table!", rulesetAvailable);
     }
 
+    @Ignore
```
Beacze: What do you mean with ignore for later development?

PR #2400, opened 2019-03-28
Fix batch labels

Beacze:


Kitodo/src/main/java/org/kitodo/production/services/data/BatchService.java
```diff
@@ -303,39 +305,45 @@ public int size(Batch batch) {
      * batch.
      *
      */
-    public String toString(Batch batch) {
+    public void toString(Batch batch) {
```
Kathrin-Huber: shouldn't this be named "createLabel" or something?

Kitodo/src/main/java/org/kitodo/production/helper/tasks/ExportSerialBatchTask.java
```diff
@@ -80,7 +80,7 @@
      *            batch holding a serial publication
      */
     public ExportSerialBatchTask(Batch batch) {
-        super(batch.getLabel());
+        super(ServiceManager.getBatchService().getLabel(batch));
```
Kathrin-Huber: getLabel is not using the label you created

Kitodo/src/main/java/org/kitodo/production/forms/BatchForm.java
```diff
@@ -161,6 +164,9 @@ public void filterBatches() {
                 currentBatches.add(batch);
             }
         }
+        for (Batch batch : this.currentBatches) {
```
Kathrin-Huber: can't you create the label in "getLabel" if there is no label?
I think it's not intuitive to create the label beforehand

Kitodo/src/main/java/org/kitodo/production/helper/tasks/ExportSerialBatchTask.java
Beacze: Yeah, it is different label. The one which overwrites toString is used only for frontend.

Kitodo/src/main/java/org/kitodo/production/forms/BatchForm.java
Beacze: Yeah, but it is different label. I don't know why, but this class creates different label in getLabel and different in toString.

Kitodo/src/main/java/org/kitodo/production/services/data/BatchService.java
Beacze: renamed

PR #2446, opened 2019-04-09
Unify outcome links for tab views

Beacze:


solth:
Is the getter `BaseForm.getActiveTabIndex()` still required if those calls in the templates are removed?

Beacze:
It must be there.
debug

PR #2451, opened 2019-04-11
Move dupliacte code to private method in ImageManagement

solth:


solth:
@matthias-ronge could you review this PR?

PR #2464, opened 2019-04-17
Add missing method

matthias-ronge:
Meta-data editor does not work if there are pages in the process, because there is a method missing. The method is invoked only form JSF so the compiler does not notice. Adds the missing method.


solth:
Since this only affects the _old_ MetadataEditor (e.g. `MetadataProcessor.java`) there shouldn't be any problems with the developments for the new editor.

PR #2465, opened 2019-04-17
Add a missing semicolon in an SQL script

matthias-ronge:
There was a semicolon missing in one SQL script. Adds the missing semicolon.


Kathrin-Huber:
I would rather not change a flywayscript if it is working for flyway. We are going to respect the database migration without flyway in another part.

PR #2471, opened 2019-04-18
Reactivate Kitodo scripts

Beacze:


Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -870,46 +877,89 @@ private void reload() {
...
+        executeKitodoScriptForProcesses(lazyDTOModel.getEntities(), this.kitodoScriptAll);
     }
 
     /**
      * Execute Kitodo script for processes displayed on the page.
      */
     @SuppressWarnings("unchecked")
     public void executeKitodoScriptPage() {
-        executeKitodoScriptForProcesses(lazyDTOModel.getEntities());
+        executeKitodoScriptForProcesses(lazyDTOModel.getEntities(), this.kitodoScriptPage);
```
Kathrin-Huber: Shouldn't there be a difference between executeKitodoScriptAll and executeKitodoScriptPage?
They both take lazyDTOModel.getEntities()

Beacze: There should be. It is also a case for all other stuff. Problem is that I know how to access current page lazyDTOModel.getEntities(), but not how to access all of the objects.

PR #2472, opened 2019-04-18
Save parent for process if you save process itself

Beacze:


Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -203,6 +203,14 @@ public Long countResults(Map filters) throws DataException {
         throw new UnsupportedOperationException();
     }
 
+    @Override
+    public void save(Process process) throws DataException {
```
Kathrin-Huber: A test would be nice :)

Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
```diff
@@ -100,6 +100,28 @@ public void shouldCountAllDatabaseRowsForProcesses() throws Exception {
...
+        process.setTitle("Child");
+        process.setParent(parent);
+        parent.getChildren().add(process);
+
+        processService.save(process);
+
+        Process foundProcess = processService.getById(process.getId());
+
+        assertEquals("Child process was not found in database!", "Child", foundProcess.getTitle());
+        assertEquals("Parent process was not assigned to child!", "Parent", process.getParent().getTitle());
```
Kathrin-Huber: I think you should use "foundProcess" here instead of process

Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
```diff
@@ -100,6 +100,28 @@ public void shouldCountAllDatabaseRowsForProcesses() throws Exception {
...
+
+        Integer parentId = foundProcess.getParent().getId();
+
+        processService.remove(parentId);
```
Kathrin-Huber: Why are you only removing the parent?

Beacze: Parent is the owner, it removes own children.

Kathrin-Huber: Oh well.. i'm not sure, if this is the behaviour we want :(
I'm going to check this

Beacze: I also don't know, but it is default behavior for Hibernate. When parent is removed, child are automatically gone also.

PR #2479, opened 2019-04-23
Execute actions for all for 100000 records

Beacze:


Kathrin-Huber:
Is 100000 enough?

Beacze:
I don't know, for now we can hope that yes. If you start export for 100 000 it will take days.

Kathrin-Huber:
@henning, maybe you can evaluate it from a practical point of view?

henning-gerhardt:
In normal cases 100 000 processes are enough or at least we did not work which such a high number as processing more then 10 000 processes through KitodoScript actions take ages. But this should be really solved later and until this 100 000 should be enough.

PR #2482, opened 2019-04-24
2.x: Add a GoobiScript function to rewrite process meta-data

matthias-ronge:
This script is suitable for cases in which you would otherwise always have to open the meta-data editor for each individual process and click on Save. That makes it much faster.


matthias-ronge:
The codec check fails because it is not configured correctly. I do not have access to it, you have to do that.

henning-gerhardt:
1) What is the usage scenario for this script? I can not image who this could be helpful.
2) If I activate Codacy for 2.x branch then any pull request will fail until the quality pattern of 3.x are reached. I don't know how to disable this check for the 2.x branch.

Goobi/src/de/sub/goobi/helper/GoobiScript.java
```diff
@@ -265,6 +267,30 @@ private void deleteMetadataDirectory(Prozess p) {
         }
     }
 
+    private void rewriteProcessMetadata(List<Prozess> processes) {
+        String currentProcessTitele = null;
```
henning-gerhardt: You must not initialise this property `currentProcessTitele` (contains even a spelling error) as this property will be overwritten in any iteration of the loop.

matthias-ronge: I must, otherwise the Java compiler complains. This is because—in abstract terms—method `process.getTitel()` might already throw an exception, and then the variable would not have been initialized at the `!= null` check. In fact, a variable in Java could always be implicitly initialized with `null`, but that is the educational part of Java. But that's what you only understand today if you've worked with such historical languages as C before.
Thanks for the typo, I have corrected.

matthias-ronge: “Titele”—that was Swabian. :wink:

matthias-ronge:
1. We need this for the migration from XStream to METS-internal. This can be done by selecting METS in the project settings for the internal format, and the next time you save, the respective process is saved in METS/MODS-Goobi. But do that with 17,000 processes! That's why the script.
1. I have no idea what happens then or what quality Codacy expects then. For example, there is no checkstyle configuration file in the branch. However, Codacy only expects the code to be adjusted, which has actually been edited—thank goodness!

matthias-ronge:
Ah, apparently it fits now with Codacy. Is there anything else to do or can this be merged?

PR #2483, opened 2019-04-24
Replace priority with correction for task bean

Beacze:
PO decided that priority is useless and should be removed. 

But priority had one function - marking correction tasks. So here priority is removed with persisting correction functionality.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Task.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Task.java
@@ -42,9 +42,6 @@ public class Task extends BaseIndexedBean {
     @Column(name = "title")
     private String title;
 
-    @Column(name = "priority")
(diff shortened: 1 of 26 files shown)


sebastian-meyer:
PO didn't decide that 'priority is useless', but that priority shouldn't be a property of tasks, but instead of processes. :)

Beacze:
So we have removed priority from tasks. What should we do with priority for processes? What does it actually mean for processes?

sebastian-meyer:
It means the same as for tasks:
If a process needs corrections, only the correction tasks get a higher priority. Those tasks are than sorted to the top of the task list and are marked as "correction".
Additionally there should be an option to prioritize a whole process throughout the complete workflow. Basically that's the same as prioritizing every task of the process, but it shouldn't be implemented the same way, because then prioritizing an already prioritized task in case of a correction wouldn't be possible.

PR #2499, opened 2019-04-29
Hierarchical METS export (1)

matthias-ronge:
**Extend the API for writing links in the METS export**
The core provides the data format module with the linked METS resources. For this, a bean with the necessary fields has been added. The children are a map, where the string determines the insertion position for the link.


matthias-ronge:
I'm not sure but we probably do not need that anymore − at least not like _that_

PR #2500, opened 2019-04-29
Hierarchical METS export (2)

matthias-ronge:
Adds methods to write links in METS

Based on PR #2499, [immediate diff](https://github.com/matthias-ronge/kitodo-production/compare/KIT-1073_1...matthias-ronge:KIT-1073_2)

### DIFF
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
@@ -76,7 +76,7 @@ class MptrXmlElementAccess {
      *            {@code <mets:div>}, which may contain information about a link
      * @return information to the link, if indicated
      */
-    static LinkedMetsResource readLink(DivType div) {
+    static LinkedMetsResource getLinkFromDiv(DivType div) {
         if (div.getMptr().isEmpty()) {
             return null;
         }
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
@@ -12,10 +12,7 @@
 package org.kitodo.dataformat.access;
 
 import java.net.URI;
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Objects;
-import java.util.Set;
 
 import org.kitodo.api.dataformat.mets.LinkedMetsResource;
 import org.kitodo.dataformat.metskitodo.DivType;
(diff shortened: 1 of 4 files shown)


Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
```diff
@@ -399,4 +420,123 @@ private StructLink createStructLink(LinkedList<Pair<String, String>> smLinkData)
         }).collect(Collectors.toList()));
         return structLink;
     }
+
+    private void addChildLinks(Map<String, LinkedMetsResource> childLinks, Mets mets) {
+        Map<String, Collection<Entry<String, LinkedMetsResource>>> linksByParent = collectTheLinksPerParent(childLinks);
```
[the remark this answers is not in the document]
matthias-ronge: I think that is always a bit superfluous for a local variable, which specifies the results of a private method, which is used in just one other method, to specify its own object ... but as you like.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
```diff
@@ -399,4 +420,123 @@ private StructLink createStructLink(LinkedList<Pair<String, String>> smLinkData)
...
+    }
+
+    /**
+     * Summarizes the links per parent.
+     *
+     * @param children
+     *            children to be linked
+     * @return the children, grouped according to parents
+     */
+    private Map<String, Collection<Entry<String, LinkedMetsResource>>> collectTheLinksPerParent(
```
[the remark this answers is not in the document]
matthias-ronge: renamed.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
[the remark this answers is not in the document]
Beacze: I find it more easy to read in the future. At this moment it is used only in this place, but can you be 100% sure that it will be not extended in the future?

Thank you for change.

matthias-ronge:
Writing links must be different then, too.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * String indicating that the LOCTYPE value is other than one of the allowed
+     * constants and has been placed in the OTHERLOCTYPE element.
+     */
+    private static final String OTHER = "OTHER";
+
+    /**
+     * Set of allowed LOCTYPE values.
+     */
+    private static final Set<String> ALLOWED_LOCTYPE_VALUES = new HashSet<>(
```
Kathrin-Huber: Please use an enum instead

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
```diff
@@ -0,0 +1,80 @@
...
+    static void addMptrToDiv(LinkedMetsResource link, DivType div) {
+        div.setORDER(link.getOrder());
+        Mptr mptr = new Mptr();
+        if (ALLOWED_LOCTYPE_VALUES.contains(link.getLoctype())) {
+            mptr.setLOCTYPE(link.getLoctype());
+        } else {
+            mptr.setLOCTYPE(OTHER);
+            mptr.setOTHERLOCTYPE(Objects.toString(link.getLoctype()));
+        }
+        mptr.setHref(link.getUri().toASCIIString());
```
Kathrin-Huber: Why do you convert it to asciistring?

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Reads the information about a link from a {@code <mets:div>}.
+     *
+     * @param div
+     *            {@code <mets:div>}, which may contain information about a link
+     * @return information to the link, if indicated
+     */
+    static LinkedMetsResource readLink(DivType div) {
```
Kathrin-Huber: Is
getLinkFromDiv
a better name?
Or do I get the method wrong? becuase you are not actually reading the link, right? You extract the information about the link from the div.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
matthias-ronge: So that the URL is escaped correctly.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MptrXmlElementAccess.java
matthias-ronge: Can one also say so.

PR #2502, opened 2019-04-29
Hierarchical METS export (3)

matthias-ronge:
**Add a field for the position to the process**
The position indicates at which position a child is linked within the logical structure of his parent. The syntax for this string is: `(\d+,)*\d+\.\d+`

Example (for a year of a newspaper): `2,15,0.5` = in the second month, in the 15th day, before the first non-linked child. (Child count is 1-based.)

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -85,6 +85,9 @@ public class Process extends BaseTemplateBean {
     @JoinColumn(name = "parent_id", foreignKey = @ForeignKey(name = "FK_process_parent_id"))
     private Process parent;
 
+    @Column(name = "position")
+    private String position;
+
     @OneToMany(mappedBy = "parent", cascade = CascadeType.ALL, orphanRemoval = true)
     private List<Process> children;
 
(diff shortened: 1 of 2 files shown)


Kathrin-Huber:
I think we already have the attribute "ordering" for order of the processes.
So order in month would be 2, order in day 15 and so on...

matthias-ronge:
I think that ordering takes the `ORDER` that you need to create the processes, and that we should also output to the METS. But that does not indicate the position of a child within its parenting process. Suppose you have an issue of a newspaper and it has an annual process as its parenting process. Since you can not even specify the position with a single int, that is too complex.

Kathrin-Huber:
> But that does not indicate the position of a child within its parenting process.

What else does it do?

matthias-ronge:
If I understand Sebastian correctly, the idea is this: For example, if you create several volumes of a multi-volume work, but not necessarily in their proper order. When you create the process, you enter the `ORDER`, so the seventh volume, for example, the 7. If you then later create the third volume, it is sorted before the 7th. If you create the 5th, it will be sorted after the 3rd but before the 7th. During creating. As soon as the processes have been created, the value has no meaning any more. It should then be output only with the export.

matthias-ronge:
We will not use a database field for positioning.

PR #2503, opened 2019-04-29
Hierarchical METS export (4)

matthias-ronge:
That would be my suggestion to encode a step in the database that causes the exportability of a child when it is completed. The condition must then be that `task.processingStatus == TaskStatus.DONE`. What do you think about?

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -47,6 +47,9 @@ public class Process extends BaseTemplateBean {
     @Column(name = "sortHelperImages")
     private Integer sortHelperImages;
 
+    @Column(name = "exportable")
+    private boolean exportable;
+
     @Column(name = "sortHelperArticles")
     private Integer sortHelperArticles;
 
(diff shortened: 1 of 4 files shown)


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Task.java
```diff
@@ -91,6 +91,9 @@
     @Column(name = "typeAcceptClose")
     private boolean typeAcceptClose = false;
 
+    @Column(name = "typeExportable")
+    private boolean typeExportable = false;
```
Kathrin-Huber: why typeExportable and not "exportable"?

matthias-ronge: All these boolean properties within Task are called starting with type. I just wanted to follow the convention. Should it only be called exportable? I think the name is confusing, because the task itself is not exportable. Maybe we need a better name. Maybe `typeExportableWhenDone`? I was here but first to the concept of how to save it. Do you think that's right?

matthias-ronge: After consultation with Sebastian that should be solved differently. The “exportable” property is said to be a process property, and the step should only set this property to `true` when it is reached. (It should also be possible to manually change this process property.) So I would suggest here as a name: `typeSetExportableToTrue`.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
```diff
@@ -47,6 +47,9 @@
     @Column(name = "sortHelperImages")
     private Integer sortHelperImages;
 
+    @Column(name = "exportable")
+    private boolean exportable;
```
Beacze: You add property but no setter and getter for it... Hibernate will not work correctly in such a case.

matthias-ronge:
The part can stay, isn’t it?

Kathrin-Huber:
After consultation with Matthias, we are closing this for now

PR #2546, opened 2019-05-16
Display linked processes in the meta-data editor

matthias-ronge:

PR #2550, opened 2019-05-16
Update the hierarchical links in the database when saving

matthias-ronge:
Follow-up pull request to #2546, [immediate diff](https://github.com/matthias-ronge/kitodo-production/compare/KIT-1069...matthias-ronge:KIT-1070)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -276,16 +280,90 @@ public String save() {
...
+            }
+            if (Objects.nonNull(referencedChildId) && Objects.isNull(savedChildId)
+                    || referencedChildId < savedChildId) {
+                Process childToAdd = ServiceManager.getProcessService().getById(referencedChildId);
+                childToAdd.setParent(process);
+                process.getChildren().add(childToAdd);
+                save = true;
+                nextReferencedChild = true;
+            } else if (Objects.isNull(referencedChildId) && Objects.nonNull(savedChildId)
+                    || referencedChildId > savedChildId) {
```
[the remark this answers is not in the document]
Kathrin-Huber: And if I am not referencing a child at all, the saved Child is deleted, too?
I am confused, do I understand this right? I think this is not what you intended to do, right?

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -276,16 +280,90 @@ public String save() {
...
+                Process childToRemove = ServiceManager.getProcessService().getById(savedChildId);
+                childToRemove.setParent(null);
+                process.getChildren().remove(childToRemove);
+                save = true;
+                nextSavedChild = true;
+            } else if (Objects.nonNull(referencedChildId) && Objects.nonNull(savedChildId)) {
+                nextReferencedChild = true;
+                nextSavedChild = true;
+            }
+        } while (nextReferencedChild || nextSavedChild);
```
Kathrin-Huber: Do you need those flags, or is a check for NonNull on th id's sufficient? Because as I see it, this dependent

Kitodo/src/main/java/org/kitodo/production/helper/HelperComparator.java
```diff
@@ -48,6 +52,52 @@ public void setSortType(SortType sortType) {
...
+     * @param left
+     *            the one list
+     * @param right
+     *            the other list
+     * @return a map with the differences, mapped as T on boolean, where
+     *         {@code true} if it appears in the first list and not in the
+     *         second, {@code false} if it appears in the second list and not in
+     *         the first.
+     */
+    public static Map<Integer, Boolean> compareLists(TreeSet<Integer> left, TreeSet<Integer> right) {
```
Beacze: It should be in different class, maybe there where it is used or in ProcessListComparator.

HelperComparator is old utility class which compares legacy stuff. It is going to be removed.

Kitodo/src/test/java/org/kitodo/production/helper/HelperComparatorTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.Map;
+import java.util.TreeSet;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class HelperComparatorTest {
+    @Test
+    public void testCompareLists() {
```
Kathrin-Huber: _please test edges:_

empty left list
empty right list
both lists empty
both lists identical
duplex entries
no entries which are in both lists
lists with different sizes
...

matthias-ronge: Input is TreeSet. Duplex entries is not possible.

Kathrin-Huber:
Just a general question:

What about:

```
ArrayList newRight = new ArrayList(right);
ArrayList newLeft = new ArrayList(left);

newRight.removeAll(left);
newLeft.removeAll(right);
```

This should do exactly the same.

Kathrin-Huber:
Update:
I ran all your tests against those 4 lines of code and they are green ;)

matthias-ronge:
The process list comparator has an algorithmic complexity of _O(n)_, i.e. linear. That is, you have at most as many comparisons as the union of both lists. What you suggest would probably be _O(2·n²)_, that is, 1250 newspaper issues in one year (Mon−Sat every week with 4 regional issues) would result (in the worst case) in 3 million comparisons. This means that 3 million objects are loaded from the database in order to compare them. At first, I wanted to do the same thing, but then I thought that’s probably too inefficient and makes things comparatively slow. I am very worried that editing the newspaper hierarchies will be incredibly slow, if only because we have METS in the background with a slow parser. We cannot change that, but here we can write good code.

Kathrin-Huber:
Why do you need to load them from the database?
You only need to make database updates for the result lists, right?

matthias-ronge:
I meant, if you do that on `process.getChildren()` directly.
As it is realized at the moment at least all database children have to be loaded once to get their IDs. It also causes me a headache, but at least they can be forgotten right afterwards. We have to wait and see how it works in the real world. Otherwise we would have to obtain the IDs, for example, with an SQL query, or index hierarchical relationships, maybe. But I would first wait and see how it develops.

Beacze:
We don't discuss here loading of Process objects. The only part of code which should be changed is comparing of ids, which are Integer objects. Writing own implementations for something what can be done with simple existing Java class is waste of time which you don't have.

matthias-ronge:
To change code that works, just because there is another solution, is also a waste of time. The thing here was also that the affected objects must be processed. With `retainAll(currentState)` and `addAll(currentState)` the other list is correct afterwards. But `retainAll()` does not give me back the removed objects, which I still have to edit and save, with `addAll()` I do not know which objects were added and which were already in there. That’s why I wrote the linear list comparison. Now that we’ve been philosophizing about it for a whole day, it may seem like the other solution might be easier, but in the beginning, when I asked myself, _how do I effectively remove elements from a collection, when I only know which objects should be in it later, and at the same time find out which were removed_, it was not so easy.

Beacze:
This solution is hard to read and over complicated for given problem.

> when I asked myself, how do I effectively remove elements from a collection, when I only know which objects should be in it later, and at the same time find out which were removed, it was not so easy.

It sounds for me very easy. I have written solution for such a question around 2,5 years ago. Here is code from our code base: 
```
    {
        ...  
        List<Integer> missingInIndex = findMissingValues(database, index);
        List<Integer> notNeededInIndex = findMissingValues(index, database);
        ...
    }
```

    /**
     * Compare two list and return difference between them.
     *
     * @param firstList
     *            list from which records can be remove
     * @param secondList
     *            records stored here will be removed from firstList
     * @return difference between two lists
     */
    private List<Integer> findMissingValues(List<Integer> firstList, List<Integer> secondList) {
        List<Integer> newList = new ArrayList<>(firstList);
        newList.removeAll(secondList);
        return newList;
    }

matthias-ronge:
A simple solution would be like this:
```
private void updateChildLinks() throws DataException, DAOException {
    for (Process child : process.getChildren()) {
        child.setParent(null);
        process.getChildren().remove(child);
        ServiceManager.getProcessService().save(child);
    }
    for (URI uri : getAllChildrenRecursive(workpiece.getRootElement())) {
        int childId = ServiceManager.getProcessService().processIdFromUri(uri);
        Process child = ServiceManager.getProcessService().getById(childId);
        child.setParent(process);
        process.getChildren().add(child);
    }
    ServiceManager.getProcessService().save(process);
}

private Collection<URI> getAllChildrenRecursive(IncludedStructuralElement includedStructuralElement) {
    Collection<URI> result = new ArrayList<>();
    if (Objects.nonNull(includedStructuralElement.getLink())) {
        result.add(includedStructuralElement.getLink().getUri());
    }
    for (IncludedStructuralElement childIncludedStructuralElement : includedStructuralElement.getChildren()) {
        result.addAll(getAllChildrenRecursive(childIncludedStructuralElement));
    }
    return result;
}
```
At the same time, it is horribly inefficient in the default case, namely that the user does not change anything on the links. But it would be simple and understandable.
My opinion is that we are being paid by the DFG to find a solution to problems. Not to find three solutions to problems, to discuss which is the best and to delete two of them. Me, I was assigned to find a solution to this problem and I found a working solution. Unlike you, I have never dealt with this topic before and did not know your solution. But I found a solution that works. If you had gotten this job, you would have solved it differently. So what?
When dealing with a problem, one usually finds out in the end that it would have been better otherwise. In pedagogy, this is called _learning by doing_, and in industry it's a no-show development: A complete development, just for the purpose of learning how to do it the best, which you then throw away. That makes an expert an expert, that he or she has made many of these experiences and knows how to do it better. I'm not an expert on Hibernate, I freely admit that. But I want to emphasize once again that my solution works. It is highly efficient, even on many links, and most efficient in the standard case, that nothing is changed. And now it is so often refactored that is is quite understandable, I think. If you do not like it, just rewrite it. But I did my part, and: I did it in time.

Beacze:
You are forgetting about one important thing. You are not going to be the only person which is going to work on this code. Your code should be easy to maintain. It works but it is hard to understand for people which will come and extend software after you. I personally needed some time to understand what you wanted to achieve with this code and after that I was able to see how big code overhead you created here.

> If you do not like it, just rewrite it. 
I have definitely no time to forever rewrite and simplify code of other people.

Erikmitk:
I want to emphasise that this is a _team effort_ and not a bunch of single developers that happen wot work on a single code base. We should work together like that!

The mindset has to change from

> when I asked myself, how do I [do something], it was not so easy.

to

> This is not so easy. Let's ask the other ones for suggestions.

We would get different solution _before_ any implementation is done and can choose one that's best for our purposes. Now we are working backwards by being given a complicated solution and we'll find an easier to read and maintain one after the fact. That's completely backwards.

matthias-ronge:
I think I get paid to find solutions here, and I think, with a full degree in computational linguistics, plus computer science and years of software development experience, even years in kitodo, I have enough knowledge to come up with solutions. I do not see it as my job, in an algorithmically new problem, to ask around here _how would you do that_ and then do nothing for a week because I'm waiting for answers and a vote if I know I'm just have two days left to deliver the solution and implement two more solutions that need to build on that.
After I have an additional day left today, which was not clearly foreseen before, I can take the topic again. I will implement it again. I know it will be algorithmically more inefficient, and I'm afraid that will get us on our feet when we get newspaper hierarchies. But that will be a problem of other days.

PR #2560, opened 2019-05-21
Resolves Hibernate classpath dependencies

matthias-ronge:
Newer Hibernate version requires explicit import of the "hibernate-ehcache" dependency. This is based on major version 2 of ehcache, therefore downgrading the ehcache version. Ehcache version 2 requires old format configuration file, uses ehcache.xml from Kitodo 2.x Branch.

Advantage: When upgrading from 2.x existing ehcache configuration can continue to be used.

### DIFF
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -127,14 +127,27 @@
             <version>${hibernate.version}</version>
         </dependency>
         <dependency>
+         <groupId>org.hibernate</groupId>
+         <artifactId>hibernate-ehcache</artifactId>
+         <version>${hibernate.version}</version>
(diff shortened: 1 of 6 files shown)


henning-gerhardt:
Why do you downgrade ehCache from version 3 to version 2?

matthias-ronge:
Using Ehcache 3.0 and above with Hibernate-ehcache is not possible.

henning-gerhardt:
Why did this work not any more with the current configuration. So far as I can see from the logs ehcache is still used. Can you prove a test or something else which shows that ehcache is not used or working any more?

matthias-ronge:
No bug. Error was caused by missing a changed configuration.

PR #2566, opened 2019-05-23
Adds the ability to create hierarchical links manually

matthias-ronge:
1. Right-click on the structure where you want to add the hierarchical link.
capture_0001
2. In the window that appears, select default “link”.
capture_0002
3. Type the process ID, or a part of the process title, into the search box and click “search”.
capture_0003
4. Select process to add from the drop-down menu, type in order value, and click “apply”.
capture_0004

Follow-up pull request to #2550, [immediate diff](https://github.com/matthias-ronge/kitodo-production/compare/KIT-1070...matthias-ronge:KIT-1071)


Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -1157,13 +1157,13 @@ public URI getProcessURI(Process process) {
      * Returns a URI that identifies the process. The URI has the form
      * {@code mysql://?process.id=42}, where {@code 42} is the process ID.
      *
-     * @param id
+     * @param aLongerVariableNameForTheIdToSatisfyCodacy
      *            process ID for which a URI is to be formed that identifies it
      * @return a URI that identifies the process
      */
-    public URI getProcessURI(Integer id) {
+    public URI getProcessURI(Integer aLongerVariableNameForTheIdToSatisfyCodacy) {
```
henning-gerhardt: Really? What is wrong with `processId`?

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -1157,13 +1157,13 @@ public URI getProcessURI(Process process) {
      * Returns a URI that identifies the process. The URI has the form
      * {@code mysql://?process.id=42}, where {@code 42} is the process ID.
      *
-     * @param id
+     * @param aLongerVariableNameForTheIdToSatisfyCodacy
```
Erikmitk: WTF?!

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
Erikmitk: Or simply `identifier`?

matthias-ronge: Yes, `id` is too short and was rejected.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
matthias-ronge: If I have a supervisor who is a program with the intelligence of a monkey, then I have to submit to him.

Erikmitk: Come on… That’s a pretty stupid response and the code is not for Codacy.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
matthias-ronge: I have renamed it now in `processId` and hope that works. But if we no longer want to use `id` in the code, we should also rename all the corresponding fields, database rows, and methods, such as `getById()`. But maybe this rule is too strict. The use of `id` is widespread in Java.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
matthias-ronge: I'm a bit on pins and needles because I'm away for weeks now and I have to deal with such nonsense. I try to take it with humor, but it’s annoying.
In any case, if `id` is no longer allowed as a name in the future, you should also adapt the existing code accordingly. It is often used.

matthias-ronge:
Closed because the base is being reworked.

PR #2602, opened 2019-05-29
Update the hierarchical links in the database when saving

matthias-ronge:
When saving, the links in the database are updated based on the root
element.

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
@@ -21,6 +21,7 @@ import static org.junit.Assert.assertTrue;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
```diff
@@ -63,7 +66,10 @@
     private static final String firstProcess = "First process";
     private static final String processNotFound = "Process was not found in index!";
 
-    @BeforeClass
+    /**
+     * Sets up the database for the tests.
+     */
+    @Before
```
Beacze: If you really need to recreate the whole database after your test is executed, do it in your test. Don't enforce recreation of data for every test in this class.

matthias-ronge: Changed; but shouldn’t it be the case that the setup is prepared individually for each test? Otherwise you have dependencies between the tests, which should not be.

Beacze: Thanks you.

Tests which are introducing some dependencies have own setup inside them (look remove test). 90% of tests only reads data, without modification. We need to remember about time of execution for all tests. Set up database and index takes a lot of time.

matthias-ronge: @Beacze I delete the processes locally at the end of the test, but still the other tests are influenced, which should not be.
```
[ERROR]   ProcessServiceIT.shouldCountAllProcesses:97 Processes were not counted correctly! expected:<3> but was:<7>
[ERROR]   ProcessServiceIT.shouldGetAllProcesses:150 Not all processes were found in database! expected:<3> but was:<8>
[ERROR]   ProcessServiceIT.shouldRemoveProcess:165 Additional process was not inserted in database! expected:<[To Remove]> but was:<[HierarchyParent]>
```
Am I doing something wrong when deleting?

PR #2620, opened 2019-06-03
Validate metadata

IkramMaalej:


Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/physicalMetadata.xhtml
```diff
@@ -37,28 +37,29 @@
                                  styleClass="#{not item.editable or readOnly ? 'disabled' : ''}"
                                  style="width:100%;"
                                  validator="#{item.validatorQuery}"
-                                 rendered="#{item.input eq 'inputText'}">
-                        <p:ajax event="blur"/>
+                                 rendered="#{item.input eq 'inputText' || item.input eq 'spinner'}">
```
Kathrin-Huber: Why are you using the spinner elment on an input here? Does this work?

PR #2627, opened 2019-06-05
Fix deletion of processes with comments

oliver-stoehr:
Processes with comments could not be deleted, because they would be referenced by their comments.


Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -300,14 +300,16 @@ private void deleteProcess(Process processToDelete) {
             processToDelete.setProject(null);
             processToDelete.getTemplate().getProcesses().remove(processToDelete);
             processToDelete.setTemplate(null);
+            ServiceManager.getCommentService().removeAll(
```
Kathrin-Huber: Why don't you use 
`processToDelete.getComments()` here?

oliver-stoehr: It turns out that my fix was more complex than necessary. I removed this part.

PR #2641, opened 2019-06-12
Make template title unique for client

Beacze:


Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
```diff
@@ -153,12 +156,37 @@ public String save() {
...
+        String templateTitle = this.template.getTitle();
+        if (StringUtils.isNotBlank(templateTitle)) {
+            try {
+                long count = ServiceManager.getTemplateService().countTemplatesWithTitleAndClient(templateTitle,
+                    ServiceManager.getUserService().getSessionClientId());
+                if (count > 1) {
+                    Helper.setErrorMessage(INCOMPLETE_DATA, TITLE_USED);
+                    return false;
+                } else if (count == 1) {
+                    if (Objects.nonNull(this.template.getId())) {
```
Kathrin-Huber: If I open an existing template and rename it to an existing name, than count is 1 and template.id is not null, but afterwards I have two templates with the same name.

PR #2643, opened 2019-06-12
Remove singleton from WorkflowControllerService

Beacze:


solth:
Can you add some context or explanation why WorkflowControllerService is changed from singleton to normal class while the other service classes remain singletons?

Beacze:
I have assumed that racing problem comes from singleton nature of this service. Singleton is one and shared between all sessions, so when 2 or more people use this service in the same time they overwrite attributes of this service. I suspect that it causes overwrite of list with automatic tasks. Probably other solution would be to rewrite this service to make automaticTasks and tasksToFinish local variables instead of class attributes.

In data services the most important objects (e.g. task object, process object) are passed as parameters to the method, not attribute of the class. We can't exclude that there are also some attributes which could be a problematic, but until now it was not observed.

Kathrin-Huber:
And does it fix the problem of parallel automatic tasks? Can you write a test for it?

Beacze:
I don't know. I suspect that it could fix. I can't really reproduce the problem. Even if I have opened 3 sessions and clicked almost in the same time in every window I still can't say if it works, as my automatic tasks must be incorrect. First automatic task in row for each process changes to in progress, it doesn't get closed (so second automatic doesn't get executed). I get the same result when I execute this only in one session. So it must be definitely problem with my tasks.

I could try to write parallel tests but I don't know how automatic tasks need to be defined to really be executed, not hanging in progress.

Beacze:
Problem needs further consideration

PR #2646, opened 2019-06-14
Add new zoom controls to workflow editor

Erikmitk:
This PR adds a new [workflow editor build](https://github.com/Erikmitk/kitodo-workflow-editor/commits/master) with additional viewer functions. The frontend page of the workflow editor in Kitodo.Production is extended to make use of those features as well.

You can control the zoom, reset the diagram position and enter fullscreen mode for the editor itself to make room for intense focused workflow modelling!

Screenshot 2019-06-14 at 08 08 45


Erikmitk:
The build for openJDK has failed on tests regarding adding batches. Maybe worth investigating since my PR has zero influence on this part and it worked the second time without a problem.

PR #2649, opened 2019-06-17
Fix bug with indexingAll flag

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -792,29 +792,31 @@ public SortBuilder sortByCreationDate(SortOrder sortOrder) {
...
-        processDTO.setBatches(getBatchesForProcessDTO(jsonObject));
-        if (!related) {
-            convertRelatedJSONObjects(jsonObject, processDTO);
-        } else {
-            ProjectDTO projectDTO = new ProjectDTO();
-            projectDTO.setId(ProcessTypeField.PROJECT_ID.getIntValue(jsonObject));
-            projectDTO.setTitle(ProcessTypeField.PROJECT_TITLE.getStringValue(jsonObject));
-            projectDTO.setActive(ProcessTypeField.PROJECT_ACTIVE.getBooleanValue(jsonObject));
-            processDTO.setProject(projectDTO);
+        if (jsonObject.size() != 0) {
```
Beacze: It is better to use !jsonObject.isEmpty()

config-local-server/hibernate.cfg.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Beacze: Probably files from config-local-server got here by accident.

PR #2657, opened 2019-06-18
Change Index and Search rest clients to not singletons

Beacze:
It caused replace of the type in the rest client if 2 or more threads in the same time wanted to index something.


Beacze:
This solution creates too many http connections which are preventing tomcat container from stop. Other solution is developed.

PR #2668, opened 2019-06-20
Fix method for returning relative path

Beacze:
- remove creation of new not used String
- extract nested ternary operation

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Folder.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Folder.java
@@ -295,9 +295,10 @@ public class Folder extends BaseBean {
     @Transient
     public String getRelativePath() {
         int lastDelimiter = path.lastIndexOf(File.separatorChar);
-        return path.substring(lastDelimiter + 1).indexOf('*') > -1
-                ? lastDelimiter >= 0 ? path.substring(0, lastDelimiter) : ""
-                : path;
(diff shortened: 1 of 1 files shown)


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Folder.java
```diff
@@ -295,9 +295,8 @@ public Project getProject() {
     @Transient
     public String getRelativePath() {
         int lastDelimiter = path.lastIndexOf(File.separatorChar);
-        return path.substring(lastDelimiter + 1).indexOf('*') > -1
-                ? lastDelimiter >= 0 ? path.substring(0, lastDelimiter) : ""
-                : path;
+        String relativePath = lastDelimiter >= 0 ? path.substring(0, lastDelimiter) : "";
+        return path.indexOf('*', lastDelimiter + 1) > -1 ? relativePath : path;
```
Erikmitk: I know you haven't written this code in the first place, but if you have to touch and fix it can you please clean it up as well? Those short notations are awful to read; please write them out. I get it if I read it a couple of times but that should be obvious and understandable in the first place.

Also why is the relativePath itself created first? You don't even know if its needed yet before the if/else at the return statement is evaluated. So why bother computing it all the time? Originally, the value was only in the branch it was needed for.

Beacze: Changed.

PR #2671, opened 2019-06-20
Shorten methods in ExportXmlLog

Beacze:


Kitodo/src/main/java/org/kitodo/production/exporter/ExportXmlLog.java
```diff
@@ -378,6 +363,12 @@ private void prepareMetadataElements(List<Element> metadataElements, Map<String,
         }
     }
 
+    private Element getElement(String name, String text, Namespace xmlns) {
```
Erikmitk: You don't just *get* an element here, right? You create an entirely *new* one. The method should be called `createElement` or `newElement` or `initializeElement` or something in that direction to reflect that.

For me the same applies to the method below `getMetadataElements` should be called something else to be clear.

What do you think, @Beacze?

PR #2675, opened 2019-06-21
Prevent double save of folders added to project

Beacze:
Getting object before final save to index and database prevents that newly inserted folders are saved double.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProjectForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProjectForm.java
@@ -234,6 +234,7 @@ public class ProjectForm extends BaseForm {
                 }
                 this.deletedTemples = new ArrayList<>();
 
+                this.project = ServiceManager.getProjectService().getById(this.project.getId());
                 ServiceManager.getProjectService().save(this.project);
 
                 return projectListPath;


Erikmitk:
You update the project by getting the project with the same id? Why is it different after that?

Kitodo/src/main/java/org/kitodo/production/forms/ProjectForm.java
```diff
@@ -234,6 +234,7 @@ public String save() {
                 }
                 this.deletedTemples = new ArrayList<>();
 
+                this.project = ServiceManager.getProjectService().getById(this.project.getId());
```
Erikmitk: You update the `this.project` by getting the project with the same id as `this.project`? What is different after that?

Beacze: First project (this project contains already newly created folders) is saved, next added / removed templates are updated. Next I query this project to avoid that in the same session new folders will be saved again and again. I know it sounds weird, but it is how hibernate works.

Difference is that queried this.project has now folders already persisted and doesn't persist them again when the save to index (therefor update flag in database) is performed.

```
                ServiceManager.getProjectService().saveToDatabase(this.project);
                if (this.copyTemplates) {
                    for (Template template : this.baseProject.getTemplates()) {
                        template.getProjects().add(this.project);
                        this.project.getTemplates().add(template);
                    }
                    setCopyTemplates(false);
                }
                for (Template template : this.project.getTemplates()) {
                    ServiceManager.getTemplateService().saveToDatabase(template);
                }
                for (Template template : this.deletedTemples) {
                    ServiceManager.getTemplateService().saveToDatabase(template);
                }
                this.deletedTemples = new ArrayList<>();

                this.project = ServiceManager.getProjectService().getById(this.project.getId());
```

PR #2676, opened 2019-06-21
Fix icons in confirm dialog and add-new-element menu

solth:


Kitodo/src/main/webapp/WEB-INF/templates/base.xhtml
```diff
@@ -78,15 +78,17 @@
                         id="yesButton"
                         value="#{msgs.yes}"
                         type="button"
-                        styleClass="ui-confirmdialog-yes primary right"
+                        styleClass="ui-button-text-icon-right ui-confirmdialog-yes primary right"
```
oliver-stoehr: `ui-button-text-icon-right` is not needed as it's added automatically by `iconPos="right"`.

PR #2677, opened 2019-06-21
Fix some popup dialog issues

solth:
- fix styling of new comment popup dialog
- replace alert box with confirm dialog in project edit page

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -1190,11 +1190,11 @@ form#metadata div label,
     padding: 0;
 }
 
-.new-comment-box {
+#newCommentForm\:newCommentTextArea {
     width: 100%;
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
```diff
@@ -1179,17 +1179,22 @@ form#metadata div label,
...
 
+#newCommentForm h3 {
+    padding: 0;
+}
+
 .new-comment-box {
     width: 100%;
     background-color: var(--trans-blue);
     margin-bottom: 32px;
+    height: calc(var(--default-double-size) * 3) !important;
```
Erikmitk: What's overwriting the height of the element so you have to use `!important`? At least `.new-comment-box` sounds distinct enough so that it should not be an issue.

solth: Using the ID instead of the class did the trick.

PR #2679, opened 2019-06-23
Rename callback functions of metadata editor buttons

solth:


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java
```diff
@@ -89,7 +89,7 @@
      * This method is invoked if the user clicks on the add multi doc struc
      * submit btn command button.
      */
-    public void addMultiDocStrucSubmitBtnClick() {
+    public void addMultiDocStrucSubmit() {
```
Beacze: Is submit needed in the name?

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java
```diff
@@ -103,7 +103,7 @@ public void addMultiDocStrucSubmitBtnClick() {
      * This method is invoked if the user clicks on the add single doc struc
      * submit btn command button.
      */
-    public void addSingleDocStrucSubmitBtnClick() {
+    public void addSingleDocStrucSubmit() {
```
Beacze: Is submit needed in the name?

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java
solth: You are right, it's probably not needed!

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java
solth: You are right, it's probably not needed!

PR #2684, opened 2019-06-24
Import via adis interface and mods file upload (2.x)

chrizzor:
This PR adds the possibility to configure an import from adis interface or import mods directly from a file.


Kathrin-Huber:
Please solve Codacy warnings

PR #2692, opened 2019-06-26
Remove type property from Batch bean

Beacze:
@matthias-ronge please review this changes.


Kathrin-Huber:
Do we need the value of the type column for migration purposes?
What was the original use for it?

matthias-ronge:
Yes, we need the value for the migration. The value expresses what the batch describes. For all processes in a series-type batch, an overall parent-series process with links must be generated. For all processes in a batch of the type Newspaper, a total newspaper process and annual processes must be generated.

For the sake of completeness: A batch of the type logistics means that the acceptance or completion of a task of the type batch takes over or completes all the tasks of the same name of all other processes within the batch.

Kathrin-Huber:
So we cannot drop the column, right?

Beacze:
Should the migration for removing column be removed?

matthias-ronge:
That depends on how it will work during migration. Well, I would leave it in there. When migrating, care must be taken to ensure that the migration program that will do so will have to run before the database update.

Kathrin-Huber:
If we are not removing it from the database but from the bean, then there will be a hibernate missmatch error, right?

Beacze:
Actuall not ;) Missmatch appears when column is missing, when you have additional there is no problem.

Kathrin-Huber:
Ok, so I'm voting for removing the migration :)

PR #2693, opened 2019-06-26
Remove ArrayListMap - it was used by export which is removed

Beacze:
Depends on #2692


Erikmitk:
Is see that this PR is based upon https://github.com/kitodo/kitodo-production/pull/2692. Is that necessary or can this one be independent?

Beacze:
Previous one is necessary. This class is used only once in the removed export class.

PR #2695, opened 2019-06-28
Fix filters

oliver-stoehr:
The "filters" field in ProcessService and TaskService was applied to all users since those service classes are singletons.
Filters are now directly passed to the loadData method by PrimeFaces dataTables' filter functionality.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
@@ -1059,25 +1059,19 @@ public class FilterService extends SearchService<Filter, FilterDTO, FilterDAO> {
      *      a Map containing only one String value
      * @return
      *      the only Entry's value as java.lang.String
-     * @throws IllegalArgumentException
-     *      when the passed Map contains more than one Entry or when the Entry does not contain a value of type String
-     *
      */
-    String parsePrimeFacesFilter(Map filters) throws IllegalArgumentException {
-        if (Objects.nonNull(filters) && filters.entrySet().size() > 0) {
-            if (filters.entrySet().size() > 1) {
-                throw new IllegalArgumentException("Filter map contains to many entries (only 0 or 1 allowed)!");
+    String parsePrimeFacesFilter(Map filters) {
+        if (Objects.nonNull(filters) && filters.entrySet().size() == 1) {
+            Map.Entry filter = (Map.Entry) filters.entrySet().stream().findAny().get();
+            if (filter.getValue() instanceof String) {
+                return (String) filter.getValue();
             } else {
-                List<Object> filterList = new ArrayList<Object>(filters.values());
-                if (filterList.get(0) instanceof String) {
-                    return (String) filterList.get(0);
-                } else {
-                    throw new IllegalArgumentException("Given filter is not a String!");
-                }
+                logger.warn("Given filter is not a String. Ignoring filter.");
             }
-        } else {
-            return "";
+        } else if (Objects.nonNull(filters) && filters.entrySet().size() > 1) {
+            logger.warn("Filter map contains to many entries (only 0 or 1 allowed). Ignoring filter map.");
         }
+        return "";
     }
 
 
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
@@ -1047,4 +1047,38 @@ public class FilterService extends SearchService<Filter, FilterDTO, FilterDAO> {
         String[] strArray = parameter.split("-");
         return Integer.parseInt(strArray[1]);
     }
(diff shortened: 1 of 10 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
```diff
@@ -1047,4 +1047,38 @@ private Integer getTaskEnd(String parameter) {
...
+     *      when the passed Map contains more than one Entry or when the Entry does not contain a value of type String
+     *
+     */
+    String parsePrimeFacesFilter(Map filters) throws IllegalArgumentException {
+        if (Objects.nonNull(filters) && filters.entrySet().size() > 0) {
+            if (filters.entrySet().size() > 1) {
+                throw new IllegalArgumentException("Filter map contains to many entries (only 0 or 1 allowed)!");
+            } else {
+                List<Object> filterList = new ArrayList<Object>(filters.values());
+                if (filterList.get(0) instanceof String) {
```
solth: Won't this result in an `IndexOutOfBoundsException` if `filters` has no entries? Perhaps you should check a null check here?

Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
```diff
@@ -1047,4 +1047,38 @@ private Integer getTaskEnd(String parameter) {
...
+     * @param filters
+     *      a Map containing only one String value
+     * @return
+     *      the only Entry's value as java.lang.String
+     * @throws IllegalArgumentException
+     *      when the passed Map contains more than one Entry or when the Entry does not contain a value of type String
+     *
+     */
+    String parsePrimeFacesFilter(Map filters) throws IllegalArgumentException {
+        if (Objects.nonNull(filters) && filters.entrySet().size() > 0) {
```
Beacze: Maybe this method could be simplified:

```
    /**
     * Parses the passed Map and returns the value. The Map should contain only one or no Entry.
     * This method should be used to parse a Map passed by PrimeFaces' DataTable default filter functionality.
     * It passes currently only one Entry because our custom filter input passes the complete filter string to
     * the first input of PrimeFaces' default filter functionality.
     * This triggers PrimeFaces to pass the filter string and update the DataTable automatically.
     *
     * @param filters
     *      a Map containing only one String value
     * @return
     *      the only Entry's value as java.lang.String
     */
    String parsePrimeFacesFilter(Map filters) {
        if (Objects.nonNull(filters) && filters.entrySet().size() == 1) {
            Object filter = filters.entrySet().stream().findAny();
            if (filter instanceof String) {
                return (String) filter;
            }
        }
        return "";
    }
```

Like I see in cases where this method is used, exception is caught and next filter is set up empty String.

This change could decrease amount of code needed later. Not needed catch of exceptions and check if filter is null.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -254,6 +253,12 @@ public static LegacyMetadataTypeHelper getMetadataType(Process inProzess, String
     @Override
     public List<ProcessDTO> loadData(int first, int pageSize, String sortField,
             org.primefaces.model.SortOrder sortOrder, Map filters) throws DataException {
+        String filter = "";
```
Beacze: `String filter = ServiceManager.getFilterService().parsePrimeFacesFilter(filters);`

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -114,11 +113,14 @@ public static TaskService getInstance() {
      *
      * @return query to retrieve tasks for which the user eligible.
      */
-    private BoolQueryBuilder createUserTaskQuery() throws DataException {
+    private BoolQueryBuilder createUserTaskQuery(String filter) throws DataException {
         User user = ServiceManager.getUserService().getAuthenticatedUser();
 
         BoolQueryBuilder query = new BoolQueryBuilder();
         query.must(getQueryForTemplate(0));
+        if (Objects.isNull(filter)) {
```
Beacze: `ServiceManager.getFilterService().parsePrimeFacesFilter(filters);`

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -184,7 +186,13 @@ public Long countResults(Map filters) throws DataException {
     @Override
     public List<TaskDTO> loadData(int first, int pageSize, String sortField, SortOrder sortOrder, Map filters)
             throws DataException {
-        return findByQuery(createUserTaskQuery(), getSortBuilder(sortField, sortOrder), first, pageSize, false);
+        String filter = "";
```
Beacze: `ServiceManager.getFilterService().parsePrimeFacesFilter(filters);`

PR #2696, opened 2019-06-28
Remove not used DmsImportThread as it does nothing

Beacze:
This thread in the Kitodo 2.x was used for printing errors. Now this functionality doesn't exists anymore. Removed here: https://github.com/kitodo/kitodo-production/commit/e5f4360d932739738528ce8f3da1d3a369a3b3b6#diff-5b57fa457ffbfc2c7bd38e2628319cf4 . The only what this thread does now is waiting.

Default value exportWithoutTimeLimit should be true. False calls this old code:

> if this parameter is missing or 'false' the old export mechanism is used,
> otherwise there is no timelimit for export


henning-gerhardt:
From reading the code and current usage of this functionality this code can be removed.

PR #2697, opened 2019-06-28
Add possibility for migrating template tasks to workflow

Beacze:
Idea:

- user with authority for adding workflow can see icon for converting template to workflow (icon appears only when template doesn't have yet workflow)
- after user click icon, new workflow is created (workflow xml file is create and workflow is inserted to database as draft) and user is redirect to workflow edit page, there he can check if everything is ok and set workflow as active

kitodo
kitodo2


Kathrin-Huber:
I got an error on my first try, which forced my to set the dynamic flag in the mapping to false instead of "strict". But I cannot recreate this error.
I'm going to show it to the POs (y)

PR #2698, opened 2019-06-28
Fix randomly failing selenium test

Erikmitk:
I ran the selenium tests locally and discovered that the try-catch-block in these methods was always called. Without them the test `addUserTest()` would fail completely.

My guess is that the webdriver tries to act on these elements although they are not ready by a tiny fraction of a second. So it works sometimes but not always.

Therefore, `addUserTest()` seemingly fails at random. Subsequent tests will fail as well because the browser throws an alert to warn you about unsaved input when leaving the page.

Screenshot 2019-06-28 at 14 55 09

This change fixes the test for me reliably.*

*Now something different will break in Travis when I publish this PR. 😝


Erikmitk:
Well… that didn't work. :)

Beacze:
I'm afraid that there happens some JS magic which we can't prevent any way ;)

PR #2700, opened 2019-07-01
Trying to improve reliability of Selenium tests

Erikmitk:
This PR addresses two issues with our Selenium tests.

1. Chrome WebDriver version severely outdated

There were sometimes errors in the log that the browser window could not be closed. We used the hard-coded version 2.40 of the Chrome WebDriver. This one is so old that it's not even listed on the current [release site](https://sites.google.com/a/chromium.org/chromedriver/downloads) and does not reliably support the latest Chrome release. In Travis we use the latest stable Chrome version so it makes sense to use the corresponding ChromeDriver as well. I added a call to fetch the latest release from the site and use that to download the ChromeDriver for use. Per documentation we have to make a call with the latest major release version of the Browser so when Chrome 76 is release we have to adjust that and so on. It might just work with an older version but in case Selenium tests get flaky again this would be a good place to start checking for adjustments.

2. Randomly failing test

I noticed that the test `addUserTest` is seemingly randomly failing on Travis. I could reproduce this locally as well. There seems to be some sort of race-condition or general flakiness in the test implementation.

First of all I replaced some try-catch-Blocks that were run *all the time* anyway. The `StaleElementReferenceException` is a sign that we tried to get a hold of an element that was later replaced via JS or something in the DOM and is no longer the real deal. So instead of getting an exception anyway and making a second selection, I added a waiting call to make sure the elements are ready for the test.

Nevertheless the test sometimes failed to close the dialog that pops up for selection roles/clients/projects. The following tests would fail because they would try to navigate away from the screen but the browser throws up an alert warning that changes will be lost.

Screenshot 2019-06-28 at 14 55 09

My guess is that there's something going wrong with the dialog when a row is clicked and selected since the form has to be rebuild with updated information. That basically works but sometimes the browser tries to act too fast and gets a stale reference of the `close`-button and fails to click it. My fix is to wait a second before clicking and that seems to work. All my Travis-builds run on the first try.*

*I will get mad if this PR is not green on the first try though!

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/Browser.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/Browser.java
@@ -93,7 +93,7 @@ public class Browser {
 
         if (!driverFile.exists()) {
             logger.debug(driverFile.getAbsolutePath() + " does not exist, providing chrome driver now");
-            WebDriverProvider.provideChromeDriver(CHROME_DRIVER_VERSION, DOWNLOAD_DIR, DRIVER_DIR);
+            WebDriverProvider.provideChromeDriver(DOWNLOAD_DIR, DRIVER_DIR);
         }
 
         ChromeDriverService service = new ChromeDriverService.Builder()
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/Browser.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/Browser.java
@@ -45,7 +45,6 @@ public class Browser {
     private static RemoteWebDriver webDriver;
     private static Actions actions;
     private static final String GECKO_DRIVER_VERSION = "0.19.1";
-    private static final String CHROME_DRIVER_VERSION = "2.40";
     private static boolean onTravis = false;
     private static BrowserType browserType = BrowserType.CHROME;
     private static final String USER_DIR = System.getProperty("user.dir");
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
@@ -35,7 +35,7 @@ public class WebDriverProvider {
     // https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection
     // Please don't rely on the LATEST_RELEASE file without a version suffix.
     // It exists for backward compatibility only, and will be removed in the near future.
-    public static final String CHROME_DRIVER_LATEST_RELEASE_URL = "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_75";
+    private static final String CHROME_DRIVER_LATEST_RELEASE_URL = "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_75";
 
     private static UnArchiver zipUnArchiver = new ZipUnArchiver();
     private static UnArchiver tarGZipUnArchiver = new TarGZipUnArchiver();
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
@@ -171,7 +171,7 @@ public class WebDriverProvider {
 
     private static String fetchLatestStableChromeDriverVersion() {
 
-        String version = null;
+        String version = "";
         try {
 
             URL url = new URL(CHROME_DRIVER_LATEST_RELEASE_URL);
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
@@ -11,9 +11,12 @@
 
 package org.kitodo.selenium.testframework.helper;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.net.URL;
+import java.net.URLConnection;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.SystemUtils;
@@ -27,6 +30,12 @@ import org.kitodo.ExecutionPermission;
 public class WebDriverProvider {
 
     private static final Logger logger = LogManager.getLogger(WebDriverProvider.class);
+
+    // https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection
+    // Please don't rely on the LATEST_RELEASE file without a version suffix.
+    // It exists for backward compatibility only, and will be removed in the near future.
+    public static final String CHROME_DRIVER_LATEST_RELEASE_URL = "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_75";
+
     private static UnArchiver zipUnArchiver = new ZipUnArchiver();
     private static UnArchiver tarGZipUnArchiver = new TarGZipUnArchiver();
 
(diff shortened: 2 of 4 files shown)


Kitodo/src/test/java/org/kitodo/selenium/testframework/Browser.java
```diff
@@ -93,7 +93,7 @@ private static void provideChromeDriver() throws IOException {
 
         if (!driverFile.exists()) {
             logger.debug(driverFile.getAbsolutePath() + " does not exist, providing chrome driver now");
-            WebDriverProvider.provideChromeDriver(CHROME_DRIVER_VERSION, DOWNLOAD_DIR, DRIVER_DIR);
+            WebDriverProvider.provideChromeDriver(DOWNLOAD_DIR, DRIVER_DIR);
```
Beacze: If you remove usage of CHROME_DRIVER_VERSION, remove also declaration of this variable (line 48).

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/EditPage.java
```diff
@@ -35,6 +37,15 @@ void addRow(List<WebElement> tableRows, String title, WebElement dialog) {
         for (WebElement tableRow : tableRows) {
             if (Browser.getCellDataByRow(tableRow, 0).equals(title)) {
                 clickLinkOfTableRow(tableRow);
+
+                // Hold up a second…
+                try {
+                    Thread.sleep(1*1000);
+                } catch (InterruptedException e) {
+                    Logger logger = LogManager.getLogger(Browser.class);
+                    logger.error(e.getMessage());
```
Beacze: Stacktrace would be also useful:
logger.error(e.getMessage()), e;

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
```diff
@@ -159,4 +167,30 @@ private static void extractTarFileToFolder(File file, File destinationFolder) {
...
+            StringBuilder content = new StringBuilder();
+            content.append(bufferedReader.readLine());
+
+            bufferedReader.close();
+
+            version = content.toString();
+
+            logger.info("Latest Chrome Driver Release found: " + version);
+
+        } catch(Exception exception){
```
Beacze: Is there possibility to use here some more specific exception ot it must be catch all?

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
```diff
@@ -159,4 +168,32 @@ private static void extractTarFileToFolder(File file, File destinationFolder) {
...
+
+    private static String fetchLatestStableChromeDriverVersion() {
+
+        String version = null;
+        try {
+
+            URL url = new URL(CHROME_DRIVER_LATEST_RELEASE_URL);
+            URLConnection urlConnection = url.openConnection();
+
+            BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(urlConnection.getInputStream()));
```
henning-gerhardt: Please use a `try-with-resources` statement instead of open and closing the streams by yourself.

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/EditPage.java
Erikmitk: What for? That's a super rare special case that this exception gets thrown, right?

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
Erikmitk: Codacy already remarked on that one and I changed it. :)

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
```diff
@@ -159,4 +168,32 @@ private static void extractTarFileToFolder(File file, File destinationFolder) {
...
+
+            logger.info("Latest Chrome Driver Release found: " + version);
+
+        } catch (MalformedURLException exception) {
+            logger.error("Failed to fetch latest Chrome Driver Release");
+        } catch (IOException exception) {
+            logger.error("Failed to fetch latest Chrome Driver Release");
+        }
+
+        return version;
```
henning-gerhardt: You are returning a `null` value in error case or after fetching the Exception. But this result is not checked and / or direct used later in method `provideChromeDriver`.

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/EditPage.java
Beacze: But if it is thrown it would be nice to be able just click in console to get to the place where it was thrown :)

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
```diff
@@ -27,6 +31,12 @@
 public class WebDriverProvider {
 
     private static final Logger logger = LogManager.getLogger(WebDriverProvider.class);
+
+    // https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection
+    // Please don't rely on the LATEST_RELEASE file without a version suffix.
+    // It exists for backward compatibility only, and will be removed in the near future.
+    public static final String CHROME_DRIVER_LATEST_RELEASE_URL = "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_75";
```
henning-gerhardt: For current usage is it not needed to made this class constant public. Maybe you can move this URI to this one used place and did not define it as a class constant.

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
Beacze: Now just make them together in one call :D

> 'catch' branch identical to 'MalformedURLException' branch

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/EditPage.java
Erikmitk: Are you telling me you run the Selenium tests locally? :P

Beacze: Of course :) Quiet often to find out why certain specific selenium test fails. I do it for me and for other developers :D

Erikmitk:
Haha… That didn't work. Which is madness because Travis on my repo works now all the time with this. :(

Erikmitk:
Is somebody able to explain why this works 100% of the time in my repo? I rebuilt the same branch multiple times without an issue.

https://travis-ci.org/Erikmitk/kitodo-production/builds/552683657

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
Erikmitk: They could be handled in the same way since they are both IOExceptions in the end. But I kept them separate to distinguish between a wrongly URL and a failed download by changing the log message.

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
Erikmitk: There's no default value possible and if this part of the test suite crashes that's fine. There's no way for the tests to handle this scenario anyway so I think that's fine to possibly crash fast. The reason has been logged.

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
Erikmitk: I can add this but I don't think it really matters in lifecycle of the test run.

This part is executed once at the beginning of the selenium tests and then it's done. If this fails the test run is over anyway.

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/EditPage.java
```diff
@@ -35,6 +39,15 @@ void addRow(List<WebElement> tableRows, String title, WebElement dialog) {
         for (WebElement tableRow : tableRows) {
             if (Browser.getCellDataByRow(tableRow, 0).equals(title)) {
                 clickLinkOfTableRow(tableRow);
+
+                // Hold up for some seconds…
+                try {
+                    Thread.sleep(3 * 1000);
+                } catch (InterruptedException e) {
+                    Logger logger = LogManager.getLogger(Browser.class);
```
henning-gerhardt: Logger class name is wrong: It should be `EditPage.class` instead of `Browser.class`. Maybe you can create the logger instance like on all other places in the code and not introduce a new variant.

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
Erikmitk: I just run the tests with both `null` and `""` for the `version` and it doesn't make a difference. The tests output exceptions because there's no ChromeDriver to be instantiated due to the missing downloaded file. The test run does not crash but tries its best and fails.

I changed it to the empty string nonetheless since that's a better practice overall.

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/EditPage.java
Erikmitk: > Maybe you can create the logger instance like on all other places in the code and not introduce a new variant.

That's an unnecessary snarky comment.

PR #2704, opened 2019-07-01
Validation module

solth:
Activates existing Validation module in metadata editor.


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -275,19 +282,20 @@ public void validate() {
...
-            structurePanel.preserve();
-            ServiceManager.getProcessService().updateChildrenFromRootElement(process, workpiece.getRootElement());
+        metadataPanel.preserve();
+        structurePanel.preserve();
+        if (validate()) {
             try (OutputStream out = ServiceManager.getFileService().write(mainFileUri)) {
                 ServiceManager.getMetsService().save(workpiece, out);
+                return close();
+            } catch (IOException e) {
+                Helper.setErrorMessage(e.getLocalizedMessage(), logger, e);
```
Beacze: Is it passing of logger inside the error message not equal to log stuff? Is the second logger below needed?

PR #2717, opened 2019-07-03
Fix query for distinct values from ES index

solth:
Query for distinct values in `SearchService` would only return the first 10 elements of a type (for example `Task`) which wasn't noticed because most workflows right now have less than 10 tasks. This change sets the size of the result set to the number of corresponding objects in the database.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/base/SearchService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/base/SearchService.java
@@ -777,13 +777,16 @@ public abstract class SearchService<T extends BaseIndexedBean, S extends BaseDTO
      *            by which aggregation is going to be performed
      * @param sort
      *            asc true or false
+     * @param size
+     *            number of rows returned by query
      * @return sorted list of distinct values
      */
-    protected List<String> findDistinctValues(QueryBuilder query, String field, boolean sort) throws DataException {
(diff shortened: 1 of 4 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/base/SearchService.java
```diff
@@ -784,9 +784,15 @@ protected Double findSumAggregation(QueryBuilder query, String field) throws Dat
     protected List<String> findDistinctValues(QueryBuilder query, String field, boolean sort, long size) throws DataException {
         List<String> distinctValues = new ArrayList<>();
         try {
-            Aggregations jsonObject = searcher.aggregateDocuments(query,
-                AggregationBuilders.terms(field).field(field).order(Terms.Order.aggregation("_term", sort))
-                        .size(Math.toIntExact(size)));
+            Aggregations jsonObject;
```
Beacze: To increase readability I would write it in that way:

```
            TermsAggregationBuilder termsAggregation = AggregationBuilders.terms(field).field(field)
                    .order(Terms.Order.aggregation("_term", sort));
            if (size > 0) {
                termsAggregation.size(Math.toIntExact(size));
            }
            Aggregations jsonObject = searcher.aggregateDocuments(query, termsAggregation);
```

PR #2685, opened 2019-06-25
Add XSLT for export internal format to DFG Viewer

Beacze:
Script contains all obligatory fields for DFG Viewer. Names of kitodo metadata elements are based on ruleset https://github.com/kitodo/kitodo-production/blob/master/Kitodo/rulesets/subhh.xml .

@solth and @sebastian-meyer : it would be nice if you could leave here your reviews :)

PR #2719, opened 2019-07-04
Copy all nodes from kitodo xml to mets-mods xml

Beacze:
PR #2685 has added file for converting Kitodo Intern format to MODS, but without copying the whole METS structure. This PR copies all existing nodes and includes previously added kitodo-mods conversion.

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -561,6 +561,16 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.xmlunit</groupId>
+            <artifactId>xmlunit-core</artifactId>
+            <version>2.6.3</version>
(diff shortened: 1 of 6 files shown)


Kathrin-Huber:
Please write a test! The previous pr should have had a test, so the fix would not have been necessary.
Convert a file with the xslt and check it against an existing file, which should be created intellectually  (not automated)

Beacze:
Actually test would not give anything, I just didn't know that previous one needed to copy this nodes. 

I can't make a test as it is only transformation file. This transformation must be run first to give result (I tested it by online transformer), if I understood correctly, we are just providing this file so librarians can adjust it to own needs (every ruleset has own names of metadata). I have used name of tags from ruleset called DFG Viewer.

PR #2722, opened 2019-07-05
 Prevent loops in the workflow

Beacze:
There was possibility that task can have two or more incoming nodes (example in the test svg diagram), it was creating never ending loop on save of workflow (displayed message about redirection, but no redirection as it was forever iterating over the nodes).

### DIFF
--- a/Kitodo/src/main/resources/messages/errors_de.properties
+++ b/Kitodo/src/main/resources/messages/errors_de.properties
@@ -155,6 +155,7 @@ errorVolume=Bandnummer ist keine g\u00DCltige Zahl.
 
 # W
 workflowExceptionParallelBranch=Eine Aufgabe in paralleler Verzweigung kann keine zweite Aufgabe haben. Bitte entfernen Sie die Aufgabe nach der Aufgabe mit dem Namen ''{0}''.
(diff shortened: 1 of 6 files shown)


Kitodo/src/main/resources/messages/errors_en.properties
```diff
@@ -166,6 +166,7 @@ errorVolume=Volume number is not a valid number.
 
 # W
 workflowExceptionParallelBranch=Task in parallel branch can not have second task. Please remove task after task with name ''{0}''.
+workflowExceptionLoop=Task with name ''{0}'' has than one more incoming elements - probably workflow contains not allowed loop.
```
Kathrin-Huber: "has more than one"

Kitodo/src/main/resources/messages/errors_de.properties
```diff
@@ -166,6 +166,7 @@ errorVolume=Bandnummer ist keine g\u00DCltige Zahl.
 
 # W
 workflowExceptionParallelBranch=Eine Aufgabe in paralleler Verzweigung kann keine zweite Aufgabe haben. Bitte entfernen Sie die Aufgabe nach der Aufgabe mit dem Namen ''{0}''.
+workflowExceptionLoop=Der Workflowaufgabe mit dem Namen ''{0}'' hat mehr als ein eingehendes Element - wahrscheinlich enth\u00E4lt der Workflow eine nicht zul\u00E4ssige Schleife.
```
Kathrin-Huber: "Die Workflowaufgabe..."

PR #2723, opened 2019-07-05
Remove duplicate smLinks

oliver-stoehr:
Loading the meta.xml in the metadata editor will now check if the smLink is linking the same physical node like another smLink on a deeper level. All smLinks except the lowest for each physical element are not added to the workpiece.
Saving the metadata editor will remove the smLinks from the meta.xml.


Kathrin-Huber:
Please add a test, which is actually removing the smLinks from a testfile

PR #2733, opened 2019-07-10
Add children’s views to parents when exporting

matthias-ronge:
The feature to add the views from the children to the parents has been fixed and integrated into the export. The button for manually triggering the function has been removed.


Kitodo/src/main/java/org/kitodo/production/services/schema/SchemaService.java
```diff
@@ -85,6 +86,26 @@
...
+     *            included structural element on which the recursion is
+     *            performed
+     */
+    private void assignViewsFromChildrenRecursive(IncludedStructuralElement includedStructuralElement) {
+        List<IncludedStructuralElement> children = includedStructuralElement.getChildren();
+        if (!children.isEmpty()) {
+            for (IncludedStructuralElement child : children) {
+                assignViewsFromChildrenRecursive(child);
+            }
+            MetadataEditor.assignViewsFromChildren(includedStructuralElement);
```
Kathrin-Huber: this is also calling "assignViewsFromChildrenRecursive". I think there is something wrong here
Why are there two different methods for regular use and for recursive?

matthias-ronge: This method recursively iterates through all children, calling `assignViewsFromChildren()` for each child. The method in the class `MetadataEditor` then assigns the views of all their children to the respective level. This must happen at all levels, not just at the top level. No duplicate references are generated, they are filtered out. I do not see where that should produce a wrong result.

PR #2736, opened 2019-07-11
Some Javadoc clean-up

matthias-ronge:
Here, a bit of the Javadoc is tidied up. Initially, I just wanted to throw the `@author` references from the Javadoc, but then I came across a few things. It should be an improvement whatsoever, so I’ll put it.


Kitodo-API/src/main/java/org/kitodo/api/Metadata.java
```diff
@@ -30,9 +30,9 @@ public MdSec getDomain() {
     }
 
     /**
-     * Get the key o the metadata.
+     * Get the key o the meta-data.
```
Beacze: This change doesn't make sense - metadata is a correct word (https://whatis.techtarget.com/definition/metadata, https://www.wikiwand.com/en/Metadata)

Here something more: https://www.epublishing.com/blogs/1-the-epublishing-blog/post/143-using-metadata-for-profit
_Is it metadata or is it meta data? What is the difference between metadata versus meta data? I get this question all the time._ 

Important point meta-data is incorrect, so please don't fix something what doesn't need fix.

Kitodo-API/src/main/java/org/kitodo/api/rulesetreader/RulesetReaderInterface.java
```diff
@@ -33,45 +33,45 @@
...
 
     /**
-     * Gets the sorted visible metadata.
-     * 
+     * Gets the sorted visible meta-data.
+     *
      * @param rulesetRequestData
      *            The needed data for retrieval.
-     * @return a Collection of Metadata.
+     * @return a Collection of Meta-data.
```
Beacze: Here it is even more pointless - Collection of Metadata - means collection of Metadata class objects :)

Kathrin-Huber:
Why do you change metadata to meta-data?

oliver-stoehr:
According to New Oxford American Dictionary "metadata" is spelled _without_ a hyphen. Changing it to "meta-data" is not standard American English.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
```diff
@@ -102,7 +102,7 @@ public Batch(String title, Collection<? extends Process> processes) {
     }
 
     /**
-     * The function getTitle() returns the batch title. Using titles for batches
+     * Returns the batch title. Using titles for batches
```
Beacze: I see here inconsistency - here Returns, few line lower Return.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
```diff
@@ -102,7 +102,7 @@ public Batch(String title, Collection<? extends Process> processes) {
     }
 
     /**
-     * The function getTitle() returns the batch title. Using titles for batches
```
Beacze: Actually if we want to be so English correct - there is simple rule: _The definition of a sentence is a subject and verb._ So _The function getTitle() returns the batch title_ is correct and _Returns the batch title._ is grammatically incorrect. There is one exception when you can make sentence without subject - imperative, but imperative here would be _Return_, not _Returns_.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
```diff
@@ -112,7 +112,7 @@ public String getTitle() {
     }
 
     /**
-     * The method setTitle() can be used to set a batch title. This function is
+     * Can be used to set a batch title. This function is
```
Beacze: Word _Can_ can't be used as imperative so here it is always incorrect.

matthias-ronge:
I found [this blog post](http://tdan.com/the-meta-data-fiasco/18502) that (if I understand it correctly) considers the hyphenated form to be correct. I do not care what spelling we use, I would only like it if we made it consistent.
As for the first sentence of Javadoc comments, I oriented myself to [this how-to](https://ase.in.tum.de/paid.globalse.org/PAID/Misc-Documents/javadoc/howto.html). See the example under “Method descriptions begin with a verb phrase”.
These were my thoughts. As I said, we can do it differently, my goal here was to make it the same everywhere.

Beacze:
This blog post considers word metadata (meta-data) itself meaningless ;) I think we are not going to discuss here if metadata word makes sense or not.

_The terms meta-data and metadata need to be abandoned!  The cause has been lost!_

matthias-ronge:
I checked the website of the LOC and got <200 uses of “meta-data” while >63.000 uses of “metadata”, so obviously the latter is the correct one. I will change that.

matthias-ronge:
The Java runtime Javadoc (example taken from Javadoc of `String` class) also uses verb phrases for the first sentence:

Example

“Returns the char value …”, not “The function charAt() returns the char value…”. The screenshot is the overview where the first sentence of each method Javadoc is repeated, and this—by the way—is the cause why a Javadoc comment must have a sentence ending with a full stop. (the parameter descriptions don’t need to have) Just FYI

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
matthias-ronge: The `@return` is an instruction to the Javadoc generator. It translates to „Returns:“

Anmerkung 2019-07-12 121221

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
matthias-ronge: This is not an imperative, but a verb phrase. A verb phrase (VP) is the part of the sentence without the subject (without the noun phrase, NP). If you want to go deeper, see: [phrase structure grammar](https://en.wikipedia.org/wiki/Phrase_structure_grammar).

PR #2741, opened 2019-07-12
Clean-up Batch class

matthias-ronge:
The class `Batch` would benefit from a manual clean-up.

* improve Javadoc to better explain
* fix broken [contract](https://www.baeldung.com/java-equals-hashcode-contracts#hashcode) in `hashCode()`

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
@@ -29,25 +29,25 @@ import javax.persistence.Transient;
 import org.kitodo.data.database.persistence.BatchDAO;
 
 /**
- * The class Batch represents a user-definable, unordered collection of
- * processes that methods can be applied on in batch processing.
- *
- * @author Matthias Ronge &lt;matthias.ronge@zeutschel.de&gt;
+ * A user-definable, unordered collection of processes whose batch-type tasks
+ * can be taken over and completed with a single operator action. This depicts
+ * taking over and completing tasks when the tasks of multiple processes refer
+ * to the same physical object, for example, a multi-content box or a
+ * multi-journal binding unit.
  */
 @Entity
 @Table(name = "batch")
 public class Batch extends BaseIndexedBean {
 
     /**
-     * The field title holds the batch title. Using titles for batches is
-     * optional, the field may be null. If so, the id will be shown to the user
-     * instead.
+     * The batch title. Using titles for batches is optional, the field may be
+     * {@code null}. If so, the ID will be shown to the user instead.
      */
     @Column(name = "title")
     private String title;
 
     /**
-     * The field processes holds the processes that belong to the batch.
+     * Holds the processes that belong to the batch.
      */
     @ManyToMany(cascade = CascadeType.PERSIST)
     @JoinTable(name = "batch_x_process", joinColumns = {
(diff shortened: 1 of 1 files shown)


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Batch.java
```diff
@@ -29,81 +29,89 @@
...
     @JoinTable(name = "batch_x_process", joinColumns = {
             @JoinColumn(name = "batch_id", foreignKey = @ForeignKey(name = "FK_batch_x_process_batch_id")) }, inverseJoinColumns = {
             @JoinColumn(name = "process_id", foreignKey = @ForeignKey(name = "FK_batch_x_process_process_id")) })
     private List<Process> processes;
 
+    /**
+     * The display label of this batch.
+     *
+     * <p>
+     * Note: This is a strange hack that seems to be related to the lazy loading
```
Beacze: It is not a strange hack, if you would look into the method which creates label, you would know that label needs access to translation files. Translation files are accessible only in Core.

matthias-ronge: Nevertheless, this should not be cached in the bean without being in the database and in the index. Or do you like that?

Beacze: This field is transient - have you ever heard about transient fields / methods in bean classes? ;)

matthias-ronge: It works. What's unattractive about this design is that after loading the batch you have to make sure that a certain function has been called manually that composes the JSP output, just to be able to output the batch “label” via `toString()` in the JSP. It would be better to put the JSP output together in JSP, accessing the fields of the bean. In the JSP, we also have access to the translation files. Also, this will not work properly if you switch the language of the interface, or if a user using a different language in the interface loaded the batch first.

matthias-ronge: ServiceManager.getBatchService().createLabel(batch);
    selectItems.add(new SelectItem(batch, batch.toString(), null));

should be replaced to

    selectItems.add(
        new SelectItem(batch, ServiceManager.getBatchService().createLabel(batch), null)
    );

then we will no longer need a `@transient` field. I will add a separate pull request for this.

Beacze: It is not the only place where this toString() is used and you, as a person which has written it, should know it.

matthias-ronge: I took your note seriously and double checked my assumption. It's true that in the code, including JAR files, >1,500 uses of `Object.toString()` occur. It happens exactly five times in the codebase itself, and only in one place does it call on the class batch. No, unfortunately I do not fully remember every change I made to the code in the last seven years. But the relevant code found coincides with my memory. Please let me know if I missed something.

Please clarify whether this is a reason that this pull request is not an improvement on the state before and whether you agree or disagree with a merge.

Beacze: It is also used in frontend pages e.g. batchList.xhtml - selectManyListbox - batch object is there displayed (it uses implicitly toString()).

I find interesting that title can be empty. If I recall correctly when new batch is created, appears window with title to type and next save it. It looks very not intuitive to open dialog window for typing title and next allow to save it empty.

I agree with your changes except removing the label / commenting it as hack.

PR #2756, opened 2019-07-17
Delete a structure element and assign its pages to the parent element.

IkramMaalej:


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -114,11 +115,32 @@ void deleteSelectedStructure() {
...
+
         parent.getChildren().remove(selectedStructure.get());
+        this.dataEditor.setExpandTree(true);
         show();
+        this.dataEditor.setExpandTree(false);
         dataEditor.getGalleryPanel().updateStripes();
     }
 
+    private Collection<View> getAllSubViews(IncludedStructuralElement selectedStructure, Collection<View> views) {
+        if (Objects.nonNull(selectedStructure.getViews())) {
```
Kathrin-Huber:
```
        if (Objects.nonNull(selectedStructure.getViews())) {
            views.addAll(selectedStructure.getViews());
        }
        for (IncludedStructuralElement child : selectedStructure.getChildren()) {
            getAllSubViews(child, views);
          }
```
Should be enough

PR #2764, opened 2019-07-19
Prevent correction comments in parallel tasks

solth:
The option to create parallel tasks with the new workflow editor mostly renders the previous function to determine the one `currentTask` that is `INWORK` useless. 
Since now mutliple parallel tasks can be `INWORK` at the same time, the option to write a correction comment (which closes all tasks that come after a selected "correction task" in the workflow) needs to be deactivated if separate tasks are currently in work by different users.

The changes in this Pull Request perform this check and deactivate the switch for correction comments, adding a tooltip to the deactivated switch informing the user which other parallel task is in work by which user.

Additionaly, this Pull Request changes the way the MetadataEditor is opened. Instead of setting the ID of a process to load in the backend, the task for which the metadata is opened is passed as an URL parameter. This way, the Metadata Editor can be opened with appropriate URL parameters (e.g. `taskId=41` etc.) and does not rely on specific states in the backend, e.g. it becomes "bookmarkable".

Bildschirmfoto 2019-07-19 um 10 27 31

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/CommentForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/CommentForm.java
@@ -15,8 +15,10 @@ import java.text.MessageFormat;
 import java.text.ParseException;
 import java.util.Collections;
 import java.util.Date;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 import javax.enterprise.context.SessionScoped;
 import javax.faces.context.FacesContext;
@@ -29,6 +31,7 @@ import org.kitodo.data.database.beans.Comment;
 import org.kitodo.data.database.beans.Process;
 import org.kitodo.data.database.beans.Task;
 import org.kitodo.data.database.enums.CommentType;
+import org.kitodo.data.database.enums.TaskStatus;
 import org.kitodo.data.database.exceptions.DAOException;
 import org.kitodo.data.exceptions.DataException;
 import org.kitodo.production.enums.ObjectType;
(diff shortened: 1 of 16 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/CommentForm.java
```diff
@@ -238,67 +235,37 @@ public String solveProblemForAllBatchProcesses(Comment comment) {
...
-        this.process = process;
-    }
-
     /**
      * refresh the process in the session.
      *
      * @param process Object process to refresh
      */
-    public void refreshProcess(Process process) {
+    private void refreshProcess(Process process) {
```
Beacze: Is the refresh of process still needed if this form actually operates now on task object?

Kitodo/src/main/java/org/kitodo/production/forms/CommentForm.java
```diff
@@ -323,4 +290,40 @@ public void newComment() {
...
         setCorrectionComment(false);
     }
+
+    /**
+     * Check whether there are concurrent tasks in work or not.
+     *
+     * @return whether there are concurrent tasks in work or not
+     */
+    public boolean isConcurrentTaskInWork() {
+        return this.currentTask.getId() < 0 || !getConcurrentTasksInWork().isEmpty();
```
Beacze: Can id be smaller than 0?

Kitodo/src/main/java/org/kitodo/production/forms/CommentForm.java
solth: I think this is still required because `getAllComments` uses the task's process to convert all old comments (e.g. "WikiField") into the new format. But I would prefer if @IkramMaalej could clear this up.

Kitodo/src/main/java/org/kitodo/production/forms/CommentForm.java
solth: You are correct, this is some residue from an earlier version of the code, where I saved the current tasks ID instead of the whole task in the CommentForm. I removed the check.

Kitodo/src/main/java/org/kitodo/production/forms/CommentForm.java
IkramMaalej: This method is still required because we need to refresh the process after every reporting and solving a problem.

PR #2774, opened 2019-07-30
Allow a workpiece to have multiple root elements

matthias-ronge:
Prerequisite for storing multiple newspaper editions (or palimpsests) in one-and-the-same process.


Kathrin-Huber:
A Workpice is not allowed to have multiple root elements

Issue #2785, opened 2019-08-06
2.x: Thumbnail creation fails in metadata editor if TIFF contains big metadata fields


ponchofiesta:
We have a problem in the metadata editor in Kitodo.Production 2.2. Some of our TIFF master images have very big metadata fields filled by the scanner software ("Cruse"). When K.Prod tries to read them, it fails with this error:

```
ERROR 2019-06-25 14:53:50,707 (TiffInterpreter.java:read:148)                                                                                                                                                                                
        something went wrong during reading of image                                                                                                                                                                                         
java.lang.ClassCastException: [J cannot be cast to com.sun.media.imageioimpl.plugins.tiff.TIFFIFD                                                                                                                                            
        at com.sun.media.imageioimpl.plugins.tiff.TIFFImageMetadata.getIFDAsTree(TIFFImageMetadata.java:175)                                                                                                                                 
        at com.sun.media.imageioimpl.plugins.tiff.TIFFImageMetadata.getNativeTree(TIFFImageMetadata.java:205)                                                                                                                                
        at com.sun.media.imageioimpl.plugins.tiff.TIFFImageMetadata.getAsTree(TIFFImageMetadata.java:193)
        at com.sun.media.imageioimpl.plugins.tiff.TIFFImageReader.getImageMetadata(TIFFImageReader.java:831)
        at de.unigoettingen.sub.commons.contentlib.imagelib.TiffInterpreter.read(TiffInterpreter.java:141)
        at de.unigoettingen.sub.commons.contentlib.imagelib.TiffInterpreter.<init>(TiffInterpreter.java:99)
        at de.unigoettingen.sub.commons.contentlib.imagelib.ImageFileFormat.getInterpreter(ImageFileFormat.java:134)
        at de.unigoettingen.sub.commons.contentlib.imagelib.ImageFileFormat.getInterpreter(ImageFileFormat.java:117)
        at de.unigoettingen.sub.commons.contentlib.imagelib.ImageManager.init(ImageManager.java:126)
        at de.unigoettingen.sub.commons.contentlib.imagelib.ImageManager.<init>(ImageManager.java:99)
        at de.sub.goobi.metadaten.MetadatenImagesHelper.scaleFile(MetadatenImagesHelper.java:357)
        at de.sub.goobi.metadaten.Metadaten.BildErmitteln(Metadaten.java:1697)
        at de.sub.goobi.metadaten.Metadaten.BildBlaettern(Metadaten.java:1484)
[...]

PR #2795, opened 2019-08-12
#2785 Fix JAI ImageIO TIFF metadata loading by using no metadata for thumbnails

ponchofiesta:
To avoid exceptions in metadata loading as described in #2785 I added a simple method to create thumbnails without reading the metadata. This works for our images and should for everyone else too. The old method still exists for compatibility.

Tested with AdoptOpenJDK 8.


ponchofiesta:
Updated the commit to correctly create JPG (or fallback to PNG).

henning-gerhardt:
I have some issues with your changes:
- remove functionality of rotating images inside meta data editor
- remove support of external Goobi / Kitodo content server usage
- left over of unused code through out commenting this single function call (method should be removed instead)
- fixing code on the wrong place: your change or needed change belongs to [Kitodo.ContentServer](https://github.com/kitodo/kitodo-contentserver) library as a fall back solution if needed

As already written in issue #2785 your image data is corrupted so fixing the image data would result in not changing "working" code beside the fact that the error handling can be improved.

ponchofiesta:
No problem. I'll close the PR. We will use this fix until we find a solution for the wrong metadata tags. Thanks

PR #2822, opened 2019-09-03
Include migration tab

Kathrin-Huber:


Kathrin-Huber:
> 
> 
> For a first step, it looks okay, but I think in a second pull request, the migration logic needs to be supplemented:
> 
>     * Execution as a long-running task in the task manager. Even if conversion for one process only takes 1 sec, for 150,000 processes it takes almost 2 days and there is no feedback yet. Also logging.
> 
>     * First, inspection, which processes need migration (for example, after a crash), then apply only to processes that need migration.

Yes of course! The Method is only a placeholder for future development!

PR #2823, opened 2019-09-04
Add MODS XML SchemaConverter module

solth:
- add new `SchemaConverterInterface` implementation for MODS-XML-Import
- move MODS to Kitodo conversion parts from `SRUImport` module to new SchemaConverter
- add new `MetadataFormat` and `FileFormat` enums and `DataRecord` class


Kitodo/src/main/java/org/kitodo/production/services/dataformat/MetsService.java
```diff
@@ -85,6 +95,25 @@ public Workpiece loadWorkpiece(URI uri) throws IOException {
...
+     * Create and return Workpiece from given Document 'document'.
+     *
+     * @param document Document from which a Workpiece is created
+     * @return Workpiece created from given Document
+     * @throws TransformerException
+     *          thrown when XML transformation fails
+     * @throws IOException
+     *          thrown when unable to read inputStream
+     */
+    public Workpiece loadWorkpiece(Document document) throws TransformerException, IOException {
```
henning-gerhardt: In this method a lot of stream objects are created / opened but not explicit closed. This can result in open stream handles which are only will closed if the application will be stopped. Maybe use of try-with-resources statement could improve automatic closing of stream objects.

solth: I commited this method by accident, it wasn't supposed to be included in this Pull Request, therefore I removed it for now.

Kitodo-API/src/main/java/org/kitodo/config/KitodoConfig.java
```diff
@@ -89,6 +89,7 @@ public static String getParameter(String key) {
         try {
             return getConfig().getString(key);
         } catch (NoSuchElementException e) {
+            e.printStackTrace();
```
henning-gerhardt: Use logger instead of `printStackTrace()` calls.

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
```diff
@@ -35,6 +35,10 @@
     private OPACConfig() {
     }
 
+    public static String getConfigValue(String catalogName, String parameter) {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
```diff
@@ -135,4 +139,8 @@ private static XMLConfiguration getConfig() {
         config.setReloadingStrategy(new FileChangedReloadingStrategy());
         return config;
     }
+
+    public static String getParentElementXPath(String catalogName) {
```
henning-gerhardt: Public method without JavaDoc

Kitodo-MODSXML-SchemaConverter/pom.xml
```diff
@@ -1,54 +1,44 @@
...
+            <version>2.0.6</version>
+            <scope>compile</scope>
         </dependency>
         <dependency>
-            <groupId>org.junit.vintage</groupId>
-            <artifactId>junit-vintage-engine</artifactId>
-            <version>${junit.version}</version>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
```
henning-gerhardt: Why did you use version 4.12 if other modules are using 5.4.1 of junit? In the global pom file is a maven variable defined for junit.

Kitodo-MODSXML-SchemaConverter/src/main/java/org/kitodo/modsxmlschemaconverter/ModsXMLSchemaConverter.java
```diff
@@ -0,0 +1,138 @@
...
+                DataRecord resultRecord = new DataRecord();
+                resultRecord.setOriginalData(transformXmlByXslt(xmlString, fileStream));
+                resultRecord.setFileFormat(targetFileFormat);
+                resultRecord.setMetadataFormat(targetMetadataFormat);
+                return resultRecord;
+            } catch (IOException e) {
+                logger.error(e.getMessage());
+            }
+        }
+        return null;
```
henning-gerhardt: I'm not a fan of returning `null` values. Is there no other way to avoid this? Inside the if condition before you are catching all `IOException` and logging this. After this the method is returning `null`. I don't think this is a good way / behaviour. Last remark to this: In the JAvaDoc of this method is no information that this method returns a `null` value.

Kitodo-SRU-Import/pom.xml
```diff
@@ -40,6 +40,22 @@
             <version>2.0.6</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
```
henning-gerhardt: Same here: Why you are using junit 4.12 instead of 5.4.1 and more not using the maven variable for this?

Kitodo/src/main/java/org/kitodo/production/helper/TempProcess.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * This class is used during import.
+ */
+public class TempProcess {
+
+    private Workpiece workpiece;
+
+    private Process process;
+
+    public TempProcess(Process process, Workpiece workpiece) {
```
henning-gerhardt: Public method without JavaDoc.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -11,45 +11,33 @@
 
 package org.kitodo.production.services.data;
 
-import java.io.File;
-import java.io.FileOutputStream;
+import com.sun.jersey.api.NotFoundException;
```
henning-gerhardt: An other `NotFoundException` class? One from Jersey and the previous one from Spring.

Kitodo-MODSXML-SchemaConverter/src/main/java/org/kitodo/modsxmlschemaconverter/ModsXMLSchemaConverter.java
```diff
@@ -0,0 +1,138 @@
...
+import org.kitodo.api.schemaconverter.MetadataFormat;
+import org.kitodo.api.schemaconverter.SchemaConverterInterface;
+import org.kitodo.exceptions.ConfigException;
+import org.xml.sax.InputSource;
+
+public class ModsXMLSchemaConverter implements SchemaConverterInterface {
+
+    private static final Logger logger = LogManager.getLogger(ModsXMLSchemaConverter.class);
+
+    private static final String XSLT_FILEPATH = "/xslt/mods2kitodo.xsl";
```
henning-gerhardt: Is this hard coded value intent? Later this value is used to retrieve a file from class path.

Kitodo-MODSXML-SchemaConverter/src/main/java/org/kitodo/modsxmlschemaconverter/ModsXMLSchemaConverter.java
```diff
@@ -0,0 +1,138 @@
...
+                && supportsSourceFileFormat(record.getFileFormat())
+                && supportsTargetMetadataFormat(targetMetadataFormat)
+                && supportsTargetFileFormat(targetFileFormat))) {
+            throw new UnknownFormatConversionException("Unable to convert from " + record.getMetadataFormat() + "/"
+                    + record.getFileFormat() + " to " + targetMetadataFormat + "/" + targetFileFormat + "!");
+        }
+
+        if (record.getOriginalData() instanceof InputStream) {
+            try (InputStream inputStream = (InputStream)record.getOriginalData();
+                 InputStream fileStream = getClass().getResourceAsStream(XSLT_FILEPATH)) {
```
henning-gerhardt: Did I interpret this correct, that this module can only handle one xsl for all possible transformation? If I'm correct then every catalogue which is using mods encoded data can or must relay on one transformation file? Why not add this file as a parameter to the method and let the usage of this module decide which transformation file is used.

This is similar to the MODS catalogue usage in 2.x but there you have the possibility to define a transformation file per catalogue and this is heavy used at SLUB.

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
solth: Is the missing JavaDoc a bigger problem than the method not being used at all? ;)

Kitodo-SRU-Import/pom.xml
```diff
@@ -40,6 +40,22 @@
...
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>${junit.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.xebialabs.restito</groupId>
+            <artifactId>restito</artifactId>
+            <version>0.9.3</version>
```
Kathrin-Huber:
```suggestion
            <version>0.9.3</version>
            <scope>test</scope>
```

Kitodo-MODSXML-SchemaConverter/pom.xml
```diff
@@ -60,16 +50,22 @@
...
                 <configuration>
+                    <archive>
+                        <manifest>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+                        </manifest>
+                        <manifestEntries>
+                            <Implementation-Build-Date>${maven.build.timestamp}</Implementation-Build-Date>
+                        </manifestEntries>
+                    </archive>
```
Kathrin-Huber: why do you add the timstamp to the manifest? Is it used by the serviceloader?

Kitodo-MODSXML-SchemaConverter/pom.xml
```diff
@@ -1,49 +1,39 @@
...
-    <organization>
-        <name>kitodo.org</name>
-        <url>http://kitodo.org</url>
-    </organization>
-    <name>Kitodo - Schema Converter</name>
-    <url>https://github.com/kitodo/kitodo-production</url>
 
-    <properties>
-        <maxAllowedViolations>4</maxAllowedViolations>
-    </properties>
```
Kathrin-Huber: nice ;)

Kitodo-MODSXML-SchemaConverter/pom.xml
solth: Removed.

Kitodo-SRU-Import/pom.xml
solth: Added.

PR #2824, opened 2019-09-06
Examine JSONArray content and treat it accordingly

matthias-ronge:
The previous code assumed JSON array only to contain JSON objects, which may not be. This will check contents of JSON array entries and handles them accordingly.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -2217,7 +2217,6 @@ public class ProcessService extends ProjectSearchService<Process, ProcessDTO, Pr
 
     private Map<String, Object> iterateOverJsonObject(JSONObject xmlJSONObject) {
         Iterator<String> keys = xmlJSONObject.keys();
-
         Map<String, Object> json = new HashMap<>();
         while (keys.hasNext()) {
             String key = keys.next();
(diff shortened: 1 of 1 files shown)


matthias-ronge:
By the way, it does not reveal what this code does. It looks like it’s going to produce just that as output, which it gets as input. Is that even necessary?

matthias-ronge:
Unfortunately, no. It can happen that a JSON array contains strings. The previous version made it imperative that a JSON array only contain JSON objects. This fails with some METS files. I have not added unnecessary checks.

Beacze:
It can't happend. If it really happens for some object, it means that Json object for this type is built incorrectly. If it is so this part of the code should be fixed, not the one which reads this data. 

Could you tell me for which type you have encountered this problem?

matthias-ronge:
The error occurs with an `agent` with a repeated `note` element. Example:

    <metsHdr CREATEDATE="2019-09-02T11:03:56.453+02:00" LASTMODDATE="2019-09-06T13:36:12.731+02:00">
        <agent ROLE="OTHER" OTHERROLE="PROGRAMMER" TYPE="INDIVIDUAL">
            <name>Ronge, Matthias</name>
            <note>Matthaeum scripsit hoc</note>
            <note>Matthaeum scripsit hoc</note>
        </agent>
    </metsHdr>

matthias-ronge:
What will happen to this now? @Beacze will you fix it your way? Or should this bugfix be merged?

matthias-ronge:
@Kathrin-Huber Could you please clarify with Beatricze how to continue here? In my understanding, the JSON is correct and my solution as well. If not, @Beacze would need to provide an example of what the JSON _should_ actually look like in the case (and maybe you can tell me where it's generated).

PR #2850, opened 2019-09-24
Scroll to corresponding TreeNode when selecting a page in gallery

solth:


Erikmitk:
I don't know why but the PR is currently not visible in Codacy and therefore nothing is getting checked.

solth:
> I don't know why but the PR is currently not visible in Codacy and therefore nothing is getting checked.

It seems to work now!

PR #2853, opened 2019-09-26
Fix metadata editor rights

solth:
- evaluate read and write rights in metadata editor gallery
- fix `readyOnly` right evaluation in metadata editor structure trees


Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/gallery.xhtml
```diff
@@ -19,6 +19,9 @@
         xmlns:a="http://xmlns.jcp.org/jsf/passthrough">
     <p:panel id="galleryWrapperPanel" styleClass="wrapperPanel">
 
+        <!--@elvariable id="mayWrite" type="boolean"-->
+        <ui:param name="mayWrite" value="#{SecurityAccessController.hasAuthorityToEditProcessImages()}"/>
```
Kathrin-Huber: maybe call it
```
editProcessImage
editImageAuthority
```

The one most fitting the other params in the xhtml would be:
`editImages`

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/structure.xhtml
```diff
@@ -27,12 +27,11 @@
                       styleClass="ui-widget-content columnHeadingWrapper">#{msgs.logical} #{msgs.structure}</h:panelGroup>
         <p:tree id="logicalTree"
                 value="#{DataEditorForm.structurePanel.logicalTree}"
-                disabled="#{readOnly}"
                 var="logicalNode"
                 selectionMode="single"
                 selection="#{DataEditorForm.structurePanel.selectedLogicalNode}"
-                draggable="true"
-                droppable="true"
+                draggable="#{not readOnly}"
```
Kathrin-Huber: So here it is "readOnly" Is there a difference between readOnly and editImages?
Is it possible, that i am allowed to edit metadate but NOT images? is changing the order of images a "image edit" or a "metadata edit"?

solth: Yes, all three columns - structure, metadata and gallery - have separate read and write rights. So there are six different right combinations for using the metadata editor. "image edit" is moving images in the gallery (third column) via drag'n'drop, "metadata edit" is changing metadata field values in the second column.

PR #2899, opened 2019-09-27
Fix vertical position of loading screen key animation

solth:


solth:
@Kathrin-Huber could you review this Pull Request?

PR #2902, opened 2019-09-30
Create template after workflow

Kathrin-Huber:
After the workflow is created from the aggregated processes (see #2828 and #2833) a new template needs to be created with the workflow/ruleset/docket contained in the aggregated processes.
If a template already exists (which can only happen if you use an existing workflow) you can choose between using the existing template and creating a new one.


Kitodo/src/main/webapp/WEB-INF/templates/includes/system/confirmWorkflowPopup.xhtml
```diff
@@ -21,17 +21,16 @@
...
                   dynamic="true">
-            <h:form id="confirmWokflowForm">
-                <p:panelGrid id="passwordGrid" columns="1" layout="grid">
+            <h:form id="confirmWorkflowForm">
+                <p:panelGrid columns="1" layout="grid">
                     <p:row>
 
                         <div>
                             <p:commandButton id="useExistingWorkflow"
-                                             value="#{msgs.cancel}"
```
solth: Don't we still need a `cancel` button in the dialog if we decide to neither use an existing workflow nor create a new one but really just want to cancel the migration process and close the dialog? We don't have "close"-buttons on our dialogs after all.

Kitodo/src/main/java/org/kitodo/production/migration/TemplateComparator.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Objects;
+
+import org.kitodo.data.database.beans.Template;
+
+public class TemplateComparator implements Comparator<Template> {
+
+    @Override
+    public int compare(Template firstTemplate, Template secondTemplate) {
+        if (Objects.isNull(firstTemplate) || Objects.isNull(secondTemplate)) {
+            return 1;
```
solth: This `compare` method will return `1` (e.g. "unequal") if both templates are `null`. Is that intended?

Kitodo/src/main/java/org/kitodo/production/forms/MigrationForm.java
```diff
@@ -282,6 +282,147 @@ public String createNewWorkflow() {
             return this.stayOnCurrentPage;
         }
 
-        return MessageFormat.format(REDIRECT_PATH, "workflowEdit") + "&id=" + workflowToUse.getId();
+        return MessageFormat.format(REDIRECT_PATH, "workflowEdit") + "&id=" + workflowToUse.getId() + "&migration=true";
+    }
+
+    /**
+     * When the navigation to the migration form is coming from a workflow
+     * creation the URL contains an WorkflowId.
```
solth:
```suggestion
     * creation the URL contains a WorkflowId.
```

Kitodo/src/main/java/org/kitodo/production/migration/TemplateComparator.java
Kathrin-Huber: Yes, null-templates are not allowed in the matching context

Kitodo/src/main/webapp/WEB-INF/templates/includes/system/confirmWorkflowPopup.xhtml
Kathrin-Huber: added

Kitodo/src/main/java/org/kitodo/production/migration/TemplateComparator.java
solth: That may be true, but a check comparing `null` with `null` should normally return `true`, shouldn't it? That the templates aren't `null` - because that's not allowed here - should be checked somewhere else.

PR #2905, opened 2019-10-01
Upgrade flyway version from 5.1.4 to 6.0.4

henning-gerhardt:


Kathrin-Huber:
I'd rather not update flyway.

henning-gerhardt:
This is bad, as flyway 6.0.x is needed for migrating database on a MariaDB 10.4.x server :-(

PR #2906, opened 2019-10-02
Index Process when comment is saved

Kathrin-Huber:
Search also for commentmessages when searching in Searchbar


Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/index/type/enums/ProcessTypeField.java
```diff
@@ -33,6 +33,7 @@
     RULESET("ruleset"),
     BATCHES("batches"),
     COMMENTS("comments"),
+    COMMENTS_MESSAGE("comments.message"),
```
Beacze: I think that in that case you need to make something like comments.id and comments.message. It is probably like JSON standard of ES needs it.

Kathrin-Huber:
Changes were only requested for the searchbar for now. Making it searchable in the extended search should happen in a separat PR

PR #2921, opened 2019-10-10
Fix migration bug

matthias-ronge:
Production 2.x databases can not be migrated if there are different properties (workpiece properties, templates properties ...) that have the same ID since they are merged into one table. This causes the error "Duplicate entry 'nnn' for key 'PRIMARY'". To fix this, the IDs are recalculated so that they are unique.

Previous migration (on test system) does not have to be redone. If migration succeeded, the problem in the data was not given. This fixes the problem only for database dump, where different properties types with same ID occur.


Kathrin-Huber:
The problem is an empty benutzereigenschaften table. I would rather add something to this table, than changing the flywayscripts

PR #2925, opened 2019-10-14
Fix changing language

IkramMaalej:


Kitodo/src/main/java/org/kitodo/production/forms/LanguageForm.java
```diff
@@ -181,7 +181,7 @@ public void switchLanguage(String langCodeCombined) throws IOException {
         context.getExternalContext().getSessionMap().put(SESSION_LOCALE_FIELD_ID, locale);
         // Reload current page to make language change effective
         context.getExternalContext()
-                .redirect(((HttpServletRequest) context.getExternalContext().getRequest()).getRequestURI());
+                .redirect(((HttpServletRequest) context.getExternalContext().getRequest()).getHeader("referer"));
```
henning-gerhardt: This will introduce a easy to exploit security issue as the referer information can contain any kind of string which must not be a valid URI nor a URI leading to the application.

solth: Perhaps you could extract the `referer` header and validate it before using it for the redirect?

IkramMaalej: I didn't find another method to get the request URI including parameters. Do you have any better solution?

henning-gerhardt: You can get the requested URI by `getRequestURI()` and potential parameters by `getQueryString()`. Calling `getQueryString()` can return a empty / null value if there is no query string. This both values can be connected to get the URI with the full query.

IkramMaalej: I have already tried `getQueryString()` but it returns always `null` even if there are multiple parameters in the request.

henning-gerhardt: My fault: `getQueryString()` did only return the query string of current request and not the former request or the request before you clicked on language change. I see there at least two solutions:

1. Store the query inside a user session specific object as long as the user is logged in and access them on request. This object must be updated on every request an user is doing as long as he is working inside the application.
2. Use the header `referer` information as you verified that the data inside `referer` is safe to use. Safe should be at least a valid URI, URI pointing to running installation including correct path and port. Validation of the referer information should be confirmed inside a proper testing suite including some bad referer information.

IkramMaalej: What do you think about this idea?

solth: If you only use `stayOnCurrentPage` once in this class, you don't need to save this variable but can call `context.getExternalContext().redirect(null);` directly instead of `context.getExternalContext().redirect(stayOnCurrentPage);`, can't you?

Kitodo/src/main/java/org/kitodo/production/forms/LanguageForm.java
```diff
@@ -50,6 +49,7 @@
 
     private static final String SESSION_LOCALE_FIELD_ID = "lang";
     private static final Logger logger = LogManager.getLogger(LanguageForm.class);
+    protected String stayOnCurrentPage = null;
```
solth:
```suggestion
```

Kitodo/src/main/java/org/kitodo/production/forms/LanguageForm.java
```diff
@@ -180,8 +180,7 @@ public void switchLanguage(String langCodeCombined) throws IOException {
         context.getViewRoot().setLocale(locale);
         context.getExternalContext().getSessionMap().put(SESSION_LOCALE_FIELD_ID, locale);
         // Reload current page to make language change effective
-        context.getExternalContext()
-                .redirect(((HttpServletRequest) context.getExternalContext().getRequest()).getRequestURI());
+        context.getExternalContext().redirect(stayOnCurrentPage);
```
solth:
```suggestion
        context.getExternalContext().redirect(null);
```

PR #2928, opened 2019-10-15
Post login checks

solth:
These changes add a check after login that tests wether the contents of the Elastic Search index are up to date with those in the database. If the two data sources are not synchronous, and the user does _not_ have the authority to manipulate and update the index, he is instead shown a dialog informing him that the system needs mainanence. 

Bildschirmfoto 2019-10-15 um 09 21 36

If the index is currently being updated, the user is instead shown a message informing him that the system is under maintanence and that he should try again later.

Bildschirmfoto 2019-10-15 um 09 22 59

If the index is out of date the user _does_ indeed have the authority to update the index, he is instead redirected to the indexing page after login and a message is displayed informing him that the index needs to be updated.

Bildschirmfoto 2019-10-11 um 16 40 13

These changes also move the client select dialog from the desktop page to the post login checks, so that a user with multiple clients is _always_ presented with the client selection dialog, even when directly calling a deep link via URL before login, e.g. a page inside the Kitodo system that is normally only accessible after login. 

Before these changes, a user opening such a URL before login would have been forwarded to that page after successfully login into the system, effectively bypassing the client selection dialog (since that was only called on the default landing page `desktop.jsf`) 

Bildschirmfoto 2019-10-10 um 10 15 10

Additionally the `IndexingForm` has been refactored and program logic has been moved to a new class `IndexingService`.


Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
```diff
@@ -0,0 +1,567 @@
...
+                Helper.setErrorMessage("Mapping not found!");
+                return "";
+            }
+        } catch (IOException e) {
+            Helper.setErrorMessage(e.getLocalizedMessage(), logger, e);
+            return "";
+        }
+    }
+
+    private static IndexRestClient indexRestClient = IndexRestClient.getInstance();
```
henning-gerhardt: Can you move this on top of the class?

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
```diff
@@ -0,0 +1,567 @@
...
+        for (Map.Entry<ObjectType, List<IndexWorker>> workerEntry : indexWorkers.entrySet()) {
+            List<IndexWorker> indexWorkerList = workerEntry.getValue();
+            for (IndexWorker worker : indexWorkerList) {
+                worker.setIndexAllObjects(false);
+            }
+        }
+        startAllIndexing(pushContext);
+    }
+
+    class IndexAllThread extends Thread {
```
henning-gerhardt: Should inner classes not avoided?

IkramMaalej:
When I click on  "Wiederholen" button, the two buttons "Wiederholen" and "Logout" are redoubled.

Screenshot 2019-10-15 at 10 45 41

IkramMaalej:
As an "Admin" user, the client select dialog is not displayed even after the correction of index.

Kitodo/src/main/java/org/kitodo/production/forms/LoginForm.java
```diff
@@ -11,21 +11,30 @@
...
+import org.primefaces.PrimeFaces;
 
 @Named("LoginForm")
 @SessionScoped
 public class LoginForm implements Serializable {
     private User loggedUser;
-    private boolean alreadyLoggedIn = false;
     private boolean firstVisit = true;
+    private static final String INDEXING_PAGE = "system.jsf?tabIndex=2";
+    private static final String DESKTOP_VIEW = "desktop.jsf";
```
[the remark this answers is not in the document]
solth: Since these two are constants the better way is to declare them at the beginning of the class instead of inserting some static Strings into the middle of the code.

Kitodo/src/main/java/org/kitodo/production/security/CustomLoginSuccessHandler.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
+import org.springframework.security.web.savedrequest.DefaultSavedRequest;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CustomLoginSuccessHandler implements AuthenticationSuccessHandler {
+
+    private static final Logger logger = LogManager.getLogger(CustomLoginSuccessHandler.class);
+    private static final String DESKTOP_LANDING_PAGE = "/pages/desktop.jsf";
+    private static final String EMPTY_LANDING_PAGE = "/pages/checks.jsf";
```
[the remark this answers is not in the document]
solth: See above.

Kitodo/src/main/java/org/kitodo/production/forms/LoginForm.java
```diff
@@ -51,7 +60,7 @@ public void setLoggedUser(User myClass) {
      * @return true or false
      */
     public boolean isAlreadyLoggedIn() {
-        return this.alreadyLoggedIn;
+        return false;
```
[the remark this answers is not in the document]
solth: Yes, you are right. I only simplified the current code, where the `alreadyLoggidIn` property would always contain the value `false`. My guess is that this check should be fixed or a new concept for this should be found in the future, but right now it's not supported by the new version of Kitodo.Production.

Kitodo/src/main/java/org/kitodo/production/forms/IndexingForm.java
```diff
@@ -200,13 +108,12 @@ public long getTotalCount() {
      * @return number of indexed objects
      */
     public long getNumberOfIndexedObjects(ObjectType objectType) {
-        long numberOfIndexedObjects = 0;
         try {
-            numberOfIndexedObjects = searchServices.get(objectType).count();
+            return ServiceManager.getIndexingService().getNumberOfIndexedObjects(objectType);
         } catch (DataException e) {
             Helper.setErrorMessage(e.getLocalizedMessage(), logger, e);
```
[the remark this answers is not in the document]
solth: Yes, that's true, but it's not part of my changes.

solth:
> As an "Admin" user, the client select dialog is not displayed even after the correction of index.

Thanks for pointing this out. It should work now. Could you test it again?

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
solth: Done.

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
solth: Moved to separate class.

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
Erikmitk: That line is still right here in the middle of the class and not at the top where declarations should be.

solth: Done (for real, this time!)

PR #2939, opened 2019-10-23
Show and control the migration of metadata in the task manager

matthias-ronge:
This is solved so that for each project to be migrated a separate thread is created. This was the easiest way to do it and will (hopefully) make the migration even faster.


Kathrin-Huber:
please write tests!

Kitodo/src/main/java/org/kitodo/production/forms/MigrationForm.java
```diff
@@ -57,25 +59,28 @@
...
     public void migrateMetadata() {
+        migrateWorkflow = false;
         try {
-            migrationService.migrateMetadata();
+            allProjects = ServiceManager.getProjectService().getAll();
+            projectListShown = true;
         } catch (DAOException e) {
-            Helper.setErrorMessage(ERROR_DATABASE_READING, new Object[] {ObjectType.PROCESS.getTranslationSingular() },
-                logger, e);
+            Helper.setErrorMessage("Error during database access");
```
henning-gerhardt: Why did you remove the `logger` and `e` parameters from method call? Maybe / helpful information get lost.

matthias-ronge: I deleted the whole method and put another method here, which has a catch in the same place, but another helper call. This only looks like the diff view, as if I had just changed the one line. The modification process was different, so I did not notice.

matthias-ronge:
Should everything be fixed now.

PR #2945, opened 2019-10-25
Parse German-style date, too

matthias-ronge:
Legacy wiki field data may have date be formatted German style, too.


henning-gerhardt:
Can you add a test inside `WikiFieldHelperIT` class to provide a good test for German encoded date?

PR #2946, opened 2019-10-25
Add test for MigrationTask

matthias-ronge:


Kitodo/src/test/java/org/kitodo/production/helper/tasks/MigrationTaskIT.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @AfterClass
+    public static void cleanDatabase() throws Exception {
+        MockDatabase.stopNode();
+        MockDatabase.cleanDatabase();
+    }
+
+    @Test
+    public void testMigrationTask() throws InterruptedException {
```
Kathrin-Huber: test should test, if metadata is migrated ;-)

matthias-ronge: This is covered by MetsKitodoConverterTest (in Kitodo-DataEditor module)

Kathrin-Huber: More specific: The test should test, if the metadata migration is actually triggered by the task. if I remove everything out of the run() method (but the setProgress100) the test is still green.

Kitodo/src/main/java/org/kitodo/production/services/ServiceManager.java
```diff
@@ -111,6 +111,12 @@ private static void initializeClientService() {
         }
     }
 
+    public static boolean discardDataEditorService() {
```
Kathrin-Huber: please do not add a public method for tests

Kitodo/src/test/java/org/kitodo/production/helper/tasks/MigrationTaskIT.java
Kathrin-Huber: can't you reuse the test logic from the converter test? A single migration should be enough to test, if everything is triggered alright

PR #2953, opened 2019-11-04
add migration script

Kathrin-Huber:


Kitodo-DataManagement/src/main/resources/db/migration/V2_88__Delete_templates.sql
```diff
@@ -0,0 +1,21 @@
...
+--
+-- For the full copyright and license information, please read the
+-- GPL3-License.txt file that was distributed with this source code.
+--
+-- 1. set templateIds to null
+-- 2. delete template Tasks from task_x_role
+-- 3. delete template Tasks
+-- 4. truncate template
+
+UPDATE process SET template_id = NULL;
```
matthias-ronge: `UPDATE process SET template_id = NULL WHERE template_id IS NOT NULL;`

solth:
Please update the name of the migration file because by now we have another migration file with number "2_89": `V2_89__Remove_dms_import_timeout.sql`

PR #2956, opened 2019-11-05
Allow to edit an existing user without login name check

matthias-ronge:
An existing user cannot be edited (for example, assign another project) if the login name is fails in validation. That makes no sense, because the login has already been assigned and cannot be changed anymore. This change causes that the login is only validated when a new user is created.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
@@ -132,7 +132,7 @@ public class UserForm extends BaseForm {
     public String save() {
         String login = this.userObject.getLogin();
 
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
```diff
@@ -132,7 +132,7 @@ public String newUser() {
     public String save() {
         String login = this.userObject.getLogin();
 
-        if (!isLoginValid(login)) {
+        if (Objects.isNull(userObject.getId()) && !userService.isLoginValid(login)) {
```
Kathrin-Huber: Please move this to the `isLoginValid` method instead of removing it

PR #2963, opened 2019-11-06
select workflowToUse in Frontend

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/forms/MigrationForm.java
```diff
@@ -233,31 +233,8 @@ public int getNumberOfProcesses(String tasks) {
...
-        for (Template template : allTemplates) {
-            if (migrationService.tasksAreEqual(template.getTasks(), processTasks)) {
-                workflowToUse = template.getWorkflow();
-                return true;
-            }
-        }
-        return false;
-
+        PrimeFaces.current().executeScript("PF('confirmWorkflowPopup').show();");
+        return this.stayOnCurrentPage;
```
solth: Since you always return `null` here, I would suggest to change to function to return type `void` and remove the return statement entirely.

Kitodo/src/main/java/org/kitodo/production/forms/MigrationForm.java
```diff
@@ -390,4 +367,26 @@ public void createNewTemplate(Template template) {
         }
         templatesToCreate.remove(template);
     }
+
+    public List<Workflow> getAllWorkflows() {
```
solth: Missing java doc in new public method.

PR #2964, opened 2019-11-06
Deactivated action icons in list views

solth:
To clean up the list views in Kitodo.Production, action icons are now always displayed if the current user generally has the authority to perform the corresponding action. It is now deactivated instead of hidden if the state of the list item prohibits the action at the moment, though.

Currently the only example for this is the workflow list where - among others - the `edit` symbol is deactivated for workflows that are already in use.

deactivated_symbols


Kathrin-Huber:
The tooltip should state something like "You are not able to edit, because the Workflow is already active"

PR #2973, opened 2019-11-11
Create process quicklink

solth:
Adds a direct link as a new action to the project list and the process widget on the desktop. This action can be used to directly create a process for the corresponding project without expanding the row in the project list to click the button next to the configured template. 

Bildschirmfoto 2019-11-11 um 17 51 37

If more than one template is configured for the selected project, a popup dialog is displayed to let the user select the template that he wants to use to create a new process for the project.

Bildschirmfoto 2019-11-11 um 17 51 52

Addtionally, a similar link is added to the template list to directly let the user create a process with a specific template. If the template is used in more than one project, a popup dialog is displayed that lets the user select the project for which a new process is to be created.


Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/dialogs/selectTemplate.xhtml
```diff
@@ -0,0 +1,67 @@
...
+                                       var="template"
+                                       itemLabel="#{template.title}"
+                                       itemValue="#{template.id}"/>
+                        <p:ajax update="selectTemplateForm:setTemplateButton"/>
+                    </p:selectOneMenu>
+                </p>
+                <p:panelGrid>
+                    <p:row>
+                        <p:commandButton id="setTemplateButton"
+                                         action="#{SelectTemplateDialogView.createProcessForProject(SelectTemplateDialogView.project.id, SelectTemplateDialogView.selectedTemplateId)}"
```
Kathrin-Huber: same here

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/SelectTemplateDialogView.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param projectId id of project for which a new process is created.
+     * @param templateId id of template from which new process is created for Project with given ID.
+     */
+    public void createProcessForProject(int projectId, int templateId) {
+        try {
+            this.project = ServiceManager.getProjectService().findById(projectId);
+            if (templateId > 0 || this.project.getTemplates().size() == 1) {
+                TemplateDTO template;
+                if (templateId > 0) {
```
Kathrin-Huber: if the project only has one template (with id > 0 which will be the most expected ase)
`template = this.project.getTemplates().get(0);` should be executed, but instead the template service is called with findById, which is unnecessary.

(same for select project dialog view)

Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/dialogs/selectProject.xhtml
```diff
@@ -0,0 +1,67 @@
...
+                                       var="project"
+                                       itemLabel="#{project.title}"
+                                       itemValue="#{project.id}"/>
+                        <p:ajax update="selectProjectForm:setProjectButton"/>
+                    </p:selectOneMenu>
+                </p>
+                <p:panelGrid>
+                    <p:row>
+                        <p:commandButton id="setProjectButton"
+                                         action="#{SelectProjectDialogView.createProcessForTemplate(SelectProjectDialogView.selectedProjectId, SelectProjectDialogView.template.id)}"
```
Kathrin-Huber: I think you don't need to explicitly give project id and selected template here, as it is know within the class, right?

PR #2979, opened 2019-11-13
OPAC configuration for K10Plus catalog

solth:


henning-gerhardt:
Why not using existing message translation for displaying the value in different languages instead of defining them multiple times? Maybe in its own message file?

PR #2986, opened 2019-11-14
Hierarchical import

solth:
Add hierarchical OPAC import to Kitodo.Production. Number of imported processes can be controlled by parameter `importDepth` in OPAC import mask. 

To import multiple hierarchical processes, the system tries to extract the ID of the parent data record from the imported metadata and queries the configured OPAC search interface again to retrieve the parent data record. This is repeated until the configured import depth is reached or no valid parent data record can be retrieved from the interface anymore, e.g. the top of the hierarchy is reached.

If necessary, extracted parent IDs can be processed before using them for subsequent OPAC queries, by adding a `trimMode` attribute to the `parentElement` node in the corresponding catalog configuration in `kitodo_opac.xml`. At the momemt, only one trim mode - `parenthesis` - is available, which removes all substrings enclosed in parenthesis from an ID.

Every data record is transformed into a Kitodo process and classified using the existing docType classification rules. In addition, the imported processes are linked with parent-child relations when saving.


Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -238,4 +237,297 @@ public String getRecordDocType(Document record) {
         }
         return "";
     }
+
+    private String getParentID(Document document) throws XPathExpressionException {
+
```
oliver-stoehr:
```suggestion
```

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -238,4 +237,297 @@ public String getRecordDocType(Document record) {
...
+                                    element.getChildNodes(), ruleset, docType, stage, languages, true);
+                        } else if (node.getLocalName().equals("metadata")) {
+                            setProcessDetailValue(detail, element.getTextContent());
+                        }
+                    }
+                    break;
+                }
+            }
+            if (!filled) {
+
```
oliver-stoehr:
```suggestion
```

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -238,4 +237,297 @@ public String getRecordDocType(Document record) {
...
+     *
+     * @param recordId identifier of the process to import
+     * @param createProcessForm CreateProcessForm instance containing import configuration
+     * @return List of TempProcess
+     */
+    public LinkedList<TempProcess> importProcessHierarchy(String recordId, CreateProcessForm createProcessForm,
+                                                          int importDepth)
+            throws IOException, ProcessGenerationException, XPathExpressionException, ParserConfigurationException,
+            NoRecordFoundException, UnsupportedFormatException, URISyntaxException, SAXException {
+
```
oliver-stoehr:
```suggestion
```

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -238,4 +237,297 @@ public String getRecordDocType(Document record) {
...
+            return String.valueOf(((ProcessBooleanMetadata) processDetail).isActive());
+        } else if (processDetail instanceof ProcessSelectMetadata) {
+            return String.join(", ", ((ProcessSelectMetadata) processDetail).getSelectedItems());
+        } else if (processDetail instanceof ProcessFieldedMetadata && processDetail.getMetadataID().equals(PERSON)) {
+            value = getCreator(((ProcessFieldedMetadata) processDetail).getRows());
+        }
+        return value;
+    }
+
+
```
oliver-stoehr:
```suggestion
```

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
```diff
@@ -43,8 +44,18 @@ private OPACConfig() {
      * @param parameter String identifying the parameter by name
      * @return value of parameter
      */
-    public static String getConfigValue(String catalogName, String parameter) {
-        return getCatalog(catalogName).getString(parameter);
+    public static String getConfigValue(String catalogName, String parameter) throws ParameterNotFoundException {
+        HierarchicalConfiguration opacConfiguration = getCatalog(catalogName);
+        if (Objects.isNull(opacConfiguration)) {
+            throw new ParameterNotFoundException("No configuration foudn for catalog '" + catalogName + "'!");
```
IkramMaalej:
```suggestion
            throw new ParameterNotFoundException("No configuration found for catalog '" + catalogName + "'!");
```

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
```diff
@@ -43,8 +44,18 @@ private OPACConfig() {
...
+        } else {
+            String parameterValue = opacConfiguration.getString(parameter);
+            if (parameterValue.isEmpty()) {
```
IkramMaalej: The use of StringUtils.isBlank(parameterValue) is better in this case because getString() method can return a null.

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
```diff
@@ -83,6 +94,25 @@ public static String getXsltMappingFile(String catalogName) {
...
 
+    /**
+     * Retrieve the "importDepth" of the catalog identified by its title.
+     * @param catalogName String identifying the catalog by its title
+     * @return int value of importDepth
+     */
+    public static int getImportDepth(String catalogName) {
+        return getCatalog(catalogName).getInt("importDepth");
+    }
+
```
IkramMaalej: This method is not used.

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
```diff
@@ -43,8 +44,18 @@ private OPACConfig() {
      * @param parameter String identifying the parameter by name
      * @return value of parameter
      */
-    public static String getConfigValue(String catalogName, String parameter) {
-        return getCatalog(catalogName).getString(parameter);
+    public static String getConfigValue(String catalogName, String parameter) throws ParameterNotFoundException {
+        HierarchicalConfiguration opacConfiguration = getCatalog(catalogName);
+        if (Objects.isNull(opacConfiguration)) {
```
IkramMaalej: This test is already done in getCatalog method.

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
IkramMaalej: This thrown exception is the same exception with what we already have in ConfigException.java.
```suggestion
            throw new ConfigException(catalogName);
```

Kitodo-API/src/main/java/org/kitodo/exceptions/ParameterNotFoundException.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Standard constructor with a message String.
+     *
+     * @param message exception message
+     */
+    public ParameterNotFoundException(String message) {
+        super(message);
+    }
+}
```
IkramMaalej: I think there is no need to add this Exception class. You can use ConfigException.java or, if necessary, you can add a method to it.

Kitodo/src/main/java/org/kitodo/production/helper/TempProcess.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param process Process of this TempProcess
+     * @param nodeList Metadata NodeList of this TempProcess
+     * @param docType Document type of process
+     */
+    public TempProcess(Process process, NodeList nodeList, String docType) {
+        this.process = process;
+        this.metadataNodes = nodeList;
+        this.workpiece = new Workpiece();
+        this.workpiece.getRootElement().setType(docType);
```
IkramMaalej: You have to make a test hier if the imported docType exists in the process ruleset.
you can use `docTypeExistsInRuleset(String docType)` method in `ProcessDataTab.java`.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessDataTab.java
```diff
@@ -183,79 +185,41 @@ public void setAllDocTypes(List<SelectItem> allDocTypes) {
...
-        this.titleDefinition = titleDefinition;
-    }
-
     /**
      * Get titleDefinition.
      *
      * @return value of titleDefinition
      */
-    public String getTitleDefinition() {
+    String getTitleDefinition() {
```
IkramMaalej: Method should be moved to `ProcessService.java`.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessDataTab.java
```diff
@@ -183,79 +185,41 @@ public void setAllDocTypes(List<SelectItem> allDocTypes) {
...
-        this.tiffDefinition = tiffDefinition;
-    }
-
     /**
      * Get tiffDefinition.
      *
      * @return value of tifDefinition
      */
-    public String getTiffDefinition() {
+    String getTiffDefinition() {
```
IkramMaalej: Same here.

Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/import.xhtml
```diff
@@ -18,48 +18,64 @@
...
+                <p:outputLabel for="catalogueSelectMenu"
+                               value="#{msgs['newProcess.catalogueSearch.catalogue']}"/>
+                <p:selectOneMenu id="catalogueSelectMenu"
+                                 required="#{not empty param['editForm:processFromTemplateTabView:performCatalogSearch']}"
+                                 immediate="true"
+                                 value="#{CreateProcessForm.importTab.hitModel.selectedCatalog}">
+                    <f:selectItem itemValue="#{null}" itemLabel="-- #{msgs.selectCatalog} --" noSelectionOption="true"/>
+                    <f:selectItems value="#{CreateProcessForm.importTab.catalogs}" var="step" itemLabel="#{step}"
+                                   itemValue="#{step}"/>
+                    <p:ajax update="editForm:processFromTemplateTabView:fieldSelectMenu,
```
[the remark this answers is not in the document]
solth: This would update the whole `tabView` including all five tabs and their content. I just want to update the three input components on the `import` tab, so I think the update should the stay the way it is.

Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/import.xhtml
```diff
@@ -18,48 +18,64 @@
...
+                                    editForm:processFromTemplateTabView:searchTerm,
```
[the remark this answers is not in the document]
solth: See above.

Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/import.xhtml
```diff
@@ -18,48 +18,64 @@
...
+                                    editForm:processFromTemplateTabView:importDepth,
```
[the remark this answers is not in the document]
solth: See above.

Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/import.xhtml
```diff
@@ -18,48 +18,64 @@
...
+                                    editForm:processFromTemplateTabView:catalogSearchButton"/>
```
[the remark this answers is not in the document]
solth: See above.

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
solth: `getCatalog` throws a `ConfigException` - which is a `RuntimeException` - and does not return anything when a configuration for the given catalog couldn't be found. In that case `opacConfiguration` is `null`, so the nullcheck here is required.

Kitodo-API/src/main/java/org/kitodo/exceptions/ParameterNotFoundException.java
solth: I think it's preferable to have multiple Exception classes to better handle individual error types that can occur instead of just having one general exception for everythin.

Kitodo/src/main/java/org/kitodo/production/helper/TempProcess.java
solth: I think creating an instance of `TempProcess` should be independent of a specific ruleset. Whether the ruleset contains the docType of one particular TempProcess should be checked somewhere else during validation and not in the TempProcess itself.

Kathrin-Huber:
If a "Titelsatzverknüpfung" is choosen in the last tab, the parent id of the main process is set correctly to the choosen one, but a new parent process is created anyhow, which is not connected to a process.
Please don't create a parent process if  one is explicitly choosen

PR #3001, opened 2019-11-29
Support defaultDisplay on metadata groups

matthias-ronge:
The `defaultDisplay` attribute on metadata groups was not supported by the front-end. This is hereby fixed.

### DIFF
--- a/Goobi/src/de/sub/goobi/metadaten/Metadaten.java
+++ b/Goobi/src/de/sub/goobi/metadaten/Metadaten.java
@@ -17,6 +17,7 @@ import org.goobi.io.SafeFile;
 import java.io.*;
 import java.net.URISyntaxException;
 import java.util.*;
+import java.util.Map.Entry;
 import java.util.concurrent.locks.ReentrantLock;
 
 import javax.faces.context.FacesContext;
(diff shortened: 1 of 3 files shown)


henning-gerhardt:
@matthias-ronge : do you have an configuration example how to use the `defaultDisplay` property in meta data groups? I'm was not aware that this is possible.

matthias-ronge:
```
        ...
        <group DefaultDisplay="true" num="*">Proveniece</group>
    </DocStrctType>
```

matthias-ronge:
No, the example rulesets do not contain any metadata groups yet.

Both exceptions would indicate a programming error, which could not be fixed by the user, so rethrowing them as unchecked exceptions here seems appropriate to me.

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -3098,15 +3099,59 @@ boolean canCreate(MetadataGroupType type) {
...
-            result.add(new RenderableMetadataGroup(record, this, language, projectName));
+            String name = record.getType().getName();
+            if (displayGroups.get(name) == null) {
+                displayGroups.put(name, new ArrayList());
+            }
+            displayGroups.get(name).add(record);
+        }
+        for (Entry<String, List<MetadataGroup>> entry : displayGroups.entrySet()) {
+            records = entry.getValue();
+            if (records != null) {
```
Kathrin-Huber: can be combined with following if statement

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -3098,15 +3099,59 @@ boolean canCreate(MetadataGroupType type) {
...
+                        result.add(new RenderableMetadataGroup(record, this, language, projectName));
+                    }
+                } else {
+                    for (MetadataGroupType groupType : myDocStruct.getType().getAllMetadataGroupTypes()) {
+                        if (groupType.getName().equals(entry.getKey())) {
+                            RenderableMetadataGroup newMetadataGroup = new RenderableMetadataGroup(groupType, language, projectName);
+                            result.add(newMetadataGroup);
+                            try {
+                                myDocStruct.addMetadataGroup(newMetadataGroup.toMetadataGroup());
+                            } catch (DocStructHasNoTypeException | MetadataTypeNotAllowedException e) {
```
Kathrin-Huber: why change exceptions here? either handle them here or hand them to the calling method

Goobi/src/de/sub/goobi/metadaten/RenderableMetadataGroup.java
```diff
@@ -157,6 +157,34 @@ public RenderableMetadataGroup(MetadataGroup data, Metadaten container, String l
...
+    public RenderableMetadataGroup(MetadataGroupType type, String language, String projectName)
+            throws ConfigurationException {
+        super(type.getAllLanguages(), null);
+        this.possibleTypes = Collections.emptyMap();
+        this.type = type;
+        this.projectName = projectName;
+        this.metadataGroup = null;
+        this.container = null;
+        updateMembers(type);
+        setLanguage(language);
```
Kathrin-Huber: why use setter here, when you use `this` for the others?

Goobi/src/de/sub/goobi/metadaten/RenderableMetadataGroup.java
```diff
@@ -157,6 +157,34 @@ public RenderableMetadataGroup(MetadataGroup data, Metadaten container, String l
...
+     * document structure hierarchy.
+     *
+     * @param type
+     *            type of metadata group
+     * @param language
+     *            display language to use
+     * @param projectName
+     *            project that the act whose metadata group is to edit belongs
+     *            to
+     * @throws ConfigurationException
```
Kathrin-Huber: why is it not possible to show a single selection in a multi-select?

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -3098,15 +3099,59 @@ boolean canCreate(MetadataGroupType type) {
...
+        if (myDocStruct == null) {
             return Collections.emptyList();
         }
+
+        List<MetadataGroup> records = myDocStruct.getAllMetadataGroups();
+        if (records == null) {
+            records = Collections.emptyList();
+        }
+        List<MetadataGroupType> defaultGroupTypes = myDocStruct.getType().getAllDefaultDisplayMetadataGroups();
+        if (records.isEmpty() && defaultGroupTypes.isEmpty()) {
```
Kathrin-Huber: shouldn't defaultGroupTypes be checked for null?

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
```diff
@@ -3098,15 +3099,59 @@ boolean canCreate(MetadataGroupType type) {
...
+            return Collections.emptyList();
```
Kathrin-Huber: no need to create a new list instance, just return records or defaultGroupTypes.

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
matthias-ronge: I think that's the best way to treat them. Both exceptions cannot occur at this point. A DocStruct with no type cannot occur here, because the metadata group is defined by the type of DocStruct in the first place. The MetadataType cannot be not allowed either, otherwise the field would not have been previously generated by the code. The most correct treatment of these exceptions would be an empty catch-block, but I reject that in principle. That's why this treatment.

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
matthias-ronge: Yes, but then I just have to check for `records != null` twice. I do not see any improvement in that, but I changed it.

Goobi/src/de/sub/goobi/metadaten/RenderableMetadataGroup.java
matthias-ronge: This is not a setter, but a function. It sets the language in the metadata group as well as in the subordinate metadata fields.

Goobi/src/de/sub/goobi/metadaten/RenderableMetadataGroup.java
matthias-ronge: You have misread. There must be no **multiple** selection in the metadata group.

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
matthias-ronge: No.

Goobi/src/de/sub/goobi/metadaten/Metadaten.java
matthias-ronge: `records` is a `List<ugh.dl.MetadataGroup>`, `defaultGroupTypes` is a `List<ugh.dl.MetadataGroupType>`, but this method must return a `List<de.sub.goobi.metadaten.RenderableMetadataGroup>`, so everything should be right here. By the way, `Collections.emptyList()` does **not** create a new instance of a list but uses an unmodifiable constant.

PR #3004, opened 2019-12-03
Define process title in ruleset

matthias-ronge:


Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Ruleset.java
```diff
@@ -146,21 +146,31 @@ public String getDefaultLang() {
...
+                for (UniversalDivision subdivision : universalDivision.getUniversalDivisions()) {
+                    if (all || subdivision.getProcessTitle().isPresent()) {
+                        universalDivisions.add(subdivision);
+                    }
+                }
             }
         }
-        return Labeled.listByTranslatedLabel(this, universalDivisions, UniversalDivision::getId,
-            UniversalDivision::getLabels, priorityList);
+        return all || !universalDivisions.isEmpty() ? Labeled.listByTranslatedLabel(this, universalDivisions,
```
Kathrin-Huber: Don't put logic in return statement

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Ruleset.java
```diff
@@ -146,21 +146,31 @@ public String getDefaultLang() {
...
+            UniversalDivision::getId, UniversalDivision::getLabels, priorityList)
+                : getDivisions(priorityList, true, subdivisionsByDate);
```
Kathrin-Huber: you are calling getDivions recursive here with the exact same attributes. looks like a possible infinite loop

Kitodo/rulesets/dfg-viewer.xml
```diff
@@ -44,7 +43,23 @@
...
+                        the possibility to automatically form process titles
+                        according to a schema. This is a string of metadata
+                        keys and constants separated by plus signs. Constants
+                        are identified by enclosing them in single quotes. The
+                        special metadata value TSL_ATS is given by the system
+                        with the author title key or, if no author is
+                        specified, the title key (4/2/2/1).
+
+                        When templates are used to form process titles in a
+                        rule set, only those divisions that have a template for
```
Kathrin-Huber:
```suggestion
                        ruleset, only those divisions that have a template for
```

Kitodo/rulesets/dfg-viewer.xml
```diff
@@ -44,7 +43,23 @@
...
+                        forming process titles are proposed as the top division
+                        types for new processes. Otherwise, all divisions are
```
Kathrin-Huber: why is there a restriction to the top level divisions depending on the fact that they are defining a processTitle or not?

Kitodo/rulesets/subhh.xml
```diff
@@ -44,7 +43,23 @@
...
+                        the possibility to automatically form process titles
+                        according to a schema. This is a string of metadata
+                        keys and constants separated by plus signs. Constants
+                        are identified by enclosing them in single quotes. The
+                        special metadata value TSL_ATS is given by the system
+                        with the author title key or, if no author is
+                        specified, the title key (4/2/2/1).
+
+                        When templates are used to form process titles in a
+                        rule set, only those divisions that have a template for
```
Kathrin-Huber:
```suggestion
                        ruleset, only those divisions that have a template for
```

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -408,8 +409,12 @@ private void processProcessHierarchy() throws ProcessGenerationException {
                 }
                 processDetails = metadata.getRows();
                 try {
+                    StructuralElementViewInterface docTypeView = rulesetManagementInterface
+                            .getStructuralElementView(docType, acquisitionStage, priorityList);
+                    String processTitle = docTypeView.getProcessTitle().orElseThrow(
+                        () -> new ProcessGenerationException(docType + " does not declare a process title."));
```
Kathrin-Huber: I would prefere an empty process title instead of throwing an exception

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessDataTab.java
```diff
@@ -189,8 +190,12 @@ public void generateProcessTitleAndTiffHeader() {
         List<ProcessDetail> processDetails = this.createProcessForm.getProcessMetadataTab().getProcessDetailsElements();
         Process process = this.createProcessForm.getMainProcess();
         try {
+            StructuralElementViewInterface docTypeView = createProcessForm.getRuleset().getStructuralElementView(
+                docType, createProcessForm.getAcquisitionStage(), createProcessForm.getPriorityList());
+            String processTitle = docTypeView.getProcessTitle()
+                    .orElseThrow(() -> new ProcessGenerationException(docType + " does not declare a process title."));
```
Kathrin-Huber: same here

Kitodo/src/main/java/org/kitodo/production/process/TitleGenerator.java
```diff
@@ -77,13 +77,28 @@ public String generateTitle(String titleDefinition, Map<String, String> genericF
...
+     *
+     * <p>
+     * <u>Examples:</u><br>
+     * {@code createAtstsl("Twenty Thousand Leagues Under the Sea", "Verne")}:
+     * {@code VernTwen}<br>
+     * {@code createAtstsl("Oxford English Dictionary", null)}: {@code OxfoEnDi}
      *
      * @param title
-     *            String
+     *            the title of the works
```
Kathrin-Huber: whats ```the works ``` ?

Kitodo/src/main/java/org/kitodo/production/process/TitleGenerator.java
```diff
@@ -77,13 +77,28 @@ public String generateTitle(String titleDefinition, Map<String, String> genericF
     }
 
     /**
-     * Create Atstsl.
+     * Forms the author title key, or the title key (4/2/2/1) if no author is
+     * given. The author title key is a librarian sort criteria, composed out of
+     * the first four letters of the first author’s last name, followed by the
```
Kathrin-Huber: how can it be formed out of the first author's last name if no author is given?

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Ruleset.java
matthias-ronge: No: if `all` was `false` and no element was found, the function is called again with `all` `true`.

Kitodo/rulesets/dfg-viewer.xml
matthias-ronge: Otherwise you cannot create a process title. That's what Sebastian had suggested in his email from 02.12.2019 12:34 o‘clock, and I think that's a good idea. That there were always all elements available was also a thorn in my side. In 2.x it is also the case that only the explicitly defined DocTypes are available for selection, for which a rule for the creation of the title has been defined.

Kitodo/src/main/java/org/kitodo/production/process/TitleGenerator.java
matthias-ronge: If an author is specified, the author title key will be rendered. If no author is specified, the title key (4/2/2/1) is formed. Also, look at the examples below.

Kitodo/src/main/java/org/kitodo/production/process/TitleGenerator.java
matthias-ronge: Whatever should be digitized in there. (German: “das Werk”)

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessDataTab.java
matthias-ronge: That means you click on the button and nothing happens. It is a matter of taste. I found it better to get a message than to have an application whose buttons just do not work, but ok, I change it.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
matthias-ronge: I cannot imagine that this can be wanted, but if you want it that way...

Kitodo/src/main/java/org/kitodo/production/process/TitleGenerator.java
Kathrin-Huber: Ah, so it should be "work" (even if its plural)

Kathrin-Huber:
```suggestion
     *            the title of the work
```

PR #3005, opened 2019-12-03
Unique process title configuration parameter

solth:
Add configuration parameter to control whether process titles must be unique in Kitodo or not. Default is `true`.


Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -261,12 +262,17 @@ public void setProject(Project project) {
     public String createNewProcess() {
         try {
             createProcessHierarchy();
+            if (Objects.nonNull(PrimeFaces.current())) {
```
Kathrin-Huber:
```suggestion
            if (Objects.nonNull(PrimeFaces.current())  && Objects.nonNull(FacesContext.getCurrentInstance()) {
```

PR #3006, opened 2019-12-03
Change from development to production project stage

henning-gerhardt:
Current MyFaces stage for application is set to development which is a bad for upcoming and working releases as many MyFaces security features are disabled and introducing "easy" ways to take over the application. Setting to production stage enable different security features and reduce the amount of log messages for different outputs.

Review from @solth and / or @oliver-stoehr ?


henning-gerhardt:
I'm was not sure if this context param is needed by someone but if then he / she can add this param again.

PR #3008, opened 2019-12-03
Check project assignment on url

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
```diff
@@ -101,17 +101,17 @@ public String replace(String inString) {
...
-            String imagePath = replaceSlashAndSeparator(fileService.getImagesDirectory(this.process));
-            String origPath = replaceSlashAndSeparator(processService.getImagesOriginDirectory(false, this.process));
-            String processPath = replaceSlashAndSeparator(processService.getProcessDataDirectory(this.process));
-            String metaFile = replaceSlash(fileService.getMetadataFilePath(this.process, false, false));
-            String ocrBasisPath = replaceSlashAndSeparator(fileService.getOcrDirectory(this.process));
-            String ocrPlaintextPath = replaceSlashAndSeparator(fileService.getTxtDirectory(this.process));
-            String sourcePath = replaceSlashAndSeparator(fileService.getSourceDirectory(this.process));
-            String importPath = replaceSlashAndSeparator(fileService.getImportDirectory(this.process));
-            String prefs = ConfigCore.getParameter(ParameterCore.DIR_RULESETS) + this.process.getRuleset().getFile();
+            final String imagePath = replaceSlashAndSeparator(fileService.getImagesDirectory(this.process));
```
solth: Don't we have naming convetions for constants? Capital letters only, with underscores between individual words?

Kitodo/src/main/webapp/pages/searchResult.xhtml
```diff
@@ -17,7 +17,9 @@
         xmlns:h="http://xmlns.jcp.org/jsf/html"
         xmlns:f="http://xmlns.jcp.org/jsf/core"
         xmlns:p="http://primefaces.org/ui">
-
+    <f:metadata>
+        <f:viewParam name="id"/>
```
solth: I am a little confused here. How can the access rights to a process be determined on a per-process basis if you pass just one `id` parameter to this view that is than used in every row of the datatable? Shouldn't the ID instead be determined by each specific record of the datatables values, e.g. something like `process.id`?

Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
Kathrin-Huber: Actually it's not a constant

solth: Yes, you are right, apparently it's a `final variable`, and not a `constant`. And according to [Wikipedia](https://en.wikipedia.org/wiki/Final_(Java)), the naming convention I mentioned only applies to the later:
> Unlike the value of a constant, the value of a final variable is not necessarily known at compile time. It is considered good practice to represent final constants in all uppercase, using underscore to separate words.

PR #3013, opened 2019-12-04
Use letters for METS ID attributes

matthias-ronge:


Kitodo-API/src/main/java/org/kitodo/api/dataformat/mets/KitodoUUID.java
```diff
@@ -0,0 +1,66 @@
+package org.kitodo.api.dataformat.mets;
```
henning-gerhardt: License header is missing.

matthias-ronge: Thanks for the hint. Actually, this should be intercepted by Checkstyle, which is why I am surprised that it passed all checks.

henning-gerhardt: Maybe because there was less errors then allowed errors and so this error was not visible. I reduced the amount of errors and allowed errors in last pull request, so rebasing this branch should be a good idea.

PR #3026, opened 2019-12-16
Declare special fields in rule set

matthias-ronge:


Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
```diff
@@ -65,6 +68,32 @@
...
+                    continue;
+                }
+                Set<SpecialField> uses = SpecialField.valuesOf(key.getUse());
+                if (uses.contains(specialField)) {
+                    idsOfKeysForSpecialField.add(key.getId());
+                }
+            } else {
+                List<String> idsOfKeysOfKey = getIdsOfKeysForSpecialField(key.getKeys(), specialField);
+                for (String idOfKeyOfKey : idsOfKeysOfKey) {
+                    idsOfKeysForSpecialField.add(key.getId() + '/' + idOfKeyOfKey);
```
Kathrin-Huber: Is this a convention? Where do I know, that a slash separates a key from the keyOfKey

Kitodo-DataEditor/src/test/java/org/kitodo/dataeditor/ruleset/RulesetManagementIT.java
```diff
@@ -704,6 +706,15 @@ public void testFieldsWithMinOccursGreaterZeroAreAlwaysShown() throws IOExceptio
...
     }
 
+    @Test
+    public void testGettingOfSpecialFields() throws Exception {
+        RulesetManagementInterface underTest = new RulesetManagement();
+        underTest.load(new File("src/test/resources/testAnExtensiveRulesetCanBeLoaded.xml"));
+
+        assertThat(underTest.getIdsOfKeysForSpecialField(SpecialField.TITLE), contains("TitleDocMain"));
+        assertThat(underTest.getIdsOfKeysForSpecialField(SpecialField.AUTHOR_LAST_NAME), contains("Person/LastName"));
+    }
```
Kathrin-Huber: please test also for not existing uses
test, that the result is exactly the ids not only "contains" because if the method always returns all ids it will still be true
please test for multiple uses separated by comma in one key

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SpecialField.java
```diff
@@ -0,0 +1,109 @@
...
+     * The name of the data source from which the record was imported. This is
+     * saved for later comparison of the data records.
+     */
+    DATA_SOURCE("dataSource"),
+
+    /**
+     * The key of a higher-level data record in a hierarchical data structure of
+     * 1:n relationships, which are stored from bottom to top.
+     */
+    HIGHERLEVEL_IDENTIFIER("higherlevelIdentifier"),
```
Kathrin-Huber: why not
`higherLevelIdentifier`?

matthias-ronge: Because it is an identifier of a higher level, not a higher identifier for a level. In natural language, I do use a [hyhen](https://www.grammarbook.com/punctuation/hyphens.asp) here to clarify the same: “The key of a **higher-level** data record”, because it is not a ~~level data record~~.

Kitodo/src/test/java/org/kitodo/DummyRulesetManagement.java
```diff
@@ -19,6 +19,7 @@
 import java.util.Map;
 
 import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
+import org.kitodo.api.dataeditor.rulesetmanagement.SpecialField;
 import org.kitodo.api.dataeditor.rulesetmanagement.StructuralElementViewInterface;
 
 public class DummyRulesetManagement implements RulesetManagementInterface {
```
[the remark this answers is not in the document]
matthias-ronge: It returns a `DummyStructuralElementView` for `StructurePanelTest.testBuildStructureTreeRecursively()`.

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SpecialField.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Certain fields in the rule set that have a special function.
+ */
+public enum SpecialField {
```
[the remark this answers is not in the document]
matthias-ronge: That depends on the perspective. In the XML file, it is the content of an element. What is it? So, it's the metadata that a function is attached to. They are used to make certain functions possible. For example, the author's last name and title are used to generate the ATS. That depends on the perspective. In the XML file, it is the content of an element. What is it? So, it's the metadata that a function is attached to. They are used to make certain functions possible. For example, the author's last name and title are used to generate the ATS. Maybe `FunctionalMetadata`?

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
matthias-ronge: I will change this to `@`, then it is identical to the spelling in the Metadata copy rules.

Kitodo-DataEditor/src/test/java/org/kitodo/dataeditor/ruleset/RulesetManagementIT.java
matthias-ronge: Added tests for not-existing uses and multiple uses of one key. `contains()` here defines equality, so if there would be more elements returned, this would cause the test to fail. Multiple uses are to be separated by space, not comma. This is how the [NMTOKENS is defined](https://www.w3.org/TR/xmlschema11-2/#NMTOKENS) in XML.

Kitodo-DataEditor/src/test/java/org/kitodo/dataeditor/ruleset/RulesetManagementIT.java
```diff
@@ -704,6 +706,15 @@ public void testFieldsWithMinOccursGreaterZeroAreAlwaysShown() throws IOExceptio
         assertThat(ids(mvwviList), contains("test1", "test2", "test2", "test2options"));
     }
 
+    @Test
+    public void testGettingOfSpecialFields() throws Exception {
+        RulesetManagementInterface underTest = new RulesetManagement();
+        underTest.load(new File("src/test/resources/testAnExtensiveRulesetCanBeLoaded.xml"));
+
+        assertThat(underTest.getIdsOfKeysForSpecialField(SpecialField.TITLE), contains("TitleDocMain"));
+        assertThat(underTest.getIdsOfKeysForSpecialField(SpecialField.AUTHOR_LAST_NAME), contains("Person/LastName"));
```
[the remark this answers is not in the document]
matthias-ronge: Added error messages.

PR #3039, opened 2020-01-07
Migrate serial publications

matthias-ronge:
This implements the migration of serial publications. For serial publications (series, multi-volume works), a parent process is created and the children found are linked to it. The METS structure is adjusted.


Kitodo/src/main/java/org/kitodo/production/forms/MigrationForm.java
```diff
@@ -132,6 +133,11 @@ public void showAggregatedProcesses() {
         processListRendered = true;
     }
```
Kathrin-Huber: Missing Javadoc

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -426,6 +426,7 @@ granularity.years=Jahrg\u00E4nge
 guessImages=gesch\u00E4tzte Seitenanzahl
 help=Hilfe
 hideCorrectionTasks=Entferne Korrekturschritte
+HierarchyMigrationTask=Migriere serielle Publikationen
```
Kathrin-Huber: please use lowecase

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -547,6 +547,20 @@ public void moveFile(URI sourceUri, URI targetUri) throws IOException {
...
+     * Process owns anchor XML.
+     *
+     * @param process
+     *            whose metadata file path to use
+     * @return whether an anchor file was found
+     * @throws IOException
+     *             if Io failed
+     */
+    public boolean processOwnsAnchorXML(Process process) throws IOException {
+        URI yearFile = FileService.this.createAnchorFile(FileService.this.getMetadataFilePath(process, false, false));
```
Kathrin-Huber:
```suggestion
        URI yearFile = createAnchorFile(FileService.this.getMetadataFilePath(process, false, false));
```

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -547,6 +547,20 @@ public void moveFile(URI sourceUri, URI targetUri) throws IOException {
...
+        return FileService.this.fileExist(yearFile);
```
Kathrin-Huber:
```suggestion
        return fileExist(yearFile);
```

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
```diff
@@ -0,0 +1,358 @@
...
+    private final ProcessGenerator processGenerator = new ProcessGenerator();
+
+    /**
+     * Service to read and write Process objects in the database or search
+     * engine index.
+     */
+    private static final ProcessService processService = ServiceManager.getProcessService();
+
+    /**
+     * Parent processes have already been created. Key is the identifier, the
```
Kathrin-Huber: Parent Processes have already been created? Really? When?

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Links parent process and child process in the database. The processes are
+     * saved.
+     *
+     * @param parentProcess
+     *            parent process to link
+     * @param childProcess
+     *            child process to link
+     */
+    private static void linkInDatabase(Process parentProcess, Process childProcess) throws DataException {
```
Kathrin-Huber: Its not only in Database but in elasticsearch and Object level.
```suggestion
    private static void linkParentProcessWithChildProcess(Process parentProcess, Process childProcess) throws DataException {
```

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
```diff
@@ -0,0 +1,358 @@
...
+
+    /**
+     * Service to read and write Process objects in the database or search
+     * engine index.
+     */
+    private static final ProcessService processService = ServiceManager.getProcessService();
+
+    /**
+     * Parent processes have already been created. Key is the identifier, the
+     * value is the process ID, then the current numbers of child links.
```
Kathrin-Huber: You are mixing up purposes of integers in the List.
Why not using the process ID as a key instead of the String?
Shouldn't the list of childs also be found in the Database?

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
```diff
@@ -0,0 +1,358 @@
...
+    private static Integer convertChildMetsFile(URI metadataFilePath) throws IOException {
+        Workpiece workpiece = metsService.loadWorkpiece(metadataFilePath);
+        IncludedStructuralElement childStructureRoot = workpiece.getRootElement().getChildren().get(0);
+        workpiece.setRootElement(childStructureRoot);
+        metsService.saveWorkpiece(workpiece, metadataFilePath);
+        return getCurrentNo(childStructureRoot);
+    }
+
+    /**
+     * Extracts the CurrentNo from the metadata.
```
Kathrin-Huber: What does currentNO stands for in the metadata?
Please give a more explicit description

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
```diff
@@ -0,0 +1,358 @@
...
+
+    /**
+     * Renames the anchor file. It is clear that this has been migrated.
+     *
+     * @param process
+     *            process to migrate
+     */
+    private static void renameAnchorFile(Process process) throws IOException {
+        URI anchorFile = fileService.createAnchorFile(fileService.getMetadataFilePath(process));
+        fileService.renameFile(anchorFile, "meta_anchor.migrated");
```
Kathrin-Huber: Is kitodo still working with this renamed anchor file?
@henning-gerhardt  what do you think?

henning-gerhardt: I don't understand why this file is still existing after migration and why it get a new name. Maybe @matthias-ronge should explain this in more detail. Maybe direct in German between @Kathrin-Huber and me.

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
matthias-ronge: If they were added to this map, they were created beforehand. It happens in line 169, which calls `createParentProcess()` itself.

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
matthias-ronge: What I wanted to say is that the link is created at the database level (object, Elasticsearch, is synonymous) in contrast to METS pointer link. _Link parent process with child process_ is not clear, which of the two links is created here.

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
matthias-ronge: * I need several integers to do this job. These integer values are saved in the list for each parent process.
* I do not use the ID as a key, because in further child processes I can only find the string as the parent, but not the ID. So with the ID as a key, I cannot check whether it has already created the parent or not. This is basically what the map is for.
* The list of children can be found in the database and is not necessary here. The list contains the current number of the children, not their object ID.

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
matthias-ronge: The current number is an integer sorting criterion, which specifies the order of the subordinate units within their superordinate entirety. In case of journal issues, this can be the same as the issue number if the issue number continues to be counted at the turn of the year. In the case of multi-volume works, this number can correspond to the part number, or (for lexica, for example) it is counted up according to the alphabetical order of the volumes, supplementary volumes are counted on afterwards (thus, in the order in which the books are usually placed on a shelf).

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
matthias-ronge: Simple, I‘m not deleting the file for safety‘s sake if anything goes wrong, but I‘m renaming it so that the migration cannot be accidentally repeated (principle of [idempotence](https://de.wikipedia.org/wiki/Idempotenz#Informatik)). I had followed the same procedure for the newspaper process migration. In principle, the file could also be deleted, but then there is no going back if something went wrong during the migration. So, consider this file as a backup, similar to backup file rotation.

Kitodo/src/main/resources/messages/messages_de.properties
matthias-ronge: I have now changed this consistently so that all long-running task classes now use their class name for the name, starting with a lowercase letter.

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
Kathrin-Huber: "because in further child processes I can only find the string as the parent"  What does that mean? The string of the ID? Or what string?
Is it correct, that the list always only contains two entries? parentID and number of processes? maybe use a Map?

matthias-ronge: The string is the PPN of the superior unit (the journal, or the multi-volume works as its whole). It is found in the anchor.xml file as `CatalogIDDigital`. This is used to identify the shared parent of several processes. First, this string is read from the file. Then the migration looks if this string is in the map. If yes, then the parent process has been created (there was already another child for this parent before). If not, a parent process is created and written into the map.

The number of integers in the list is always one greater as the number of issues already found for the parent, because first there is the parent record number. So it is two after one child, but more after more children.

**An example:**
- The migration finds a hierarchial process (database no. 22) and reads parent PPN `1234567X` from the anchor file, and `CurrentNo` 2 from the meta.xml file. The map is empty. So it creates a new process as parent process. That process gets database number 46. Then it writes the following entry into the map: `"1234567X" → <46, 2>`. It adds a link in the logical tree from the parent to the child no. 22.
- Then it finds a hierarchical process (database no. 35) and reads parent PPN `1234567X` from the anchor file, and `CurrentNo` 3 from the meta.xml file. There is an entry for string `1234567X` in the map, so there is already a parent for this and no new process is created. It adds a link in the logical tree from the parent to the child no. 35 _after_ the first link, because it knows (from the int list) that the first link is current no. 2 and this is current no. 3 and must therefore be added after the first link. Then it updates the map entry to look like this: `"1234567X" → <46, 2, 3>`
- Then it finds a hierarchical process (database no. 19) and reads parent PPN `23456790` from the anchor file, and `CurrentNo` 3 from the meta.xml file. There is no entry for string `23456790` in the map, so it creates a new process as parent process. That process gets database number 47. Then it writes the following entry into the map: `"23456790" → <47, 3>`. It adds a link in the logical tree from the parent to the child no. 19.
- Then it finds a hierarchical process (database no. 17) and reads parent PPN `23456790` from the anchor file, and `CurrentNo` 1 from the meta.xml file. There is an entry for string `23456790` in the map, so there is already a parent for this and no new process is created. It adds a link in the logical tree from the parent to the child no. 17 _before_ the first link, because it knows (from the int list) that the first link is current no. 3 and this is current no. 1 and must therefore be added before the first link. Then it updates the map entry to look like this: `"23456790" → <47, 1, 3>`
- Then it finds a hierarchical process (database no. 45) and reads parent PPN `1234567X` from the anchor file, and `CurrentNo` 5 from the meta.xml file. There is an entry for string `1234567X` in the map, so there is already a parent for this and no new process is created. It adds a link in the logical tree from the parent to the child no. 45 _after_ the two other links, because it knows (from the int list) that the first link is current no. 2 and the second link is current no. 3 and this is current no. 5 and must therefore be added after the two links. Then it updates the map entry to look like this: `"1234567X" → <46, 2, 3, 5>`
- Then it finds a hierarchical process (database no. 18) and reads parent PPN `23456790` from the anchor file, and `CurrentNo` 2 from the meta.xml file. There is an entry for string `23456790` in the map, so there is already a parent for this and no new process is created. It adds a link in the logical tree from the parent to the child no. 18 _in between_ the first and second link, because it knows from the int list that the first link is current no. 1 and the second link is current no. 3 and this is current no. 2 and must therefore be added between the two links. Then it updates the map entry to look like this: `"23456790" → <47, 1, 2, 3>`
- Then it finds a hierarchical process (database no. 40) and reads parent PPN `1234567X` from the anchor file, and `CurrentNo` 4 from the meta.xml file. There is an entry for string `1234567X` in the map, so there is already a parent for this and no new process is created. It adds a link in the logical tree from the parent to the child no. 40 between the second and third link, because it knows from the int list that the second link is current no. 3 and the third link is current no. 5 and this is current no. 4 and must therefore be added between the second and third link. Then it updates the map entry to look like this: `"1234567X" → <46, 2, 3, 4, 5>`
- Then it finds a hierarchical process (database no. 14) and reads parent PPN `1234567X` from the anchor file, and `CurrentNo` 1 from the meta.xml file. There is an entry for string `1234567X` in the map, which looks like this: `"1234567X" → <46, 2, 3, 4, 5>`. Should be clear now where to add the link …

PR #3050, opened 2020-01-13
Add ORDER to logical structure

oliver-stoehr:
Adds the ORDER attribute to the divs of the logical structMap. This is the basis for ordering the logical and physical nodes in the combined tree.

### DIFF
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/DivXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/DivXmlElementAccess.java
@@ -112,15 +112,15 @@ public class DivXmlElementAccess extends IncludedStructuralElement {
         metsReferrerId = div.getID();
         BigInteger order = div.getORDER();
         if (Objects.nonNull(order) && order.intValue() > 0) {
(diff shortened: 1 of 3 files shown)


Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/DivXmlElementAccess.java
```diff
@@ -105,9 +110,17 @@ public DivXmlElementAccess() {
             super.getMetadata().addAll(readMetadata((MdSecType) mdSecType, amdSecTypeOf(mets, (MdSecType) mdSecType)));
         }
         metsReferrerId = div.getID();
+        BigInteger order = div.getORDER();
+        if (Objects.nonNull(order) && order.intValue() > 0) {
+            super.setOrder(order.intValue());
```
Kathrin-Huber:
```suggestion
            setOrder(order.intValue());
```

Kathrin-Huber: You don't need `super `here. As far as I see thats counts for every `super` in this class. Remove at least those you added ;-)

PR #3051, opened 2020-01-14
Investigate working st

Kathrin-Huber:
This does not fix the Selenium Problems (which I was hoping for) but it speeds up the test by 15 minutes.

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/TaskServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/TaskServiceIT.java
@@ -216,8 +216,7 @@ public class TaskServiceIT {
 
     @Test
     public void shouldGetCurrentTasksOfBatch() {
-        List<Task> tasks = taskService.getCurrentTasksOfBatch("Task", 1);
-        System.out.println("shouldGetByTitleAndBatches: " + tasks.size());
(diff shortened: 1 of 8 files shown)


Kitodo/src/test/java/org/kitodo/production/services/data/TaskServiceIT.java
```diff
@@ -216,8 +216,7 @@ public void shouldGetTitleWithUserName() throws Exception {
 
     @Test
     public void shouldGetCurrentTasksOfBatch() {
-        List<Task> tasks = taskService.getCurrentTasksOfBatch("Task", 1);
-        System.out.println("shouldGetByTitleAndBatches: " + tasks.size());
+        taskService.getCurrentTasksOfBatch("Task", 1);
     }
```
henning-gerhardt: There is no test at all or what is this test is checking for? Even old code did not test anything :-(

Kathrin-Huber: I was just removing the system out. But you are right, i'm adding a check

Kathrin-Huber: Ok, there is something wrong with this method. I'm going to fix it in a separat Pull Request

PR #3060, opened 2020-01-16
Display information about process hierarchy

oliver-stoehr:
Display information about process hierarchy in the process list:
- whether the process has child/parent processes or both
- list of the parent processes
- number of direct child processes with a link to filter for these processes


Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -785,6 +789,7 @@ shortcuts.dataEditor.detailView.keys=Ctrl Shift +
 showAutomaticTasks=show automatic tasks
 showClosedProcesses=Show closed processes
 showDataForProcess=show process data
+showHierarchicalChildren=Show
```
Kathrin-Huber: If the message is just "show" so maybe the key shold be "show" ;-)

Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -1330,4 +1330,34 @@ public void setFilter(String filter) {
...
+    /**
+     * Get number of direct child processes for the given process.
+     *
+     * @param processId
+     *          process id for given process
+     * @return number of child processes
+     */
+    public int getNumberOfChildProcesses(int processId) {
+        try {
+            return ServiceManager.getProcessService().getById(processId).getChildren().size();
```
Kathrin-Huber: Please write a method `ProcessService.getNumberOfChildren(process)`
which is calling the ProcessDAO to count on database

PR #3061, opened 2020-01-16
List sorting & styling

solth:
- enable sorting lists by first column
- fix list layout errors
- update list row expansion design


Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/projectsWidget.xhtml
```diff
@@ -19,16 +19,20 @@
     <p:dataTable id="projectTable"
                  value="#{DesktopForm.projects}"
                  scrollable="true"
+                 styleClass="default-layout"
                  var="project">
         <p:column headerText="#{msgs.title}"
+                  sortBy="#{project.title}"
```
Kathrin-Huber:
```suggestion
                  sortBy="#{project.title.keyword}"
```

solth: The records for the desktop widgets are not retrieved from the index like the normal list views, but instead from the DB, so the index specific `keyword` does not exist here. Instead, sorting projects in the `projectsWidget` should already work without it!

PR #3063, opened 2020-01-17
Image processing with conditions

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/forms/CurrentTaskForm.java
```diff
@@ -403,13 +405,28 @@ public void generateMissingImages() {
...
      */
     private void generateImages(GenerationMode mode, String messageKey) {
+        Folder generatorSource = myProcess.getProject().getGeneratorSource();
+        List<Folder> contentFolders = currentTask.getContentFolders();
+        if (Objects.isNull(generatorSource)) {
+            Helper.setErrorMessage("noSourceFolderConfiguredInProject");
+        }
+        if(Objects.isNull(contentFolders)){
+            Helper.setErrorMessage("noImageFolderConfiguredInProject");
+        }
```
solth: Shouldn't we leave the method if either of these conditions is met, instead of executing the `try` block?

PR #3064, opened 2020-01-17
fix condition

Kathrin-Huber:


Kitodo-Validation/src/main/java/org/kitodo/validation/metadata/MetadataValidation.java
```diff
@@ -179,7 +179,7 @@ private static ValidationResult checkForStructuresWithoutMedia(Workpiece workpie
             warning = true;
         }
 
-        if (treeStream(workpiece.getRootElement(), IncludedStructuralElement::getChildren)
+        if (!treeStream(workpiece.getRootElement(), IncludedStructuralElement::getChildren)
                 .flatMap(structure -> structure.getViews().stream()).map(View::getMediaUnit)
                 .anyMatch(workpiece.getMediaUnits()::contains)) {
```
matthias-ronge: You also have to replace `anyMatch` with `allMatch` here, otherwise there is only a validator error if _none_ of the media are in the physical structure tree. However, there should already be an error message if _only one_ is not in the physical tree.

PR #3091, opened 2020-01-23
Fix search result page

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/forms/SearchResultForm.java
```diff
@@ -156,6 +158,22 @@ private void refreshFilteredList() {
         filteredList.addAll(resultList);
     }
 
+    /**
+     * delete process.
+     * @param processDTO to delete
+     */
+    public void delete(ProcessDTO processDTO){
```
henning-gerhardt: Missing white space between `)` and `{`.

Kitodo/src/main/java/org/kitodo/production/forms/SearchResultForm.java
```diff
@@ -156,6 +158,22 @@ private void refreshFilteredList() {
...
+       try {
+           ProcessForm processForm = new ProcessForm();
+           processForm.setProcess(ServiceManager.getProcessService().getById(processDTO.getId()));
+           processForm.delete();
```
henning-gerhardt: Should method `delete` not inside `ProcessService` class so deleting a process is not depending on a form class?

Kitodo/src/main/java/org/kitodo/production/forms/SearchResultForm.java
```diff
@@ -156,6 +158,22 @@ private void refreshFilteredList() {
...
+       }
+        catch (DAOException e) {
+            Helper.setErrorMessage("errorOnSearch", searchQuery);
+        }
+       filteredList.remove(processDTO);
```
henning-gerhardt: Is it a good idea to remove this process from the `filteredList` if deleting is causing a `DAOException`?

Kitodo/src/main/java/org/kitodo/production/forms/SearchResultForm.java
```diff
@@ -22,7 +22,9 @@
 import javax.enterprise.context.SessionScoped;
 import javax.inject.Named;
 
+import org.kitodo.data.database.beans.Process;
```
henning-gerhardt: Not needed import statement.

PR #3092, opened 2020-01-23
Pass LDAP password

matthias-ronge:


matthias-ronge:
@Kathrin-Huber & @henning-gerhardt: I can now confirm that this works:
Anmerkung 2020-01-24 094512

PR #3114, opened 2020-01-29
replace FLocat on export

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/services/schema/SchemaService.java
```diff
@@ -129,6 +132,29 @@ private void addVirtualFileGroupsToMetsMods(Workpiece workpiece, Process process
...
+                                .substring(lastSeparator + 1);
+                        mediaFileForMediaVariant
+                                .setValue(new URI(replacePlaceholder(folder.getUrlStructure() + lastSegment, process)));
+                    }
+                }
+            }
+        }
+    }
+
+    private String replacePlaceholder(String uri, Process process) {
```
henning-gerhardt: Why you are not using the `replace` method from ` VariableReplacer` class? This method is doing the same.

Kathrin-Huber: Thanks for noticing!

PR #3144, opened 2020-02-05
Add missing strings to translation files

matthias-ronge:
Comparing the language files showed that some keys were not translated. That will now be added.


Kitodo/src/main/resources/messages/errors_en.properties
```diff
@@ -130,6 +130,8 @@ processTitleEmpty=No process title stated.
 processTitleInvalid=Invalid title for process. It should match ''{0}'' regular expression.
 projectTitleAlreadyInUse=The project title is already in use.
 propertiesForProcessNotSaved=Properties for process {0} could not be saved.
+prozesskopieForm.createNewProcess.noInsertionPositionSelected=No bibliographic link position was selected.
+prozesskopieForm.createNewProcess.titleRecordOpen=The process cannot be created because the bibliographic record is being edited by {0}.
```
solth: It seems these message keys are not used any more (instead, `createProcessForm.createNewProcess.noInsertionPositionSelected` etc. are used now). 
Rather than adding the deprecated keys to the file where they are missing, you should remove from them from the file they still exist in.

PR #3147, opened 2020-02-05
No longer disclose credentials on the start screen

matthias-ronge:
Currently the application discloses the credentials of the default users on the start screen, which is really bad. This is hereby fixed.


henning-gerhardt:
You should even update any existing documentation as without this removed information no one can login any more.

matthias-ronge:
Is there any existing documentation that deals with this?

matthias-ronge:
@henning-gerhardt How do you want this to be resolved?

henning-gerhardt:
* If there is no documentation you should create them or add them at least to the installation guide of 2.x
* I did not understand and maybe never understand why you are starting now at the end of life this version to change something like this.

matthias-ronge:
An updated [installation instructions (temporarily named version 2.3.1)](https://github.com/kitodo/kitodo-production/wiki/Installation-instructions-for-Kitodo.Production-2.3.1) has been provided which mentions how to log in (with the changed credentials).

PR #3148, opened 2020-02-05
Remove default users, rename admin and LDAP group

matthias-ronge:


Goobi/setup/default.sql
```diff
@@ -16,27 +16,17 @@
...
 INSERT INTO `ldapgruppen` (
   `ldapgruppenID`, `titel`) VALUES 
-  (2, 'test'); /* id 2 has to be inserted to match users (benutzer)*/
+  (2, 'Example LDAP group'); /* id 2 has to be inserted to match users (benutzer)*/
 /*!40000 ALTER TABLE `benutzer` ENABLE KEYS */;
 
 --
--- Password for test users is "test"
--- NOTICE: Disable those users in production environment!
+-- Password for admin user is "goobi"
```
henning-gerhardt: Why you are using `goobi` as login password? Application and documentation is already adjusted to use Kitodo in many places.

matthias-ronge: Changed

matthias-ronge:
- `schritteberechtigtegruppen` references `BenutzerGruppenID` from table `benutzergruppen` which is unchanged.
- The login is mentioned in the file. Is there any documentation that deals with this?

If you see a problem with this, you may want to comment on the ticket. It may be a good reason to reject the ticket and not work it. Then it would be done, too.

matthias-ronge:
@henning-gerhardt How do you want this to be resolved?

henning-gerhardt:
Like on your other fix to an very old "issue": I did not understand and maybe never understand why you are starting now at the end of life this version to change something like this.

matthias-ronge:
An updated [installation instructions (temporarily named version 2.3.1)](https://github.com/kitodo/kitodo-production/wiki/Installation-instructions-for-Kitodo.Production-2.3.1) has been provided which mentions how to log in with the changed credentials.

Issue #320, opened 2015-03-18
Investigate and fix issues reported by static code analyzers


stweil:
Static code analyzers for Java report some issues in the current code: FindBugs reports more than 500, Coverity around 440. As there are real errors among those issues, they need more investigation.

FindBugs is Freeware and included in Linux Distributions, so it is easy to run those checks in your own build environment.

Coverity is closed source, but they offer a scan service for free software. They scanned Goobi.Production, and a summary is available here: https://scan.coverity.com/projects/4522. The full report is available on demand - just use the link on the website.

claussni:
I agree that static analyzers should be used in the development process. However, the code base is old and mostly untested, but stable. Affected and erroneous parts are all over the class path. Some bugs might as well be features (and the other way around). Almost all parts of the application need a complete renovation, which is costly.

Introducing proper development mechanisms like code analysis, automated integration testing and style checking is common sense. But as long as there is only feature-driven funding we will need to take a pragmatic road.

henning-gerhardt:
General notice for Coverity: You can use your GitHub account for access to it. Afterwards you should apply for more access. Without you see no Coverity reports / detailed information.

matthias-ronge:
Hi @stweil, The Coverty report doesn't seem to have been updated for more than three years. If we still follow this approach, how will it be updated? I logged in, but I see no way. Is it possible at all?

stweil:
Hi Matthias, yes, that's confusing. https://scan.coverity.com/projects/kitodo-kitodo-production?tab=overview is a dead end and won't be updated because it was superseded by https://scan.coverity.com/projects/kitodo-production-master-branch?tab=overview. @henning-gerhardt, are you planning an update now as 3.0.0 was released?

henning-gerhardt:
Coverity is up again? I did not check this tool anymore after the had a lot of security issues, a new owner and their command line tool works not anymore. I have no plan to re-activate it because there are many other tools in the wild with more features and a better UI. I used a long time [SonarCloud ](https://sonarcloud.io/) as a cloud based solution or Vulas (currently named as [steady](https://github.com/eclipse/steady) as a local working solution for security checks). But no one was really working on the issues discovered by Coverity or SonarCloud :-( 

If @Kathrin-Huber will use Coverity I can transfer it to her or we abandon it.

stweil:
Issue #2349 has also be fixed before external code analyzers can access the whole Kitodo code base.

stweil:
[...]

PR #3162, opened 2020-02-10
Fix a possible integer overflow

matthias-ronge:
An [integer overflow](https://en.wikipedia.org/wiki/Integer_overflow) can occur when a number becomes larger than the maximum allowed, in which case the rest is added to the minimum. This typically only happens with large numbers and is therefore usually inconspicuous in development.

[The weakness](https://lgtm.com/projects/g/kitodo/kitodo-production/snapshot/05c6a0fab4e82861d27f146b8ec4f992c5c55e35/files/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java#x9b6a8b81dfc6fa67:1) was discovered by the LGTM scanner (see issue #320)

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
@@ -191,11 +191,13 @@ public class IndexingService {
      * progress.
      *
      * @return int number of all currently indexed objects
+     * @throws ArithmeticException
+     *             if the value will not fit in a {@code long}
      */
-    public int getAllIndexed() throws DataException {
-        int allIndexed = 0;
+    public long getAllIndexed() throws DataException {
+        long allIndexed = 0;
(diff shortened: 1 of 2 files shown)


henning-gerhardt:
Why not change the return type from `int` to `long` and do not cast the big value to integer back. If you have a big data set then you would to see this number and not a maybe not catched exception.

matthias-ronge:
This should fix both

matthias-ronge:
This should fix both.

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
```diff
@@ -191,11 +191,13 @@ public boolean isIndexCorrupted() throws DAOException, DataException {
      * progress.
      *
      * @return int number of all currently indexed objects
```
henning-gerhardt:
```suggestion
     * @return long number of all currently indexed objects
```

PR #3174, opened 2020-02-12
Log Hibernate exceptions during indexing

matthias-ronge:
It is easily possible to have a Hibernate exception when indexing a database for the first time after the migration from Production 2.x, because the database often still contains legacy errors. To improve troubleshooting, the exception should be logged.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/IndexWorker.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/IndexWorker.java
@@ -15,6 +15,7 @@ import java.util.List;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.hibernate.HibernateException;
 import org.kitodo.config.ConfigCore;
 import org.kitodo.config.enums.ParameterCore;
 import org.kitodo.data.database.exceptions.DAOException;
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/IndexWorker.java
```diff
@@ -76,7 +77,7 @@ public void run() {
                     indexChunks(batchSize);
                 }
             }
-        } catch (CustomResponseException | DAOException e) {
+        } catch (CustomResponseException | DAOException | HibernateException e) {
```
Kathrin-Huber: Log it yes, but catch it?
maybe make an extra catch clause for it, log it and rethrow it.
Shouldn't hibernate exceptions be logged by hibernate anyway?

matthias-ronge: Basically, the result is the same, the thread ends in both cases, and the thrown exception is evaluated nowhere. No, Hibernate only logs very generic message ("LoadContexts - HHH000100: Fail-safe cleanup (collections) : org.hibernate.engine.loading.internal.collectionloadcontext@5d6148a<rs=com.mchange.v2.c3p0.impl.newproxyresultset@29c7bd3 [wrapping: null]>", "HHH000160: On CollectionLoadContext#cleanup, localLoadingCollectionKeys contained [1] entries") that don't really give you the reason for the error, unfortunately. The question would rather be whether the exception should not also be displayed to the user in the user interface (in all three cases)? And, would there something have to be reset so that the indexing buttons are re-enabled?

Issue #3193, opened 2020-02-14
DB Migration V2_6 fails because of wrong IDs


ponchofiesta:
I started to work on our migration from v2 to v3. At first I try to migrate the DB using Flyway. On step [V2_6](https://github.com/kitodo/kitodo-production/blob/kitodo-production-3.0.0/Kitodo-DataManagement/src/main/resources/db/migration/V2_6__Store_properties_only_in_one_table.sql) the IDs of all the property tables are changed to prevent duplicate entries.

I figured out that LAST_INSERT_ID() gets wrong IDs.

The migration step fails with message *Duplicate entry '2' for key 'PRIMARY'* and ends with this state:
- workpieceProperty starting with ID 1 containing 19228 records (OK)
- userProperty is empty (OK)
- templateProperty starting with ID 2 containing 19865 records (WRONG)
- processProperty starting with ID 19868 (OK)

I'm not sure if this can be fixed in our DB before all migration steps or if this is a bug in this migration step.

Flyway error:
```
[ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:5.1.4:migrate (default-cli) on project kitodo-production: org.flywaydb.core.internal.command.DbMigrate$FlywayMigrateException: 
[ERROR] Migration V2_6__Store_properties_only_in_one_table.sql failed
[ERROR] -------------------------------------------------------------
[ERROR] SQL State  : 23000
[ERROR] Error Code : 1062
[ERROR] Message    : Duplicate entry '2' for key 'PRIMARY'
[ERROR] Location   : ./Kitodo-DataManagement/src/main/resources/db/migration/V2_6__Store_properties_only_in_one_table.sql (/home/mri/projects/kitodo-production-tub/./Kitodo-DataManagement/src/main/resources/db/migration/V2_6__Store_properties_only_in_one_table.sql)
[ERROR] Line       : 149
[ERROR] Statement  : INSERT INTO property (id, title, value, obligatory, dataType, choice, creationDate, container)
[ERROR]        SELECT id, title, value, obligatory, dataType, choice, creationDate, container
[ERROR]        FROM workpieceProperty
```

ponchofiesta:
I don't know what the JOINs are for but without them and LAST_INSERT_ID() it works:
```sql
set @rank := 0;

-- for workpieceproperty
ALTER TABLE workpieceProperty
	CHANGE id id INT(11) NOT NULL;
ALTER TABLE workpieceProperty
[...]

PR #3201, opened 2020-02-17
#3193 Use static counter instead of LAST_INSERT_ID()

ponchofiesta:
See #3193


Kathrin-Huber:
I finally could test your PR and it looks good! It actually reseolves the problem of empty tables.
Just a small change, which breaks my system.

PR #3224, opened 2020-02-21
fix task indexing

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
```diff
@@ -37,6 +37,7 @@
 import org.kitodo.production.helper.SelectItemList;
 import org.kitodo.production.model.LazyDTOModel;
 import org.kitodo.production.services.ServiceManager;
+import org.kitodo.production.services.data.TemplateService;
```
solth: `TemplateService` wasn't imported in `TemplateForm` until now? That should have made us suspicous!

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -293,7 +293,7 @@ public TaskDTO convertJSONObjectToDTO(Map<String, Object> jsonObject, boolean re
          * never-ending loops as the list of templates reads the list of tasks.
          */
         int process = TaskTypeField.PROCESS_ID.getIntValue(jsonObject);
-        if (process > 0 && !related) {
+        if (Objects.nonNull(process) && process > 0 && !related) {
```
solth: `int` cannot become `null`. Why do you we need the null check here?

Kathrin-Huber: This is true indeed. I have no idea why I added it.

PR #3248, opened 2020-02-28
use workpieceproperty instead of reading meta.xml

Kathrin-Huber:


Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -262,8 +261,14 @@ public boolean createProcessesWithCalendar(ProcessDTO processDTO) {
         try {
             Process process = ServiceManager.getProcessService().getById(processDTO.getId());
             Integer rulesetId = process.getRuleset().getId();
-            String docType = ServiceManager.getMetsService()
-                    .getBaseType(ServiceManager.getProcessService().getMetadataFileUri(process));
+            List<Property> workpieceProperties = process.getWorkpieces();
+            String docType = "";
+            for (Property workpieceProperty : workpieceProperties) {
```
solth:
```suggestion
            String docType = "";
            for (Property workpieceProperty : process.getWorkpieces()) {
```

PR #3268, opened 2020-03-06
Add exported flag

Kathrin-Huber:


Kitodo-DataManagement/src/main/resources/db/migration/V2_91__Add_exported_flag.sql
```diff
@@ -0,0 +1,11 @@
...
+-- (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+--
+-- This file is part of the Kitodo project.
+--
+-- It is licensed under GNU General Public License version 3 or later.
+--
+-- For the full copyright and license information, please read the
+-- GPL3-License.txt file that was distributed with this source code.
+
+ALTER TABLE process ADD COLUMN exported TINYINT(1) DEFAULT 1;
```
henning-gerhardt: Is setting this flag to 1 (aka true?) as default value a good decision? This would mean that every new process is marked already as exported which is not true.

If default value 1 means false then ignore this comment.

Kathrin-Huber: You are right, I changed it (y)

PR #3287, opened 2020-03-13
Check for existing parent process during import even with import depth 1

### LINKED ISSUE #3257: Handle parent search even on "Importtiefe 1"
The check for an existing parent in Database should happen, even if the import depth is set to 1.

solth:
Always try to find a parent process in the database for the last process in the process hierarchy to link to during OPAC import, even if `importDepth=1`.

Fixes #3257


subhhwendt:
The specific case when "Importtiefe=1" was chosen with a volume from a multivolumeWork, which is already imported, fails in linking to the multivolumeWork  - in my experience.
The relation to the parent id is detected in metadata - but the volume shows no connection to the parentelement in the database.
grafik

grafik

PR #3324, opened 2020-03-20
Improve parsing of data copying rules

matthias-ronge:
In particular, this allows spaces to appear within quotation marks. Basically, this is more stable against leading, double or trailing spaces. Still, a space is needed between the segments.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/metadata/copier/DataCopier.java
+++ b/Kitodo/src/main/java/org/kitodo/production/metadata/copier/DataCopier.java
@@ -12,8 +12,9 @@
 package org.kitodo.production.metadata.copier;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/metadata/copier/DataCopier.java
```diff
@@ -42,13 +45,39 @@
...
-        List<String> commands = Arrays.asList(program.split(";"));
+        List<List<String>> commands = parse(program);
         rules = new ArrayList<>(commands.size());
-        for (String command : commands) {
+        for (List<String> command : commands) {
             rules.add(DataCopyrule.createFor(command));
         }
     }
 
+    private static List<List<String>> parse(String input) {
```
Kathrin-Huber: please give a more descriptive name. what are you parsing and what are you parsing it for?

PR #3345, opened 2020-03-24
Remove unused project fields

matthias-ronge:
The project properties `useDmsImport` and `dmsImportCreateProcessFolder` should always be true. Removes the fields from the database, removes all corresponding if’s and a lot of dead code.

Depends on prerequisite branch PR: #3344


Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -15,6 +15,7 @@
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.rmi.server.ExportException;
```
henning-gerhardt: Why do you not create a new exception class as this [class ](https://docs.oracle.com/javase/7/docs/api/java/rmi/server/ExportException.html)has nothing to do with Kitodo.Production export.

Kitodo/src/main/java/org/kitodo/exceptions/ExportException.java
```diff
@@ -0,0 +1,27 @@
...
+public class ExportException extends RuntimeException {
+
+    /**
+     * Constructs an {@code ExportException} with the specified detail message.
+     *
+     * @param s the detail message
+     */
+    public ExportException(String s) {
+        super(s);
+    }
```
Kathrin-Huber:
```suggestion
     * @param exceptionMessage the detail message
     */
    public ExportException(String exceptionMessage) {
        super(exceptionMessage);
    }
```

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -70,23 +64,22 @@ public ExportDms(boolean exportImages) {
...
+     *            process to export
+     * @param unused
+     *            user home directory
      */
     @Override
-    public boolean startExport(Process process, URI destination) {
-        if (process.getProject().isUseDmsImport()
-                && ConfigCore.getBooleanParameterOrDefaultValue(ParameterCore.ASYNCHRONOUS_AUTOMATIC_EXPORT)) {
-            TaskManager.addTask(new ExportDmsTask(this, process, destination));
+    public boolean startExport(Process process, URI unused) {
```
Kathrin-Huber: if it is unused, remove it

matthias-ronge: This method overrides a method in a superclass which uses it. Therefore I cannot remove it.

PR #3374, opened 2020-04-01
Update import module to make it usable for arbitrary search queries by URL

### LINKED ISSUE #3373: Generalize SRUImport module
At the moment, the SRUImport module can only be used to query SRU interfaces. This should be changed so it can be used for arbitrary interfaces that can be queried via search URLs.

solth:
- rename `SRUImport` to `QueryURLImport`
- add attribute `prefix` to element `identifierParameter` in `kitodo_opac.xml` (specifies an optional prefix to be prepended before IDs when submitting queries by identifier)
- add enum `InterfaceType` to allow distinguishing between multiple types and available interface parameters (right now `SRU` and `OAI`)

Closes #3373


Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
```diff
@@ -130,17 +133,24 @@ public DataRecord getFullRecordById(String catalogId, String identifier) throws
...
+                if (Objects.nonNull(interfaceType)) {
+                    if (Objects.nonNull(interfaceType.getStartRecordString())) {
+                        queryString = queryString + interfaceType.getStartRecordString() + equalsOperand + "0&";
+                    }
+                    if (Objects.nonNull(interfaceType.getMaxRecordsString())) {
+                        queryString = queryString + interfaceType.getMaxRecordsString() + equalsOperand + rows + "&";
+                    }
+                    if (Objects.nonNull(interfaceType.getQueryString())) {
+                        queryString = queryString + equalsOperand;
+                    }
```
Kathrin-Huber, line 151: How does the condition effect the execution of the concatenation?

solth: Well spotted, this is one place where I forgot to add the return value of `interfaceType.getQueryString()` itself to complete query string. Fixed it now.

PR #3375, opened 2020-04-02
do passwort Validation onSave

Kathrin-Huber:
Due to the fact, that the setter is called many times from frontend, the passwortd-validation always hooks in, even if the passwort is not changed.
This PR moves the validation to the save method.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
@@ -20,8 +20,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-
 import java.util.Set;
+
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.SessionScoped;
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
```diff
@@ -130,8 +134,15 @@ public String newUser() {
      * @return page or empty String
      */
     public String save() {
-        String login = this.userObject.getLogin();
+        Set<ConstraintViolation<KitodoPassword>> passwordViolations = getPasswordViolations();
+        if(passwordViolations.isEmpty()) {
```
oliver-stoehr, line 138:
```suggestion
        if (!passwordViolations.isEmpty()) {
```
I think this should be negated.

Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
```diff
@@ -21,12 +21,17 @@
 import java.util.Map;
 import java.util.Objects;
 
+import java.util.Set;
```
oliver-stoehr, line 24:
```suggestion
import java.util.Set;

```

PR #3395, opened 2020-04-06
fix error and add validation for password

Kathrin-Huber:


Kathrin-Huber:
> Why did you move these message keys from the error resource bundle to the message resource bundle?

Because, the error message was not found. It was already broken in the "change password" dialog and it looked something like this before:
Unbenannt

solth:
I think it was just not found because the wrong resource bundle name was used. Instead of 
`msgs.passwordsDontMatch` it should have been `err.passwordsDontMatch`, shouldn't it?
See also: 
https://github.com/kitodo/kitodo-production/blob/14a33e629a9680614745a57e3d81038b2c3a7e28/Kitodo/src/main/webapp/WEB-INF/faces-config.xml#L27

PR #3404, opened 2020-04-08
Fix 'metsFromKitodo.xml' test file

### LINKED ISSUE #3329: Test METS file not well-formed and without license header
The file [Kitodo/src/test/resources/metsFromKitodo.xml](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/test/resources/metsFromKitodo.xml) ist not well-formed. The error message in line 24 reads: _cvc-complex-type.2.4.b: The content of element 'mets:xmlData' is not complete. One of '`WC[##any]`' is  expected._

The file is also missing the license header.

solth:
Fixes #3329


solth:
You are right, manually running `XsltHelperTest.shouldTransformKitodoToMods` - the only test where this XML file is used - shows the result produced by the XSL transformation is not equal to the contents of the updated file. I will check the transformation result again and make the necessary adjustments.

solth:
The license header _did_ make a difference, even though it is just a comment.

@matthias-ronge could you review the changes again?

PR #3457, opened 2020-04-16
Sort processes for desktop window right after index query

### LINKED ISSUE #3296: Desktop should select the correct process
From the desktop page, when you open a process in the metadata editor, the wrong process opens.

**Goal:** The selected process should be opened. (The same should apply to deletion; needs review)

matthias-ronge:
Fixes #3296

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/processesWidget.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/processesWidget.xhtml
@@ -20,8 +20,6 @@
                  value="#{DesktopForm.processes}"
                  scrollable="true"
                  styleClass="default-layout"
-                 sortBy="#{process.id}"
-                 sortOrder="descending"
                  var="process">
         <p:column headerText="#{msgs.id}"
                   width="50"


Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
```diff
@@ -99,6 +100,7 @@ public DesktopForm() {
         try {
             if (ServiceManager.getSecurityAccessService().hasAuthorityToViewProcessList() && processList.isEmpty()) {
                 processList = ServiceManager.getProcessService().loadData(0, 10, SORT_ID, SortOrder.DESCENDING, null);
+                Collections.sort(processList, (one, other) -> other.getId() - one.getId());
```
solth, line 103: Why is this neccessary? The processes are already returned in a sorted list, as indicated by the `SORT_ID` constant passed as the third parameter to `loadData` in the previous line, aren't they?

If you just want to reverse the order and start with the lowest ID, you should use `SortOrder.ASCENDING` instead of `DESCENDING`, I guess.

Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/processesWidget.xhtml
```diff
@@ -20,7 +20,6 @@
                  value="#{DesktopForm.processes}"
                  scrollable="true"
                  styleClass="default-layout"
-                 sortBy="#{process.id}"
```
solth, line 23: If you remove `sortBy`, you should probably also remove the following line, because `sortOrder` doesn't make sense without `sortBy`.

PR #3460, opened 2020-04-16
Localized Workflow Editor

### LINKED ISSUE #3338: Translate workflow editor
The workflow editor is largely not translated. This affects both the tooltips for the mouse menu and the small symbols to insert further steps, as well as most of the "Properties" dialog with its content.

Screenshot

Erikmitk:
This PR resolves #3338. I added a new editor build that supports localization. I extended the `WorkflowForm` to get the current user's language so the editor knows which translations to use.

This value only changes when the user logs out and in again?

Currently, all localization strings are in the editor's repo https://github.com/Erikmitk/kitodo-workflow-editor/tree/master/src/language.

Screen Shot 2020-04-16 at 20 56 59

Screen Shot 2020-04-16 at 20 57 25


solth:
Fixes #3338

PR #3461, opened 2020-04-16
Ignore empty <fileGrp> when reading METS file

### LINKED ISSUE #3173: Empty METS file cannot be read when saved in 2.x
If a process without images was saved in 2.x, it contains an empty `LOCAL` `<fileGrp>`. (Unlike freshly imported processes which do not even contain a `<fileSec>`.) Then Production 3 can no longer open it (also not automatically, for example during the migration).
An empty `<fileGrp>` must be ignored, but not crash.

matthias-ronge:
Fixes #3173


Kathrin-Huber:
Please resolve conflicts

PR #3495, opened 2020-04-21
Add cancel button to 'changeDocType' dialog

solth:
Cancel button added:

Bildschirmfoto 2020-04-21 um 14 19 45

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/dialogs/changeDocStrucType.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/dialogs/changeDocStrucType.xhtml
@@ -54,7 +54,7 @@
                                      oncomplete="updateMetadataEditorView(false);"/>
                     <p:commandButton value="#{msgs.cancel}"
                                      icon="fa fa-times fa-lg"
-                                     iconPos="secondary"
+                                     iconPos="right"
                                      styleClass="secondary right"
                                      onclick="PF('dialogEditDocStrucType').hide();"/>
                 </h:panelGroup>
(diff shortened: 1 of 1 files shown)


solth:
> I don’t see the use case for a “cancel” button, as it is sufficient not to change the drop-down and click OK

Yes, I thought you might say that, but having to click 'Ok' or 'Apply' even if I changed my mind and do not want to change is counterintuitive. It is better to make it explicit that nothing will be changed. 

The additioal button also brings the dialog in line with all the other popup dialogs we have in the system for performig specific action, that all have a dedicated cancel button to give the user an unambiguous option to cancel the action.

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/dialogs/changeDocStrucType.xhtml
```diff
@@ -28,28 +27,38 @@
...
+                                     value="#{msgs.apply}"
+                                     action="#{DataEditorForm.changeDocStrucTypeDialog.editDocStruc}"
+                                     styleClass="primary right"
+                                     icon="fa fa-check fa-lg"
+                                     iconPos="right"
+                                     update="structureTreeForm:logicalTree,paginationForm:paginationWrapperPanel,metadataAccordion:logicalMetadataWrapperPanel,commentWrapperPanel,galleryWrapperPanel"
+                                     oncomplete="updateMetadataEditorView(false);"/>
+                    <p:commandButton value="#{msgs.cancel}"
+                                     icon="fa fa-times fa-lg"
+                                     iconPos="secondary"
```
oliver-stoehr, line 57: Should this be `right`?

solth: You are right, it should of course be `right`, not `secondary`. 

Funny enough, it seems any value but `left` results in the icon on the right side:
- `iconPos="left"` -> Icon on the **left**
- `iconPos="right"` -> Icon on the **right**
- `iconPos=""` -> Icon on the **right**
- `iconPos="foobar"` -> Icon on the **right**
- no `iconPos` attribute at all -> Icon on the **left**

PR #3513, opened 2020-04-22
remove unused userHome URI

Kathrin-Huber:
If the export is started from an automatic task, no user is found because everything happens in a Thread. an NPE occured
As the URI is not used in the method, null is enough.


matthias-ronge:
This is also the error (`NullPointerException`) that Travis ran into.

PR #3534, opened 2020-04-24
Show metadata for orientation

### LINKED ISSUE #3406: Metadata for “Select insertion position”
### Problem
If a new process for a periodical volume is created, it is necessary to define the position for its insertion in the list of other volumes. To support that task, curently the ID is shown as mouse over-effect. 

zeitschriftenbandeinfügen02

Practically, the user does not know the internal IDs of processes. If the processes for the periodical volumes are generated in the correct order, it works. Otherwise (the more likely case) not. 

### Solution
It would be helpful, if a metadata field is used for the mouse over effect. The opportunity to enable a configuration by the users would be the best solution – for example by an attribute in the ruleset use=”sortingNumber”.

matthias-ronge:
This displays metadata for "select insertion position". The metadata can be selected in the ruleset with `use="displaySummary"`.

Fixes #3406

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/TitleRecordLinkTab.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/TitleRecordLinkTab.java
@@ -260,8 +260,8 @@ public class TitleRecordLinkTab {
                 String value = MetadataEditor.getMetadataValue(rootElement, key);
 
                 if (Objects.nonNull(value)) {
-                    Optional<String> label = ruleset.getTranslationForKey(rootElement.getType(), priorityList);
-                    toolTip.add(label.orElse(rootElement.getType()) + ": " + value);
+                    Optional<String> label = ruleset.getTranslationForKey(key, priorityList);
+                    toolTip.add(label.orElse(key) + ": " + value);
                 }
             }
         }
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/TitleRecordLinkTab.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/TitleRecordLinkTab.java
@@ -14,18 +14,21 @@ package org.kitodo.production.forms.createprocess;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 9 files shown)


Kathrin-Huber:
Where exactly can I configure it in the ruleset? I was trying:

        ```
<restriction division="Volume" unspecified="forbidden">
            <permit key="TSL_ATS" maxOccurs="1" minOccurs="1" use="displaySummary"/>
</restriction>
```

But it's always showing the processtitle and Id in square-brackets like
`WiDuMi_23988394938 [42]`

**Comment by matthias-ronge** [comment]
The `use` attributes are always on the `<key>`s (or `<division>`s), not on the `<permit>`s. For example:
```
<key id="TSL_ATS" use="displaySummary">
```
This reminds me that I have to extend the XSD file ...

**Comment by matthias-ronge** [comment]
I find it annoying that we have the file in the repository twice. The two files were also available in different versions. I have adjusted them.

File: Kitodo/src/main/java/org/kitodo/production/forms/createprocess/TitleRecordLinkTab.java [hunk 1]
```diff
@@ -225,6 +233,47 @@ private void createInsertionPositionSelectionTreeRecursive(String positionPrefix
...
+
+            final String metadataLanguage = ServiceManager.getUserService().getCurrentUser().getMetadataLanguage();
+            List<LanguageRange> priorityList = Locale.LanguageRange.parse(metadataLanguage);
+
+            for (String key : summaryKeys) {
+                String value = MetadataEditor.getMetadataValue(rootElement, key);
+
+                if (Objects.nonNull(value)) {
+                    Optional<String> label = ruleset.getTranslationForKey(rootElement.getType(), priorityList);
+                    toolTip.add(label.orElse(rootElement.getType()) + ": " + value);
```
Kathrin-Huber, line 264:
```suggestion
                    toolTip.add(key + ": " + value);
```

Kathrin-Huber, line 264:
```suggestion
                    toolTip.add(key + ": " + value);
```

Kathrin-Huber:
I would suggest to use the key as label instead of the type. Because otherwise I have the type information several times but not about the key
Unbenannt1
Unbenannt

PR #3553, opened 2020-04-29
use contains instead of startswith

### LINKED ISSUE #3551: search for image-number and page-number in "new docstruct"
### Problem
When a new docstruct is created, the images are usually assigned by the page-number of the original. In Kitodo.Production 2.x it was therefore possible to search in the fields "first page" and "last page" for the image-number and the page number. 
In Kitodo.Production 3.x, only the image-number can be searched. Thus, the image-number needs to be derived from the structure tree. 

### Solution
It is more convenient, when both image-number and page-number are indexed in "first page" and "last page". 

### Example
Search for 10 -> Only image numbers containing 10 are found: 
seitenzuweisen01

Page number 10 does exist: 
seitenzuweisen02

In Kitodo.Production 2.x both numbers are offered: 
seitenzuweisen03

Kathrin-Huber:
fixes #3551


Kathrin-Huber:
of course! (facepalm)

PR #3582, opened 2020-05-05
Add MetadataGroup to index

### LINKED ISSUE #3565: metadata of metadata group cannot be retrieved by search or filter
### Problem
The metadata of metadata groups, mainly person or corporate metadata is neither entered in the properties nor indexed. Thus, the metadata cannot be retrieved. 

### Solution
The metadata of metadata groups should be retrievable via search and filter.  

### Example
Name in metadata group
- process: DerMe_165881360X_Einfacher_Strukturbaum (ID: 950)
- search term: Händel, Georg Friedrich

Series in metadata group
- process: NeunJa_142865559X_0002_komplett_Einfacher_Strukturbaum (ID: 1218)
- search term: Beiträge zur Kulturgeschichte

Kathrin-Huber:
Makes Metadatagroups searchable by searchbar
fixes #3565  (the part "retrivable by filter" is not fixed, because filters don't respect metadata at all)


Kathrin-Huber:
Thanks for noticing, I only adjusted the problems in test/mapping.json 🤦

PR #3587, opened 2020-05-05
avoid NPE on missing process

Kathrin-Huber:
I cannot determine, how I'm producing the NPE, but it occurs from time to time, so I added a nullcheck.


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -172,6 +173,8 @@
      */
     private List<Pair<MediaUnit, IncludedStructuralElement>> selectedMedia;
 
+    private static final String DESKTOP_LINK = "/pages/desktop.jsf";
```
solth, line 176: We do already have a `referringView` in `DataEditorForm.java`. Perhaps it would be better to redirect there (which is on default also set to `desktop`) instead of statically always leading to `DESKTOP_LINK`?

PR #3609, opened 2020-05-08
Remove unused code

matthias-ronge:
I thought there was a lot of dead code here. These were remnants of old code, which allowed the selection for internal file format and export format. Since the internal file format is now only METS, and the same applies to export, this is no longer necessary. I removed this.

Follow-up pull request to #3607


Kathrin-Huber:
please rebase against master to fix failing tests

Kathrin-Huber:
The migration file has the wrong number, please change to V2_95

matthias-ronge:
Another selenium test fails here and I don't know why. What else would I have to adjust?

PR #3617, opened 2020-05-12
add id to index to enable sorting by id

### LINKED ISSUE #3417: Sorting of process list by ID does not work
The sorting of the process list by ID does not work. It is not clear by which criteria the processes are sorted

Kathrin-Huber:
To sort by id, it needs to be added to the indexed object. Sorting by object id in index (_id) is not possible.
fixes #3417

### DIFF
--- a/Kitodo/src/main/resources/mapping.json
+++ b/Kitodo/src/main/resources/mapping.json
@@ -302,6 +302,9 @@
         "process": {
             "dynamic": "strict",
             "properties": {
+                "id": {
+                    "type": "text"
+                },
                 "batches": {
                     "properties": {
                         "id": {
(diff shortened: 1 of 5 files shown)


henning-gerhardt:
Tests are failing
```
Failed tests:   shouldCreateDocumentWithCorrectAmountOfKeys(org.kitodo.data.elasticsearch.index.type.ProcessTypeTest): Amount of keys is incorrect! expected:<27> but was:<28>
```

Kitodo/src/main/resources/mapping.json
```diff
@@ -302,6 +302,9 @@
         "process": {
             "dynamic": "strict",
             "properties": {
+                "id": {
+                    "type": "text"
```
oliver-stoehr, line 306:
```suggestion
                    "type": "long"
```
`id` should be of type `long` because `text` does not work with default settings:

> Fielddata is disabled on text fields by default. Set fielddata=true on [id] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory.

PR #3620, opened 2020-05-12
Select node on secondary click and update context menu

### LINKED ISSUE #3470: Metadata editor: Context menu might open for wrong node
Opening the context menu in the metadata editor via secondary click does not select the node clicked on. A preceding primary click on the node is necessary to select the node.
Otherwise the node is not selected and the context menu shows options for the previously selected node.

A secondary click should update the node selection before displaying the context menu.

### LINKED ISSUE #3477: Correction of the click way for the allocation of one page to two structure elements
If you want to get the proper menu item ("Mit nächstem Element verknüpfen") by the right mousekey in the structure tree - you need to click at first at another place in the structure tree or to use the left mousekey for the chosen page before you can use the right mousekey with the complete expected menu
grafik

oliver-stoehr:
Resolves #3470 and resolves #3477


solth:
Fixes #3470 and fixes #3477, doesn't it?

Issue #3640, opened 2020-05-18
Import breaks user interface when person has multiple roles


matthias-ronge:
Example: GBV PPN 1041657277
Here the person has 2 roles, author and artist.

Error message:
```
Type error evaluating ($role) in xsl:when/@test on line 209 column 70 
  XPTY0004: A sequence of more than one item is not allowed as the first argument of
  fn:starts-with() ("VerfasserIn", "KünstlerIn") 
  In template rule with match="element(Q{http://www.loc.gov/mods/v3}name)[xs:string(@type) eq "personal"]" on line 196 of 
     invoked by xsl:apply-templates at #28
  In template rule with match="element(Q{http://www.loc.gov/mods/v3}mods)" on line 24 of 
     invoked by built-in template rule (text-only)
```

**Goal:** XSLT has to be adjusted accordingly.

In addition: The error message must be caught and displayed correctly.

solth:
> In addition: The error message must be caught and displayed correctly.

This part is already working (screenshot from preview system that has the current version of the software installed):
![xslt-error-message](https://user-images.githubusercontent.com/19183925/82285041-bc37ac80-999a-11ea-86ef-7bd9739f3bc6.png)

Kathrin-Huber:
please reopen, if this is still a problem

PR #3643, opened 2020-05-18
Show error message in case that XSLT fails

matthias-ronge:


Kathrin-Huber:
please fix codacy

matthias-ronge:
As @solth says [that already works](https://github.com/kitodo/kitodo-production/issues/3640#issuecomment-630573588), the change is then superfluous.

PR #3684, opened 2020-05-22
Ensure that no bogus warning is displayed

matthias-ronge:


Kathrin-Huber:
There is no difference in execution here!

matthias-ronge:
The warning below (_Ruleset declares undefined field use…_) also appeared when the correct value was in the rule set. The reason for this was that only the inner (searching) loop was `break`ed, so the following command was executed and the log message printed. In this case, this is bogus because the search was successful.

The solution is that the outer loop is `continue`d so that the log command is skipped. That does work, I checked it.

Kathrin-Huber:
so please extract the inner loop in a separate method to enforce readability.

solth:
thumb_i-killed-a-man-reworked-old-meme-tuse-goto-dude-67046876

Issue #3671, opened 2020-05-20
After a script task has finished, the process is updated slowly


matthias-ronge:
Extracted from [comment on issue #3279](https://github.com/kitodo/kitodo-production/issues/3279#issuecomment-626210008) by @mnscholz:
> Although the script finishes instantly (< 1s), the process is updated/advanced in Kitodo only after a very long time (10-30min). Why is there so much idle time?

solth:
Could this perhaps be related to and fixed by #3687 ?

matthias-ronge:
Needs to check whether that is currently still the case.

andre-hohmann:
@matthias-ronge: Does the problem still persists?

"Our" automatic tasks (Image validation, generation of Images, ...) are processed quickly and also the status of the process is quickly updated.

matthias-ronge:
I don't know at the moment. There are no shell scripts in the Production 3 systems that I maintain. The problem here was with external shell scripts in automatic tasks, not with internal functions like generation of images.

@m-kotzyba, you use a lot of scripts in your workflow, what have you observed in this regard?

m-kotzyba:
Currently we run just a “MD5 Check” and  a “Remove all images from thumbs, max and in the meta.xml” (in case of faulty scans) as external shell scrip but these scripts are not automated but manually triggered. 

However, while the (productive) work with Kitodo and also (all) my tests, I did not experienced the described delay of 10-30 min. in the process status update (yet).

matthias-ronge:
So I assume #3687 did fix this. I will close the ticket.

PR #3687, opened 2020-05-22
Prevent automatic tasks from generating images if it is not their task

matthias-ronge:


solth:
Does this solve #3671 ?

PR #3698, opened 2020-05-26
Catch IllegalArgumentException

matthias-ronge:


henning-gerhardt:
I'm a little bit unsure: 

In the thrown IllegalArgumentException message in https://github.com/kitodo/kitodo-production/blob/master/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/UseXmlAttributeAccess.java#L53-L69: there is no information about path to the meta data file or any process information which would be help to locate the corrupt meta data file.

Maybe I must test this pull request with real data but this will take time.

matthias-ronge:
Yes, thanks for the hint. I believed that the process number would be logged.

henning-gerhardt:
Ok, I get this change a try on a copy of productive migrated data.

Good thing are that with this change the exception message is catched, logged into the defined log file and indexing of the next process is startet instead of skipping to the next index batch.

Bad thing is: without setting log level to at least `info` you have still no context information where this error happening. So the error message needs to improved to work as stand alone message. Even with a log level on `info` you can not assume that message before belongs to the next line as concurrent actions can create at the same time an other log message.

With log level `info` and no other action running, the log is the following:

```
[INFO ] 2020-05-27 10:15:34.223 MetsService - Reading 5243/meta.xml
[ERROR] 2020-05-27 10:15:34.249 ProcessService - Could not determine base type for process: Illegal character in path at index 28: file://./deutreubd_30162372Z _1880_004_tif/00000001.tif
[INFO ] 2020-05-27 10:15:34.254 MetsService - Reading 5244/meta.xml
[INFO ] 2020-05-27 10:15:34.306 MetsService - Reading 5245/meta.xml
[ERROR] 2020-05-27 10:15:34.341 ProcessService - Could not determine base type for process: Illegal character in path at index 28: file://./deutreubd_30162373Z _1880_002_tif/00000001.tif
```

matthias-ronge:
I just changed that. I also lowered the log type to _info_ because this is normal operation.

Background: The DocType is now also indexed (if it is readable). To do this, the METS file must be readable. If it cannot be read, nothing will be indexed.

matthias-ronge:
By the way, you have a space in your image paths, that's why the error occurs. As far as I can see, paths with spaces are not yet fully supported (also in other places, like the scripts). But that would be a separate problem that we would have to address in a separate issue.

henning-gerhardt:
Your last change changed the log output to

```
[INFO ] 2020-05-27 10:44:16.757 ProcessService - Could not determine base type for process verhdesal_20091718Z_1929_001 [685]: Corrupt file: <mets:fileGrp USE="LOCAL"> contains differing MIMETYPE values: , image/tiff
```

which is helpful. I'm not sure if changing the log level from `error` to `info` is a good idea. In normal case no one is running the application on productive environments on `info` log level.

matthias-ronge:
But normal case is to use "" in this case. From this perspective, there is no error here.

henning-gerhardt:
I have a different opinion regarding to empty attributes (should not exist at all or the XML schema is wrong!) or attributes with wrong content should be corrected  (`MIMETYPE` attribute should contain correct content).

Issue #3307, opened 2020-03-18
Higher-level processes should not have a workflow


matthias-ronge:
Currently, higher-level processes get the same workflow as their children, and tasks, and the tasks are open for operator processing, as is usual at the beginning of workflow.

**Goal:** There is nothing to do with higher-level processes. They should be hidden (like completed processes) and have no tasks, especially no open tasks. Solution is subject to discussion, no production template seems the most logical. But can also be a dummy workflow without steps, or whatever.

matthias-ronge:
Twinned with issue #3190

matthias-ronge:
I'm reopening this ticket because newspaper process has tasks when created. And, progress bar is shown, as for a normal process. It is configured correctly and the calendar is offered as an option. Nevertheless, the newspaper process has received the workflow tasks and these can be taken over.

**Goal:** An overall newspaper process should not produce any tasks that can be taken over.

PR #3707, opened 2020-05-27
create processes with noworkflow configured

Kathrin-Huber:
This is the first part of the issue #3307 
Works now for creation of processes with the configured doctype.

ToDo: make it work on automatic generated processes (e.g. parent processes on hierarchical import)


matthias-ronge:
The integration test fails because `ProcessValidator.isContentValid()` becomes `false`.

Screenshot

What I don't understand is why it hasn't happened before. If I see it correctly, you haven't changed anything at all.

PR #3753, opened 2020-06-10
respect noWorkflow in automatic generated processes

### LINKED ISSUE #3307: Higher-level processes should not have a workflow

Kathrin-Huber:
fixes #3307

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
@@ -1032,6 +1032,12 @@ public class ImportService {
         updateTasks(process);
     }
 
+    /**
+     * Checks if an imported Process should be created with Tasks and removes them if not,
+     * depending on the configuration of the doctype.
+     * @param process the process to check.
+     * @param docType the doctype to check in the ruleset.
+     */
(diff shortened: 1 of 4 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -1020,16 +1020,27 @@ public static void updateTasks(Process process) {
...
                 acquisitionStage, priorityList);
         String docType = tempProcess.getWorkpiece().getRootElement().getType();
         createProcessTitle(tempProcess, managementInterface, acquisitionStage, priorityList, processDetails);
         Process process = tempProcess.getProcess();
         addProperties(tempProcess.getProcess(), template, processDetails, docType, tempProcess.getProcess().getTitle());
+        checkTasks(process, docType);
         updateTasks(process);
     }
 
+    public static void checkTasks(Process process, String docType) throws IOException, RulesetNotFoundException {
```
solth, line 1035: Add missing Javadoc

PR #3769, opened 2020-06-17
Prevent ViewExpiredException by saving ViewState in client

### LINKED ISSUE #3444: Establishing a new thumbnail view in the gallery during the tagging process is buggy
You enter a steady "F5-situation" with incomplete thumbnails when you have chosen images in the gallery and the proper structure element  - completing the action by the button "Übernehmen".

solth:
Fixes #3444 by saving the `ViewState` of a page in the client instead of the server. (see last sentence in https://stackoverflow.com/a/8474609/5467214) 

Since this is quite a fundamental configuration change it should be scrutinized and discussed by as many developers as possible.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/web.xml
+++ b/Kitodo/src/main/webapp/WEB-INF/web.xml
@@ -24,7 +24,7 @@
 
     <context-param>
         <param-name>javax.faces.STATE_SAVING_METHOD</param-name>
-        <param-value>server</param-value>
+        <param-value>client</param-value>
     </context-param>
 
     <context-param>


solth:
I don't think security is an issue here since
- only the `ViewState` is saved in the client, not some critical secret data
- the value of the `ViewState` in the client is encrypted (see comment of BalusC in linked stack overflow answer)
- submitted form data is always validated on the server side in JSF, independent of the `ViewState` (setting the value of a menu to a fabricated option that wasn't provided by the server does not work for this reason, for example, because JSF detects that the option submitted by the client does not correspond to any option on the menu in the backing bean)

> How many times this `ViewExpiredException` is thrown? Why not catch this exception and handle this exception in a clean way? Maybe later someone can fix the reason why this exception is thrown.

The problem here is that we _can't_ catch the `ViewExpiredException`. It happens when clicking on a thumbnail in the gallery after creating a structural element from selected pages (see @subhhwendt s screencast) and before any of our code is invoked (where we could catch the exception). We have tried to find cause for the problem but were unsuccessful.

henning-gerhardt:
I still thinking that this change is a security relevant change. Encryption used in MyFaces (or any other) is a synchronous encryption which works on both ways. If you know the used algorithm and parameters for this encryption (see [StateUtils.java](https://github.com/apache/myfaces/blob/master/impl/src/main/java/org/apache/myfaces/application/viewstate/StateUtils.java) for default MyFaces values)  you can easily decrypted any encrypted string, modify this data and send it back. So encryption is not strong and not a security feature. It is only hiding some information on a non human readable way.

If you store the data on server side only a reference name (like `EA9626C8741F7DAE00000001`) is send to client and send back. But if you store the data on client side a serialized object is send to client and will be de-serialized on server side and interpreted / executed. If you are good in your job and knows some security issues in used libraries you will found a way to execute other code then application code.

henning-gerhardt:
Regarding to security: 

1) The argument from Arved with pointing to the same stackoverflow article that someone is saying that it is "It's encrypted/enctyptable." and a few month later someone else is adding an [article](http://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/) how to miss use this client state save - which is still possible if you are using the right security holes - did not speak in my opinion for saving the state on client site.
2) There is practical no safe security if you store the data on client side with the [default MyFaces values](https://github.com/apache/myfaces/blob/master/impl/src/main/java/org/apache/myfaces/application/viewstate/StateUtils.java#L92-L93) as there is no secure token or anything else is in use.
3) You can change it to other values with maybe some initialization vectors but this must be done on every instance of Kitodo.Production separately without any public documentation what you change. I don't think that this will be done everywhere and that the amount of useful combination is limited it is only a matter of time to break it.

matthias-ronge:
I just looked at that in the debugger. A new key is generated each time Tomcat is started. In fact, it happens [here](https://github.com/apache/myfaces/blob/master/impl/src/main/java/org/apache/myfaces/application/viewstate/StateUtils.java#L667). A different key is also generated for several starts, I tried that.

Erikmitk:
I'm pretty unsure.

I understand @henning-gerhardt's concern regarding sending serialized objects back and forth while giving the client the possibility to change that encoded state value (though encrypted). Certainly, things can go wrong if a malicious piece of data gets de-serialized again.

On the other hand, I don't see the significance in this context. I cannot pinpoint this to any formal proof though. The handling of tokens ad cryptography seems to be robust enough?

solth:
Tbh., I do not consider setting the `STATE_SAVING_METHOD` to `client` being a fix rather than a workaround. The reason for the view state being lost is still in there and needs to be fixed. 

> If I understand [the view management of JSF](https://stackoverflow.com/a/16050424) correctly, I suspect that the view of the metadata editor will be thrown out of the list when the thumbnails are reloaded. (After confirming the dialog box, [the animation](https://github.com/kitodo/kitodo-production/issues/3444#issuecomment-636175420) shows the loading wheel for all thumbnails for a short time!) There are two options: saving the view on the client side or increasing the number of views held on the server. In the second case, the number would then have to be greater than the largest expected number of all thumbnails of all simultaneous users combined. Depending on the number of users, this could be a lot.

I don't think that is what is happening here. If I understand it correctly, the individual thumbnails - even though they are loaded asynchronously via AJAX - are not individual views and do not count towards the total number of views saved on the server. You can easily verify this be reproducing the error with a process that only has a handful of images. IIRC, @oliver-stoehr was able to trigger the `VIEW_EXPIRED_EXCEPTION` with a process containing as few as 5 or 6 thumbnails.

Instead, I believe the `VIEW_EXPIRED_EXCEPTION` just masks the real error occuring when the `Add doc struct` dialog is closed. We had this problem with various bugs lately: the error page shows a `VIEW_EXPIRED_EXCEPTION`, but when inspecting the logs, there was always another error happining earlier, provoking a redirect to the `error` page (where, for some reason, the previous view always throws a `VIEW_EXPIRED_EXCEPTION`). Only, the log does not seem to contain anything helpful this time. :weary:

matthias-ronge:
#3884 may be related

henning-gerhardt:
> Instead, I believe the VIEW_EXPIRED_EXCEPTION just masks the real error occuring when the Add doc struct dialog is closed. We had this problem with various bugs lately: the error page shows a VIEW_EXPIRED_EXCEPTION, but when inspecting the logs, there was always another error happining earlier, provoking a redirect to the error page (where, for some reason, the previous view always throws a VIEW_EXPIRED_EXCEPTION). Only, the log does not seem to contain anything helpful this time.

My question is: did changing the saving state to client help to find the real issue or was this still hidden and / or not visible in the log file?

Erikmitk:
> Tbh., I do not consider setting the STATE_SAVING_METHOD to client being a fix rather than a workaround.

That would mean that this PR is just a basis of discussion to find the real problem, right? And as @henning-gerhardt said this change should be used temporally in a development context to narrow the issue down.

solth:
> That would mean that this PR is just a basis of discussion to find the real problem, right? And as @henning-gerhardt said this change should be used temporally in a development context to narrow the issue down.

Yes, it is mainly thought as to start a discussion - that's why this PR is still set to "draft" and why I requested reviews from so many developers. It may also be used a temporary last resort if we can't figure out what causes the problem.

Erikmitk:
Has someone tried this approach for catching the exception, yet? -> https://stackoverflow.com/questions/4992526/how-to-handle-session-expiration-and-viewexpiredexception-in-jsf-2

solth:
> My question is: did changing the saving state to client help to find the real issue or was this still hidden and / or not visible in the log file?

No, when setting the `STATE_SAVING_METHOD` to `client`, neither the `VIEW_EXPIRED_EXCEPTION` nor any other error occurs. The metadata editor seems to be working normally but there is probably some error perhaps hidden in the AJAX communication, that does not cause Kitodo to break.

PR #3788, opened 2020-06-26
Make shortcuts configurable per user

oliver-stoehr:
Shortcuts are no longer hard-coded in `metadata_editor.js` but saved in the database instead.
They can be configured per user on the user edit page.


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/User.java
```diff
@@ -242,6 +246,14 @@ public void setWithMassDownload(boolean withMassDownload) {
         this.withMassDownload = withMassDownload;
     }
 
+    public String getShortcuts() {
```
solth, line 249: Missing javadoc

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/User.java
```diff
@@ -242,6 +246,14 @@ public void setWithMassDownload(boolean withMassDownload) {
...
+        return shortcuts;
+    }
+
+    public void setShortcuts(String shortcuts) {
```
solth, line 253: Missing javadoc

PR #3799, opened 2020-07-01
Add options to log catalog input to debug folder

### LINKED ISSUE #3386: Missing way to see imported data before mapping
In production version 2, imported data from catalog / database was somehow visible (in version 2, `opacBeautifyBefore.xml` file was written in file system). In version 3, such a possibility is missing, it is nowhere possible to see. If data is not visible after import: Does the data not come from the database? Is it lost in the mapping? Or is it not visible in the display?

**Goal:** to reinstall the function, to be able to look at raw data (at least administrative users).

matthias-ronge:
Fixes #3386


matthias-ronge:
Added for debugging #3800

PR #3835, opened 2020-07-14
Temporary hotfix: ignore schema converter test

oliver-stoehr:
Ignore `XmlSchemaConverterTest.shouldConvertMarcToInternalFormat()` until a sensible solution is found.


Kitodo-XML-SchemaConverter/src/test/java/org/kitodo/xmlschemaconverter/XmlSchemaConverterTest.java
```diff
@@ -99,6 +100,7 @@ public void shouldConvertModsToInternalFormat() throws IOException, ParserConfig
         Assert.assertEquals("PublicationYear after conversion is wrong!", "1999", year);
     }
 
+    @Ignore
```
solth, line 103:
```suggestion
    @Ignore("Disabled due to problems with downloading marc xsl transformation file from loc.gov")
```

PR #3845, opened 2020-07-16
Refactor SearchResultForm and ProcessForm to align available actions

### LINKED ISSUE #3429: Limited functionality for processes in search-result list
### Problem
For the processes in the result list after a search (not filter) is conducted, only a small set of functions is available. As well as regarding the icons on the rights side (export, ...), the selection box on the left side and the menus "possible actions" and "Statistische Auswertung" at the bottom. 

searchlist02

### Solution
To improve the processing of the retrieved processes, these functions should be made available. Otherwise, the IDs have to be copied and filtered, if the retrieved processes should be anylysed statistically or the status should be set.

solth:
Fixes #3429


Kathrin-Huber:
There are two suspicious tests failing:
EditingST.editProcessTest
SearchingST.searchForProcesses

PR #3872, opened 2020-07-29
Prevent NumberFormatException in task filter

solth:
Prevent NumberFormatException when entering filter values with leading or trailing whitespaces in task filter (e.g. "id: 3")

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
@@ -837,8 +837,10 @@ public class FilterService extends SearchService<Filter, FilterDTO, FilterDAO> {
         Set<Integer> ids = new HashSet<>();
         List<String> stringIds = getFilterValuesFromFilterString(filter, filterString);
         for (String tempId : stringIds) {
-            Integer id = Integer.parseInt(tempId);
-            ids.add(id);
+            if (!tempId.isEmpty()) {
+                Integer id = Integer.parseInt(tempId);
+                ids.add(id);
+            }
         }
         return ids;
     }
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
```diff
@@ -837,8 +837,10 @@ private QueryBuilder filterAutomaticTasks(String filter, ObjectType objectType)
         Set<Integer> ids = new HashSet<>();
         List<String> stringIds = getFilterValuesFromFilterString(filter, filterString);
         for (String tempId : stringIds) {
-            Integer id = Integer.parseInt(tempId);
-            ids.add(id);
+            if (!tempId.isEmpty()) {
+                Integer id = Integer.parseInt(tempId);
```
henning-gerhardt, line 841: `parseInt` is throwing an `NumberFormatException` if `tempId` does not contain a parsable integer. See https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#parseInt-java.lang.String-

Maybe you should catch this exception and do not add this value to the list in this case?

solth: AFAIK in Java the idiom of "look before you leap" is preferred over willingly risking and catching exceptions, which would probably be a more "pythonic" way to deal with exceptions (see https://docs.python.org/3.5/glossary.html#term-eafp).

So I guess you are right that different potential reasons for `NumberFormatException`s should be prevented, but I would prefer doing that by explicitly testing whether a String is numeric or not instead of willingly provoking an Exception in Java. 

Please check if you can agree with my updated changes.

henning-gerhardt: Your last change (from `!tempId.isEmpty()` to `!tempId.isEmpty() && StringUtils.isNumeric(tempId)`) is ok for me.

PR #3878, opened 2020-07-30
Disable autocomplete for calendar input

### LINKED ISSUE #3867: Suppress browser suggestions in the date output
When entering the date in Firefox, the calendar is covered with internal suggestions when you start writing. At least that looks bad.

Browser suggestions

**Goal:** Suppress browser suggestions for fields with JSF calendar that folds out.

oliver-stoehr:
Resolves #3867


henning-gerhardt:
Why did you not move the pop out to the right side so browser autocomplete feature and calendar pop out are both usable?

oliver-stoehr:
`p:datePicker` does not offer any option to change the popup's position. The popup is always placed with its top left corner at the input's bottom left corner.
The only option would be to move the popup when its shown via JavaScript. But that seems to me like a very dirty solution.

henning-gerhardt:
Thank you for this information. This is bad to know and I'm using the autocomplete feature many times.

PR #3891, opened 2020-08-03
Set booleans not null in the database schema

### LINKED ISSUE #3175: Set booleans not null in the database schema
There are boolean columns in the database, but they are defined _nullable_ in the database schema. This is not necessary and a potential source of errors and should be changed to `TINYINT(1) NOT NULL DEFAULT 0` everywhere.

matthias-ronge:
Fixes #3175


matthias-ronge:
After the Travis build went through and it worked for me locally, the syntax error in SQL, which Codacy complains about, is obviously a false positive.

Kathrin-Huber:
Please change `MODIFY` to `MODIFY COLUMN`
We do it that way in all the other migrations.

PR #3897, opened 2020-08-05
restore findByProperty

### LINKED ISSUE #3888: "ProcessService.findByProperty" does not work
The method `findByProperty` in the class `ProcessService` seems to be broken and always returns empty lists.

Kathrin-Huber:
Fixes #3888


solth:
Fixes #3888, doesn't it?

Issue #3160, opened 2020-02-07
Clean up package names


matthias-ronge:
There are still a handful of packages called `goobi`. These should be renamed.

PR #3899, opened 2020-08-05
Rename package org.goobi.mq to org.kitodo.production.interfaces.activemq

matthias-ronge:
Part of #3160


Kathrin-Huber:
why `org.kitodo.production.interfaces.activemq` And not
`org.kitodo.production.activemq` ?

matthias-ronge:
Because it is an interface, I thought.

PR #3901, opened 2020-08-05
Move SearchResultGeneration into the package which uses it

matthias-ronge:
Part of #3160


Kathrin-Huber:
it's not a Service, so maybe move it to `org.kitodo.helper`

PR #3902, opened 2020-08-05
Remove the broken rest interface

### LINKED ISSUE #3873: Remove the broken rest interface
The rest interface offered read access to a selection of process, project and catalog configuration data in version 2.x. In essence, it provided the process title with identifier (whatever was used as the identifier) and the order of the steps with title (without their ID) and status, the status being an undocumented integer. Information about the catalog configuration and the projects with a number of properties could be queried.

The interface has stopped working. If you try to access it, you will be redirected to the login page. The data offered by the interface is no longer available in this form in version 3.x.

The source code classes of the interface, which are still present, access classes for reading configuration files that are no longer available (for example `projects.xml`), so that the dependant classe cannot be removed yet.

**Goal:** The rest of the rest interface should be removed.

matthias-ronge:
Fixes #3873


matthias-ronge:
Some dependencies cannot be removed because they are used elsewhere:

**Kitodo − DataManagement**

Cannot remove artifact: jersey-server
Povides package: `javax.ws.rs`
Used by: `org.kitodo.data.elasticsearch.search.SearchRestClient`, `org.kitodo.data.elasticsearch.KitodoRestClient`, `org.kitodo.data.elasticsearch.index.IndexRestClient` and `org.kitodo.data.elasticsearch.index.Indexer`

**Kitodo**

Cannot remove artifact: jersey-json
Povides packages: `org.codehaus.jackson.map` and `org.codehaus.jackson.type`
Used by: `org.kitodo.production.forms.UserForm`

henning-gerhardt:
1) Package `javax.ws.rs` can be used from Javax WS RS API
```
<dependency>
    <groupId>javax.ws.rs</groupId>
    <artifactId>javax.ws.rs-api</artifactId>
    <version>2.1.1</version>
</dependency> 
```
2) Clas `CreateProcessForm` and used `com.sun.jersey.api.NotFoundExceptionNotFoundException` can be replaced by self-written Exception `org.kitodo.exceptions.NotFoundException`.
3) `jersey-json` in class `UserForm` and a lot other classes is a little bit to hard coded and more difficult to replace. But there are a lot of `org.codehaus.jackson` maven packages available to find a possible solution.

henning-gerhardt:
It is more easy as I thought: `jersey-json` can be replaced with package
```
<dependency>
    <groupId>org.codehaus.jackson</groupId>
    <artifactId>jackson-xc</artifactId>
    <version>1.9.13</version>
</dependency>
```

matthias-ronge:
1./3. works, thank you!
2. There is no specific catch for that exception, so just removing the block which throws it is sufficient. Instead, `get(0)` will throw an `IndexOutOfBoundsException` in that case, which will result in the same behaviour. I fixed the places where the getter is invoked during exception logging, where it would hide the exception to be logged if it threw an exception itself.

henning-gerhardt:
Changes looks nice but I'm unsure about the change regarding to the exception handling.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -296,8 +289,14 @@ public String createNewProcess() {
...
         } catch (RulesetNotFoundException e) {
-            Helper.setErrorMessage("rulesetNotFound", new Object[] {this.getMainProcess().getRuleset().getFile()},
-                    logger, e);
+            String rulesetFile;
+            try {
+                rulesetFile = getMainProcess().getRuleset().getFile();
+            } catch (IndexOutOfBoundsException noProcess) {
+                logger.catching(Level.TRACE, noProcess);
+                rulesetFile = "Process list is empty!";
+            }
```
Kathrin-Huber, line 298:
```suggestion
            String rulesetFile = "Process list is empty";
            if (!this.processes.isEmpty() && Objects.nonNull(getMainProcess().getRuleset())) {
                rulesetFile = getMainProcess().getRuleset().getFile();
            }
```

Kitodo/src/main/java/org/kitodo/production/services/data/ProjectService.java
```diff
@@ -249,7 +249,7 @@ public boolean isProjectComplete(Project project) {
         boolean projectsXmlExists = KitodoConfigFile.PROJECT_CONFIGURATION.exists();
         boolean digitalCollectionsXmlExists = KitodoConfigFile.DIGITAL_COLLECTIONS.exists();
 
-        return Objects.nonNull(project.getTitle()) && Objects.nonNull(project.template)
+        return Objects.nonNull(project.getTitle())
```
Kathrin-Huber, line 252: Why did you remove this?

Kitodo/src/test/java/org/kitodo/production/services/data/ProjectServiceTest.java
```diff
@@ -52,11 +49,6 @@ public void testProjectForCompletness() throws IOException {
         project.setTitle("testProject");
         Assert.assertFalse("Project shouldn't be complete", projectService.isProjectComplete(project));
 
-        // Add templates, still not complete
```
Kathrin-Huber, line 55: why did you remove this?

henning-gerhardt:
A lot of the changes in classes `ConfigOpac.java ` and `ConfigOpacDoctype.java` did not look as they are needed for jersey remove as they change other things. Looks for me that this changes belongs to an other pull request or?

Kitodo/src/main/java/org/kitodo/production/services/data/ProjectService.java
matthias-ronge: The only use of the `project.template` field was to list in the web service which production templates were assigned to the project. Since the web service is removed, it is no longer needed.

Kitodo/src/test/java/org/kitodo/production/services/data/ProjectServiceTest.java
matthias-ronge: Since the F field is removed, the test had to be adapted.

matthias-ronge:
> 
> 
> A lot of the changes in classes `ConfigOpac.java ` and `ConfigOpacDoctype.java` did not look as they are needed for jersey remove as they change other things. Looks for me that this changes belongs to an other pull request or?

These values were only in the web service read from the file and echoed in the generated XML response. They no longer have any meaning in 3.x. This change is intentional and related to the change.

PR #3907, opened 2020-08-06
change comparators

### LINKED ISSUE #3825: Misuse of Comparator-Interface
Both [Task-](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/migration/TaskComparator.java) and [TemplateComparator](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/migration/TemplateComparator.java) in the package `org.kitodo.production.migration` should either be removed or stop implementing the Comparator-Interface.

A Comparator imposes the ordering of objects. That's not happening since `-1` is never returned. What these "Comparators" actually do is ✌️compare✌️two objects to check if they are equal or not. The return value serves no other purpose.

That's what an `.equals`-method is supposed to be. Therefore, the logic should move into the `equals()`-method of the corresponding classes where it belongs.

If that's not possible for whatever reason, the Comparator interface should be removed from these classes since it's clearly implemented in the wrong way.

Kathrin-Huber:
fixes #3825

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/migration/TaskComparer.java
+++ b/Kitodo/src/main/java/org/kitodo/production/migration/TaskComparer.java
@@ -11,55 +11,53 @@
 
 package org.kitodo.production.migration;
 
-import java.util.Comparator;
 import java.util.Objects;
 
 import org.kitodo.data.database.beans.Task;
 
-public class TaskComparator implements Comparator<Task> {
+public class TaskComparer {
 
-    @Override
-    public int compare(Task firstTask, Task secondTask) {
+    public boolean isEqual(Task firstTask, Task secondTask) {
         if (Objects.isNull(firstTask) || Objects.isNull(secondTask)) {
-            return 1;
+            return false;
         }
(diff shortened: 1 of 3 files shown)


Kathrin-Huber:
We need a separate comparator because we are not comparing for java but for migration purposes, so using equals is not an option. But as statet correctly, the Comparator interface was misused

Kitodo/src/main/java/org/kitodo/production/migration/TaskComparer.java
```diff
@@ -11,55 +11,57 @@
...
                 : !firstTask.getScriptPath().equals(secondTask.getScriptPath())) {
-            return 1;
+            return false;
         }
         if (firstTask.isBatchStep() != secondTask.isBatchStep()) {
-            return 1;
+            return false;
         }
-        return 0;
+        return true;
```
solth, line 64: The last `if` condition an the `return true;` at the end can be simplified to 
`return firstTask.isBatchStep() == secondTask.isBatchStep();`

PR #3909, opened 2020-08-06
Execute tests in random order

henning-gerhardt:
This pull request change the order of test execution from default "filesystem" to "random". This can improve the quality of tests as they are not executed in a system specific ordering and bad written tests will be visible.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -128,6 +128,14 @@
                     </filesets>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.2</version>
+                <configuration>
+                    <runOrder>random</runOrder>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
@@ -140,6 +148,9 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-failsafe-plugin</artifactId>
                         <version>${maven-failsafe-plugin.version}</version>
+                        <configuration>
+                            <runOrder>random</runOrder>
+                        </configuration>
                         <executions>
                             <execution>
                                 <goals>


henning-gerhardt:
Any comments on this change @kitodo/kitodo-production-developers ?

matthias-ronge:
The Travis tests fail. Please fix that. 😉

matthias-ronge:
I reject nondeterministic tests because they cause errors that only occur occasionally, are not easily reproducible, and therefore cannot be easily fixed. The purpose of the tests is to find errors in branches. If the branch is correct but the test (sometimes) fails, this is not effective.

I also believe—I’m not sure, but I think—that the Selenium tests build on one another, i.e. that elements are added in the AddingST that are deleted in the DeletingST.

henning-gerhardt:
> The Travis tests fail. Please fix that.

That is the reason why I change the run order of the tests because they are failing on my systems (two at least) since weeks but pass on Travis or other systems.

> I reject nondeterministic tests because they cause errors that only occur occasionally, are not easily reproducible, and therefore cannot be easily fixed. The purpose of the tests is to find errors in branches. If the branch is correct but the test (sometimes) fails, this is not effective.

Tests should be written and executed independent of other tests. They should not relay on other previous or later executed tests. Even their execution order should not determinate their pass or failing.

> I also believe—I’m not sure, but I think—that the Selenium tests build on one another, i.e. that elements are added in the AddingST that are deleted in the DeletingST.

I don't hope that this is true and they are (hopefully) independent.

matthias-ronge:
You are welcome to do that, fix all tests and if the tests have run 1000 times without errors, we can gladly take over. At the moment we have [completely different problems](https://github.com/kitodo/kitodo-production/issues?q=is%3Aopen+is%3Aissue+label%3A3.x+-label%3Adocumentation+-label%3Aduplicate+-label%3Afeature+sort%3Acomments-desc).

henning-gerhardt:
Closing as I have no time to fix all "bad written tests" nor there is no general agreement to run tests in random order.

PR #3911, opened 2020-08-06
remove placeholder

### LINKED ISSUE #3905: placeholder text is used for process titles of newspaper issues
### Problem
The field of the name of the newspaper issue contains the placeholder text "Issue 1". If the text is not deleted, that text is used for the generation of the newspaper issue, as for example: 
- AnzefRiS_1667232932-18800531is_issu 

If the text is deleted, it is still shown, but is not used for the process titel: 
- AnzefRiS_1667232932-18800426

The description of that behavior is not easy. If the deletion of the text is forgotten, there are quickly hundreds of wrong processes. The creation of the CatalogIDDigital in the XSLT-export file has to be adjusted, too. 

### Solution
The placeholder text should be removed. The probability that the placeholder text is the desired one for the process title is low. 

### Example
issuetitle01

Kathrin-Huber:
fixes #3905

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/CalendarForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/CalendarForm.java
@@ -597,7 +597,6 @@ public class CalendarForm implements Serializable {
     public void addIssue(Block block) {
         if (Objects.nonNull(block)) {
             Issue issue = block.addIssue();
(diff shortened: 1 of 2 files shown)


Kathrin-Huber:
I think the placeholder should be removed too, so it's explicit, that there is no title.

PR #3914, opened 2020-08-07
Use Kitodo - Command to create process root directory

### LINKED ISSUE #3196: 3.0: Change permissions for /usr/local/kitodo/metadata/<processid>
For each process a directory /usr/local/kitodo/metadata/<processid> is created with following permissions: 
```
drwxr-x---  4 tomcat tomcat 4096 Feb 14 14:06 16/
```
Is there a way to set other permissions?
Why is this dir created with java, but its subdirs are created with the shell script script_createDirMeta.sh? (In the shell script I can set the permissions are desired.)

This is important in cases where you want to allow a user to access the image subdirs, but the user is not in the tomcat group.

matthias-ronge:
Fixes #3196


henning-gerhardt:
For error `Caused by: java.util.NoSuchElementException: No configuration found in kitodo_config.properties for key script_createDirMeta!`  you should adjust the file `Kitodo-FileManagement/src/test/resources/kitodo_config.properties` to include the missing key.

Kitodo-DataEditor/src/test/resources/kitodo_config.properties
```diff
@@ -13,3 +13,4 @@ MetsEditorDefaultPagination=uncounted
 directory.modules=../Kitodo/modules/
 xsltFilenameMetadataTransformation=MetsModsGoobi_to_MetsKitodo.xsl
 directory.metadata=../../../../src/test/resources/metadata/
+script_createDirMeta=src/test/resources/scripts/script_createDirMeta.sh
```
Kathrin-Huber, line 16: I think it's only needed in Kitodo-FileManagement where you also added it, istn't it?

matthias-ronge: I thought the same thing at first, but as you can see from [this Travis run](https://travis-ci.org/github/kitodo/kitodo-production/builds/716495360) (that was right before I added it here), several tests require it.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -365,7 +371,8 @@ public String getFileNameWithExtension(URI uri) {
     @Override
     public URI createProcessLocation(String processId) throws IOException {
         File processRootDirectory = new File(KitodoConfig.getKitodoDataDirectory() + File.separator + processId);
-        if (!processRootDirectory.exists() && !processRootDirectory.mkdir()) {
+        String command = scriptCreateDirMeta + ' ' + processRootDirectory.getPath();
+        if (!processRootDirectory.exists() && !commandService.runCommand(command.hashCode(), command).isSuccessful()) {
```
Kathrin-Huber, line 375:
```suggestion
        String scriptCreateDirMeta = KitodoConfig.getParameter("script_createDirMeta");
        String command = scriptCreateDirMeta + ' ' + processRootDirectory.getPath();
        if (!processRootDirectory.exists() && !commandService.runCommand(command.hashCode(), command).isSuccessful()) {
```

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -45,6 +47,10 @@
 
     private static final String IMAGES_DIRECTORY_NAME = "images";
 
+    private final CommandInterface commandService = new KitodoServiceLoader<CommandInterface>(CommandInterface.class)
+            .loadModule();
+    private final String scriptCreateDirMeta = KitodoConfig.getParameter("script_createDirMeta");
+
```
Kathrin-Huber, line 53:
```suggestion

```

Kathrin-Huber (changes requested):
You need to remove the script from the instantiation of FileManagement, then you don't need  the file in the DataEditor. (see suggestions)

PR #3920, opened 2020-08-11
Remove deprecated and error-prone Workpiece.getMediaUnits()

matthias-ronge:
Before the introduction of editability of the physical structure tree, the physical structure tree was flat. `List<MediaUnit> getMediaUnits()` offered access to the first child level of the physical structure tree, which contains the pages of digitized books. The method has been deprecated for a long time and its use is prone to errors because the other parts of the physical structure tree are not evaluated. The method is removed and the now correct `List<MediaUnit> getAllMediaUnits()` is used where it was still used.


matthias-ronge:
Currently postponed, because API methods do not do what they are intended to do.

PR #3942, opened 2020-08-17
Remove usages of org.joda.time

oliver-stoehr:
Remove all usages of org.joda.time package and remove the unused class `Citation`.


solth:
Just out of curiosity: why is `org.joda.time` removed? Did it cause any problems?

Issue #3546, opened 2020-04-27
Folder labels are not validated


matthias-ronge:
Production allows the creation of folders in the project configuration where the internal designation “METS filegroup” is empty or the same name has been assigned to several folders. However, this designation must not be empty because the internal storage format is METS and the assignment of internal picture groups is then not possible. However, no error message is displayed.

**Goal:** When configuring the project and trying to create processes, an error message should appear if a folder has no “METS filegroup” label or several folders have the same label.

Kathrin-Huber:
fixed by #3951

PR #3935, opened 2020-08-13
Fix several folder issues

matthias-ronge:
Fixes #3187, fixes #3340, fixes #3731, fixes #3775

PR #3951, opened 2020-08-18
Move foldergenerator from Folder to ProjectForm

### LINKED ISSUE #3950: Image derivatives of certain width
Kitodo offers these two posibilities to create derivatives:
- Derivat erstellen: scale the original image by percent
- Web-Bild erstellen: set a width by pixels

For the latter, however, I cannot change the default 150px. If I change the value in the field, after saving the value is set to 150px again.
Thus, I cannot generate derivatives with a custom width.

Kathrin-Huber:
fixes #3950


matthias-ronge:
Please rebase your branch so that the fixes from #3935 are included.

matthias-ronge (changes requested):
Works for editing existing folders.

When trying to add a new folder to a project, without another folder was edited before, this gives a
```
    at org.kitodo.forms.FolderGenerator.getMethod(FolderGenerator.java:101)
```

When adding a new folder after another folder was edited before, the image generation settings of the folder which was edited before are shown and edited! The new folder will be created, but will have no image generation settings. Instead the settings are changed on the folder which was edited before.

matthias-ronge:
If you add a new folder, edit it again and leave the dialog a second time with “Apply”, a NullPointerException occurs. I think I would have already fixed that one. Could you rebase the branch?

When adding a new folder with a setting for generating images, the entered value (enter a value ǂ 100% or ǂ 150 px) is forgotten, after clicking on Edit the default value is showing again. But this can be related to the above problem.

matthias-ronge (changes requested):
I found two problems:

1.) If you add a new folder, edit it again and leave the dialog a second time with “Apply”, a NullPointerException occurs.

        at org.kitodo.production.forms.ProjectForm.saveFolder(ProjectForm.java:334)

The reason is that the last entry in the `getFolders()` list is the newly created folder that does not yet have an `id`, so `equals()` cannot be called on the `getId()`.

I understand what you are up to: You want to compare the current folder `myFolder` with all other folders to see if they do not have the same filegroup. If I understand it correctly, the comparison on object identity (` == `) would be the right one at this point in order to ignore the occurrence of the same folder. The following should do it:

    if (folder != myFolder && folder.equals(myFolder)) {
        Helper.setErrorMessage("errorDuplicateFilegroup", new Object[] {ObjectType.FOLDER.getTranslationPlural() });
    }

2.) When adding a new folder with a setting for generating images, the entered value (enter a value ǂ 100% or ǂ 150 px) is forgotten when clicking on the edit button, and the default value is showing instead. (If you don’t click on edit, the correct value is written to the database.)

matthias-ronge:
1. Click "add subfolder"

Bildschirmfoto_0008

2. Type in folder and file group, select "create derivative", set value to 50, click "accept"

Bildschirmfoto_0009

3. Click "edit" of the newly created folder

Bildschirmfoto_0010

Value of derivative is wrong (100%), should be 50%.

Bildschirmfoto_0011

matthias-ronge:
I found another problem: If you add a folder, then delete it, then add it again with the same filegroup value, you get false error "filegroup already in use" and the folder cannot be created again.

matthias-ronge:
Fixes #3546, too

PR #3965, opened 2020-08-21
Fix some code issues reported by inspection tools

solth:
- remove unused imports
- remove some unused variables
- make variables `final` that are never re-assigned
- replace `Arrays.asList` with only one parameter with `Collections.singletonList`
- inline some redundant variables
- remove exceptions that are never thrown from method signatures


Kitodo/src/main/java/org/kitodo/production/forms/LoginForm.java
```diff
@@ -42,12 +42,10 @@
      * @return The user object or null if no user is authenticated.
      */
     public User getLoggedUser() {
-        if (Objects.nonNull(loggedUser)) {
-            return this.loggedUser;
-        } else {
+        if (!Objects.nonNull(loggedUser)) {
```
Kathrin-Huber, line 45:
```suggestion
        if (Objects.isNull(loggedUser)) {
```

PR #3986, opened 2020-09-01
Ignore duplicate keys

oliver-stoehr:
The "ORDERLABEL" attribute produces an exception because of duplicate keys when collecting the stream.


Kathrin-Huber:
Please fix Codacy issue

PR #4023, opened 2020-10-07
close httpClient before recreation

Kathrin-Huber:
Recreate httpClient also for querys to single records, because calling it multiple times for parent levels can also lock httpClient.


solth (approved):

henning-gerhardt:
Why you are using a class global variable instead of a method local variable as you recreating this variable every time? If you are using a method local variable then you can even use a try with resources statement and the http client is closed every time.

solth:
I agree with @henning-gerhardt. I think for starters the variables `httpClient` and `ftpClient` shouldn't be static. Additionally using try-with-resources could help avoiding unclosed connections.

Kathrin-Huber:
I'm going to change  this in a separate PR

PR #4028, opened 2020-10-07
Suggestion for Windows and Linux file extensions

matthias-ronge:
Here is my suggestion for solving the problem with different script file extensions on Linux and Windows.


henning-gerhardt:
Changes are ok and did work. But your explanation is not correct. First extension is used if on the used operation system the `/` character is the directory separator and second extension is used on the other case. It is not depending on `.sh` or `.bat` and this should be documented in a understandable way. If you are using a Python based script `.py` instead of a Shell script `.sh` this did work too.

PR #4048, opened 2020-10-09
Remove wrong static modifier from variables in QueryURLImport

solth:
Remove wrong static modifier from some variables and methods in QueryURLImport that could potentially lead to some nasty and difficult to debug problems when multiple users use the catalog import at the same time.


Kathrin-Huber:
please resolve conflicts

PR #4049, opened 2020-10-13
Migrate more variants of databases

matthias-ronge:
We had to make these adjustments as part of the migration workshop so that the database in the example could be migrated.

### DIFF
--- a/Kitodo-DataManagement/src/main/resources/db/migration/V2_18__Change_process_id_to_not_null_for_template.sql
+++ b/Kitodo-DataManagement/src/main/resources/db/migration/V2_18__Change_process_id_to_not_null_for_template.sql
@@ -13,4 +13,6 @@
 -- Migration: Changing the columns for process id in template table to not null
 --
 
+SET FOREIGN_KEY_CHECKS = 0;
 ALTER TABLE template MODIFY process_id INT(11) NOT NULL;
+SET FOREIGN_KEY_CHECKS = 1;
--- a/Kitodo-DataManagement/src/main/resources/db/migration/V2_29__Split_processes_and_templates.sql
+++ b/Kitodo-DataManagement/src/main/resources/db/migration/V2_29__Split_processes_and_templates.sql
@@ -100,10 +100,10 @@ DELETE batch_x_process FROM batch_x_process
 --
 
 SET FOREIGN_KEY_CHECKS = 0;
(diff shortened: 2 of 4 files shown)


matthias-ronge:
Hi @Kathrin-Huber, we have to migrate another system tomorrow. Could you merge these fixes today? Then we could use the official version.

henning-gerhardt:
I migrated many times our big database and your proposed changes are not needed. If this changes go in, I need to migrate our database again as we use flyway to migrate our database. In the past I was me told not to change released migration files.

matthias-ronge:
There are databases where these changes are necessary, or migration fails. You may consider this as as “messy” 2.x databases, and if the migration worked for you, you don’t have to migrate your database again, as these changes will not result in any changes with your database. Everybody is using Flyway to migrate databases, so this argument doesn’t count.
I am not happy that we need to fix released migration files, but these changes do some clean-up which is necessary to migrate the one or other 2.x database which contains messy data. So this cannot be done _later_ as migration will fail with an error it these changes aren’t run _before_ the migration.

henning-gerhardt:
What about fixing the 2.x database before you start migration? I must do this even as our database is even old and messed up (wrong engine, collation, additional tables, additional views, ...).

matthias-ronge:
I cannot do this anymore now. We did run these chanegs in between in places wehere migration failed when doing the migration workshop which was part of the development project last year, and this PR is part of the documentation. I don’t know if running these commands as-is before the migration is the same.

henning-gerhardt:
If the old migration files already applied to this database then your suggested changes would not be applied anymore by flyway. So you must do this in a manual way.

I don't know how it was possible to add template processes some kind of properties in 2.x. The only way I can image is to change a normal process to a process template but who - in hell - is doing this?

matthias-ronge:
Yes, we are talking about any “new” 2.x databases that need to be migrated. This change is uninteresting for any database that already have been migrated successfully.

henning-gerhardt:
Good thing is: our current productive 2.x database has not any kind of such wrong data - if I translated correct your additional delete statements to select statements of 2.x. So any migration in future should hopefully not be affected and I will refuse any kind of change of existing processes to process templates.

Question is or remain: should the migration support any kind of messed up 2.x databases? I don't think so.

Kathrin-Huber:
Sorry I'm not going to change flyway migrations if not inevitable.
Databases in a wrong structure must be cleaned-up beforehand.

PR #4054, opened 2020-10-15
remove unneccessary assignment of processes

Kathrin-Huber:
the double assignment of processes leads to a hibernate exception when saving the template.


matthias-ronge (commented):
To be honest, I have no idea what that does or in what context. You yourself inserted this line [here](https://github.com/kitodo/kitodo-production/commit/c4fabba0083c4f5beecf0825248b84478564e8b6#diff-cd5aafc341aa137f324b8d4e7b604d56b6d5bb68f5657393a26b89ba7f393826R412). What was it about?

PR #4058, opened 2020-10-21
Use a supported AJAX event for selectBooleanCheckbox

### LINKED ISSUE #4051: Unable to save boolean metadata in metadata editor
In the current version of the master branch (commit faeecab9f629ae96a966d590cf93ab265fc944f3) it is impossible to save boolean metadata - e.g. "activating" the corresponding "switch" -  in the metadata editor if it isn't already "checked". 

I have a metadata key in the ruleset configured like this:
```
<key id="testBooleanMetadata">
    <label>TestBooleanMetadata</label>
    <codomain type="boolean"/>
</key>
```
and a corresponding restriction on a `division` like this:
```
<restriction division="testDivision" unspecified="forbidden">
      <permit key="testBooleanMetadata" maxOccurs="1" minOccurs="1"/>
</restriction>
```

Unchecking the resulting checkbox in the metadata table of the metadata editor works and can be saved to the `meta.xml` file. However, setting the checkbox to "checked" again afterwards cannot be saved. (tried without `minOccurs="1"` and adding the metadata via the "+" button as well).

Presumably this is caused by the current implementation not distinguishing between a boolean metadata missing from the metadata file (e.g. "Unknown state") and the metadata value being set to `false` explicitely ("unchecking" a checkbox and saving the process in the editor results in the whole `kitodo:metadata` entry being remove from the corresponding `dmdSec` in the METS file). 

@matthias-ronge do I need to configure something else to make this work? Or can you confirm that this is working as intended in the mentioned commit?

matthias-ronge:
Fixes the bug described in #4051; won’t cure the XSD documentation issue.

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
@@ -13,7 +13,10 @@ package org.kitodo.api.dataeditor.rulesetmanagement;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
+
+import org.apache.commons.lang.StringUtils;
 
 /**
  * Provides an interface for the metadata key view service. The metadata key
@@ -29,10 +32,11 @@ public interface SimpleMetadataViewInterface extends MetadataViewInterface {
      *         metadata entry.
      */
     default Optional<String> convertBoolean(boolean value) {
-        if (getSelectItems().isEmpty()) {
+        if (value) {
+            return getSelectItems().entrySet().stream().map(Entry::getKey).filter(StringUtils::isNotEmpty).findAny();
+        } else {
             return Optional.empty();
         }
-        return value ? Optional.of(getSelectItems().entrySet().iterator().next().getKey()) : Optional.empty();
     }
 
     /**
@@ -89,7 +93,7 @@ public interface SimpleMetadataViewInterface extends MetadataViewInterface {
 
     /**
      * Returns whether values under this key can be edited in this view.
-     * 
+     *
      * @return whether values can be edited
      */
     boolean isEditable();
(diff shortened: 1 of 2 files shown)


solth:
> Fixes the bug described in #4051

With this syntax the issue won't be automatically closed when this PR gets merged ;) (there shouldn't be any text between "Fixes" and the issue number, I think)

Bildschirmfoto 2020-10-23 um 09 06 38

matthias-ronge:
That was intentional in this case, because I am also interested in how best to describe in the XSD that a `<codomain type="boolean"/>` needs exactly one `<option value="..."/>`. At least with a comment, but I'm currently still looking for a formal description of this so that an editor will criticize it if it is wrong. So in this case I prefer to close the ticket manually.

solth (changes requested):
It still does not work for me. 

When checking the checkbox and clicking "save", the corresponding `kitodo:metadata` element is written to the `meta.xml` file, but without a value. As I understand it, that works as you intended it to work (e.g. saving the metadata without explicit value equates to `true` and removing it equates to `false`).

The problem seems to be that _loading_ the `meta.xml` does not reflect this concept/design, as the saved `kitodo:metadata` element without an explicit value is not evaluated to `true` (a "checked" checkbox), but rather to `false` (e.g. an "unchecked" checkbox.)

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
```diff
@@ -29,10 +32,11 @@
      *         metadata entry.
      */
     default Optional<String> convertBoolean(boolean value) {
-        if (getSelectItems().isEmpty()) {
+        if (value) {
+            return getSelectItems().entrySet().stream().map(Entry::getKey).filter(StringUtils::isNotEmpty).findAny();
```
solth, line 36:
```suggestion
            return getSelectItems().keySet().stream().filter(StringUtils::isNotEmpty).findAny();
```

Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
```diff
@@ -13,8 +13,11 @@
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Map.Entry;
```
solth, line 16:
```suggestion
```

solth (approved):
That did the trick!

Just one little suggestions to slightly simplify the "filter stream" expression.

Issue #3718, opened 2020-06-03
VariableReplacer does not work properly in the script step


matthias-ronge:
Reported by @stefanCCS 
> I have tried out the following as "Script Path":
> /home/workshop/myScripts/mySimpleScript.sh (processpath)
> I would have expected to find in my Script as command line parameter the "process path".
> Unfortunately, I have not found anything (no command line parameter at all

stefanCCS:
In  version 3.2 build 09.06.2020 I can confirm that at least the parameter 
(processid) (processtitle) (metaFile)
are given to the script als command line parameter.
Others (ocrbasispath) (ocrplaintextpath) have NOT given any value to the script.
More parameter I have tried out so far.

stefanCCS:
Finally, I have checked all variables available in [VariableReplacer](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java) as follows:
Putting a script task in a workflow with the following parameters:
`myScript.sh "1:(tifurl)" "2: (tifpath)" "3:(origurl)" "4:(origpath)" "5: (imageurl)" "6:(imagepath)" "7:(processpath)" "8:(importpath)" "9:(sourcepath)" "10:(ocrbasispath)" "11:(ocrplaintextpath)" "12:(metaFile)" "13:(prefs)" "14:(processtitle)" "15:(processid)" "16: (stepid)" "17: (stepname)" 
`

Result is:
`
"1:file://202006291236_media"
"2:
202006291236_media"
"3:file://master_202006291236_media"
"4:master_202006291236_media"
"5:
file://"
"6:"
"7:"
"8:"
"9:"
"10:"
"11:"
"12:meta"
"13:/usr/local/kitodo/rulesets/ruleset_CCS_min.xml"
"14:202006291236"
"15:24"
"16:164"
"17:printAllPossibleParameters"
`
--> with this you can identify, which paratemer creates which value, and evaluate, if this is correct or not (I personally cannot do this evaluation, as I do not know, what is intended).

matthias-ronge:
Thanks for the work! I see I misunderstood the ticket. The placeholders are not passed to the script, the placeholders are replaced, but the value to be used has not been determined, therefore an empty string is put in place.  
The `file://` URIs with relative paths are nonsensical.

matthias-ronge:
I think the correct answer here is that only the following variables should be used. These also work and, in my opinion, are sufficient.

```
[...]

PR #4078, opened 2020-10-29
Simplify variable replacer

matthias-ronge:
See #3718

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
@@ -11,9 +11,6 @@
 
 package org.kitodo.production.helper;
 
-import java.io.File;
-import java.io.IOException;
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
Please do not change this! All path variables must no hard coded into the database and as the `scriptPath` database field could only contain 500 characters. This is not enough if you using a lot of path variables and your base path is really long. Even if you change your storage location you must update all entries inside the database and maybe re-index everything.

Sorry, I don't not understand why a base path defined in `kitodo_config.properties` and the other values from folder information are not used in this way as I understand to use them and made this available through the variable replacer and must be replaced by hard coded database values.

matthias-ronge:
At the moment these values don't work. Not erasing them doesn't make the problem any better. But if you want, I'll close this ticket.

henning-gerhardt:
Instead of deleting them I would fix them and maybe only remove this variables where not fixable.

PR #4084, opened 2020-11-05
replace projectId with variablereplacer

Kathrin-Huber:


matthias-ronge (approved):
Looks correct. 👍 

However, you could simplify these two lines, if you like:
```java
inString = replaceString(inString, "(processid)", process.getId().toString());

inString = replaceString(inString, "(projectid)", process.getProject().getId().toString());
```

PR #4086, opened 2020-11-06
Document setup of systemd demon

matthias-ronge:


henning-gerhardt (commented):
I can not say if this explanation is correct or not. In our productive environment (Debian 9) this is not necessary but maybe this is correct for more up to date or other Linux distributions.

matthias-ronge:
Yes, this is new e.g. with the new Ubuntu LTS. The note is intended more as a tip or a reminder in case it doesn't work for someone.

I'm thinking about what it would take to move the entire file management into its own dæmon, so that the web app doesn't have to deal with it anymore at all. To be honest, I don't like this mishmash. Also in order to be able to scale here over several file servers / NAS’es. With the introduction of the file management interface, we have already taken an important step in this direction. But I haven't quite finished thinking up the idea.

Issue #456, opened 2016-03-03
Replace current mechanism to store encrypted passwords


stweil:
Currently, Goobi.Production stores DES encrypted passwords with a well known pass phrase. This approach has several problems:
- As the pass phrase is well known, the encryption is not really useful.
- The DES algorithm is no longer a good choice because of its weakness.
- The source of class DesEncrypter has no known origin, so it should be removed.

The solution could be implemented in two steps.
1. Replace DesEncrypter by a new implementation, maybe based on [JCA](https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html). This can be tested by the existing test code, so it would be safe for the next version.
2. Replace the DES encryption by something more secure. This involves an irreversible conversion of existing passwords and might be done in the next version.

See also previous discussion in PR #454.

henning-gerhardt:
Some small hints and minds:
- extending test scenarios to ensure correct usage of en- and decryption of passwords
- changing pass phrase and / or encryption algorithm needs a migration tool as stored passwords must be migrated
- ldap password transfer must be respected as know it used stored password

stweil:
I suggest adding tags security and enhancement.

matthias-ronge:
Minimum requirements to stored passwords are that they are stored [hashed](https://en.wikipedia.org/wiki/Cryptographic_hash_function) by a hash function that is [cryptographically safe](https://en.wikipedia.org/wiki/Security_of_cryptographic_hash_functions). (This is different from (reversible) _encryption_.) In addition, each password must be stored with an _individual_ [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)), making the generation of [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table) useless. Also, dumping and re-importing the user table must not break the hashes. (Both MySQL and Java handle illegal string content, but it cannot be restored from a MySQL dump, which is a text file.)

[...]

PR #4091, opened 2020-11-11
Make LDAP group optional and authenticate users without one against DB

### LINKED ISSUE #4014: Enable individual users are authenticated against database even when LDAP is on
**Goal:** It is possible to select **no** LDAP group for a user. The user is then authenticated against the database, regardless whether LDAP is "on" or "off".

matthias-ronge:
Fixes #4014, part of #456


henning-gerhardt (commented):
I can not test your changes and your changes are not using any kind of test, so I can not judge if your changes are correct or not.

Issue #3304, opened 2020-03-18
Improve example LDAP group


matthias-ronge:
Example LDAP group in the base database contains some very nonsensical values.

**Goal:** to exchange the following values:
Title: `Local LDAP`
User DN: `cn={login},ou=users,dc=host,dc=example,dc=org`
gidNumber: `100`
gecos: `{user full name}`
loginShell: `/bin/false`

PR #4093, opened 2020-11-11
Remove the LDAP on/off switch

### LINKED ISSUE #3336: What about the LDAP on/off switch?
There is a switch to enable or disable LDAP authentication. The switch is at the bottom of the list of LDAP groups, and there is a comment in the HTML code that recommends finding a better place for it. The switch is not permanent, after a restart of Tomcat the status is forgotten and the configuration entry from the configuration file is adopted. Anyone who has never worked with it should know that it is a good thing to have this switch **outside of** the web application in the configuration file, otherwise you could lock everyone (including yourself) out of the system forever when something doesn't fit while setting up authentication.

_Suggestion:_ It makes sense to have this switch on the LDAP server page read-only. If you click on it, you may see a message that the change can only be made in the configuration file.

matthias-ronge:
Fixes #3336, part of #456

**Note:** The `ldap_use parameter` in the `kitodo_config.properties` file should initially be retained so that users can initially remove their LDAP group after a migration. However, the default for this attribute can be set to true for new installations.

Follow-up pull request to #4092 ([immediate diff](https://github.com/matthias-ronge/kitodo-production/compare/issue-456_2...issue-456_3+3336))


Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -465,8 +465,8 @@ exportWithoutTimeLimit=true
 # LDAP Configuration
 # -----------------------------------
 
-# Logins ueber LDAP verwenden
-ldap_use=false
+# Use logins via LDAP
+ldap_use=true
```
henning-gerhardt, line 469: Why are you enabling LDAP authentication by default?

matthias-ronge: See [here](https://github.com/kitodo/kitodo-production/issues/456#issuecomment-723027253):

> Procedure proposal:
> 
>     1. Enable individual users are authenticated against database even when LDAP is on. (#4014)
> 
>     2. Deactivate the LDAP group for the users supplied with the basic database. (Cf. also #3304)
> 
>     3. Remove the LDAP on/off switch. (#3336)
>        (The ldap_use parameter in the kitodo_config.properties file should initially be retained
>        so that users can initially remove their LDAP group after a migration. However, the
>        default for this attribute can be set to true for new installations.)

henning-gerhardt: Than add this as explanation. Without this I would deactivate this again after this pull request is merged!

I hope that your migration way of LDAP usage will be documented somewhere in an official way as issues and pull requests are not a long term documentation.

matthias-ronge: As I wrote: You would want to leave the setting unchanged after a migration. It is only about empty new installations, where this can be `true` in the future (does not have to be. Anyone can set it to `false` by hand. If someone did that, they should be aware that LDAP does not work.)

henning-gerhardt (commented):
I can not test your changes and your changes are not using any kind of test, so I can not judge if your changes are correct or not.

Issue #4090, opened 2020-11-10
Script condition needs to be explicitly selected


Kathrin-Huber:
"Script" is the default selection for a condition, when creating a workflow. But to make it work, the dropdown needs to be triggered and script needs to be selected explicitly.

![selectScript](https://user-images.githubusercontent.com/22615787/98663243-1eb9a600-2349-11eb-896e-f6a6851fe51b.gif)

a "noSelection" - Option is required, because no selection is an option.

PR #4094, opened 2020-11-12
add none workflow condition

Kathrin-Huber:
Pre-work for #4090

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/WorkflowCondition.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/WorkflowCondition.java
@@ -44,12 +44,21 @@ public class WorkflowCondition extends BaseBean {
      *            of workflow condition
      */
     public WorkflowCondition(String type, String value) {
-        if (type.equalsIgnoreCase("script")) {
-            this.type = WorkflowConditionType.SCRIPT;
-        } else if (type.equalsIgnoreCase("xpath")) {
-            this.type = WorkflowConditionType.XPATH;
-        } else {
-            throw new InvalidParameterException("Type should be script or XPath, but was " +  type);
+        switch (type.toLowerCase()) {
+            case "none": {
+                this.type = WorkflowConditionType.NONE;
+                break;
+            }
+            case "script": {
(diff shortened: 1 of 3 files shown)


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/WorkflowCondition.java
```diff
@@ -48,8 +48,10 @@ public WorkflowCondition(String type, String value) {
             this.type = WorkflowConditionType.SCRIPT;
         } else if (type.equalsIgnoreCase("xpath")) {
             this.type = WorkflowConditionType.XPATH;
+        } else if (type.equalsIgnoreCase("none")) {
+            this.type = WorkflowConditionType.NONE;
         } else {
-            throw new InvalidParameterException("Type should be script or XPath, but was " +  type);
+            throw new InvalidParameterException("Type should be none, script or XPath, but was " +  type);
```
Erikmitk, line 54: I would put the `none`-condition first like it is defined in `WorkflowConditionType` for consistency. Also, it might be the most likely case anyway. We wouldn't have to evaluate all options all the time.

Might make sense to change the if/else if block to a switch-statement now.

```
switch (type.toLowerCase()) {
    case "none": { … }
    case "script": {…}
    case "xpath": {…}
    default: {Exception…}
}
```

Erikmitk (changes requested):

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/WorkflowCondition.java
Kathrin-Huber: Actually it's the least likely case, because it's only set if you actively select the noOption in the frontend, otherwise it's null.
But 👍  for the switch statement, I'm going to change this.

Erikmitk: > Actually it's the least likely case, because it's only set if you actively select the noOption in the frontend, otherwise it's null.

Ah… yeah… you're right. The difference between theory and practice.😄

Erikmitk (approved):

PR #4105, opened 2020-11-24
Add option to apply multiple XSLT files during import

solth:
These changes add the option to apply multiple XSL transformations sequentially to an imported XML document, similar to how the current import of MARCXML documents works (first map MARC to MODS, then MODS to Kitodo). 

The XSLT files need to be ~~separated by a comma in the `<mappingFile>` element of~~ added to the corresponding OPAC configuration in `kitodo_opac.xml`  as individual `<file>` elements, contained in a new `mappingFiles` grouping element in the order in which they should be applied like this:
```
<mappingFiles>
    <file>marc21slim2mods.xsl</file>
    <file>mods2kitodo.xsl</file>
</mappingFile>
```
Note that existing configurations with a single mapping file ~~are not affected and do not need to be changed!~~ need to be updated to this new syntax to continue to work (with just a single `<file>` element in the `<mappingFiles>` container)

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/SchemaConverterInterface.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/SchemaConverterInterface.java
@@ -14,6 +14,7 @@ package org.kitodo.api.schemaconverter;
 import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
+import java.util.List;
 
 /** Enables the conversion of a DataRecord from one format to another. */
 public interface SchemaConverterInterface {
@@ -24,11 +25,11 @@ public interface SchemaConverterInterface {
      * @param record DataRecord to be converted
      * @param targetMetadataFormat MetadataFormat to which the given DataRecord is converted
      * @param targetFileFormat FileFormat to which the given DataRecord is converted
-     * @param mappingFile mapping file; if null, the schema converter module uses a default mapping
+     * @param mappingFiles list of mapping files; if empty, the schema converter module uses a default mapping
      * @return The result of the conversion as a DataRecord.
      */
     DataRecord convert(DataRecord record, MetadataFormat targetMetadataFormat, FileFormat targetFileFormat,
-                       File mappingFile) throws IOException, URISyntaxException;
+                       List<File> mappingFiles) throws IOException, URISyntaxException;
 
     /**
      * Check and return whether the current SchemaConverter supports the given MetadataFormat as a target format or not.
--- a/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
+++ b/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
@@ -107,7 +107,7 @@ public class OPACConfig {
      * @param catalogName String identifying the catalog by its title
      * @return HierarchicalConfiguration for catalog's "mappingFile"
      */
-    public static String getXsltMappingFile(String catalogName) {
+    public static String getXsltMappingFiles(String catalogName) {
         return getCatalog(catalogName).getString("mappingFile");
     }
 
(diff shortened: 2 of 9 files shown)


matthias-ronge:
Could you please change that so that the XML element is repeated?

```xml
<!-- 'mappingFile' is repeatable. Multiple XSLT files are applied in descending
     order. -->
<mappingFile>marc21slim2mods.xsl</mappingFile>
<mappingFile>mods2kitodo.xsl</mappingFile>
```

That could be something like (untested):

```java
public static List<String> getXsltMappingFiles(String catalogName) {
    List<String> xsltMappingFiles = getCatalog(catalogName).getList("mappingFile").stream()
            .filter(String.class::isInstance).map(String.class::cast).collect(Collectors.toList());
    return xsltMappingFiles;
}
```

solth:
@matthias-ronge yes, I also thought about implementing this by allowing multiple `mappingFile` elements in an OPAC configuration instead of multiple filenames in a single element. That said, I would prefer those `mappingFile` elements to grouped in one `mappingFiles` (plural) parent element:
```
<mappingFiles>
    <file>formatA2formatB.xsl</file>
    <file>formatB2kitodo.xsl</file>
</mappingFiles>
```
I think this would make the configuration much easier to read and prevent confusion when there are multiple mapping files configured without necessarily being direct sibling elements: 
```
    <catalogue title="TestCatalog" description="TestCatalog description">
        <interfaceType>sru</interfaceType>
        <returnFormat>xml</returnFormat>
        <metadataFormat>TESTFORMAT_A</metadataFormat>
        <mappingFile>testformata2testformatb.xsl</mappingFile>
        <parentElement xpath="//test:format/parent:element" />
        <config>
            <param name="host" value="test.url.de" />
            <param name="scheme" value="https" />
            <param name="path" value="/sru" />
        </config>
        <mappingFile>testformatb2testformatc.xsl</mappingFile>
        <identifierParameter value="identifier" />
        <urlParameters>
            <param name="version" value="1.2" />
            <param name="operation" value="searchRetrieve" />
            <param name="recordSchema" value="testformat" />
        </urlParameters>
        <searchFields>
            <searchField label="ID" value="identifier"/>
        </searchFields>
        <mappingFile>testformatc2kitodo.xsl</mappingFile>
    </catalogue>
```

The downside of introducing a grouping element like `mappingFiles` would be that existing OPAC configuration would need to be adapted to this changes. I think it would be preferable to keep these changes backwards compatible, wouldn't it?
 
Because of these reasons I opted for containing all mapping file names in one single `mappingFile` element, separated by a comma. This way, existing configuration do not need to be changed and all mapping files are in one place and easily readable.

About your second question: there are several metadata formats - MARC, MODS etc. - for which default mapping files are shipped with Kitodo.Production and that are used if an OPAC configration in `kitodo_opac.xml` if the corresponding metadata format is used and no custom mapping file is specified. See `MetadataFormatConversion.java` for more information.

henning-gerhardt:
My 2 cents: 

Why not use both: If you need only one mapping file, than use the `<mappingFile>...</mappingFile>` entry. If you need more, then you must use `<mappingFiles><mappingFile>...</mappingFile><mappingFile>...</mappingFiles></mappingFiles>`. If you have both entries, throw a runtime exception ;) So you must not change your existing `kitodo_opac.xml` file and you have still the possibility to use more then one mapping file.

solth:
@henning-gerhardt I think that unnecessarily complicates things since then we would need to define some precedence order - which configuration applies if both, an individual `mappingFile` _and_ an grouping `mappingFiles` element are present? In my opinion there should be exactly one way to configure something like the mapping files in Kitodo.Production, since configuration is already complicated and confusing enough as it is in this software.

I would therefore opt to implement just the grouping element. This means exisiting configurations need to be slightly adapted, but since 3.x isn't widely in use, yet, and this is a new feature, a minor and easy change in the settings which only needs to be applied once is not only feasible but also preferable to changes that would make the OPAC configuration more confusing and difficult to understand, IMHO.

solth:
Loading default mapping files (for metadata formats like MARC and MODS) is not triggered anymore if no valid custom mapping file configuration was found. I changed this PR to "draft" until I have updated that.

Issue #4081, opened 2020-11-03
Not filled DialogBox in Metadata-Editor "Validate" 


stefanCCS:
Version used: 3.2.1-build 02.11.2020
If I press "Validate" in Metadata-Editor I get an empty dialog box like this (expected would be warning, that not all logical elements have a physical page assigned).
![error-validate](https://user-images.githubusercontent.com/61045115/97990969-04c81280-1de1-11eb-80a1-26cc4d3480b5.png)

andre-hohmann:
The green dialog box is empty, too.

Kathrin-Huber:
fixed by #4117

Issue #4082, opened 2020-11-03
Not filled dialog box when pressing "Import to DMS"


stefanCCS:
Version used: 3.2.1-build 02.11.2020
If I press "Import to DMS" in non-automatic export task I get an empty dialog box like this ("null"):
![error-null](https://user-images.githubusercontent.com/61045115/97991120-48228100-1de1-11eb-99c7-9841e7f99c50.png)
Maybe this is related to #4081

matthias-ronge:
Maybe related to: #3420

Kathrin-Huber:
fixed by #4117

PR #4117, opened 2020-12-03
fix spelling and error message

### LINKED ISSUE #4108: Saving a workflow without setting permissions throws empty error message
Saving a workflow without setting a permission throws an empty error message.

CleanShot 2020-11-25 at 15 08 30

Kathrin-Huber:
fix #4108


Kitodo/src/main/webapp/pages/workflowEdit.xhtml
```diff
@@ -36,13 +36,7 @@
     <!--@elvariable id="isViewMode" type="boolean"-->
     <ui:param name="isViewMode"
               value="#{not isEditMode and not isDuplicateMode and not isCreateMode and SecurityAccessController.hasAuthorityToViewWorkflow()}"/>
-
-    <ui:define name="messages">
```
solth, line 40: This was specifically added in #3558 to prevent the same error message from being displayed twice!

Kathrin-Huber: But this was presumably leading to #4108 so both solutions aren't really satisfactory....
I'm trying to find a way to fix both...

Kathrin-Huber:
workflowErrorMessage

Both is working now :-)

solth (approved):

solth:
This also seems to fix #4081 and #4082

PR #4124, opened 2020-12-11
enable parentExtraction from record

Kathrin-Huber:
If parent records are delivered in the child record xml, it is possible to use a parentMapping.xsl to extract the parent record from the childs xml.

This is one part of #3990

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
+++ b/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
@@ -13,8 +13,6 @@ package org.kitodo.config;
 
 import java.io.FileNotFoundException;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -445,7 +453,13 @@ public String getSearchTermWithDelimiter(String searchTerm, String catalog) {
...
                 Process parentProcess = loadParentProcess(parentIDMetadata, template.getRuleset().getId(), projectId);
                 if (Objects.isNull(parentProcess)) {
-                    parentID = importProcessAndReturnParentID(parentID, processes, opac, projectId, templateId);
+                    if (OPACConfig.isParentInRecord(opac)) {
+                        parentID = importProcessAndReturnParentID(recordId, processes, opac, projectId, templateId,
+                            true);
+                    } else {
+                        parentID = importProcessAndReturnParentID(parentID, processes, opac, projectId, templateId,
+                            false);
+                    }
```
solth, line 462:
```suggestion
                    parentID = importProcessAndReturnParentID(recordId, processes, opac, projectId, templateId,
                            OPACConfig.isParentInRecord(opac));
```

solth (changes requested):
Works well, just some small nit-picking.

solth (approved):

Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
```diff
@@ -13,6 +13,8 @@
 
 import java.io.FileNotFoundException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
```
solth, line 17:
```suggestion
```

PR #4142, opened 2021-01-12
Fix SQL migration to add system page authorities

solth:
Fix assigning new system page tab authorities to admin role


henning-gerhardt (commented):
Changing existing migration files will fail flyway migration on existing and already migrated databases.

solth:
Since this migration file is not part of an official release yet, I think adjusting it is justifiable.

henning-gerhardt:
This is in this case true but we are using this file already in our migration system to test the current functionality of the master branch as the last release is to long away and a new release is not released.

solth:
It should be possible to apply the contents of the SQL migration file manually and then call `flyway --repair` to compensate the discrepancy between the migration histories.

PR #4143, opened 2021-01-14
remove index references on migration

Kathrin-Huber:
Migration should be index-independent, as the index may be empty. This leads to triggering the whole index while migrating and therefore a huge lack in migration.
If index is filled first, the metadata of processes will not be indexed, so migrating first is recommended.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/tasks/NewspaperMigrationTask.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/tasks/NewspaperMigrationTask.java
@@ -69,14 +69,14 @@ public class NewspaperMigrationTask extends EmptyTask {
     /**
      * Creates a new newspaper migration task.
      *
-     * @param transferredBatch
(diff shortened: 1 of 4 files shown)


matthias-ronge (approved):
Looks correct

Kitodo/src/main/java/org/kitodo/production/helper/tasks/NewspaperMigrationTask.java
```diff
@@ -72,7 +72,7 @@
      * @param transferredBatch
      *            an object that transfers a newspaper batch from index
      */
-    public NewspaperMigrationTask(BatchDTO transferredBatch) {
+    public NewspaperMigrationTask(Batch transferredBatch) {
```
matthias-ronge, line 75: If you change batch transfer object (`BatchDTO`) to `Batch`, you could rename the variable `transferredBatch` to `batch`, too. Just if you want. Same for the field

PR #4144, opened 2021-01-14
check uploaded file for hierarchie

Kathrin-Huber:
Realizes the FileUpload for Files containing the parent Record.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/FileUploadDialog.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/FileUploadDialog.java
@@ -19,6 +19,7 @@ import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
+
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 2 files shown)


andre-hohmann:
The file upload works for multivolume: the parent process is created correctly with with the necessary metadata. 
Thus, the demanded functionality has been met.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/FileUploadDialog.java
```diff
@@ -70,6 +73,22 @@ public void handleFileUpload(FileUploadEvent event) {
 
             LinkedList<TempProcess> processes = new LinkedList<>();
             processes.add(tempProcess);
+
+            Collection<String> higherLevelIdentifier = this.createProcessForm.getRulesetManagement()
```
solth, line 77: The method `handleFileUpload` gets a little bloated - could you move this new part to a separate private method that is called instead to improve code readability?

solth (approved):

solth (approved):

PR #4146, opened 2021-01-18
remove more index dependencies from migration

Kathrin-Huber:
Removes more dependencies on index at migration. (Loading batches for selecting a newspaper migration)


matthias-ronge (approved):

matthias-ronge (approved):
Looks correct.

Issue #1301, opened 2018-02-24
License


stweil:
It is currently not easy to find the license used by Kitodo.Production as a whole. Ideally that information should be supplied in a GitHub compatible form, so GitHub would display it. Adding it to the README would also be fine.

In addition it would be good to have also license information for all images. Can some of those images be used for Wikipedia, for example?

stweil:
Is the license for the image files now clear? Can they be used on Wikipedia? If they also use GPL3 and not allow other licenses, that would not be possible.

Kathrin-Huber:
What image files do you referre to?

stweil:
The most important one is the Kitodo logo which would be useful for the [Wikipedia article on Kitodo](https://en.wikipedia.org/wiki/Kitodo). That requires an open license which is [compatible with Wiki Commons](https://commons.wikimedia.org/wiki/Commons:Licensing).

Other [Kitodo images for Commons](https://commons.wikimedia.org/wiki/Category:Kitodo) would also be appreciated.

Kathrin-Huber:
I think it is a question  you should discuss with the Kitodo Board, isn't it?

Erikmitk:
@stweil The logo itself is provided on the kitodo.org homepage (https://www.kitodo.org/community/logo-und-praesentationen/) as well for free. But you're right. There's also no explicit licence information available. So I'd too suggest to point this question not at the developers here but to the Geschäftsstelle of the Kitodo e.V.

stweil:
I wrote to the Geschäftsstelle now. As far as I see we have two relevant logos: the [key symbol](https://s3.amazonaws.com/cloud.ohloh.net/attachments/91091/kitodo_med.png) and the [Verein's logo](https://www.kitodo.org/fileadmin/groups/kitodo/Logo/kitodo_logo_rgb.png). For Wikimedia Commons a CC license would be best, if possible with each logo in SVG format.

Kathrin-Huber:
Not a code problem

stweil:
... but still a problem of this repository which not only contains code but also images and other non-code files.

And for the images the GPL license is a real problem because such images cannot be used in Wikipedia. Even screenshots in scientific articles or presentations might be problematic.

stweil:
> I think it is a question you should discuss with the Kitodo Board, isn't it?

Right, so let's ask the @kitodo/kitodo-board. I have these two questions:

[...]

PR #4148, opened 2021-01-18
add GPL3 as default license file in github

ulb-sa-schmilj:
Github shows the wrong license file in the about page - https://github.com/kitodo/kitodo-production/blob/master/Camunda-License.txt

To disable this wrong detection, add file with the name license with the correct GPL text.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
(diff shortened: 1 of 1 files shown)


matthias-ronge:
Related to: #1301

henning-gerhardt (changes requested):
Why did you add an other GPL3 license text? Maybe you want to rename existing file "GPL3-license.txt" to "LICENSE" to fix this issue of the GitHub software which is not configurable to choose the right license on a multi-license project like Kitodo.Production? 

I don't know if adding a new file or renaming the existing file to this file name would fix the issue, which want to be solved.

ulb-sa-schmilj:
Yes, Kitodo.Production is a mutli-license project. But the "main"/default license detected from Github is wrong.

A user should click on View license and see "Copyright (c) 2014-2016 camunda Services GmbH". This isn't good. 
Github is not perfect. So loads the Github software for all branches the same license file from the master.
But I think this solution is better as the link to the camunda file.

The fix can checked on my fork - https://github.com/ulb-sa-schmilj/kitodo-production/tree/master.

I don't know the naming convention for other license files of the project. This fix works also, if we rename the existing file "GPL3-license.txt" to "LICENSE".

henning-gerhardt:
If renaming is working to would solve this problem but renaming open a new issue: every GPL3 license header in the project files is referring to a file called "GPL3-license.txt". So all this project files must be changed too.

PR #4189, opened 2021-02-15
Delete user

### LINKED ISSUE #4184: Check if any tasks are assigned to a user before deleting him
When a user is deleted who still has tasks assigned to him, the system does not unassign him from those tasks, leaving them in limbo until an admin or user with corresponding system privileges resets the tasks processing status from "IN WORK" to "OPEN".
To fix this, the system should check if any tasks are assigned to a user before deleting him and display a warning/confirmation dialog if that is the case. If the user confirms deleting the user, any tasks that are still assigned to this user should be set to "OPEN".

IkramMaalej:
Fixes #4184


Kitodo/src/main/java/org/kitodo/production/controller/SecurityAccessController.java
```diff
@@ -422,6 +422,15 @@ public boolean hasAuthorityToEditUser() {
         return securityAccessService.hasAuthorityToEditUser();
     }
 
+    /**
+     * Check if the current user has the authority to unassign all tasks in progress of a user.
+     *
+     * @return true if the current user has the authority to to unassign all tasks in progress of a user
```
solth, line 428:
```suggestion
     * @return true if the current user has the authority to unassign all tasks in progress of a user
```

Kitodo/src/main/java/org/kitodo/production/services/security/SecurityAccessService.java
```diff
@@ -420,6 +420,15 @@ public boolean hasAuthorityToEditUser() {
         return hasAuthorityGlobalOrForClient("editUser");
     }
 
+    /**
+     * Checks if the current user has the authority to unassign all tasks in progress of a user.
+     *
+     * @return true if the current user has the authority to to unassign all tasks in progress of a user
```
solth, line 426:
```suggestion
     * @return true if the current user has the authority to unassign all tasks in progress of a user
```

solth (approved):

Kathrin-Huber (approved):
Please change migration script to number 102

PR #4209, opened 2021-02-24
make serial migration index independent

Kathrin-Huber:
more index independency on migration


matthias-ronge (approved):
I don’t have an environment to test this at the moment, but it looks your changes are OK and I assume you tested it so that it still works, so I approve it.

PR #4223, opened 2021-03-01
add metadataScript for copyToChildren

Kathrin-Huber:
Adds a MetadataScript (formerly known as CopyData) for copying Metadata from parent to children.


matthias-ronge (approved):
I am completely unhappy with what you are implementing here because it has nothing to do with what is the plan in the development of copyData, see [#3368](https://github.com/kitodo/kitodo-production/issues/3368#ref-pullrequest-754230210). However, it is overall an improvement over the current state (and I believe your test that it works, I haven't tested it) so I agree to the pull request. Overall, however, I would think it would be more successful if we develop together and not side by side. Maybe #3368 is also suitable for the <kbd>development fund</kbd>, what do you think?

Issue #3327, opened 2020-03-23
Make more functions static


matthias-ronge:
Production uses a service loader to initialize services once at runtime and to make them available later in the application. This makes sense for services that provide functionality from modules, since in this case the modules are loaded once at runtime. However, it makes no sense for static functions that are provided by services that do not use modules (and which do not have to be declared static).

**Goal:** In the services, declare all functions that work statically and do not use a module as `static`. Call these functions statically, without redirection via the service loader. If a service only provides static functions, remove the service from the service loader altogether.

matthias-ronge:
There are a number of classes in Production called `Service`. Some of them are related to database objects. In this case, the services perform actions on the database objects. The functionality is in the service, the data in the data-management module.

    class                                 | service manager | module(s)
    --------------------------------------+-----------------+---------------------------------------
    AuthorityService                      | yes             | data-management
    BatchService                          | yes             | data-management
    ClientSearchDatabaseService           | no              |
    ClientSearchService                   | no              |
    ClientService                         | yes             | data-management
    CommentService                        | yes             | data-management
    DocketService                         | yes             | data-management
    FilterService                         | yes             | data-management
    FolderService                         | yes             | data-management
    LdapGroupService                      | yes             | data-management
    LdapServerService                     | yes             | data-management
    ListColumnService                     | yes             | data-management
    ProcessService                        | yes             | data-management, docket
    ProjectSearchService                  | no              |
    ProjectService                        | yes             | data-management
    PropertyService                       | yes             | data-management
    RoleService                           | yes             | data-management
    RulesetService                        | yes             | data-management
    SearchService                         | no              |
    TaskService                           | yes             | data-management
    TemplateService                       | yes             | data-management
    TitleSearchService                    | no              |
    UserService                           | yes             | data-management
    WorkflowConditionService              | yes             | data-management
    WorkflowService                       | yes             | data-management

Other servces are related to the modules. The services call functions on the modules. The functionality is in the module, the services should do little except to access the module.

    class                                 | service manager | module(s)
    --------------------------------------+-----------------+---------------------------------------
    CommandService                        | yes             | command
    DataEditorService                     | yes             | dataeditor
    FileService                           | yes             | file-management
    ImageService                          | yes             | image-management
    ImportService                         | yes             | query-url-import, xml-schema-converter
    LongTermPreservationValidationService | yes             | longtimepreservationvalidation
    MetadataValidationService             | yes             | validation
    MetsService                           | yes             | dataformat
[...]

PR #4224, opened 2021-03-02
Deactivate export buttons when process does not contain images

### LINKED ISSUE #3292: Show existence of images in process list
In version 2, buttons to export METS, export PDF and export to DMS are deactivated if a process does not contain any image files. This is very helpful in several scenarios when image files are necessary and you want to skip processes without image files.

**Goal:** Buttons will be disabled if there are no images in the process

solth:
Fixes #3292


Kitodo/src/main/java/org/kitodo/production/forms/ProcessListBaseView.java
```diff
@@ -453,4 +454,20 @@ public void delete(ProcessDTO processDTO) {
...
+
+    /**
+     * Check and return whether process with given ID has an empty generator folder or not.
+     *
+     * @param processId process ID
+     * @return whether given URI points to empty directory or not
+     */
+    public boolean hasImages(int processId) {
+        try {
+            return FileService.hasImages(processId);
```
matthias-ronge, line 466: At the moment, we do access File Service using the Service Loader. However, I prefer static access where possible. See #3327.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1281,4 +1284,25 @@ public URI deleteFirstSlashFromPath(URI uri) {
...
+     * Check and return whether process with given ID has an empty generator folder or not.
+     *
+     * @param processId process ID
+     * @return whether given URI points to empty directory or not
+     * @throws IOException thrown if listing contents of given URI is not possible
+     */
+    public static boolean hasImages(int processId) throws IOException, DAOException {
+        Process process = ServiceManager.getProcessService().getById(processId);
+        URI processUri = ServiceManager.getProcessService().getProcessDataDirectory(process);
+        String sourceDir = processUri.toString() + process.getProject().getGeneratorSource().getRelativePath();
```
matthias-ronge, line 1298: This will fail if `generatorSource` is `null`.

matthias-ronge: This won’t work. `VariableReplacer` needs to be invoked on the path returned for generatorSource (for example “images/(processtitle)_tif” must be converted to “images/1234567X_LoremIpsum_tif”)

Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/processActionsColumn.xhtml
```diff
@@ -119,15 +119,17 @@
 
             <h:commandLink action="#{ProcessListView.exportMets(process.id)}"
                            id="exportMets"
-                           styleClass="action"
-                           title="#{msgs.exportMets}"
+                           styleClass="action #{ProcessForm.hasImages(process.id) ? '' : 'disabled'}"
+                           disabled="#{not ProcessForm.hasImages(process.id)}"
```
matthias-ronge, line 123: Should export be disabled for hierarchically superior processes, too, which do never have images?

matthias-ronge (changes requested):
First issue: Each time the user interacts with the user interface, for each process that is displayed in the list, it is checked **eight times** on the file system to determine whether the process contains images.

In the [Restore View phase](https://www.tutorialspoint.com/jsf/jsf_life_cycle.htm):
* to determine the style for the commandLink exportMets
* to determine the disabled status for the commandLink exportMets
* to determine the style for the commandLink exportDms
* to determine the disabled status for the commandLink exportDms

Four more times in the [Render Response phase](https://www.tutorialspoint.com/jsf/jsf_life_cycle.htm). Caching should be implemented here.

Second issue: The way checking for images is error-prone, as it doesn’t handle all special cases. I would suggest the following.

```java
public boolean hasImages(int processId) throws DAOException {
    Process process = ServiceManager.getProcessService().getById(processId);
    Folder generatorSource = process.getProject().getGeneratorSource();
    if(Objects.isNull(generatorSource)) {
        return false;
    }
    Subfolder sourceFolder = new Subfolder(process, generatorSource);
    return !sourceFolder.listContents().isEmpty();
}
```

BTW, I am unsure if this method should be in `FileService` or `ProcessService`.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
solth: It did work in the examples I tested because I didn't use variables in the folder configuration of my project. Thanks for pointing this out.

Kitodo/src/main/java/org/kitodo/production/forms/ProcessListBaseView.java
solth: So I take it you agree with `hasImages` remaining `static`?

matthias-ronge: @Kathrin-Huber should decide

matthias-ronge (approved):
This seems to cover all requirements.

PR #4228, opened 2021-03-03
Add PDF to ImageFileFormats that can be created with ImageMagick

solth:


matthias-ronge (approved):
I didn’t try it, but it looks correct.

PR #4256, opened 2021-03-05
make filegroup editable

### LINKED ISSUE #3964: METS-Filegroup cannot be entered
### Problem
In the project settings, it is not possible anymore to enter the METS filegroup. 

### Solution 
It should be possible to enter Text in the field. 
It should be possible to configure the menu. 

### Example
metsfilegroup01

Kathrin-Huber:
fixes #3964

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/projectEdit/projectEditMetsPopup.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/projectEdit/projectEditMetsPopup.xhtml
@@ -107,7 +107,7 @@
                             <p:column style="width:50%">
                                 <p:outputLabel for="fileGroup" value="#{msgs['editFolderDialog.fileGroup']}"/>
                                 <p:selectOneMenu id="fileGroup" value="#{ProjectForm.myFolder.fileGroup}" effect="fold"
-                                                 editable="#{ProjectForm.myFolder.fileGroup == null}">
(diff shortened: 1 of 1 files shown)


matthias-ronge (changes requested):
Being able to change that has been disabled on purpose. The `fileGroup` property is used as _business key_ to identify the folder. This is a Hibernate problem (see [here](https://docs.jboss.org/hibernate/stable/core.old/reference/en/html/persistent-classes-equalshashcode.html)). Making this editable will cause problems when deleting a folder that has not yet been saved to the database ([this line](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/forms/ProjectForm.java#L347)). This is because [`hashCode()` and `equals()` of the `Folder`](https://github.com/kitodo/kitodo-production/blob/master/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Folder.java#L472-L486) rely on it.

If you want to enable changing this, you need to define a different business key that is unique (not `null`) for **both** folders that have _not yet been saved_ to the database and folders that already have been saved to the database. (The _ID_ of the folder _does not_ do this job, since it is `null` for **all** folders that have not yet been saved to the database.)

Kathrin-Huber:
Ok you are right,
but filegroup should at least be changeable, until folder is saved

matthias-ronge (approved):
I would have sworn stone and leg that it wouldn't work, but it works. Yes, I am amazed. 👏

PR #4263, opened 2021-03-09
use keyword for wildcardsearch

### LINKED ISSUE #3938: Wildcard search does not work properly
### Problem
The wildcard search is only possible in specific parts of the process title:

1. **"7746**" has the following results: 
- Aben_**7746**18868-1873032402_02-a
- Aben_**7746**18868-1873
- ...

2. "**1873**" has the following results: 
- Aben_774618868-**1873**032402_02-a
- Aben_774618868-**1873**
- ...

3. "Aben_774618868-1873" has no results
- It seems as if "-" blocks the wildcard search. 

### Solution
It should be possible to search for "Aben_774618868-1873032" and get the following results: 
- Aben_774618868-18730321
- Aben_774618868-18730322

Kathrin-Huber:
fixes #3938

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
@@ -220,9 +220,9 @@ public class ProcessServiceIT {
 
     @Test
     public void shouldFindProcessWithUnderscore() throws DataException, DAOException {
-        Project projectDTO = ServiceManager.getProjectService().getById(1);
+        Project project = ServiceManager.getProjectService().getById(1);
         Process process = new Process();
-        process.setProject(projectDTO);
+        process.setProject(project);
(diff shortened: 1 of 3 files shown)


Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
```diff
@@ -217,6 +218,22 @@ public void shouldFindByMetadataContent() throws DataException {
         assertEquals(processNotFound, 1, processService.findByAnything("SecondMetaShort").size());
     }
 
+    @Test
+    public void shouldFindProcessWithUnderscore() throws DataException, DAOException {
+        Project projectDTO = ServiceManager.getProjectService().getById(1);
```
solth, line 223: Why is this variable called `projectDTO`? It is an instance of `Project`, not `ProjectDTO`.

solth (approved):

Issue #4208, opened 2021-02-24
Update Elasticsearch


Kathrin-Huber:
## Description
Elasticsearch is used in an old version. There are two options:
#### Option 1
Update Elasticsearch to a newer Version. 
Version 6 of the Elasticsearch index uses a new format. The so-called mapping types have been removed. This requires changing the way Elasticsearch is addressed so that we can use a more recent version of Elasticsearch.

#### Option 2
Use Elasticsearch Integration of Hibernate, this resolves a few issues like

correct deleting of data in Hibernate and ES
writing similar queries instead of two different (HQL for Hibernate and ES-dialect for ES)
not adjusting ourself the changes for new ES versions (less maintenance for Kitodo.Production)

## TODO
- [ ] evaluate Options and decide
- [ ] implement choosen Option

## Estimated Costs and Complexity
this is a **high-ranged** project with 15-20 PT

## Related Issues
* **#3153** 
* #3158 
* #3719
* #3999
* #3713 
* #3694 
* #3496
* #4331

solth:
Would there be any downside to the second option? It sounds like the a certain winner!

stweil:
Do we expect problems with the recent [license change](https://www.elastic.co/de/pricing/faq/licensing) for Elasticsearch?

stefanCCS:
As general comment: I would appreciate very much, if we can get rid off old elastic search version 5 :-) ,
as this old version can be considered as some kind of security risk.

I cannot evaluate, if option 1 or 2 (or any other option) is better, but still I would like you to please consider the following:

- As far as I know already elastic search version 7 is existing - so, why not migrate directly to 7 ?

- Please consider migration efforts, which are created out of this change. In my opinion it is worth to do a more complex software develop in KITODO, if the migration effort for all instances is lower.
  e.g. we have setup an own elasticsearch server, so it is somehow easy for us to setup a new one, with version 6 or 7, and assuming that the a new KITODO-Software version will simply support this version, the migration is somehow simple (as the KITODO-Software can already today re-create the whole index, there is no need to migrate the elasticsearch database itself).

stefanCCS:
> 
> 
> Do we expect problems with the recent [license change](https://www.elastic.co/de/pricing/faq/licensing) for Elasticsearch?

[...]

Issue #4099, opened 2020-11-17
Limitation of processes in the excel file (statistics) 


andre-hohmann:
### Problem
In the generated excel file, the amount of included processes seems to be limited to 1.000. 

### Solution
In the excel file, all retrieved processes should be included. 

See also: #3507.

matthias-ronge:
I looked into the code but found no limiting. The processes are [fetched from the index](https://github.com/kitodo/kitodo-production/blob/faeecab9f629ae96a966d590cf93ab265fc944f3/Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java#L85). May this be an ElasticSerach limit?

andre-hohmann:
As the reason for the limitation is the elastic search, i have opened a new issue: 
- #4331

PR #4277, opened 2021-03-15
raise indexlimit to given limit from elasticsearch

Kathrin-Huber:
Elasticsearch is configured not to return more than 10'000 results (or to search behind that number).

It is not advised to change this value, but we can nevertheless get amounts up to this limit.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/search/SearchRestClient.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/search/SearchRestClient.java
@@ -174,7 +174,7 @@ public class SearchRestClient extends KitodoRestClient {
         if (Objects.nonNull(size)) {
             sourceBuilder.size(size);
         } else {
-            sourceBuilder.size(1000);
+            sourceBuilder.size(10000);
         }
 
         SearchRequest searchRequest = new SearchRequest(this.index);


henning-gerhardt:
I'm unsure, if your change is really fixing this problem or only moving it to an other issue. 

Between: If this changed method is called with a value for parameter `size` then this value is used and this value can be bigger then `10.000` which will cause the next issue.

Kathrin-Huber:
> I'm unsure, if your change is really fixing this problem or only moving it to an other issue.
> 
> Between: If this changed method is called with a value for parameter `size` then this value is used and this value can be bigger then `10.000` which will cause the next issue.

But we already had this problem before, and this PR doesn't change anything on it. (a different PR may fix this Problem)
There may be a problem, if an offset is added to the max value, but no method call is adding an offset, to a `null` size, so currently we are save. (but this could also be catched in a separat PR)

henning-gerhardt:
Ok. I thought this PR should fix this issue for all time.

solth (commented):
Will _all_ documents be included in the resulting excel file if the system contains more than 10.000 documents? I obviously do not have enough processes to test this, but the linked issue description does say that _all_ processes should be included in the excel file, not that the limit should just be increased from 1.000 to 10.000.

matthias-ronge:
I don't know exactly, but somewhere there is a limit to the maximum number of rows in an Excel spreadsheet, right? Certainly not at 10,000 yet, but...

Kathrin-Huber:
That's right, only 10'000 documents will be included. but it is just not possible to get more than this (even with a loop, or paging ect...)
A filter may help, or sorting, so one can add the first 10'000 and the last 10'000 but this are only hotfixes. (like this PR)

Honestly, I hope that #4208 will fix this.

This PR will at least add significant more documents to the excel sheet than 1000.

solth:
I understand that the proposed change mitigates the problem a little, but it does not _solve_ the linked issue. Therefore I would suggest to remove the link to the issue - so that it isn't closed when this PR gets merged - if no other changes are made to the pull request.

With the limitations ElasticSearch 5 brings we should perhaps add a warning dialog that pops up if the system contains more than 10.000 processes and that informs the user that the resulting Excel file will not be complete because of technical restrictions.

andre-hohmann:
Maybe, i should create a new issue. The main problem ist, that it is not possible to retrieve more then 10.000 processes in Kitodo.Production. In some cases more hits have to be retrieved, for example: 
- Retrieval of all processes in a project to answer the question of the projects size
- Retrieval of all issues of a newspaper to add metadata like collection, rights information
- ...

It seems that the creation of the Excel file with more then 10.000 processes would be possible, if the more processes could be retrieved.

andre-hohmann:
I have created a issue: #4331. 
From my point of view, the issue #4099 should only be closed, when there is no limitation in the result list anymore. Nevertheless i mentioned this problem in #4331. 

I strongly agree with @solth:
> With the limitations ElasticSearch 5 brings we should perhaps add a warning dialog that pops up if the system contains more than 10.000 processes and that informs the user that the resulting Excel file will not be complete because of technical restrictions.

PR #4292, opened 2021-03-21
Improve pagination

IkramMaalej:


IkramMaalej:
- Fix pagination's disappearance of the last page
- Synchronise pagination panel after drag and drop in gallery 
- Synchronise selected pages in pagination panel with gallery's current selection

PR #4294, opened 2021-03-22
Let getter for password return empty string in UserForm

### LINKED ISSUE #4238: Changing password - previous changed password displayed
If you must change a few user passwords as administrator the previous used password is automatically filled out in the password change dialog!

solth:
Fixes #4238

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
@@ -594,12 +594,12 @@ public class UserForm extends BaseForm {
     }
 
     /**
-     * Gets password.
+     * Return empty string. Returning the actual password is never required, but GUI needs a getter for form fields.
      *
-     * @return The password.
+     * @return Empty string.
      */
     public String getPasswordToEncrypt() {
-        return passwordToEncrypt;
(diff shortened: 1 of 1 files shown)


henning-gerhardt (changes requested):
If you added a (selenium) test then my review would be more positive, so I must check this change in a manual way and this change may work for the next couple of time. If in the future someone is changing this method and ignoring the method comment than this issue will show up again.

matthias-ronge:
It's always true. An application should consist of 90% testing and 10% implementation. But who pays for it?

henning-gerhardt:
>  An application should consist of 90% testing and 10% implementation. But who pays for it?

You are paying now for less investments in future where the costs of repairing will be much bigger then implementing it now.

matthias-ronge:
I know that myself, I mean who is paying me / us to do this now?

solth:
Generally I agree with the idea of having as much funcitionality covered by tests, but at the same time I think we should apply some common sense when deciding in which cases they are really necessary. 

This particular change just affects a _getter_ called by the GUI - that now just returns a static value, nonetheless - for which the requirement of a dedicated Selenium test seems a bit obsessive, IMHO.

PR #4305, opened 2021-03-24
Rewrite VariableReplacer

### LINKED ISSUE #3718: VariableReplacer does not work properly in the script step

matthias-ronge:
Fixes #3718

This pull request implements the `VariableReplacer` in a new way. It was still based on the old UGH classes, and it was very inefficient because the string was parsed for every possible variable and all possible replacements were calculated even when they weren't needed. This will be remedied with this pull request. The string is examined only one more time with the regular expression, and then only the replacements that are required are determined. In addition, meaningful warnings are issued in the log if a value cannot be determined.

There were a lot of hard-coded variables from 2.x in `VariableReplacer` that no longer make sense with the flexible folder model. I would have liked to have removed these completely, but @henning-gerhardt said they should be retained for backwards compatibility (see the discussion in the issue). I have outsourced these to the configuration. With the specified configuration entries, the obsolete variables continue to function with the same semantics as in 2.x.

PS: Look at the new `VariableReplacer` as it, is and not at the diff. There is practically nothing left of the old class, and the diff tries hard to bring things together that don't belong together. It's totally misleading.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/export/ExportDms.java
+++ b/Kitodo/src/main/java/org/kitodo/export/ExportDms.java
@@ -386,7 +386,7 @@ public class ExportDms extends ExportMets {
         Collection<Subfolder> processDirs = process.getProject().getFolders().parallelStream()
                 .filter(Folder::isCopyFolder).map(folder -> new Subfolder(process, folder))
                 .collect(Collectors.toList());
-        VariableReplacer variableReplacer = new VariableReplacer(null, null, process, null);
+        VariableReplacer variableReplacer = new VariableReplacer(null, process, null);
 
         String uriToDestination = destination.toString();
         if (!uriToDestination.endsWith("/")) {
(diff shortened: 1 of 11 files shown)


henning-gerhardt:
If there would be a provided way to access the directory or any other information from the "folder" information on calling a script then it would not be necessary to stay with the old directory based variables.

matthias-ronge:
You are absolutely right about that, and I have already seen such inquiries and am thinking about adding this in a sensible way. But first, the functionality should be repaired and updated to the current code base. For example, the request was to pass all files from an images directory as parameters. But I deliberately left out this in order to make the pull request understandable and not too big.

matthias-ronge:
@Kathrin-Huber Are you missing something here, or can you merge this?

matthias-ronge:
@Kathrin-Huber Are you still missing something here?

matthias-ronge:
@Kathrin-Huber Is there anything else I should do?

PR #4307, opened 2021-03-25
Return to the correct process list's page after opening metadataeditor

### LINKED ISSUE #4246: Process list forgets current page
When going to a certain page of the processes list, like the 28th page, open the metadata editor but accidentally it was the wrong process, you close it, the process list is again on page 1 and you have to click through to page 28 again.

**Goal:** The process list should remember on which page it was when you come back to it

IkramMaalej:
Fixes #4246


IkramMaalej:
Fixes #4246

solth (approved):

PR #4325, opened 2021-04-08
Enable default values in OPAC import configuration

### LINKED ISSUE #3561: Default values in kitodo_opac.xml
The catalogue search during the creation of a new process could be more comfortable, if it was possible for "Catalogue" and "Field" to define default values. 
As most users are working with one catalogue and are searching for an identifier, the manual configuration is not necessary, but after a while quite annoying. 

The target state: 
cataloguesearch04

The current state:
cataloguesearch01

The current way to the target state: 
cataloguesearch02

cataloguesearch03

solth:
Fixes #3561 

Default values for catalogs and search fields can be configured by adding the attibute `default="true"` to the corresponding XML elements in `kitodo_opac.xml`. If multiple fields are configured to be default, the first one is selected.

See changes in `Kitodo/src/main/resources/kitodo_opac.xml` for examples!

@subhhwendt, @andre-hohmann I have added these changes to the preview system so that you can test them.

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
+++ b/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
@@ -32,6 +32,8 @@ import org.kitodo.exceptions.ParameterNotFoundException;
 public class OPACConfig {
     private static final Logger logger = LogManager.getLogger(OPACConfig.class);
     private static XMLConfiguration config;
+    private static final String TRUE = "true";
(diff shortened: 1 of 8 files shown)


andre-hohmann:
Thanks a lot - that is much more convenient. 
I changed the default-attribute to another catalogue and it works. The PPN is also set as default. 
Thus, from my point of view, it works.

subhhwendt:
These new configuration possibilities support the import process a lot!

Issue #4229, opened 2021-03-03
Enable dependent value restrictions in the ruleset


matthias-ronge:
A functionality is desired in the ruleset so that only certain values can be selected in one field, depending on what is in another field.

matthias-ronge:
There are two imaginable ways to implement this.

### Variant 1

Only in the rule set. Example:

```xml
<declaration>
    <key id="area">
        <label>Area code</label>
        <label lang="de">Flurstücknummer</label>
        <option value="1234-2a"/>
        <option value="1234-2b"/>
        <!-- ... 4.000 area codes ... -->
    </key>
    <key id="address">
        <label>Address</label>
        <label lang="de">Adresse</label>
        <key id="street">
            <label>Street</label>
            <label lang="de">Straße</label>
            <option value="Buschweg"/>
            <option value="Pastor-Luis-Straße"/>
            <option value="Rosenstraße"/>
            <option value="Vereinsstraße"/>
            <!-- ... 15.000 streets ... -->
        </key>
        <key id="house">
            <label>House no.</label>
            <label lang="de">Hausnummer</label>
            <option value="1"/>
            <option value="1a"/>
            <option value="1b"/>
            <option value="2"/>
            <option value="2 Rgb.">
                <label>2, rear building</label>
                <label lang="de">2, Rückgebäude</label>
            </option>
            <option value="2a"/>
            <option value="3"/>
            <option value="4"/>
            <option value="4a"/>
            <option value="5"/>
            <!-- ... 1000 house numbers ... -->
        </key>
    </key>
</declaration>
<correlation>
    <restriction key="address">
        <permit key="street" minOccurs="1" maxOccurs="1" unspecified="forbidden">
            <condition key="^area" equals="1234-2b">
                <permit value="Rosenstraße"/>
                <permit value="Vereinsstraße"/>
            </condition>
        </permit>
        <permit key="house" minOccurs="1" maxOccurs="1" unspecified="forbidden">
            <condition key="^area" equals="1234-2b">
                <condition key="street" equals="Rosenstraße">
                    <permit value="3"/>
                    <permit value="5"/>
                    <permit value="7"/>
                </condition>
                <condition key="street" equals="Vereinsstraße">
                    <permit value="2"/>
                    <permit value="2 Rgb."/>
                    <permit value="4"/>
                    <permit value="4a"/>
                </condition>
             </condition>
        </permit>
    </restriction>
</correlation>
```

[...]

PR #4333, opened 2021-04-12
Implement XML model for dependent value restrictions

matthias-ronge:
First part of #4229. In this pull request, the XML model for rulesets is expanded so that condition blocks can be defined.

This is a follow-up pull request to #4329, [immediate diff](https://github.com/matthias-ronge/kitodo-production/compare/matthias-ronge:metadata-in-ruleset...issue-4229_1.1)


matthias-ronge:
@Kathrin-Huber Are you missing something here, or can you merge this?

PR #4336, opened 2021-04-14
Add process title to the confirmation dialog message for deleting the…

### LINKED ISSUE #4323: Include elements name/title in 'delete' dialog
The `confirmDelete` dialog does not display the name/title of the object that is going to be deleted.

Bildschirmfoto 2021-04-06 um 16 19 55

This is impractical and also a little dangerous, as the user can easily delete the wrong object on accident.

IkramMaalej:
… process
Fixes #4323


Kitodo/src/main/java/org/kitodo/production/forms/BaseForm.java
```diff
@@ -406,4 +407,8 @@ public void resetPaginator(String keepPagination) {
             this.setFirstRow(0);
         }
     }
+
```
solth, line 410: Javadoc missing

solth (changes requested):
This is a good first step, but needs to be implemented for _all_ types of objects that can be deleted - users, projects, templates etc. - in order to fix the linked issue.

PR #4340, opened 2021-04-15
Implement business logic for dependent value restrictions

### LINKED ISSUE #4229: Enable dependent value restrictions in the ruleset

matthias-ronge:
Resolves #4229

In this pull request, the business logic to calculate dependent value restrictions is implemented. Since a large amount of values has to be checked here (> 10,000 values per entry in the present case), this was implemented with `HashMap`s (class `ConditionMap`), which is initialized once, and then always used.

A special task is to return from the output of selection elements in the ruleset which fields they depend on, so that the application knows when these text fields are changed, that it has to refresh the selection elements. This is done with the `Map` to `Boolean`, whereby the element-generating function sets the boolean to true if this metadata entry is leading, i.e. changing this metadata entry must trigger a refresh.


Kathrin-Huber:
What is happening here? Are you still working on this?

matthias-ronge:
Put on hold by management

matthias-ronge:
This pull request would then be ready for review. I could not yet understand why the selenium test fails.

matthias-ronge:
@Kathrin-Huber this branch is waiting for merge. I did a rebase, but Selenium test still fails, I think the problem is because of #5018, but unrelated to the changes in this branch. It would be very helpful if this gets merged, because Pull Request #4969 is incomplete, because it lacks handling classes that are introduced in this branch.

matthias-ronge:
The selenium test fails again, but that doesn't seem to have anything to do with the changes in this branch.

BartChris:
Since you are also changing the code for metadata validation logic maybe you can also have a look at the issue/bug which might be inside the valdation logic:

https://github.com/kitodo/kitodo-production/issues/5009

matthias-ronge:
No. We handle each bug in a separate pull request. This keeps each pull request small. The pull requests are reviewed and the reviewer should be able to understand what the pull request is doing. Resolving multiple issues in one pull request is more accidental (although it sometimes happens when different issues have the same technical root cause) than intentional.

BartChris:
Thank you, sounds reasonable.

PR #4365, opened 2021-04-29
Create structure element from consecutive pages at current position

### LINKED ISSUE #3486: Add new InsertionPosition for structural elements in metadata editor
When creating new structural elements in the metadata editor from selected consecutive pages, an additional option should be added for adding the structural element at the position of the selected pages (e.g. the index of the first selected page in the parent structural element).

solth:
Fixes #3486 

A new option "Current position" is added to the "Add new structure element" dialog if consecutive pages/media units in one existing structure element are selected. The new structure element will assume the lowest `ORDER` attribute of the selected pages/media units to include it at the correct position between other existing structure elements and media units.


Kathrin-Huber:
thanks for the renaming of `docStructPositionSelectionSelectedItem  ` ;-)

PR #4398, opened 2021-05-11
Move flag

### LINKED ISSUE #3264: Move flag 'separateStructure' from Task to Workflow
With the current structure of Kitodo.Production 3, all tasks in a workflow must have the same value for the flag `separateStructure`. This makes saving the flag for each task individually redundant. 

For checking conditions in the MetadataEditor whether the combined or the separate structure tree should be shown, the editor therefor also needs to maintain a handle on the `currentTask` of a process, which unnecessarily complicates matters.

Because of this, the flag should be moved from the Task object to the Workflow object.

Kathrin-Huber:
move separate Structure flag from task to workflow
fixes #3264


solth:
Please fix `MetadataST.hideStructureDataTest`!

PR #4399, opened 2021-05-11
Restrict query encoding to SRU interfaces

### LINKED ISSUE #4353: Problem with encoding of queries with identifier
Hello,

it seems like the fix introduced in https://github.com/kitodo/kitodo-production/pull/4187 has unintended(?) side effects since there seems to be no difference between URLs where the Identifier is embedded in another query paramter as in the HEBIS search:
http://sru.hebis.de/sru/DB=2.1?version=1.1&operation=searchRetrieve&recordSchema=pica&recordPacking=xml&maximumRecords=1&query=pica.ppn=406442703  (see https://github.com/kitodo/kitodo-production/issues/4185)

and other scenarios. 

A query like 

https://genericapi.com?id=123456
seems also to be encoded as
https://genericapi.com?id%3D123456 where id therefor cannot be processed as a CGI-parameter.

https://github.com/kitodo/kitodo-production/blob/04253d1b5c817187a55b54aced06ddb69630cc54/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java#L282

Maybe there has to be a check, wether the query is targeted at a SRU interface or not and only encode the query URL in case of a SRU interface. It seems like something like that is also done in this place:

https://github.com/kitodo/kitodo-production/blob/04253d1b5c817187a55b54aced06ddb69630cc54/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java#L436

solth:
Fixes #4353

### DIFF
--- a/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
+++ b/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
@@ -280,7 +280,7 @@ public class QueryURLImport implements ExternalDataImportInterface {
         }
         String prefix = Objects.nonNull(idPrefix) && !identifier.startsWith(idPrefix) ? idPrefix : "";
         String queryParameter = idParameter + equalsOperand + prefix + identifier;
-        if (Objects.nonNull(interfaceType) && SearchInterfaceType.SRU.equals(interfaceType)) {
+        if (SearchInterfaceType.SRU.equals(interfaceType)) {
             fullUrl += encodeQueryParameter(queryParameter, encoding);
         } else {
             fullUrl += queryParameter;
@@ -378,7 +378,7 @@ public class QueryURLImport implements ExternalDataImportInterface {
                 }
             }
             return performQuery(queryString + createSearchFieldString(searchFieldMap));
-        } catch (URISyntaxException | UnsupportedEncodingException | ResponseHandlerNotFoundException e) {
+        } catch (URISyntaxException | ResponseHandlerNotFoundException e) {
             throw new CatalogException(e.getLocalizedMessage());
         }
     }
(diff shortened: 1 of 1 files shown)


matthias-ronge (changes requested):
It's not that easy. Essentially, you are restoring the state as it was before. I had changed that on purpose because there was a catalog that hadn't worked if the parameter wasn't coded. Unfortunately I don't know which one anymore. Could have been the ZDB, but I don't remember. Maybe I can find out. In any case, we have to make that configurable if one catalog requires the URL to be encoded and the other rejects it encoded.

Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
```diff
@@ -279,7 +279,12 @@ private DataRecord performQueryToRecord(String queryURL, String identifier) thro
             }
         }
         String prefix = Objects.nonNull(idPrefix) && !identifier.startsWith(idPrefix) ? idPrefix : "";
-        fullUrl += encodeQueryParameter(idParameter + equalsOperand + prefix + identifier, encoding);
+        String queryParameter = idParameter + equalsOperand + prefix + identifier;
+        if (Objects.nonNull(interfaceType) && SearchInterfaceType.SRU.equals(interfaceType)) {
```
matthias-ronge, line 283: Unnecessary `nonNull()` check. Enum constants compare to `null` as `false` and don’t throw `NullPointerException`.

Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
```diff
@@ -434,7 +439,7 @@ private String createSearchFieldString(LinkedHashMap<String, String> searchField
                 .collect(Collectors.toList());
         String searchString = String.join(" AND ", searchOperands);
         if (Objects.nonNull(interfaceType) && SearchInterfaceType.SRU.equals(interfaceType)) {
-            return URLEncoder.encode(searchString, StandardCharsets.UTF_8.displayName());
+            return encodeQueryParameter(searchString, encoding);
```
matthias-ronge, line 442: `encodeQueryParameter()` handles the `UnsupportedEncodingException`, please remove `throws UnsupportedEncodingException` from the method signature.

matthias-ronge (approved):
Sorry, I misunderstood the problem yesterday. Your solution is correct and should work.

By the way I found two minor improvements.

matthias-ronge (approved):

Issue #4038, opened 2020-10-08
Structure tree should show image number (or name), not internal order number


matthias-ronge:
In the example, pages 2-6 are to be added to the new structure element.

![Screenshot 1](https://user-images.githubusercontent.com/3040657/95461441-865d8980-0976-11eb-9973-9defe6b6a673.png)

![Screenshot 2](https://user-images.githubusercontent.com/3040657/95461454-89587a00-0976-11eb-9c72-74e9daf553fb.png)

The number displayed has changed. Picture 2 is now picture 1, and picture 1 is now picture 6. In short, you cannot use the (changing) numbers as a guide to identify the index of the scanned picture (and the continuous book, from front to back).

![Screenshot 3](https://user-images.githubusercontent.com/3040657/95462617-0801e700-0978-11eb-970d-420f07befff5.png)

**Goal:** In front of the colon, the canonical part of the file name without leading zeros should be displayed, not the internal order number. Then you would always know where you are.

andre-hohmann:
I strongly support this suggestion. The current behaviour is very confusing. 
From my point of view, the combined structure will be much more usable if Matthias goal is reached.

BartChris:
@matthias-ronge 

We also encountered this behaviour. I made some internal changes to display the images names in the structure tree. However i am not sure in which places the image name should be displayed. Apart from that i am no expert in JSF so i am hesitating to make a pull request here. So here just for reference the changes i did. Hopefully someone can adress this soon in the master because it really improves the editing experience:

Changes in StructurePanel.java:

old:
https://github.com/kitodo/kitodo-production/blob/61f0b530b3013b7c3ffc2499be49faad0dcf9d31/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java#L540

```
    private String buildViewLabel(View view) {
        String order = view.getMediaUnit().getOrder() + " : ";
        if (Objects.nonNull(view.getMediaUnit().getOrderlabel())) {
            return order + view.getMediaUnit().getOrderlabel();
        } else {
            return order + "uncounted";
        }
    }
```
new:
```
import org.apache.commons.io.FilenameUtils;

private String buildViewLabel(View view) {
        URI uri = view.getMediaUnit().getMediaFiles().entrySet().iterator().next().getValue();
        String fileName = FilenameUtils.getBaseName(uri.getPath()) + " : ";
        if (Objects.nonNull(view.getMediaUnit().getOrderlabel())) {
[...]

PR #4401, opened 2021-05-11
Change label generation in Structure Panel

BartChris:
This pull requests changes the label generation in the structure panel following the guidelines from @matthias-ronge. The label now is constructed from the image name with leading zeros removed.

 fixes https://github.com/kitodo/kitodo-production/issues/4038


solth:
I suspect there may be some conceptional misunderstanding or conflict in the linked issue (please see my comment there) and think this pull request should not be merged, yet.

matthias-ronge (changes requested):
Nice that it did work, and a smart regular expression! I wouldn't have thought of that!

When I looked at the code again, I found three other problems:

- If a media unit has not assigned any files (for example in the case of “partial orders”) it would crash with an exception.
- Reinitializing the subfolder for each media unit (there could be 2000 in one process!) makes page loading slower. I would provide a cache here.
- If the subfolder is not found (this shouldn't be the case, then something went wrong in the program beforehand) then there should be a speaking error message.

I have exemplarly written this down here (untested):

```java
private Map<String, Subfolder> subfoldersCache = new HashMap<>();

private String buildViewLabel(View view) {
    MediaUnit mediaUnit = view.getMediaUnit();
    Iterator<Entry<MediaVariant, URI>> mediaFileIterator = mediaUnit.getMediaFiles().entrySet().iterator();
    String canonical = "-";
    if (mediaFileIterator.hasNext()) {
        Entry<MediaVariant, URI> mediaFileEntry = mediaFileIterator.next();
        Subfolder subfolder = subfoldersCache.computeIfAbsent(mediaFileEntry.getKey().getUse(),
            use -> new Subfolder(dataEditor.getProcess(), dataEditor.getProcess().getProject().getFolders()
                    .parallelStream().filter(folder -> folder.getFileGroup().equals(use)).findAny()
                    .orElseThrow(() -> new IllegalStateException("Missing folder with file group \"" + use
                            + "\" in project \"" + dataEditor.getProcess().getProject().getTitle() + '"'))));
        canonical = subfolder.getCanonical(mediaFileEntry.getValue());
    }
    return canonical.replaceFirst("^0+(?!$)", "") + " : "
            + (Objects.isNull(mediaUnit.getOrderlabel()) ? "uncounted" : mediaUnit.getOrderlabel());
}
```

solth:
> I suspect there may be some conceptional misunderstanding or conflict in the linked issue (please see my comment there) and think this pull request should not be merged, yet.

After discussing this topic with @matthias-ronge I can now better understand that these changes make sense, so I wouldn't oppose them anymore.

@BartChris can you rebase your branch against the master to get rid of failing Travis checks?

PR #4415, opened 2021-05-18
use saveOrUpdate instead of merge

Kathrin-Huber:
Due to performance problems saveOrUpdate is the faster choice


solth (commented):
I am no expert with Hibernate, but from what I gather, `saveAndUpdate` instead of `save`/`merge` at least increases the risk of running into exceptions when an object is not attached to the current session (when using `merge`, Hibernate itself handles attaching objects to session when needed).
Is the performance increase significant enough for us to introduce this new potential course for problems?

solth (approved):

PR #4416, opened 2021-05-18
add title on migration

Kathrin-Huber:
Is adding the newspaper title to the meta.xml

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
+++ b/Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
@@ -21,18 +21,21 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.TreeMap;
 import java.util.concurrent.TimeUnit;
 
+import java.util.stream.Collectors;
 import javax.naming.ConfigurationException;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.kitodo.api.Metadata;
 import org.kitodo.api.dataeditor.rulesetmanagement.DatesSimpleMetadataViewInterface;
+import org.kitodo.api.dataeditor.rulesetmanagement.FunctionalMetadata;
 import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.SimpleMetadataViewInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.StructuralElementViewInterface;
@@ -597,10 +600,7 @@ public class NewspaperProcessesMigrator {
 
         ServiceManager.getFileService().createProcessLocation(yearProcess);
 
-        Workpiece yearWorkpiece = new Workpiece();
-        yearWorkpiece.setId(yearProcess.getId().toString());
-        yearWorkpiece.setRootElement(yearToCreate.getValue());
-        metsService.saveWorkpiece(yearWorkpiece, fileService.getMetadataFilePath(yearProcess, false, false));
+        createWorkpiece(yearToCreate, yearTitle, yearProcess);
 
         Collection<Integer> childIds = yearsChildren.get(yearToCreate.getKey());
         for (Integer childId : childIds) {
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
```diff
@@ -619,6 +619,22 @@ public void createNextYearProcess() throws ProcessGenerationException, IOExcepti
         }
     }
 
+    private void createWorkpiece(Entry<String, IncludedStructuralElement> yearToCreate, String yearTitle, Process yearProcess) throws IOException {
+        Workpiece yearWorkpiece = new Workpiece();
+        yearWorkpiece.setId(yearProcess.getId().toString());
+        yearWorkpiece.setRootElement(yearToCreate.getValue());
+        RulesetManagementInterface ruleset = ServiceManager.getRulesetService().openRuleset(overallProcess.getRuleset());
```
matthias-ronge, line 626: Please do not parse the ruleset file every time. It is already opened in function `initializeMigrator()`. You better change the local variables `rulesetManagement` and `newspaperView` there (line 284, 286) into class fields, and just access them here.

Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
```diff
@@ -619,6 +619,22 @@ public void createNextYearProcess() throws ProcessGenerationException, IOExcepti
         }
     }
 
+    private void createWorkpiece(Entry<String, IncludedStructuralElement> yearToCreate, String yearTitle, Process yearProcess) throws IOException {
```
matthias-ronge, line 622: Different workpieces are assembled in this class. Could you rename this method to `createYearWorkpiece`?

Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
```diff
@@ -619,6 +619,22 @@ public void createNextYearProcess() throws ProcessGenerationException, IOExcepti
...
+        Workpiece yearWorkpiece = new Workpiece();
+        yearWorkpiece.setId(yearProcess.getId().toString());
+        yearWorkpiece.setRootElement(yearToCreate.getValue());
+        RulesetManagementInterface ruleset = ServiceManager.getRulesetService().openRuleset(overallProcess.getRuleset());
+        StructuralElementViewInterface newspaperView = ruleset.getStructuralElementView(yearWorkpiece.getRootElement().getType(), acquisitionStage, Locale.LanguageRange.parse("de"));
+        final Collection<String> processTitleKeys = ruleset.getFunctionalKeys(FunctionalMetadata.PROCESS_TITLE);
+        List<SimpleMetadataViewInterface> yearViews = newspaperView.getAllowedMetadata().parallelStream().filter(SimpleMetadataViewInterface.class::isInstance)
+                .map(SimpleMetadataViewInterface.class::cast)
+                .filter(metadataView -> processTitleKeys.contains(metadataView.getId())).collect(Collectors.toList());
+        for (SimpleMetadataViewInterface yearView : yearViews) {
```
matthias-ronge, line 632: You shouldn't need a `for` after a stream. You can combine the two expressions:
```java
newspaperView.getAllowedMetadata().parallelStream()
    .filter(SimpleMetadataViewInterface.class::isInstance).map(SimpleMetadataViewInterface.class::cast)
    .filter(metadataView -> processTitleKeys.contains(metadataView.getId()))
    .forEachOrdered(yearView ->
        MetadataEditor.writeMetadataEntry(yearWorkpiece.getRootElement(), yearView, yearTitle)
    );
```

Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
```diff
@@ -619,6 +619,22 @@ public void createNextYearProcess() throws ProcessGenerationException, IOExcepti
         }
     }
 
+    private void createWorkpiece(Entry<String, IncludedStructuralElement> yearToCreate, String yearTitle, Process yearProcess) throws IOException {
+        Workpiece yearWorkpiece = new Workpiece();
+        yearWorkpiece.setId(yearProcess.getId().toString());
+        yearWorkpiece.setRootElement(yearToCreate.getValue());
+        RulesetManagementInterface ruleset = ServiceManager.getRulesetService().openRuleset(overallProcess.getRuleset());
+        StructuralElementViewInterface newspaperView = ruleset.getStructuralElementView(yearWorkpiece.getRootElement().getType(), acquisitionStage, Locale.LanguageRange.parse("de"));
+        final Collection<String> processTitleKeys = ruleset.getFunctionalKeys(FunctionalMetadata.PROCESS_TITLE);
```
matthias-ronge, line 628: Looks to me like you've messed up something here. `yearTitle` contains the name of the year (e.g. `2021`), but `FunctionalMetadata.PROCESS_TITLE` refers to the metadata fields in which the process title (e.g. `LoreIpsu_1234567X-2021`) is to be written. If you want to save the name of the year it is much easier to get the name of the metadata field: `yearSimpleMetadataView.getId()`.

matthias-ronge (commented):
Now that I believe that I have understood what you are up to, this can be achieved with one line:
```java
MetadataEditor.writeMetadataEntry(yearWorkpiece.getRootElement(), yearSimpleMetadataView.getId(), yearTitle);
```
The more complicated way would only be necessary if you want to write the process title to the file, but I don't think that's the point.

Kathrin-Huber:
I DO want to write the processtitle to the file. Thats exactly, what I want to achieve.
And in my test processes, thats exactly what happened:
`<kitodo:metadata name="processTitle">DieEn_495715131_1928</kitodo:metadata>`
It didn't just write the name, as you implied.

Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
Kathrin-Huber: This is right for the ruleset, but not for the newspaper, because I explicitly need the yearView.

Kathrin-Huber:
Thank you for the fast review!

matthias-ronge (approved):
Okay, then I misunderstood that. In that case, the solution is the right one. yes and I see that `yearTitle` is actually the process title. I guess I was wrong. Anyway, the solution now looks totally correct!

Issue #3300, opened 2020-03-18
Optimize the performance of series migration


matthias-ronge:
The migration function for series and multi-volume works is very slow. It processes an average of 1000 processes per hour. Just checking whether a process folder contains the anchor file (and if not, skipping it) takes a whole second. This is because services are used through the service loader, it is all very complex. There was also the impression that the existence of files was repeatedly checked many times.

**Goal:** Review functionality and improve performance. Do more things on the spot, remove everything that is unnecessary to increase processing speed.

andre-hohmann:
@matthias-ronge: Does the problem still exists? 

The SLUB Dresden faced the same problem, but it works now.

matthias-ronge:
By now, I had no chance to test it again.

Performance issues typically involve various factors, so judging this is always subjective. If you can confirm that a recent hierarchy migration of the SLUB data is going ahead in a timely manner, we can also close this and, if it turns out badly again, open it again.

PR #4417, opened 2021-05-18
speed up migrationPage

### LINKED ISSUE #4384: Not useable newspaper migration site
If you have a large database with a lot of newspaper processes (over 300,000) and corresponding batches (> 150) using the newspaper migration site is not possible as it takes hours or days  to display this page. While loading this page a lot of SELECT statements - which is ok - but even the same amount of UPDATE statements are started. It is for me unclear why UPDATE statements are used / fired on displaying this page.

Displaying the batches tab on the process page takes one or two minutes and is much faster then displaying the nearly same data on the newspaper migration site.

Kathrin-Huber:
fixes #4384

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
@@ -146,7 +146,7 @@ public class HierarchyMigrationTask extends EmptyTask {
             }
             while (progress < processesList.size()) {
                 Process process = processService.getById(processesList.get(progress));
-                if (fileService.processOwnsAnchorXML(process, true) && !fileService.processOwnsYearXML(process, true)) {
+                if (fileService.processOwnsAnchorXML(process) && !fileService.processOwnsYearXML(process)) {
                     setWorkDetail(process.getTitle());
                     migrate(process);
                 }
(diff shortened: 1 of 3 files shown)


matthias-ronge (commented):
This makes method `processOwnsYearXML(Process process)` in `FileService` dead code and you could delete it. Alternative suggestion: You combine the two methods  `processOwnsYearXML(Process process)` and  `processOwnsYearXML(Process process, boolean forIndexingAll)` in `FileService` (line 589−603) to this one method:

```java
public boolean processOwnsYearXML(Process process) throws IOException {
    URI yearFile = createYearFile(getMetadataFilePath(process, false, true));
    return fileExist(yearFile);
}
```

That should have the same effect.

Kathrin-Huber:
Thats a pretty good idea! That way I also found that `processOwnsAnchorXML` was dead.

henning-gerhardt (changes requested):
Your changes did not solve the issues mentioned in #4384 as your changes are inside the migration itself and not on displaying the newspaper batches. So you may rename the pull request title and remove the "fixes #4384".

henning-gerhardt:
I must partial revert my last comment as I got confused by the changes which including some refactoring. On a new migrated database the values for `processBaseUri` did not get updated any more and this is a big improvement for migration speed. But this pull request (or at least the current version of it) did not solve the slow page displaying at all as the check for the `year.xml` files are still executed and this takes still a lot of time (even less then storing the changed value in `processBaseUri` to the database and maybe index).

So far as I know the database and its entries there is a `type` field in `batch` table which is holding the information about what kind of batch is stored. Is this not enough to decide the displayed batches on the newspaper migration page?

matthias-ronge:
Slow file system access is a different issue (#3300) which also needs to be taken looked after, but I think this pull request fixes the database updates, and this is what talks are about in #4384

Using the batch.type field should be the best solution to find newspaper batches, I agree to this.

PR #4419, opened 2021-05-19
don't save processes from project

### LINKED ISSUE #3294: Don’t check existence of files when saving project
If changes to the project settings are saved, you can see in the log file that all metadata files are checked for their existence. This makes saving the project very slow if the project has many processes.

**Goal:** When saving a project, metadata files should not be searched for

Kathrin-Huber:
fixes #3294


matthias-ronge (approved):
I have not implemented this part and I cannot 100% understand whether this certainly has no side effects. I would still agree to this change.

PR #4421, opened 2021-05-19
Use batch type to find newspaper batches

matthias-ronge:
As @henning-gerhardt suggested in https://github.com/kitodo/kitodo-production/pull/4417#issuecomment-844019633, the batch type should be used to determine whether a batch is a newspaper batch.


henning-gerhardt (approved):
I could "test" your changes only on my local system and not on our migration system (local data vs data on and over network): your changes looks good and the displaying of this newspaper migration site is now fast.

PR #4422, opened 2021-05-20
reuse yearIncludedStructuralElement

### LINKED ISSUE #4396: Hierarchy migration: Newspaper-Year-level processes with workflow
Problem
After the the migration of serial publications (hierarchy migration), a workflow is assigned to the created processes for the newspaper-year-processes, as for example. 
Additionally, the display of the processes in the process is irritating: 
- in the process list only for some processes a workflow is shown
- in the database all process have tasks (and thus a workflow)

Solution
During the migration of serial publications (hierarchy migration), no workflow should be assigned to the newspaper-year-processes and no tasks should be assigned in the database.

See also: #4268

Kathrin-Huber:
fixes #4396

@matthias-ronge I'm not sure, why there was created a new IncludedStructuralElement at this place, maybe you can explain?

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
+++ b/Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java
@@ -401,9 +401,8 @@ public class NewspaperProcessesMigrator {
             year = MetadataEditor.getMetadataValue(yearFileYearIncludedStructuralElement, FIELD_TITLE);
         }
         IncludedStructuralElement processYearIncludedStructuralElement = years.computeIfAbsent(year, theYear -> {
-            IncludedStructuralElement yearIncludedStructuralElement = new IncludedStructuralElement();
-            MetadataEditor.writeMetadataEntry(yearIncludedStructuralElement, yearSimpleMetadataView, theYear);
-            return yearIncludedStructuralElement;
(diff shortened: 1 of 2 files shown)


matthias-ronge (commented):
> @matthias-ronge I'm not sure, why there was created a new IncludedStructuralElement at this place, maybe you can explain?

It was just a decision. `years` is a map which holds the includedStructuralElement (root of the logical structure tree) for each year-level process to be generated. Creating a new (empty) logical structure tree root element for each year process to be generated seemed straightforward for me. A new process is generated, and it should be empty at the beginning, shouldn’t it? This is what I thought.

Your change will copy the structure tree of a random `meta_year.xml` (from exactly one issue of that year, but not necessarily the first in chronology, but a random one, depending on which issue of a year is processed first). This may bring in a lot of stuff (metadata, children, METS pointers…) that you need to deal with. For example, version 2 stored the value of the LABEL and ORDERLABEL as metadata field `TitleDocMain` and `TitleDocMainShort` (sort title, with typo). Version 3 doesn’t. If you reuse the logical structure tree of version 2 internal METS file, you need to delete these. Maybe there is other data in that tree (I don’t have a version 2 internal year METS file at hands at the moment) that you need to clean up. But technically, both ways are possible.

I don’t have data at hands to test this change, but if the result is what is wanted, you can also go that way.

PR #4423, opened 2021-05-20
fix test fpr saveOrUpdate

Kathrin-Huber:
fix test, which is failing after #4415


matthias-ronge (approved):
Test no longer fails.

PR #4424, opened 2021-05-20
fix adding metadata to empty collection

Kathrin-Huber:
When adding metadata through metadatascript an error occured when metadata collection was empty.
There is no need, to check for empty collection, because metadata can be added in any case.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/EditDataScript.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/EditDataScript.java
@@ -56,16 +56,7 @@ public abstract class EditDataScript {
      * @return a list of metadata of a structural element selected from the list.
      */
     public Collection<Metadata> getMetadataCollection(List<IncludedStructuralElement> allIncludedStructuralElements) {
-        Collection<Metadata> metadataCollection = Collections.emptyList();
-        if (!allIncludedStructuralElements.isEmpty()) {
-            for (IncludedStructuralElement allIncludedStructuralElement : allIncludedStructuralElements) {
-                if (!allIncludedStructuralElement.getMetadata().isEmpty()) {
-                    metadataCollection = allIncludedStructuralElement.getMetadata();
-                    break;
-                }
-            }
-        }
-        return metadataCollection;
+        return allIncludedStructuralElements.isEmpty() ? new ArrayList<>() : allIncludedStructuralElements.get(0).getMetadata();
     }
 
     /**
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/EditDataScript.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/EditDataScript.java
@@ -51,15 +51,6 @@ public abstract class EditDataScript {
     }
 
     /**
-     * gets the corresponding metadata as collection.
-     * @param allIncludedStructuralElements the structural Elements to check.
-     * @return a list of metadata of a structural element selected from the list.
-     */
-    public Collection<Metadata> getMetadataCollection(List<IncludedStructuralElement> allIncludedStructuralElements) {
-        return allIncludedStructuralElements.isEmpty() ? new ArrayList<>() : allIncludedStructuralElements.get(0).getMetadata();
-    }
-
-    /**
      * Executes the given script on the given file for the given process.
      * @param metadataFile the file to edit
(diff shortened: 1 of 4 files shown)


Kitodo/src/main/java/org/kitodo/production/services/command/EditDataScript.java
```diff
@@ -56,16 +56,7 @@ public void process(LegacyMetsModsDigitalDocumentHelper metadataFile, Process pr
...
-        if (!allIncludedStructuralElements.isEmpty()) {
-            for (IncludedStructuralElement allIncludedStructuralElement : allIncludedStructuralElements) {
-                if (!allIncludedStructuralElement.getMetadata().isEmpty()) {
-                    metadataCollection = allIncludedStructuralElement.getMetadata();
-                    break;
-                }
-            }
-        }
-        return metadataCollection;
+        return allIncludedStructuralElements.get(0).getMetadata();
```
solth, line 59: Retrieving some element by index from a list without checking whether the list contains elements at all is never a good idea. Don't rely on surrounding methods to guarantee the given list has an element at index '0'!

matthias-ronge (changes requested):
Your change changes the semantics of that method in two ways:

1. The method did collect and return all metadata from all included structural elements of that list, you only return the metadata of the first included structural element.
2. The method returned a new collection which the calling code could modify (add, delete) without side effects. You return the metadata list object of the first included structural element. If the calling code modifies that list, it will en passant edit the metadata of the structural element where it was from, which shouldn’t happen.

Double-checking the code you deleted, I don’t see a place where the code could raise an error if the metadata collection is _empty_. It may throw a `NullPointerException` if `allIncludedStructuralElements` or the result of `allIncludedStructuralElement.getMetadata()` is `null`, which—as per our coding guidelines—should never be, but should be an empty collection instead. This should be fixed in the place where the `null` comes from, not here.

By the way, this code could benefit from using streams:
```java
Collection<Metadata> metadataCollection =  allIncludedStructuralElements.stream()
    .flatMap(includedStructuralElement -> includedStructuralElement.getMetadata().stream())
```
(That will also fail on `null`.)

Kathrin-Huber:
> Your change changes the semantics of that method in two ways:
> 
> 1. The method did collect and return all metadata from all included structural elements of that list, you only return the metadata of the first included structural element.
> 2. The method returned a new collection which the calling code could modify (add, delete) without side effects. You return the metadata list object of the first included structural element. If the calling code modifies that list, it will en passant edit the metadata of the structural element where it was from, which shouldn’t happen.
> 
> Double-checking the code you deleted, I don’t see a place where the code could raise an error if the metadata collection is _empty_. It may throw a `NullPointerException` if `allIncludedStructuralElements` or the result of `allIncludedStructuralElement.getMetadata()` is `null`, which—as per our coding guidelines—should never be, but should be an empty collection instead. This should be fixed in the place where the `null` comes from, not here.
> 
> By the way, this code could benefit from using streams:
> 
> ```java
> Collection<Metadata> metadataCollection =  allIncludedStructuralElements.stream()
>     .flatMap(includedStructuralElement -> includedStructuralElement.getMetadata().stream())
>     .collect(Collectors.toList());
> ```
> 
> (That will also fail on `null`.)

The thing is: I don't want to have a new collection, but the mutual list of metadata elements from the first included structural element. So changing the semantics was intended here. Your comment is a perfekt susumption of the expected behavior :-)
The new collection was only returned, if the metadata of the structuralElement was empty (thats when we recognized the error). When there were metadata in the element it already returned the mututal list of metadata (which is expected here). 
Adding the new metadata to a new collection, didn't added it to the actual metadata, because it didn't change the metadata list, but a loose new collection, which was never saved.
Thats what is fixed by this PR.

matthias-ronge (changes requested):
If so, the whole method no longer makes sense.

The process of first _collecting_ all metadata from _all_ structural elements and then only using the first of them is inefficient. So I suggest you delete this method entirely. And at the points where it has been called like this:

```java
List<IncludedStructuralElement> allIncludedStructuralElements = workpiece
        .getAllIncludedStructuralElements();
Collection<Metadata> metadataCollection = getMetadataCollection(allIncludedStructuralElements);
```
Just do the following:
```java
Collection<Metadata> metadataCollection = workpiece.getRootElement().getMetadata();
```

PR #4432, opened 2021-05-26
Implement standard cloning

### LINKED ISSUE #4429: Implement standard cloning
Creating a copy of an object in memory, so that you have two different objects which you can modify independently, is typically referred to as cloning. This is unluckily named _duplicate_. (Affected: `LdapGroup`, `Project`, `Template`, `Workflow`)

Technically, object cloning should be done in the object by [implementing the clone method](https://www.javatpoint.com/object-cloning). For Hibernate-stored objects, the ID should be set to `null` on the clone in `BaseBean`:
```java
@Override
public BaseBean clone() throws CloneNotSupportedException {
    BaseBean clone = (BaseBean) super.clone();
    clone.id = null;
    return clone;
}
```

**Goal:** implement standard object cloning

matthias-ronge:
Fixes #4429


Kathrin-Huber (approved):

Kathrin-Huber:
Please rebase

PR #4440, opened 2021-05-28
Fix removing structure element containing links

### LINKED ISSUE #4367: Delete option for structure elements requires a check of linked pages to more than one structure element
The user should be advised to delete at first the special link(s) for this kind of page. Otherwise a confusing state in the area of unstructured pages is caused.

solth:
Fixes #4367

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -133,15 +133,15 @@ public class StructurePanel implements Serializable {
     }
 
     void deleteSelectedStructure() {
-        Optional<IncludedStructuralElement> selectedStructure = getSelectedStructure();
-        if (!selectedStructure.isPresent()) {
+        if (!getSelectedStructure().isPresent()) {
             /*
              * No element is selected or the selected element is not a structure
              * but, for example, a media unit.
              */
             return;
         }
-        LinkedList<IncludedStructuralElement> ancestors = MetadataEditor.getAncestorsOfStructure(selectedStructure.get(), structure);
+        IncludedStructuralElement selectedStructure = getSelectedStructure().get();
+        LinkedList<IncludedStructuralElement> ancestors = MetadataEditor.getAncestorsOfStructure(selectedStructure, structure);
         if (ancestors.isEmpty()) {
             // The selected element is the root node of the tree.
             return;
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -133,39 +133,49 @@ public void clear() {
...
-        if (!selectedStructure.isPresent()) {
+        if (!getSelectedStructure().isPresent()) {
             /*
              * No element is selected or the selected element is not a structure
              * but, for example, a media unit.
              */
             return;
         }
-        LinkedList<IncludedStructuralElement> ancestors = MetadataEditor.getAncestorsOfStructure(selectedStructure.get(), structure);
+        IncludedStructuralElement selectedStructure = getSelectedStructure().get();
```
matthias-ronge, line 143: Please don’t rely on `get()`. The method returns a conditional for a reason. Please add an error message:

    getSelectedStructure().orElseThrow(() -> new IllegalStateException("No structure selected"))

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -133,39 +133,49 @@ public void clear() {
...
         }
         IncludedStructuralElement parent = ancestors.getLast();
 
         Collection<View> subViews = new ArrayList<>();
-        subViews = getAllSubViews(selectedStructure.get(), subViews);
+        getAllSubViews(selectedStructure, subViews);
+
+        List<View> multipleViews = subViews.stream().filter(v -> v.getMediaUnit().getIncludedStructuralElements().size() > 1)
+                .collect(Collectors.toList());
+        for (View view : multipleViews) {
```
matthias-ronge, line 156: You are mixing recursion, a stream and iterating here. If you want to use streams, you can do all of the actions in one stream here, I think:
```java
Workpiece.treeStream(selectedStructure)
    .flatMap(logicalDivision -> logicalDivision.getViews().parallelStream()) // subViews
    .filter(view -> 
        {   List<IncludedStructuralElement> logicalDivisions = view.getMediaUnit().getIncludedStructuralElements();
            if (logicalDivisions.size() <= 1) {
                return true; // add to parent views
            }
            dataEditor.unassignView(selectedStructure, view, selectedStructure.getViews().getLast().equals(view));
            if (logicalDivisions.size() <= 1) {
                severalAssignments.remove(view.getMediaUnit());
            }
            return false; // don't add to parent views
        }
    )
    .forEachOrdered(parent.getViews()::add)
;
```

matthias-ronge (approved):
Looks correct, even though it may be simplified, but it’s an improvement whatsoever.

I assume that `dataEditor.unassignView()` does modify `view.getMediaUnit().getIncludedStructuralElements().size()`, otherwise the expression

```java
if (view.getMediaUnit().getIncludedStructuralElements().size() <= 1) {
    severalAssignments.remove(view.getMediaUnit());
}
```
would be dead code because you are iterating over a list which was previously filtered on condition `.size() > 1`.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
solth: There is already a check whether the structure element is present in line 136 and the method returns if that is not the case, so unconditionally calling `get()` should be safe here.

solth:
> I assume that `dataEditor.unassignView()` does modify `view.getMediaUnit().getIncludedStructuralElements().size()`

It does indeed.

PR #4442, opened 2021-05-28
Changed Structure tree label generation

### LINKED ISSUE #4038: Structure tree should show image number (or name), not internal order number

BartChris:
This pull requests continues the work started in pull request https://github.com/kitodo/kitodo-production/pull/4401. The generation of the labels in the structure tree should be based on the image name and not the order. 

fixes #4038

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -571,7 +571,6 @@ public class StructurePanel implements Serializable {
         MediaUnit mediaUnit = view.getMediaUnit();
         Iterator<Entry<MediaVariant, URI>> mediaFileIterator = mediaUnit.getMediaFiles().entrySet().iterator();
         String canonical = "-";
-        String order = view.getMediaUnit().getOrder() + " : ";
         if (mediaFileIterator.hasNext()) {
             Entry<MediaVariant, URI> mediaFileEntry = mediaFileIterator.next();
             Subfolder subfolder = this.subfoldersCache.computeIfAbsent(mediaFileEntry.getKey().getUse(),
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -537,13 +548,31 @@ private void orderChildrenAndViews(List<IncludedStructuralElement> children, Lis
...
+     *
+     * @param view
+     *            View which holds the MediaUnit
+     * @return the display label of the MediaUnit
+     */
     private String buildViewLabel(View view) {
+        MediaUnit mediaUnit = view.getMediaUnit();
+        Iterator<Entry<MediaVariant, URI>> mediaFileIterator = mediaUnit.getMediaFiles().entrySet().iterator();
+        String canonical = "-";
         String order = view.getMediaUnit().getOrder() + " : ";
```
matthias-ronge, line 564: This line (`String order`) has no function and can be deleted.

matthias-ronge:
@Kathrin-Huber looks like _you_ must approve the workflow first. (Looks like I am not allowed to do that.)

BartChris:
@matthias-ronge  In my tests i noticed that it might get confusing that the pagination widget still has its labels generated with the order number on the left. So in my opionion we sholud also change the labels in the pagination widget to show the image number. What do you think? Would you put this in a seperate PR or include it here?

matthias-ronge:
Glad you noticed. Yes, the two windows should contain identical displays. Yes, I would suggest that both of these should be corrected in this pull request branch. Can you, for both, call the same method internally to calculate labels?

matthias-ronge:
Rebased as #4642.

PR #4444, opened 2021-05-28
Enable default context menu for input and textarea elements

### LINKED ISSUE #4212: Missing mouse menu in metadata editor
Screenshot

I selected a value (here, the PPN) and wanted to copy it to the clipboard, by clicking the secondary (right) mouse button and selecting „Copy“, but clicking the secondary mouse button does not open the menu. Nothing happens

**Goal:** Enable mouse menu

oliver-stoehr:
Enable the browser's default context menu for HTML elements of type "input" or "textarea" in metadata editor to allow copy and paste actions via mouse.

Fixes #4212

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_editor.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_editor.js
@@ -81,8 +81,9 @@ var metadataEditor = {
 
 metadataEditor.contextMenu = {
     listen() {
-        document.oncontextmenu = function() {
-            return false;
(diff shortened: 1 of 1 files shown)


Kathrin-Huber:
It's not working in my test evironment. (Windows 10, Chrome Version 91.0.4472.77 )
Do I need to reload the js explicitly? F5 didn't change anything.

oliver-stoehr:
If your browser caches the js files you need to clear the cache explicitly. You can check in the browser's development tool, if the loaded js contains the new code. (In Chrome you should find it in the development tools under "application" > "frames".)

PR #4445, opened 2021-05-31
Restore function to copy metadata from existing processes

### LINKED ISSUE #4369: Adding new journal volumes from the process list and using existing volumes of the chosen (parent) process as a template by the tab "Suche"
We expect in the tab "Suche" a list of linked volumes for the chosen (parent) process to use as a template for my current volume.

solth:
Fixes #4369


matthias-ronge (commented):
If it works, OK for me.

PR #4462, opened 2021-06-08
Allow KitodoScript as action

matthias-ronge:
As the title says: If you specify a KitodoScript as a script in the workflow editor, this is executed using the `KitodoScriptService`. The distinguishing is done by that a KitodoScript always begins with the character string "action:" (which happens to be a semantically valid indicator for a protocol, by the way). In combination with the functionality from pull request #4450, this provides for writing media files to the METS file in the course of the workflow without having to open the metadata editor and click on <kbd>Save</kbd>, which was a long-cherished wish.

The `KitodoScriptService` was adapted so that it is provided via the `ServiceLoader` like all other services, so that we have a uniform procedure here.

_This feature is funded by Technische Universität Darmstadt._

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
@@ -17,6 +17,7 @@ import java.io.InputStream;
 import java.net.URI;
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


Kathrin-Huber:
I don't get it to work. The script is not shown in the task (as link to click) and it is not triggered by automatic tasks.
How can I test it?

matthias-ronge:
I tested it once again and I find it working when clicking on the link.

There is nothing special, just put the KitodoScript in the script box:

Configure script

PR #4469, opened 2021-06-10
Synchronise page selection in structure panel and gallery panel with …

IkramMaalej:
…selection in pagination panel

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
@@ -17,6 +17,7 @@ import java.util.Objects;
 
 import javax.faces.model.SelectItem;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.kitodo.api.dataformat.IncludedStructuralElement;
 import org.kitodo.api.dataformat.MediaUnit;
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
```diff
@@ -152,6 +152,19 @@ public void setNewPagesCountValue(int newPagesCountValue) {
      *            selected items to set
      */
     public void setPaginationSelectionSelectedItems(List<Integer> paginationSelectionSelectedItems) {
+        List<MediaUnit> mediaUnits = dataEditor.getWorkpiece().getAllMediaUnitChildrenFilteredByTypePageAndSorted();
+        MediaUnit lastItem = mediaUnits.get(paginationSelectionSelectedItems.get(paginationSelectionSelectedItems.size() - 1));
+        if (!Objects.equals(this.paginationSelectionSelectedItems.get(this.paginationSelectionSelectedItems.size() - 1 ), lastItem)) {
```
solth, line 157: `lastItem` contains a `MediaUnit`
`this.paginationSelectionSelectedItems.get(this.paginationSelectionSelectedItems.size() - 1)` returns an `Integer`

That means the comparison in this line will _always_ return `true` (because `Integer`s are always unequal to `MediaUnit's) and therefore will _always_ the metadata panel update, even if it's not necessary.

solth (changes requested):
The synchronization of the panels works as intended, but there is one problem with the code that needs to be fixed (see above).

solth (approved):

PR #4490, opened 2021-06-15
change workflowForm to ViewScoped

Kathrin-Huber:
After creating a workflow out of "migration" the migration boolean in the workflowForm is set to `true`
Every other call to workflowForm behaves incorrect after that, because it acts like being in migration.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowForm.java
@@ -31,9 +31,9 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import javax.enterprise.context.SessionScoped;
 import javax.faces.context.FacesContext;
 import javax.faces.model.SelectItem;
+import javax.faces.view.ViewScoped;
 import javax.inject.Named;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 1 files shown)


solth (changes requested):
Switching to `ViewScoped` here does seem to work on the surface, but without further adjustments it removes some functionality from the current state of the application.

For example, since the `WorkflowForm` is currently used for the pages `workflowList` _and_ `workflowEdit`, it needs to retain some information when switching between those pages, like the current page of the workflow list (stored in `BaseForm.firstRow`). 

When `WorkflowForm` is `SessionScoped`, this information is stored as long as the session lives, which means the user returns to the correct page in the workflow list when saving or canceling in the workflowEdit page.

Changing the form to `ViewScoped` means that the information of the current page is lost as soon as the user changes from the list to the edit page, so that he always returns to the _first_ page of the list when saving or canceling in the edit form.

To properly change the scope of the current `WorkflowForm` from `SessionScoped` to `ViewScoped` we should either separate the forms for the list view and the edit view (and leave the former `SessionScoped` for the time being) or pass variables that would otherwise be lost in the transition between `ViewScoped` beans as view parameters from one page to the other (like `pageIndex=...` or something like that).

Both solutions seem like quite a hassle to just solve the issue described in the pull request, so I would rather suggest to solve that issue differently. Perhaps it would be much easier and straightforward to just reset the `migration` variable whenever needed, instead of changing the whole bean management just to achieve this side effect.

PR #4492, opened 2021-06-17
Log absolute location of config file being used

matthias-ronge:
Never again ask yourself (like when Production is automatically deployed by your IDE) which copy of the configuration file is actually being read...

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/config/Config.java
+++ b/Kitodo-API/src/main/java/org/kitodo/config/Config.java
@@ -15,7 +15,9 @@ import java.util.Objects;
 
 import org.apache.commons.configuration.AbstractConfiguration;
 import org.apache.commons.configuration.ConfigurationException;
+import org.apache.commons.configuration.ConfigurationUtils;
 import org.apache.commons.configuration.ConversionException;
+import org.apache.commons.configuration.FileSystem;
 import org.apache.commons.configuration.PropertiesConfiguration;
 import org.apache.commons.configuration.reloading.FileChangedReloadingStrategy;
 import org.apache.logging.log4j.Level;
@@ -41,6 +43,10 @@ public abstract class Config {
                     AbstractConfiguration.setDefaultListDelimiter('&');
                     try {
                         initialized = new PropertiesConfiguration(configFile);
+                        if (logger.isDebugEnabled()) {
+                            logger.debug("Using configuration from "
+                                    + ConfigurationUtils.locate(FileSystem.getDefaultFileSystem(), null, configFile));
+                        }
                     } catch (ConfigurationException e) {
                         logger.warn("Loading of " + configFile + " failed. Trying to start with empty configuration.",
                             e);


Kitodo-API/src/main/java/org/kitodo/config/Config.java
```diff
@@ -41,6 +43,10 @@ static PropertiesConfiguration getConfig(String configFile) {
                     AbstractConfiguration.setDefaultListDelimiter('&');
                     try {
                         initialized = new PropertiesConfiguration(configFile);
+                        if (logger.isDebugEnabled()) {
```
Kathrin-Huber, line 46: Why check for this? just use logger.debug and the user can decide, which level he wants to see on his log

matthias-ronge: Because the line is a short form of the following:

    FileSystem fileSystem = FileSystem.getDefaultFileSystem();
    URL url = ConfigurationUtils.locate(fileSystem, null, configFile);
    StringBuilder stringBuilder = new StringBuilder();
    stringBuilder.append("Using configuration from ");
    stringBuilder.append(url);
    String message = stringBuilder.toString();
    logger.debug(message);

All of these actions are unnecessary, if debug level is disabled, and should therefore be skipped in that case.

Kathrin-Huber: Sorry I don't quite understand your answer.
why are you checking for `isDebugEnabled` ?
I don't think we do this at any other place in the code.
Just add `logger.debug` and everything is fine.

Kitodo-API/src/main/java/org/kitodo/config/Config.java
```diff
@@ -41,6 +43,10 @@ static PropertiesConfiguration getConfig(String configFile) {
...
+                            logger.debug("Using configuration from "
+                                    + ConfigurationUtils.locate(FileSystem.getDefaultFileSystem(), null, configFile));
+                        }
```
Kathrin-Huber, line 49:
```suggestion
                        logger.debug("Using configuration from "
                                + ConfigurationUtils.locate(FileSystem.getDefaultFileSystem(), null, configFile));
                        
```

Kitodo-API/src/main/java/org/kitodo/config/Config.java
matthias-ronge: The thing is, that Java invokes the commands in a specific order:

                            FileSystem.getDefaultFileSystem() 
                                                 ↓
                      ConfigurationUtils.locate(..., null, configFile)
                                   ↓
    "Using configuration from " + ...
                  ↓
    logger.debug(...);

This means, if debug level is **disabled**, Java _will still_ load the file system, resolve the absolute path of the file on disk, and build a log string, before passing it to `logger` which decides that it doesn’t need to log it. Therefore, if retrieving the information to log is “work” (not just passing existing variables) it should be checked **before**, if it shall be logged at all in the end, or not.

There **are** some other places where these checks can be found in the code, especially for the “trace” level.

- https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/helper/metadata/legacytypeimplementations/LegacyLogicalDocStructHelper.java#L252-L257
- https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java#L295-L298
- …

PR #4501, opened 2021-06-21
fix lockException

Kathrin-Huber:
After switching to `saveOrUpdate` ( #4415 )
The explicit fetching from database isn't needed anymore, but leads to "OptimisticLockExceptions"

This kind of behavior is probably on other places too. Think of it, when you get an `OptimisticLockException`


solth:
I can't trigger the mentioned `OptimisticLockException` when saving a template - is there a special way to reproduce this problem?

Kathrin-Huber:
> I can't trigger the mentioned `OptimisticLockException` when saving a template - is there a special way to reproduce this problem?

Unfortunately it's kind of a raceCondition problem.
But it was failing for me 2 out of 3 times.
The template needs to have a workflow and needs to be saved already at least 1 time (so changed code is triggered) -> after this it's failing every other time (because code is only triggered every other time, because error leads to deletion of tasks)

solth (approved):

PR #4509, opened 2021-06-23
Enable title metadata keys nested in metadata groups

### LINKED ISSUE #4500: titleMetadata does not work for metadata groups
Kitodo.Production 3 allows configuring specific metadata to be displayed in the title bars of the gallery stripes in the metadata editor. To enable this feature, the names of the metadata entries must be added to the key `metsEditor.titleMetadata` in the file `kitodo_config.properties`. 

This does not work for metadata entries nested within metadata groups, though. So if the corresponding ruleset contains title information structured similar to mods (`titleInfo/title`), this information can currently not be displayed as title metadata in the editor.

solth:
Fixes #4500

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
@@ -36,8 +36,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.kitodo.api.Metadata;
-import org.kitodo.api.MetadataEntry;
 import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
 import org.kitodo.api.dataformat.IncludedStructuralElement;
 import org.kitodo.api.dataformat.MediaUnit;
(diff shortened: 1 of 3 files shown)


matthias-ronge (commented):
You don’t need to implement the case "no @ character" separately. If you invoke `split(…)` on a string without an @ character, it will simply return a list with just one element which is your string. Then, use a loop from 0 to **before** last index to descend through the metadata groups to the entry level. If there is no @ character, last index is 0, a loop with i=0;  i<0; i++ will be skipped (because 0 < 0 is false).

Something like this:

```java
for (String titleKey : getTitleKeys()) {

    // if no '@' character is contained, this will return a list with size() == 1
    // which contains the whole string
    List<String> keySegments = Arrays.asList(titleKey.split("@"));
    int lastIndex = keySegments.size() - 1;

    // you can initialize 'metadata' from 'element'---no 'if' needed
    Collection<Metadata> metadata = element.getMetadata();

    // run loop only for indices SMALLER THAN the last index
    // this will work through all levels of metadata groups
    // if last index is 0, the loop will automatically be skipped entirely
    for (int currentIndex = 0; currentIndex < lastIndex; currentIndex++) {
        final String currentKey = keySegments.get(currentIndex);
        metadata = metadata.stream()
            // filter on key
            .filter(currentMetadata -> Objects.equals(currentMetadata.getKey(), currentKey))

            // continue only with metadata groups
            .filter(MetadataGroup.class::isInstance).map(MetadataGroup.class::cast)

            // join all metadata from the groups to one list
            // (this will also work if the first group has no such sub-key
            // but the next one has one)
            .flatMap(metadataGroup -> metadataGroup.getGroup().stream())

    }

    Optional<String> titleValue = metadata.stream()
        // filter on key
        .filter(currentMetadata -> Objects.equals(currentMetadata.getKey(), keySegments.get(lastIndex)))
        
        // only metadata entries
        .filter(MetadataEntry.class::isInstance).map(MetadataEntry.class::cast)
        
        // only values
        .map(MetadataEntry::getValue)
        
        // only non-empty
        .filter(value -> !value.isEmpty())

    // there-of the first 
    .findFirst();
    
    // return result if found
    if (titleValue.isPresent()) return " - ".concat(titleValue.get());
    
// else continue loop  
}

// nothing found at all
return "";
```

matthias-ronge (approved):

PR #4512, opened 2021-06-24
Hide process link option with selected pages

solth:
The option to link processes in the metadata editor has to be deactivated when pages are selected. Otherwise this option would suggest that the selected pages are assigned to the linked process, which is not possible in Kitodo (yet).

Bildschirmfoto 2021-06-24 um 13 06 06


matthias-ronge:
It shouldn’t be deactivated in my opinion, it should be hidden, because it doesn’t make any sense and will never.

PR #4519, opened 2021-06-24
fix project saving

Kathrin-Huber:
Another changed needed due to the switch to saveOrUpdate
(see #4501)


matthias-ronge (approved):
I do not understand why it is needed, or why it was not needed here before, or was it needed before? In any case, I have no objections in terms of the Coding Guidelines.

PR #4521, opened 2021-06-24
use related attribute on related objects

Kathrin-Huber:
speed up desktopPage by not loading related items of related items

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
@@ -17,6 +17,7 @@ import java.util.Objects;
 
 import javax.faces.model.SelectItem;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
```diff
@@ -17,6 +17,7 @@
 
 import javax.faces.model.SelectItem;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
```
solth, line 20: This should be a standalone hotfix pull request! ;)

solth (approved):
This really brings a tangible performance improvement 👍 
Can something similar be applied to the process list?

Kathrin-Huber:
> This really brings a tangible performance improvement 👍
> Can something similar be applied to the process list?

This is in fact affecting the Processlist, because Projects were loaded for Processes in list and AGAIN all Processes were loaded for that project.
Not yet the final fix, but a turning point in performance... still on it 🦸‍♀️

PR #4524, opened 2021-06-25
Show error messages only once instead of twice

solth:
The `p:message` components in Kitodo.Production 3 are configured to show message details, but currently we never provide separate message details. This results in all messages being displayed twice, once as message summary and once as message details, which doesn't look very nice:

**_Process locked:_**
Bildschirmfoto 2021-06-25 um 17 00 48

**_Password constraints not met:_**
Bildschirmfoto 2021-06-25 um 17 13 37

**_Missing form input:_**
Bildschirmfoto 2021-06-25 um 17 03 17

Therefor this PR removes the details from the error message to improve the rendering of our error messages:

**_Process locked:_**
Bildschirmfoto 2021-06-25 um 17 19 10

**_Password constraints not met:_**
Bildschirmfoto 2021-06-25 um 17 18 20

**_Missing form input:_**
Bildschirmfoto 2021-06-25 um 17 17 50

Should we ever have a case where separate message details are required, we can easily add a second `p:messages` tag with `showDetail=true` that can be targeted in such a case.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/PaginationPanel.java
@@ -17,6 +17,7 @@ import java.util.Objects;
 
 import javax.faces.model.SelectItem;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.kitodo.api.dataformat.IncludedStructuralElement;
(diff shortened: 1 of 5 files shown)


matthias-ronge:
The reason for this is that otherwise the messages do not show in the log file. Some time ago we had the situation that the logfile produced output like:

    There are unread faces messages. This means that a message had no chance to be rendered.
    -
    -
    -
    -
    -
    -
    -
    -

But you didn’t see a message. Because of this I changed this in #4072

PR #4525, opened 2021-06-28
Add option to import child processes directly from metadata editor

solth:
This adds the option to import suborditnated processes directly from the metadata editor if the selected structural element is configured accordingly in the ruleset in use. The import masks "titleRecordLink" is prepopulated with the correct process from the metadata editor and after "Saving" or "Canceling" in the import view, the user is returned to the metadata editor.

Bildschirmfoto 2021-06-28 um 14 25 31


matthias-ronge:
If someone clicks on this, the metadata editor is gone. Are unsaved changes then automatically saved, or are they forgotten, or does the user be asked to answer?

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/logicalStructure.xhtml
```diff
@@ -17,6 +17,7 @@
         xmlns:h="http://xmlns.jcp.org/jsf/html"
         xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
         xmlns:p="http://primefaces.org/ui">
+    <!--@elvariable id="readOnly" type="java.lang.Boolean"-->
```
matthias-ronge, line 20: What is this comment for, or is it accidentally?

solth:
> If someone clicks on this, the metadata editor is gone. Are unsaved changes then automatically saved, or are they forgotten, or does the user be asked to answer?

The information is retained. So after returning to the metadata editor from the import mask no unsaved changes are lost.

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/logicalStructure.xhtml
solth: This annotation helps IDEs to identify the type of Java variables in EL expressions so it can better determine available properties and methods of those variables. 

We already use this type of annotation in many other Facelets.

matthias-ronge: Yes, I've seen it several times and always wondered what kind of note that should be. ... There is a lot that is good in Eclipse, but for JSP & EL, it only provides poor support.

PR #4531, opened 2021-07-01
Fix some design issues

solth:
- use same icon for `duplicate` action in process edit property list as in other lists (template, project, workflow)
- change background color of whole process row expansion on hover (current task part remained white while everything else was light blue)
- unify order of buttons above user page, process page and task page to: filter (where applicable) -> extended search (where applicable) -> column configration button -> "new" button (where applicable)
- hide "filter" menu when opening "Batches" tab on process page (similar to filter menu being hidden on all tabs but the first on the user page)


Kathrin-Huber:
Please rebase
Can you maybe also switch title of processes page between "Processes" and "Batch"

Kathrin-Huber (approved):

solth:
> Can you maybe also switch title of processes page between "Processes" and "Batch"

You mean here?
Bildschirmfoto 2021-07-08 um 17 47 33

We don't do that on any other tab view either, so it would be a little strange to change it just here, wouldn't it?
For example:
Bildschirmfoto 2021-07-08 um 17 47 44
Bildschirmfoto 2021-07-08 um 17 47 53
Bildschirmfoto 2021-07-08 um 18 01 56

oliver-stoehr (approved):

PR #4543, opened 2021-07-14
Don't save related

Kathrin-Huber:
Related objects in Index are now only updated if explicitly called.
This hugely reduces the indexing overhead on saving.

INFO: Releated objects are always stored in Database, because this is handled by hibernate. Relations to objects which are only represented in DB (e.g. properties) are not harmed by this.


solth (approved):
As far as I can test it, it _does_ speed up some actions and does not disrupt existing functionality.

Issue #4465, opened 2021-06-10
Meta Data Editor - when adding a whole keygroup, mandatory subkeys of similar groups get lost!


stefanCCS:
Release 3.3

In my Ruleset I have (extract):

```
<key id="Person">
	<label>Person (group)</label>
	<label lang="de">Person (Gruppe)</label>
	<key id="LastName">
		<label>Last name</label>
		<label lang="de">Familienname</label>
	</key>
	<key id="RoleCode">
		<label>Relationship designation (code)</label>
		<label lang="de">Beziehungskennzeichnung (Code)</label>
		<option value="rcp">
			<label>Adressat</label>
		</option>
		<option value="aut">
			<label>Verfasser</label>
		</option>
		<preset>aut</preset>
	</key>
	<key id="Authority">
		<label>Authority</label>
		<label lang="de">Quelle (Normdaten)</label>
	</key>
</key>

<restriction division="Monograph" unspecified="forbidden">
	<permit key="Person">
		<permit key="RoleCode" maxOccurs="1" minOccurs="1"/>
	</permit>

```

Same as "ContributorPerson"

If I have a "Person" in meta.xml (lets assume from catalog), and I add manually "ContributorPerson", the entry "RoleCode" in Person gets lost!
The validation shows this strange behaviour:
![grafik](https://user-images.githubusercontent.com/61045115/121490396-3e937400-c9d5-11eb-8110-89c82716a271.png)
(this example is with other keygroups with same structure)

I have made some test:
If I remove this
`
 permit key="RoleCode" maxOccurs="1" minOccurs="1" 
`
in Rule Set it looks to be ok.
==> Sorry, I was wrong, still vanishing :-(

But anyway, please check.

andre-hohmann:
Probably related to: 
- #4588 
- [#4061 (comment)](https://github.com/kitodo/kitodo-production/issues/4061#issuecomment-717362734)

Kathrin-Huber:
@stefanCCS a lot happend here, can you check if this is still an issue?

stefanCCS:
I have checked this with Rel 3.5.0-beta - looks ok now.
I will close this issue.

Issue #3821, opened 2020-07-09
Meta data editor: At first time need to press the Save-Button twice


stefanCCS:
Version 3.2-build 16.09.2020
After starting the meta data editor, and making some editing:
If I press the "save" button, it is NOT saved (nothing happens). I need to press the button a second time.

solth:
I think this is due to the first click just triggering a Javascript `blur` event on the currently selected `input` element.

matthias-ronge:
Related: #3472

solth:
@oliver-stoehr could this be the same issue you described in #3472?

stefanCCS:
Check this with Rel 3.5.0-beta. Looks ok now.
I will close this issue.

PR #4550, opened 2021-07-20
Fix adding metadata to metadata groups

### LINKED ISSUE #4285: Adding metadata to nested metadata causes NullPointerException
If you click on the plus:

Screenshot

```
    at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.preserve(ProcessFieldedMetadata.java:519)
    at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.getMetadata(ProcessFieldedMetadata.java:449)
    at org.kitodo.production.forms.dataeditor.AddDocStrucTypeDialog.prepareSelectAddableMetadataTypesItems(AddDocStrucTypeDialog.java:510)
```

After that, no more processes can be opened in the metadata editor (not even any other), there is only one white page. You have to log out and log back in to continue.

**Goal:** adding metadata is possible.

### LINKED ISSUE #4457: When adding sub-field in Metadata-Editor, existing sub-fields are doubled
My ruleset has some complex keys/fields, that have sub-fields. E.g. personGroup or corporationGroup.

In Metadata-Editor, I can add such a key, e.g. corporationGroup - that is ok, group is added.

Now I want to add sub-fields: 
- First field (e.g. Displayform) is ok. 
- Next field (e.g. IdentifierPPN) can be added, but another field "DisplayForm" is also added

So, an already existing sub-field in metadata-editor gets doubled, when adding a new sub-field.

Update: this does NOT happen, when I first add the sub-fields one-by-one without entering values. 

The doubling only appears, if I am adding a sub-field, input a value, save - and then add the next sub-field.

Screenshot:
md-editor-corporation-field-doubling

### LINKED ISSUE #4464: Meta Data Editor - add keygroup in keygroup (which should not be possible)
Release 3.3 

If you have a structured keygroup in RuleSet (e.g. "person").
And you try to add a sub-key in this keygroup, you also get a choice in this "add-dialog" to add full keygroup (here "person"),
which leads to strange behaviour including destroying other parts of meta.xml !!!
grafik

-> Please check

solth:
Fixes #4464, fixes #4285

Edit: seems #4465 and #4457 aren't fixed after all


IkramMaalej (approved):

Kathrin-Huber:
doubleMetadata

Kathrin-Huber:
lostKeys

matthias-ronge (commented):
The right metadata appears to be available for selection at first. Adding also works. But I always have to click two times before the pop-up to open (the fundamental problem that we sometimes have in the metadata editor).

Another problem is, but I don't know whether this is related or a completely different problem, I just noticed: Whenever I click the delete metadata button, I first have to click away a Firefox warning whether I want to leave the page (no) and then again, then it usually works (but not always, and one time I got a warning for a validation error). But it can also be that this is a completely independent problem or it is due to my ruleset, maybe, I can't say that. I haven't been using the button for a long while.

solth:
> The right metadata appears to be available for selection at first. Adding also works. But I always have to click two times before the pop-up to open (the fundamental problem that we sometimes have in the metadata editor).

Yes, this has to do with the update being triggered by the javascript `blur` events (documented in #3821)

> Another problem is, but I don't know whether this is related or a completely different problem, I just noticed: Whenever I click the delete metadata button, I first have to click away a Firefox warning whether I want to leave the page (no) and then again, then it usually works (but not always, and one time I got a warning for a validation error). But it can also be that this is a completely independent problem or it is due to my ruleset, maybe, I can't say that. I haven't been using the button for a long while.

That is a different problem and occurs in the master branch, too.

solth:
> Issue #4457

I think #4457 is about `MetadataGroup` children being unintentionally duplicated. Your screencast shows `MetadataGroups` directly assigned to a structural element being duplicated on purpose. Looks like a similar problem, but not exactly what was described in the issue, IMO.

PR #4557, opened 2021-08-02
only use first year

Kathrin-Huber:
The ephemera year title should only contain the first year


henning-gerhardt (commented):
So far as I can tell from my local development system the creation of ephemera titles of a year contain now only the first year.

PR #4565, opened 2021-08-05
select level for metadataScript

Kathrin-Huber:
This enables the MetadataScript to select a specific type of structural eleent to add the metadata to.
Example:

`action:addData singleDigCollection="New Digital collection"+type=PeriodicalIssue`

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/AddDataScript.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/AddDataScript.java
@@ -12,13 +12,10 @@
 package org.kitodo.production.services.command;
 
 import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
 
 import org.kitodo.api.MdSec;
 import org.kitodo.api.Metadata;
 import org.kitodo.api.MetadataEntry;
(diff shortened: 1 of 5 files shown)


matthias-ronge:
Why are you implementing this functionality again instead of implementing the defined copyData syntax? So the question is: Is the copyData syntax no longer wanted? I'm just afraid that if there are several syntagms for the same functionality, no one will understand that afterwards.

    singleDigCollection="New Digital collection"+type=PeriodicalIssue

What do you mean with that? Is that the same as

    /PeriodicalIssue@singleDigCollection="New Digital collection";

A structured element `PeriodicalIssue`, and to which a metadata should be added? What if there are several of them? Then only the first, or all of them?

matthias-ronge:
Could it be that you pushed your commit to branch `Kathrin-Huber:add_level_to_metadata_script` to repository `kitodo` instead of `Kathrin-Huber`?

Kathrin-Huber:
> Could it be that you pushed your commit to branch `Kathrin-Huber:add_level_to_metadata_script` to repository `kitodo` instead of `Kathrin-Huber`?

I have no ides, why you get this warning, I pushed it to my repository as usual
pr_Screenshot

PR #4566, opened 2021-08-06
fix template migration

Kathrin-Huber:
Fixes error `Multiple representations of the same entity are being merged` on migration (create Template)


henning-gerhardt (approved):
So far as I can test this on my local development system this change suppress the former appearing error message at the end of assigning workflows / templates to projects and processes.

PR #4567, opened 2021-08-06
add snapshot to version

Kathrin-Huber:
add snapshot for further development


henning-gerhardt (approved):
Modules and war file are created with new version and snapshot addition.

PR #4574, opened 2021-08-11
workaround for huge result counts

Kathrin-Huber:
This is a workaround for excel and pdf sheets. Until #4099 is solved.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
@@ -23,9 +23,11 @@ import org.apache.poi.hssf.usermodel.HSSFRow;
 import org.apache.poi.hssf.usermodel.HSSFSheet;
 import org.apache.poi.hssf.usermodel.HSSFWorkbook;
 import org.elasticsearch.index.query.BoolQueryBuilder;
+import org.elasticsearch.index.query.RangeQueryBuilder;
 import org.elasticsearch.search.sort.SortOrder;
 import org.kitodo.api.dataformat.MediaUnit;
 import org.kitodo.api.dataformat.Workpiece;
+import org.kitodo.data.elasticsearch.index.type.enums.ProcessTypeField;
 import org.kitodo.data.exceptions.DataException;
 import org.kitodo.production.dto.ProcessDTO;
 import org.kitodo.production.enums.ObjectType;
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
```diff
@@ -67,10 +69,25 @@ public HSSFWorkbook getResult() {
 
     private List<ProcessDTO> getResultsWithFilter() {
         List<ProcessDTO> processDTOS = new ArrayList<>();
+        int elasticsearchLimit = 9999;
```
matthias-ronge, line 72: Just as a question: I mean we set the ElasticSearch limit in config.properties. Shouldn't you then have to use the value from the config.properties here? So, if someone has set a different value, does it still work as expected?

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
```diff
@@ -67,10 +69,25 @@ public HSSFWorkbook getResult() {
...
-                    ServiceManager.getProcessService().sortByTitle(SortOrder.ASC), false);
+            Long numberOfProcesses = ServiceManager.getProcessService().count(getQueryForFilter(ObjectType.PROCESS));
+            if (numberOfProcesses > elasticsearchLimit) {
+                int queriedIds = 0;
+                while (processDTOS.size() < numberOfProcesses) {
+                    RangeQueryBuilder rangeQueryBuilder = new RangeQueryBuilder(ProcessTypeField.ID.toString());
+                    rangeQueryBuilder.gte(queriedIds).lt(queriedIds+elasticsearchLimit);
+                    BoolQueryBuilder queryForFilter = getQueryForFilter(ObjectType.PROCESS);
+                    queryForFilter.should(rangeQueryBuilder);
+                    processDTOS.addAll(ServiceManager.getProcessService().findByQuery(queryForFilter,
```
matthias-ronge, line 83: Basic consideration: If there are more than 10,000 objects, I would not first put all objects in a `Collection` and process them later, but process the objects at this point with for-each, so that Java can then free them again from memory. But, as long as you don't run out of memory, that's fine with me.

matthias-ronge (approved):
I don't have enough processes to test whether all processes are actually being considered. If it works for you, I agree.

In the long run, we should find a better way and navigate the hit list instead of starting queries over blocks of IDs over and over again.

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
Kathrin-Huber: It's a different elasticsearch.limit. It's the 10'000 limit which comes with elasticsearch. See also #4277 
So it's not configurable.

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
Kathrin-Huber: what do you mean with "process them" ?
Putting them in the Workbook?

matthias-ronge: yes

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
matthias-ronge: OK

PR #4580, opened 2021-08-13
sort task before activating

### LINKED ISSUE #4575: Ordering of tasks is not used on closing tasks
If you have a workflow like this

Screenshot_2021-08-11_15-14-50

and did closing the task "Scannen" through the list of processes and the action "Bearbeitungsstatus hochsetzen" the next "open" task is "LZA Validierung nach Scannen" but a random task like f.e. "Export in das DMS". The reason behind is, that the next task is not selected by the `ordering` column but on the `id` column of the task:

Screenshot_2021-08-11_15-12-53

Kathrin-Huber:
fixes #4575

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
@@ -18,6 +18,7 @@ import java.net.URI;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


solth (approved):

solth:
@henning-gerhardt could you test if the issue you described is indeed resolved by the proposed fix? I wasn't able to reproduce the situation you described - the ordering of all my tasks always corresponded with their ids.

henning-gerhardt:
I did only a simple test with a few processes and theirs corresponding tasks and it looks correct now. 

Sorting the order of tasks inside the application is one way. A maybe better but maybe even more complicated way - in comparison to the current implementation - is the ordering of tasks is done inside the database.

PR #4583, opened 2021-08-13
refresh filter when filter is typed in

### LINKED ISSUE #4579: Selected processes remain selected
### Problem
Selected processes remain in some cases selected, although they have been deselected. 
This leads in some cases to the application of exports, Scripts, ... to unwanted processes. 
It is difficult to search for selected processes. The only way to make sure to deselect all processes is to logout and login again or to run a script - as after the script finished, the processes on which it was applied are listed in the message. 

### Solution
Deselected processes should remain deselected. 

### Example
The following Gif shows that the unselected process is again selected, after the complete process list is shown and the process is shown after filtering for it by its ID. 
 
markierteVorgänge01

Kathrin-Huber:
fixes part of #4579 
Nice side-effekt: The processes-amount value is also updated

### DIFF
--- a/Kitodo/src/main/webapp/pages/processes.xhtml
+++ b/Kitodo/src/main/webapp/pages/processes.xhtml
@@ -56,6 +56,7 @@
                                  value="#{ProcessForm.filter}"
                                  placeholder="#{msgs.filter}">
                         <p:ajax event="change"
+                                listener="#{ProcessForm.changeFilter(ProcessForm.filter)}"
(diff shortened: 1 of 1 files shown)


solth:
This does not fix #4579 , but only the error mentioned in the comment by @matthias-ronge concerning the filters.

solth (changes requested):
This does not fix #4579 , but only the error mentioned in the comment by @matthias-ronge concerning the filters.

PR #4584, opened 2021-08-13
use id instead of label

### LINKED ISSUE #4553: Capture of Metadata for newspaper issues incorrect - 2
### Problem
When processes for newspaper issues are created, the metadata like for example language or rights information cannot be captured correctly.
It is possible to select the metadata fields and to capture values, but in the attribute "name" of the metadata, the German label of the metadata is stored, instead of its internal name in the ruleset, for example:
`<kitodo:metadata name="Sprache">ger</kitodo:metadata>`

However, it should be: 
`<kitodo:metadata name="DocLanguage">ger</kitodo:metadata>`

### Solution
It should be possible to capture the metadata with the correct internal name in the attribute "name" of the metadata.

Kathrin-Huber:
using label in #4502  was wrong, it should be id.
fixes #4553


matthias-ronge (approved):
ID is used.

Issue #4564, opened 2021-08-04
Wrong process title generation on newspaper migration


henning-gerhardt:
If you have a newspaper with several issues per day and every issue is created in 2.x as a separate process. Process generation in 2.x was 

`<shorted title aka ATS/TLS>_<catalog id digital aka PPN>-<YEAR><MONTH><DAY><IS ... first two letters of issue>_<ISSUE ... complete issue title> `

and an example for such a process title

`StaaHofM_1031939121-1844032901_01-p`

If a such process like above is the first process on migration then the created process title for the overall newspaper process and all newspaper year processes starts with

`StaaHofM_103`

instead of the expected 

`StaaHofM_1031939121`

process title. 

Finding this processes is difficult as you can only use the shorted title and not the catalog id digital which is more common.

If you are lucky and your first process of a newspaper looks like

`RadlunRa_411907697-18990115`

then you have no issues even if there are later processes for multiple issues for a day.

So far as I can say the complex regular expression in [NewspaperProcessesMigrator.java#L92](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/migration/NewspaperProcessesMigrator.java#L92) is not cable of this kind of created process titles in 2.x. as 2.x has no restriction on how you are creating your process title for multiple issues per day of a newspaper.

matthias-ronge:
In version 2, the `projects.xml` specifies what the newspaper process title looks like for each issue. That can vary, it would even have been possible to use different titles in several projects, I think. Since you can rename the title after the migration, and that only has to be done for the newspaper and the years (not the issues, so that’s not so many processes) the easiest way would be to fix it manually after the migration, if the result is unfortunate, and it only needs to be fixed onece. Programming a generic system here that is suitable for all cases is at least ornate.

henning-gerhardt:
Manually renaming means over 100 processes to touch and this are only the processes for this newspaper title. Maybe there are more newspaper titles which get this issue.

I experimented a little bit with this regular expression and if I changed it from

`(?<=.)\p{Punct}*(?:1[6-9]|20)\d{2}\p{Punct}?(?:0[1-9]|1[012]).*$`

to

`(?<=.)\p{Punct}*(?:1[6-9]|20)\d{2}\p{Punct}?(?:0[1-9]|1[01]).*$`

it created with this newspaper title the correct or expected newspaper title. But I don't know what I'm really doing on changing this expression.

[...]

PR #4585, opened 2021-08-13
Change regular expression to needed separator before date.

henning-gerhardt:
This pull request contains the provided solution in discussion for https://github.com/kitodo/kitodo-production/issues/4564 

I have no clue how this regular expression is to read and to understand nor did I only a small test with our data. So I don't know if the provided solution is correct in any case.  ~~Providing (integration) tests is not possible as the used test environment nor the code itself could create easy and understandable test code.~~ If this pull request is not merged then it is even okay.


Kathrin-Huber:
What a perfect exemplary Pull Request! :-)

PR #4600, opened 2021-08-20
Show media file number in tree, not internal order

### LINKED ISSUE #4038: Structure tree should show image number (or name), not internal order number

matthias-ronge:
fixes #4038


matthias-ronge:
See #4642

PR #4601, opened 2021-08-20
Add confirm dialog to remove projects from user before removing client

### LINKED ISSUE #4545: Client not removed correctly from user
After removing one of two clients from my user I was still greeted by the "Select client" dialog on login (the problem persists through re-login!) and in the user menu (see screenshot).
Bildschirmfoto 2021-07-14 um 11 30 14
The mapping between the client and the user _was_ removed from the `client_x_user` table, though, so it doesn't seem to be a database problem.

solth:
This pull request adds a confirm dialog that informs the user about projects assignements associated with the client he is about to remove from a user object. The user can then decide to proceed (which means the projects in question will also be removed from the user) or cancel the action.

Bildschirmfoto 2021-08-20 um 16 12 52

Fixes #4545


Kathrin-Huber:
An ST-Test seems to be broken

markusweigelt:
Deleting the client does not work properly yet.
After I clicked on next to remove the client, the following JavaScript error occurs. 

image

After refreshing the browser, the deleted client reappears in the list.

Tested in firefox and chrome without browser caching.

Kitodo/src/main/webapp/WEB-INF/templates/includes/userEdit/removeClientFromUserDialog.xhtml
```diff
@@ -0,0 +1,70 @@
...
+              appendTo="@(body)"
+              resizable="false"
+              closable="false">
+        <h3>#{msgs.removeClientProjectsFromUser}</h3>
+        <h:form id="removeFromProjectsForm">
+            <div class="select-note ui-messages-info">#{msgs.removeClientProjectsFromUserConfirmMessage}</div>
+            <div class="select-selector">
+                <p>
+                    <p:dataTable id="clientProjects"
+                                 style="max-height: 100px; overflow-y: scroll; border: 1px solid var(--blue); border-radius: 3px;"
```
markusweigelt, line 35: Suggestion: For reusing and browser caching create css class, move style to kitodo.css and use styleClass instead or use global style if exists.

markusweigelt:
I tested the branch again because I didn't see the save button. (Sorry for that!) This also deletes the clients. I couldn't recreate the JavaScript error.

Please rebase because master includes some major changes, e.g. elastic search.

markusweigelt (approved):

Kathrin-Huber:
I think a test is still failing here

Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
```diff
@@ -339,28 +341,66 @@ public String addToRole() {
...
+
+    /**
+     * Return list of project titles that are associated with the client with ID 'removeClientId' and
+     * assigned to User 'userObject'.
+     * @return list of process titles
+     */
+    public List<String> getClientProjects() {
+        if (Objects.nonNull(this.removeClientId)) {
+            return this.userObject.getProjects().stream()
+                    .filter(p -> p.getClient().getId().equals(this.removeClientId))
```
matthias-ronge, line 382: .filter(project -> project.getClient().getId().equals(this.removeClientId))

PR #4605, opened 2021-08-23
Remove switch to create unreachable production templates

### LINKED ISSUE #3311: Remove switch to create unreachable production templates

matthias-ronge:
Fixes #3311


Kathrin-Huber:
What about the `active` attribute from the template? Can this be removed too?

matthias-ronge:
It may be intended to put production templates out of use, if they shouldn’t be used any longer?

Kathrin-Huber:
But how can I do that, if the switch is gone? ;-)

Kathrin-Huber:
Please resolve conflicts

matthias-ronge:
It looks like the switch has already gone.

solth:
> It looks like the switch has already gone.

It's still there:
Bildschirmfoto 2022-05-10 um 08 42 01

PR #4607, opened 2021-08-24
Create LDAP user on login

### LINKED ISSUE #4576: Change the creation of a user in LDAP to upon login
Part of #456.

The reason we have reversibly encrypted passwords in the database is that we want to be able to later create an LDAP entry for this user, which requires the password in plain text.

**Goal:** This change should pick up the password when logging in. Under the condition that

- the user has a password in the database,
- the user has an LDAP group,
- the LDAP group has an LDAP server,
- the LDAP server is writeable,
- the entered password is valid against the database password (i.e., not mistyped)

we use the _entered password_ to create the user in the LDAP. If this was successful, a password (obfuscated or hash) is set to `null` in the database.

**Rationale:** This work is a prerequisite for replacing the reversibly encrypted passwords in the database with irreversible hashes, as the hash cannot be used to create a user in LDAP.

matthias-ronge:
Fixes #4576


Kitodo/src/main/java/org/kitodo/production/security/password/SecurityPasswordEncoder.java
```diff
@@ -93,7 +102,22 @@ public String encrypt(String messageToEncrypt) {
         try {
             byte[] utfEight = messageToEncrypt.getBytes(StandardCharsets.UTF_8);
             byte[] enc = encryptionCipher.doFinal(utfEight);
-            return new String(Base64.encodeBase64(enc), StandardCharsets.UTF_8);
+            String encrypted = new String(Base64.encodeBase64(enc), StandardCharsets.UTF_8);
+            if (Objects.nonNull(user) && ConfigCore.getBooleanParameterOrDefaultValue(ParameterCore.LDAP_USE)
+                    && Objects.nonNull(user.getLdapGroup()) && Objects.nonNull(user.getLdapGroup().getLdapServer())
+                    && !user.getLdapGroup().getLdapServer().isReadOnly() && encrypted.equals(user.getPassword())) {
+                try {
+                    ServiceManager.getLdapServerService().createNewUser(user, messageToEncrypt);
```
henning-gerhardt, line 110: For which use case(s) do you want to create a new ldap user entry?

matthias-ronge: If the user password is no longer stored encrypted in the database, but a password hash is stored, then writing in LDAP is no longer possible (with the LDAP write button), because the password is required as clear text for the LDAP entry to write, and from a hash you cannot recalculate a plaintext password. Therefore, the LDAP entry is generated when the user logs in for the first time (successfully) because at that moment the password is known as clear text.

henning-gerhardt: But the LDAP write button did still exists on this pull request. Maybe you should move this change to the "remove LDAP button" pull request?

Personally I would implement your changes (all your 3 pull request) in an other way (goal is fine, implementation of hash algorithm too):

1) let the login do only one thing: verify login password against database or ldap depending on which the user is needed for
2) on user creation write password to database or ldap - depending on where the password should be stored. creating ldap entry in a second step like now is not needed as you have already a working ldap connection
3) add a migration function to migration tab to migrate all database logins to ldap if this migration is still needed

matthias-ronge: Your suggestion would work for creating new users, when LDAP is already set up. But it will not work, if users have already been created, and you decide to use LDAP later.

henning-gerhardt: For this is step 3) you can re-run the migration to ldap any time for any (later) created user.

matthias-ronge: No, you can't, because the user will be created with a password hash only. Or, you need, for the migration, all users to appear in person and type in their passwords…

henning-gerhardt:
This is part one of four pull requests (#4607 (this one), #4609, #4610 and #4616 (which is closed)).

Kathrin-Huber:
Please reopen, if it's not a Draft anymore

PR #4609, opened 2021-08-25
Remove "Write user to LDAP" button

matthias-ronge:
Part of #456 

Follow-up pull request to #4607 ([immediate diff](/matthias-ronge/kitodo-production/compare/issue-4576...remove-write-user-to-ldap-button))

PR #4610, opened 2021-08-25
Use secure hashing for new passwords

matthias-ronge:
Part of #456

For newly created users, only a hash of their password is saved. For the user, there is a new field `algorithm`, in which the additional parameters are saved. These are the algorithm name (to provide for future changes), the “iteration count” and the “salt”.

**Iteration count** means the CPU power needed to calculate the hash. To harden the password hashes in case of theft, the higher the needed CPU cycles are, the more resources a hacker needs to spend to brute-force the original passwords. Since CPUs improve over time, this is automatically calculated on a daily basis to be future-safe.

**Salt** is a random string that is added in the password hashing process to make the hashes unique, so the creation of rainbow tables is turned useless.

Follow-up pull request to #4609 ([immediate diff](/matthias-ronge/kitodo-production/compare/remove-write-user-to-ldap-button...use-secure-hashing-for-new-passwords))


henning-gerhardt:
This is part three of four pull requests (#4607, #4609, #4610 (this one) and #4616 (which is closed)).

Kathrin-Huber:
please reopen if it is not a draft anymore

PR #4612, opened 2021-08-26
Task filter checkboxes

### LINKED ISSUE #4159: Add filter to show only correction steps
There is a filter to hide correction steps in the tasks.

task filters

**Goal:** Another filter should be added to do the opposite and only show correction steps.

### LINKED ISSUE #4345: Change task filters to checkbox menu
Currently the text of task filters like `Show only own tasks` is changed depending on whether the filter is applied or not:

**_Filter not applied:_**
Bildschirmfoto 2021-04-16 um 15 32 47

**_Filter applied:_**
Bildschirmfoto 2021-04-16 um 15 34 41

This is not really user friendly and should be replaced with a checkbox menu that communicates much clearer whether a filter is applied or not (using something like this https://www.primefaces.org/showcase/ui/input/checkboxMenu.xhtml)

solth:
Fixes #4345, fixes #4159

Bildschirmfoto 2021-08-26 um 17 08 50

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -1294,7 +1294,7 @@ Import form
     color: var(--medium-gray);
 }
 
-.task-filter-panel .ui-selectcheckboxmenu-items-wrapper {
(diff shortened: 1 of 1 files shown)


Kathrin-Huber:
The dropDowns seems to need styling

PR #4616, opened 2021-08-30
Migrate passwords

### LINKED ISSUE #456: Replace current mechanism to store encrypted passwords

matthias-ronge:
Last part of, and finally fixes #456 after 4¼ years! :sweat_smile:

Follow-up pull request to #4610 ([immediate diff](/matthias-ronge/kitodo-production/compare/use-secure-hashing-for-new-passwords...migrate-passwords))


henning-gerhardt:
Please fix the merge conflicts as with them an usable review is not possible.

matthias-ronge:
This is no longer possible because the diff is no longer understandable. The development, which has passed half a year, is finished.

henning-gerhardt:
@matthias-ronge : So far as I understand is this pull quest the fourth of four pull requests (#4607, #4609, #4610 and this one). Should the other pull requests not closed too? So far as I was in my review notices the other three pull request change a lot and without this pull request they are breaking the login.

PR #4624, opened 2021-09-01
Issue 4621

### LINKED ISSUE #4621: Mismatching implementation and use of Helper.getTranslation(String, String)
The second parameter to the function is **a fallback value**, if the first value doesn’t resolve in the `messages.properties`:

```java
public static String getTranslation(String inParameter, String inDefaultIfNull) {
    String result = getTranslation(inParameter);
    return Objects.nonNull(result) && !result.equals(inParameter) ? result : inDefaultIfNull;
}
```

However, in many places where the function is used, the second parameter is used to pass an argument. Example:

```java
Helper.setErrorMessage(Helper.getTranslation("errorReading", file.getFileName()));
```

This leads to the argument missing. For the above example, will show / log:

    Error while reading from ''{0}''.

But should be, for example:

    Error while reading from ''meta.xml''.

**Goal:** Insert argument correctly. The `getTranslation(String)` which is called does never return null. Provide a `getTranslation(String, String...)`.

matthias-ronge:
Fix #4621


Kathrin-Huber:
There is a failing Test

Kathrin-Huber:
please resolve conflicts

PR #4629, opened 2021-09-03
Log absolute path if modules folder isn’t found

matthias-ronge:


Kathrin-Huber:
please fix checkstyle

PR #2929, opened 2019-10-15
Update tomcat

henning-gerhardt:
Updating Apache Tomcat to a more current version. Distribution of Apache Tomcat 7 is decreasing as many current Linux distribtions shipping at least Apache Tomcat 8.5.x. Apache Tomcat 8.0.x is already end of life since a few years.

# Note
* used version is similar to shipped version of Debian 9

# Advantage
* Servlet specification is increased from 3.0 to 3.1
* JSP specification from 2.2 to 2.3
* EL specification from 2.2 to 3.0

# Disadvantage
* Tomcat Maven Plugin is not developed any more and must be replaced or removed (PR is removing this)

PR #4634, opened 2021-09-06
Upgrade to Tomcat v9

thomaslow:
In reference to the recent update to Tomcat 8.5, see #2929, we could also update to Tomcat 9.

Advantages: 
- Potentially longer maintenance window (end-of-life dates for 8.5 and 9 not announced yet)
- New installations of Kitodo might be based on current LTS releases of Ubuntu and Debian,  
  which only provide packages for Tomcat 9
  - Ubuntu 18.04 (eol April 2023) provides packages for both Tomcat 8 and 9 called `tomcat8` and `tomcat9`
  - Ubuntu 20.04 (eol  April 2025) provides only packages for Tomcat 9 called `tomcat9`
  - Debian 9 (eol August 2022) provides only packages for Tomcat 8 called `tomcat8`,  
     Tomcat 9 can only be installed via backports or manually
  - Debian 10 (eol ~2024) provides only packages for Tomcat 9 called `tomcat9`

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -32,7 +32,7 @@
         <parent.basedir>${basedir}/../</parent.basedir>
         <maxAllowedViolations>15</maxAllowedViolations>
         <selenium.version>3.141.59</selenium.version>
-        <cargo.plugin.version>1.7.7</cargo.plugin.version>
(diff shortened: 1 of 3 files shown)


Kathrin-Huber:
Please rebase and resolve conflicts

PR #4667, opened 2021-09-13
Fixes #4666

### LINKED ISSUE #4666: Cannot change password for non-LDAP-bound user
With LDAP globally enabled, a user password cannot be changed for users that are not LDAP-bound (they have no LDAP group set):
```
    at org.kitodo.production.services.data.LdapServerService.changeUserPassword(LdapServerService.java:370)
    at org.kitodo.production.forms.UserForm.changePasswordForCurrentUser(UserForm.java:642)
    ...
```

**Workaround:** Temporarily set `ldap_use=false` in `kitodo_config.properties`. Here, requires a restart of Tomcat to apply.

For LDAP-bound users, the password is changed even if a password constraint is violated, and the application shows an error message that the new password is not allowed.

matthias-ronge:
- Treats a user without an LDAP group as a database user
- Order of instructions changed: If the password is not allowed, it will not be changed in the LDAP.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
@@ -639,12 +639,14 @@ public class UserForm extends BaseForm {
             Helper.setErrorMessage("passwordsDontMatchOld");
         }
         try {
-            if (DynamicAuthenticationProvider.getInstance().isLdapAuthentication()) {
-                ServiceManager.getLdapServerService().changeUserPassword(userObject, this.passwordToEncrypt);
-            }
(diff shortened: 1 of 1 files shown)


henning-gerhardt (changes requested):
I tried your changes today. With your changes to used password for LDAP users must now be valide against the used password rules. But your changes did not affect database based users. I get still the same exception or must your mentioned work-around still be applied before?

henning-gerhardt (approved):
On my first review I used a wrong change set as I needed to apply your changes in a manual way. But this time I used an other way to get your changes on my system so I can now successful confirm that your changes are working.

PR #4672, opened 2021-09-14
Fix process creation performance

Kathrin-Huber:
Requesting all processes and filtering them with streaming took a long time, when user is assigned to a lot of projects (with a lot of processes)
Speed up page loading by indexseraching.


matthias-ronge (approved):
It's good!

Issue #4241, opened 2021-03-05
LDAP server manager password visible and stored in plain text


henning-gerhardt:
Setting up a LDAP server the server manager password is stored in plain text in the database and is visible as plain text if you open this entry!

matthias-ronge:
Yes that's true. How do you want to do it differently?

henning-gerhardt:
> How do you want to do it differently?

I was against storing this password manager into the database and I did not implement this change. 

My approach would be maybe: 

1) store the password with a strong symmetric algorithm and if this algorithm need / support a salt value store this salt value in a different place, so you need two places to look and decrypt the stored password. If you need access to this password, decrypted it, use is and destroy the used variable on method exit.
2) On display: did not return the stored password into the form but use a notice that informing the user / administrator that the password is stored but not displayed again because of security reasons.

matthias-ronge:
Neither do I. But any encryption that is done, which password is publicly available on GitHub, is just as worthless in the end, so why not trust Unix file system permissions and write the password to a file where no other user can easily read it, only Tomcat , and even root has to change permissions first, that's a small hurdle, but if root already has permissions to read local LDAP content anyway, that's only a small hurdle. I always agree to include security and hurdles, but where is the reasonable limit for administrator usability?

henning-gerhardt:
>  But any encryption that is done, which password is publicly available on GitHub, is just as worthless in the end, 

That is only true if you don't use a salt and use the default initialization vectors of the encryption. If you store this outside and did not make it public then it is more complicated to get the information. The used algorithm is only one key in the chain of encryption an decryption.

> so why not trust Unix file system permissions and write the password to a file

Then the password should be moved back.

> but where is the reasonable limit for administrator usability

Because of security? The LDAP manager password is the most important passwords that you can have and if this password get public in some way then is calling the police only one step in the next couple of hours.

matthias-ronge:
That should be taken seriously.

matthias-ronge:
I've thought about it a little longer. In my opinion, there is only one sensible solution to this problem in a high-security context: **not to save the password at all.** Also not reversibly encrypted because the key would also have to be stored on the system. Also not encrypted with a salt, because the salt would also have to be stored on the system. Both can be viewed by an intruder who has gained access to the system. (In addition, both would probably never be changed in an ordinary installation, and it would create an additional semblance of security where there is no security, which is even more dangerous.)

So, in this case, that's just the suggestion:

[...]

PR #4677, opened 2021-09-16
#4241 change prime face tag to hide password

markusweigelt:
Does not fix issue #4241 atm


matthias-ronge:
Maybe you can add [the eye](https://stackoverflow.com/q/57162808) so that administrators can still copy the password if they need it?

Kathrin-Huber (approved):

markusweigelt:
> Maybe you can add [the eye](https://stackoverflow.com/q/57162808) so that administrators can still copy the password if they need it?

I added a ticket to accommodate your suggestion #4684

PR #4685, opened 2021-09-21
Change metadata script syntax

Kathrin-Huber:
This changes unintuitive syntax to new:

Example:
`action:addData key:titleDocMain value:myTitle`

possible attributes:
* key (name of new key), 
* value (new value), 
* source (when value is copied from another metadatakey), 
* variable (e.g. processid, all from VariableReplacer), 
* type (if new metadata should be added to a specific type, e.g. PeriodicalIssue)

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/AddDataScript.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/AddDataScript.java
@@ -36,7 +36,7 @@ public class AddDataScript extends EditDataScript {
 
         for (String value : metadataScript.getValues()) {
             MetadataEntry metadataEntry = new MetadataEntry();
-            metadataEntry.setKey(metadataScript.getGoal());
+            metadataEntry.setKey(metadataScript.getMetadataKey());
             metadataEntry.setValue(value);
             metadataEntry.setDomain(MdSec.DMD_SEC);
             metadataCollection.add(metadataEntry);
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
@@ -21,19 +21,18 @@ import org.apache.commons.lang.text.StrTokenizer;
 
 public class MetadataScript {
 
-    private Map<String, String> parameters;
-    private String metadataKey;
-    private String valueSource;
-    private String typeTarget;
-    private String variable;
-    private List<String> values = new ArrayList<>();
+    private final String metadataKey;
+    private final String valueSource;
+    private final String typeTarget;
+    private final String variable;
+    private final List<String> values = new ArrayList<>();
 
     /**
      * Creates a MetadataScript with given command.
      * @param command the given command.
      */
     public MetadataScript(String command) {
-        this.parameters = new HashMap<>();
+        Map<String, String> parameters = new HashMap<>();
         StrTokenizer tokenizer = new StrTokenizer(command, ' ', '\"');
         while (tokenizer.hasNext()) {
             String tok = tokenizer.nextToken();
@@ -90,7 +89,4 @@ public class MetadataScript {
         return variable;
     }
 
-    public void setVariable(String variable) {
-        this.variable = variable;
-    }
 }
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
@@ -11,13 +11,13 @@
 
 package org.kitodo.production.services.command;
 
-import org.apache.commons.lang.text.StrTokenizer;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+
+import org.apache.commons.lang.text.StrTokenizer;
 
 public class MetadataScript {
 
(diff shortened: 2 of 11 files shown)


Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
```diff
@@ -12,60 +12,62 @@
...
-        }
-        if (metadataCommand.contains("=")) {
-            String[] metadataParts = metadataCommand.split("=");
-            goal = metadataParts[0];
-            String rootOrValue = metadataParts[1];
-            if (rootOrValue.startsWith("@") || rootOrValue.startsWith("$")) {
-                root = rootOrValue;
-            } else {
-                values.add(rootOrValue);
+        this.parameters = new HashMap<>();
```
matthias-ronge, line 36: `parameters` isn’t used outside of the function. Please replace the class field by a local variable.

Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
```diff
@@ -76,19 +78,19 @@ public String getRoot() {
...
      */
     public String getTypeTarget() {
         return typeTarget;
     }
+
+    public String getVariable() {
+        return variable;
+    }
+
+    public void setVariable(String variable) {
```
matthias-ronge, line 93: The setter is not used. You can delete it, to make `variable` final.

Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
```diff
@@ -12,60 +12,62 @@
...
 public class MetadataScript {
 
-    private String goal;
-    private String root;
+    private Map<String, String> parameters;
+    private String metadataKey;
+    private String valueSource;
     private String typeTarget;
+    private String variable;
     private List<String> values = new ArrayList<>();
```
matthias-ronge, line 29: `values` is always zero or at most one element. Please change to `Optional<String> value;`

Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
```diff
@@ -12,60 +12,62 @@
...
 import java.util.Objects;
 
+import org.apache.commons.lang.text.StrTokenizer;
+
 public class MetadataScript {
 
-    private String goal;
-    private String root;
+    private Map<String, String> parameters;
+    private String metadataKey;
```
matthias-ronge, line 25: It looks like you can make all of these variables `final`.

Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
```diff
@@ -12,60 +12,62 @@
...
             }
         }
-        else {
-            goal = command;
-        }
-        if (typeCommand.contains("=")) {
-            typeTarget = typeCommand.split("=")[1];
+
+        metadataKey = parameters.get("key");
+        if (Objects.nonNull(parameters.get("value"))) {
```
matthias-ronge, line 48: With `Optional<String> value`, you can simplify this line to:

    value = Optional.ofNullable(parameters.get("value"));

Kitodo/src/main/java/org/kitodo/production/services/command/AddDataScript.java
```diff
@@ -36,7 +36,7 @@ public void executeScript(LegacyMetsModsDigitalDocumentHelper metadataFile, Proc
 
         for (String value : metadataScript.getValues()) {
```
matthias-ronge, line 37: `value` can occur at most once (see comment on `MetadataScript` class, replacement with `Optional<String>`). You can remove the loop.

Kitodo/src/main/java/org/kitodo/production/services/command/AddDataScript.java
```diff
@@ -36,7 +36,7 @@ public void executeScript(LegacyMetsModsDigitalDocumentHelper metadataFile, Proc
...
             MetadataEntry metadataEntry = new MetadataEntry();
-            metadataEntry.setKey(metadataScript.getGoal());
+            metadataEntry.setKey(metadataScript.getMetadataKey());
             metadataEntry.setValue(value);
             metadataEntry.setDomain(MdSec.DMD_SEC);
```
matthias-ronge, line 41: I would suggest making the domain configurable as a parameter, or better yet, load the ruleset and look it up.

Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
```diff
@@ -12,60 +12,62 @@
...
-            String[] metadataParts = metadataCommand.split("=");
-            goal = metadataParts[0];
-            String rootOrValue = metadataParts[1];
-            if (rootOrValue.startsWith("@") || rootOrValue.startsWith("$")) {
-                root = rootOrValue;
-            } else {
-                values.add(rootOrValue);
+        this.parameters = new HashMap<>();
+        StrTokenizer tokenizer = new StrTokenizer(command, ' ', '\"');
+        while (tokenizer.hasNext()) {
```
matthias-ronge, line 38: The script is already tokenized in the `KitodoScriptService` class. The result is available there in the class field `parameters` of the class `KitodoScriptService`. Please do not tokenize the script here again, instead, pass the `parameters` field through here.

matthias-ronge (changes requested):

Kitodo/src/main/java/org/kitodo/production/services/command/MetadataScript.java
Kathrin-Huber: It's not parsed correctly in the KitodoScriptService. I cannot use it, when multiple scripts are entered with ';'

Issue #4675, opened 2021-09-15
No need to enter password for read-only users with LDAP


matthias-ronge:
When creating a user, you have to set a password for him. That makes sense for authentication without LDAP in its whole, and it makes sense for authentication with a writeable LDAP server, where Production writes the user entry with password into LDAP. For users who authenticate themselves on external LDAP that is read-only, entering a password here makes no sense. It will never be used and should not be required.

**Goal:** If an LDAP group is selected for a new user, and the LDAP group has an LDAP server that is read-only, no password should be requested.

![No password required](https://user-images.githubusercontent.com/3040657/133409625-439e8d84-cf42-4186-be73-14b8bc6db942.png)

PR #4687, opened 2021-09-21
Disable user password field for read only LDAP servers

solth:
#4675


matthias-ronge (approved):
Looks correct.

Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
```diff
@@ -718,4 +718,18 @@ public void setFilter(String filter) {
...
+    /**
+     * Check and return whether LDAP group and LDAP server are configured for current user
+     * and if LDAP server is read only.
+     *
+     * @return whether LDAP server is configured and read only
+     */
+    public boolean isLdapServerReadOnly() {
+        try {
+            return this.userObject.getLdapGroup().getLdapServer().isReadOnly();
+        } catch (NullPointerException e) {
```
Kathrin-Huber, line 731: Please use null checks instead of Exception Handling!

Kathrin-Huber (changes requested):

Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
solth: This is the rare case where I opted for a more "phytonic" way of error handling ("asking for forgivness is easier than asking for permission"), because it seemed more appropriate and concise in this case. Still not acceptable?

matthias-ronge: I had overlooked that, but yeah, that's not how it should be. In particular, we know the case, it's well-defined: If a user doesn't have an LDAP group, then it's a database user. An LDAP group without an LDAP server should actually no longer be possible (```CONSTRAINT `FK_ldapGroup_ldapServer_id` FOREIGN KEY (`ldapserver_id`) REFERENCES `ldapserver` (`id`)```). I don't know if the case _should_ be caught at all.

PR #4699, opened 2021-09-23
Prestructured import

solth:
This pull request enables "prestructured import" of structure and metadata. 

That means the import service delegates the creation of the whole process METS structure required to the referenced XSLT mapping file (see provided example XSLT mapping file `mods2dfgviewer_prestructured.xsl` for details). The XSLT mapping is responsible for creating the METS structure (in contrast to just mapping the contained metadata in MODS/MARC/whatever to Kitodo internal format) by either copying the existing or creating new METS elements.

This feature can be activated by adding the line `<prestructured>true</prestructured>` to the corresponding OPAC configuration in `kitodo_opac.xml`.

**Note:** the provided file `mods2dfgviewer_prestructured.xsl` is just a template and should be adapted for individual use cases. Especially the mapping of METS filegroups and including files with `href` attributes without changes will not be applicable in most cases, because it means remote image files on other system will be referenced within the resulting METS file.

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
+++ b/Kitodo-API/src/main/java/org/kitodo/config/OPACConfig.java
@@ -67,8 +67,21 @@ public class OPACConfig {
     }
 
     /**
+     * Retrieve and return the value of the "prestructured" flag of a specific catalog,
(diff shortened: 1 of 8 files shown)


Kathrin-Huber:
Please add example config-parameter

PR #4701, opened 2021-09-24
add scripts to script popup

### LINKED ISSUE #4589: Adjustment of KitodoScripts
### Problem
On the front-end of Kitodo.Production 3.x not all [KitodoScripts](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java) are shown, as for example: 
- runscript
- deleteProcess
- addData
- overwriteData
- deleteData
- copyDataToChildren

On the other hand, some KitodoScripts should be examined regarding the functionality, as for example: 
- copyData

### Solution
Addition of all missing KitodoScripts and removal of unnecessary KitodoScripts.

Kathrin-Huber:
fixes #4589

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/executeScriptSelectedPopup.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/executeScriptSelectedPopup.xhtml
@@ -64,7 +64,25 @@
                                                  onclick="document.getElementById('executeScriptSelectedForm:selectionScriptFieldTextArea').value='action:exportDms exportImages:false'"/>
                                 <p:commandButton value="copyData"
                                                  update="executeScriptSelectedForm:selectionScriptFieldTextArea"
-                                                 onclick="document.getElementById('executeScriptSelectedForm:selectionScriptFieldTextArea').value='action:copyData &lt;rule(s)&gt;'"/>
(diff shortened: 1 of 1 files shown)


Kathrin-Huber:
Please add example config

andre-hohmann:
> Should we introduce individual permissions for the scripts? With deleteProcess script for example, you can create a lot serious trouble.

That is a good question, @matthias-ronge . It seems, that all users can use the KitodoScripts. I expected them only for Administration. Therefore, i think, we need permissions for the scripts and i guess @henning-gerhardt also supports the permissions. 

However, i suggest to discuss that question in a new issue. Would that be OK?

andre-hohmann:
> Question is: which permission is now needed to execute the KitodoScripts? If there is no permission then we should add at least one or more permissions, depending on how granular we need this permissions.

It really seems that there are no permissions. I think one permission for all Scripts should be fine, otherwise, the configuration is cumbersome. That could be assigned to the Administration or a new Role could be created.

andre-hohmann:
> I would suggest to add a permission "KitodoScriptExecution" which can then be assigned to any role.

In my opinion, a role for all KitodoScripts would be fine. Then everybody can assign the permission to any role, which needs it. 
I suggest "Execute KitodoScript" / "KitodoScript ausführen" so that the label-structure of the permissions are similar ("Edit task" / "Aufgabe bearbeiten").

PR #4702, opened 2021-09-27
Add new images to actual logical root

### LINKED ISSUE #3434: Validation warning: The structure has no media assigned: PeriodicalIssue "null"
### Problem
For processes of newspaper-issues the following warning is shown: 
`The structure has no media assigned: PeriodicalIssue "null"`

This leads to problems during the export. 

### Solution
The images of the process should be assigned automatically to the element "PeriodicalIssue". Otherwise, the images have to be assigned manually, which is quite time-consuming. 

### Example
validationwarning01

matthias-ronge:
When there are nameless containers on top of the logical structure, skip them. Fixes #3434

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -1184,6 +1184,11 @@ public class FileService {
     private void addNewMediaToWorkpiece(List<String> canonicals, Map<String, Map<Subfolder, URI>> mediaToAdd,
             Workpiece workpiece) {
 
+        LogicalDivision actualLogicalRoot = workpiece.getLogicalStructure();
+        while (Objects.isNull(actualLogicalRoot.getType()) && actualLogicalRoot.getChildren().size() == 1) {
+            actualLogicalRoot = actualLogicalRoot.getChildren().get(0);
+        }
+
         for (Entry<String, Map<Subfolder, URI>> entry : mediaToAdd.entrySet()) {
             int insertionPoint = 0;
             for (String canonical : canonicals) {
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1183,6 +1183,11 @@ private void addNewURIsToExistingPhysicalDivisions(Map<String, Map<Subfolder, UR
     private void addNewMediaToWorkpiece(List<String> canonicals, Map<String, Map<Subfolder, URI>> mediaToAdd,
             Workpiece workpiece) {
 
+        LogicalDivision actualLogicalRoot = workpiece.getLogicalStructure();
+        while (Objects.isNull(actualLogicalRoot.getType()) && actualLogicalRoot.getChildren().size() == 1) {
```
Kathrin-Huber, line 1187: What if there are no childrens yet? (is this possible?) Or when a pre-structured import already brings multiple children?

matthias-ronge: Okay, this isn’t fullly think about.  Case of multiple shildren should be as previous

matthias-ronge:
There is no case for which there are no children (or at least it shouldn't be); but that will now be automatically taken into account.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1183,6 +1183,16 @@ private void addNewURIsToExistingPhysicalDivisions(Map<String, Map<Subfolder, UR
...
+            actualLogicalRoot = actualLogicalRoot.getChildren().get(0);
+        }
+        // If the newspaper has multiple issues in the process, then everything
+        // stays as it was
+        if (Objects.isNull(actualLogicalRoot.getType()) && actualLogicalRoot.getChildren().size() != 1) {
+            actualLogicalRoot = workpiece.getLogicalStructure();
```
Kathrin-Huber, line 1193: isn't this obsolete with line 1186?

matthias-ronge: No. `actualLogicalRoot` is changed in line 1188, and this line sets it _back_ to what it was in line 1186 in case that the number of children is not 1 (that is, for processes with more than one newspaper issue; or (corrupted) processes with no issue at all).

PR #4703, opened 2021-09-27
No longer reinstantiate MetadataImageComparator

matthias-ronge:


Kathrin-Huber:
Seems like, checkstyle is complaining here ;-)

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -70,6 +69,8 @@
 
     private static final Logger logger = LogManager.getLogger(FileService.class);
 
+    public final MetadataImageComparator metadataImageComparator = new MetadataImageComparator(this);
```
henning-gerhardt, line 72: Public class variables? Use a getter method for accessing this variable and implement the used class as a singleton then you did not need this.

matthias-ronge: Singleton isn’t possible here, because of back-reference to non-static file service

Kathrin-Huber (approved):

Kathrin-Huber:
There seems to be something wrong with FilterServiceIT

matthias-ronge:
Probably caused by #4715

solth:
> Probably caused by #4715

Yes

Kathrin-Huber:
please rebase

Issue #4704, opened 2021-09-27
Metadata groups should be created with all contained metadata fields


andre-hohmann:
### Problem
If a metadata group is added in the metadata editor, several steps have to be followed: 
1. creation of the metadata group
2. creation of each contained and needed metadata field

When there are several metadata fields, as for example for a person or a corporation (display name, first name, last name, prefix, ID, ...) this is very cumbersome. 

### Solution
If a metadata group is added in the metadata editor, all contained metadata fields should be created immediately. Unneeded metadata fields can be left empty. 
If it is inconvenient for some users to create empty metadata fields, two buttons could be offered to create the metadata groups: 
1. creation of empty metadata group
2. creation of complete metadata group

matthias-ronge:
In the ruleset, you can activate this behaviour with the `alwaysShowing` option.

```xml
<editing>
    <setting key="Person">
        <setting key="Role" alwaysShowing="true"/>
        <setting key="IdentifierURI" alwaysShowing="true"/>
        <setting key="FirstName" alwaysShowing="true"/>
        <setting key="LastName" alwaysShowing="true"/>
        <setting key="DateOfDeath" alwaysShowing="true"/>
        <!-- ... whatever sub-keys you have/want to be showing ... ->
    </setting>
```

andre-hohmann:
Thank you very much. It works. Therefore, i close the issue.

Issue #4671, opened 2021-09-14
Handle multiple metadata


Kathrin-Huber:
follow-up to: #4670 

when importing from second import-source handle metadata, which already is imported from first import.

respect restrictions in ruleset. When key is permitted multiple times in the ruleset allow additional metadata (to the limit of ruleset).
Metadata which is not permitted by ruleset is ignored (for now)

PR #4705, opened 2021-09-27
Additional import

### LINKED ISSUE #4670: Enable additional import
when opening a new import (katalog or file) the user should be ask if he wants to import "new" or "additionaly".
When choosing "additionaly" the imported matadate should be added to the already importet metadata.

Steps:
[ ] when import is opend a popup should appear asking the user "import new or additionally"
[ ] when importing new -> usual workflow
[ ] when importing additionally -> already existing metadata should remain
[ ] for first draft: second import is not overwriting already exisiting metadata. (imported metadata for which key is already existing, newly imported metadata is ignored)

markusweigelt:
Fixes #4670


markusweigelt:
@Kathrin-Huber As required in this issue, the metadata fields are only generated if they are not yet available. Empty metadata fields are ignored. 

In view of the issue #4704, we should consider overwriting it if the metadata field is empty.

Ruleset of issue #4671 will also be considered, but it doesn't fully address the requirement yet.

PR #4707, opened 2021-09-28
#3316 switch to ajax request

### LINKED ISSUE #3316: Reports successful export only once
If the export was successful, this will be displayed three times. In addition, the green boxes at the top right ~~cannot be closed (and~~ do not disappear ~~) so the menu is no longer accessible~~  automatically.

Anmerkung 2020-03-17 143015

**Goal:** Show message only once.

markusweigelt:
Fixes #3316

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/processesWidget.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/processesWidget.xhtml
@@ -83,7 +83,7 @@
                                action="#{DesktopForm.exportMets(process.id)}"
                                styleClass="action"
                                title="#{msgs.exportMets}"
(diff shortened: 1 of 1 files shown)


matthias-ronge (approved):
I don’t understand what is happening here, but with regard to program style, etc., that is correct

matthias-ronge:
Yes, I am not deep into this, but I agree.

markusweigelt:
@solth I saw the issue and adapted it. ;) In my opinion, my fix a workaround rather than a solution. Because in some cases an Ajax variant will not be possible.

In the future, we should choose between output in flag/sticky-notification or top header message levels.

In order not to display  flag/sticky-notifications in a disruptive way, we should update prime face cause lifetime of flag is not changeable. "life" attribute of growl does not work.

PR #4733, opened 2021-10-11
Check newspaper process title duplicates

### LINKED ISSUE #4712: Duplicates are created during creation of newspaper-processes
### Problem
When newspaper-processes are created with the calendar-plugin, the created processes are not examined regarding existing processes with the same processtitle. 
This may lead to duplicates, which have to be deleted or repaired. Both are unnecessary efforts. 

### Solution
When newspaper-processes are created, the processtitles should be checked for duplicates, like during the creation of processes for other document types (monograph, volume). 
In Kitodo.Production 2.x, the creation of newspaper-processes was aborted in the taskmanager, as soon as a duplicte has been identified. 

### Examples
DubletteAusgaben

Kathrin-Huber:
this fixes #4712 
It's an easy and pragmatic solution wihtout a lot of changes in the code. The duplicates are identified when starting the thread. if there will be duplicates the thread finishes (without creating any processes).
This is not the solution suggested in the ticket, but it prevents the creation of duplicated titles, and it also doesn't create half of the processes until a duplicate is found (which would lead to incomplete editions).

Disadvantage: the duplicates are not identified in the CalendarForm, so if the user makes a mistake, he has to configure the whole newspaper again.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/CalendarForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/CalendarForm.java
@@ -22,10 +22,12 @@ import java.time.MonthDay;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


matthias-ronge (approved):
The solution is good for the problem.

PR #4757, opened 2021-10-14
Implement a switch to change, either type or metadata want to be disp…

### LINKED ISSUE #4202: Possibility to switch the structure map tree is missing
In Production 2, there were switches to influence the display of the structure map tree. So type or metadata could be displayed. The type and metadata behind it are displayed here (always), which takes some time to get used to it for people switching from Production 2.

**Goal:** It would be nice to have a switch to change, either type or metadata. For metadata, we could also offer a selection.

IkramMaalej:
…layed in the title information in the metadata editors structure and gallery panel.

Resolves #4202


Kathrin-Huber:
please resolve conflicts

Issue #4257, opened 2021-03-05
Deactivated users can still log in and working


henning-gerhardt:
Every deactivated user can log in and working independent if he / she as roles and / or projects assigned . This should not be possible!

solth:
I cannot reproduce this error. When I deactivate the following switch in the user settings:
<img width="161" alt="Bildschirmfoto 2021-03-18 um 15 58 35" src="https://user-images.githubusercontent.com/19183925/111647543-dd5b6a80-8802-11eb-9993-235e5a046721.png">
and then try to log in as that user, I get the following error message:
<img width="637" alt="Bildschirmfoto 2021-03-18 um 15 57 41" src="https://user-images.githubusercontent.com/19183925/111647607-ea785980-8802-11eb-8e2a-3fda9f8859da.png">

henning-gerhardt:
I can reproduce my own error and even your output. Difference is: if a user is disabled and LDAP authentication is used then this user can even log in. If the user is authentication is against the database then the user can not log in.

solth:
In that case I think you should update the issue title and description accordingly, to include the information about LDAP.

henning-gerhardt:
I changed the issue title and comment after your suggestions, @solth .

PR #4763, opened 2021-10-18
#4257 check for active user and throw spring authentication exception

markusweigelt:
Fixes: #4257

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/Helper.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/Helper.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Observable;
 import java.util.Observer;
+import java.util.Optional;
 import java.util.ResourceBundle;
 
 import javax.faces.application.FacesMessage;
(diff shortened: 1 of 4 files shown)


henning-gerhardt:
> > Looks good and did work for me. Only the not translated message "User is disabled" is a small needed correction.
> 
> It was a bit more complex because the FacesContext.getCurrentInstance() was null, and the helper always returns English. So locale is fetched from the LocaleContextHolder and helper had to be adjusted to support locales with a country.

Maybe this is the reason why on invalid login data only "Bad credential" independent of the browser / user locale appear. But this should be fixed in a separate pull request. The information "user is disabled" is not so good as this give the attacker the information that the login is existing.

markusweigelt:
@henning-gerhardt Yes i can change the value of "user is disabled" to more general message and log problem at debug level. I adapt the spring behavior and spring core throws the same error when DaoAuthenticationProvider is selected. There are other messages in the spring itself https://github.com/spring-projects/spring-security/blob/main/core/src/main/resources/org/springframework/security/messages.properties 
which are returned to ui on exception. 

I can catch these AuthenticationExceptions or overwrite message keys for more general messages. 
In general which error messages should we give and which shouldn't? Maybe we should create a separate ticket

PR #4808, opened 2021-11-04
Document parameter UserForImageReading

matthias-ronge:
Document parameter `UserForImageReading` ([File Management](/kitodo/kitodo-production/blob/master/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java#L497)).


markusweigelt:
Is it possible in that context to refactor this parameter in FileManagement to use Constant of ParameterCore instead of string literal?

markusweigelt (approved):

PR #4810, opened 2021-11-05
#4676 add login language selector and refactor locale functions to Lo…

### LINKED ISSUE #4676: Login page is always German
Foreign users report that the login page is always German, and there is no way to change it for the user who wants to log in.

**Goal:** The application should take into account the languages requested by the browser in the HTTP request headers, and if nothing is suitable, use the English language. Usually, JSF should do this automatically, so there is something going wrong in some place that it doesn't apply. 

It is a question of whether, on the login page, we should offer a language selection.

Login page

markusweigelt:
…caleHelper
Fixes #4676

Implementations status in frontend. Does it fit that way or are there other suggestions?
image


Kitodo/src/main/webapp/WEB-INF/resources/js/login.js
```diff
@@ -0,0 +1,18 @@
...
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+function setLanguageCookieAndReload( lang ) {
+    const d = new Date();
+    d.setTime(d.getTime() + (365 * 24 * 60 * 60 * 1000));
+    let expires = "expires="+d.toUTCString();
+    document.cookie = "kitodo_lang=" + lang + ";" + expires + ";path=/kitodo";
+    location.reload();
+}
```
solth, line 18: Missing new line at end of file

solth (changes requested):
Works well and I like the concept of having a language switch on the login patch, enabling users to change the UI language easily and even without requiring an account. 

Just a few little things concerning code style should get addressed before merging the pull request.

PR #4826, opened 2021-11-15
Fix order of thumbnails after drag'n'drop

### LINKED ISSUE #4820: Image order is reversed when dragging and dropping
If you drag several pictures with the mouse, they are ordered backwards after the action. In the animation below, the images paginated 4-7 are in the correct order, first. Then, after dragging them to another gallery strip, they are sorted backwards, you can see that the pagination is from 7-4 then:

Animation

**Goal:** The order of the pictures should not change when you drag them with the mouse.

solth:
Fixes #4820

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java
@@ -103,6 +103,9 @@ public class GalleryPanel {
 
     private String cachingUUID = "";
 
+    private static final Comparator<Pair<View, LogicalDivision>> IMAGE_ORDER_COMPARATOR = Comparator
+            .comparing(pair -> pair.getLeft().getPhysicalDivision().getOrder());
+
     GalleryPanel(DataEditorForm dataEditor) {
         this.dataEditor = dataEditor;
     }
@@ -227,8 +230,7 @@ public class GalleryPanel {
             }
         }
 
-        viewsToBeMoved = viewsToBeMoved.stream().sorted(Comparator.comparing(p -> p.getLeft().getPhysicalDivision()
-                .getOrder())).collect(Collectors.toList());
+        viewsToBeMoved.sort(IMAGE_ORDER_COMPARATOR);
 
         int toMediaIndex = getMediaIndex(event);
         try {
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java
```diff
@@ -226,6 +227,9 @@ public void onPageDrop(DragDropEvent event) {
             }
         }
 
+        viewsToBeMoved = viewsToBeMoved.stream().sorted(Comparator.comparing(p -> p.getLeft().getPhysicalDivision()
```
matthias-ronge, line 230: Coding guidelines discourage single letter variable names. Replace `p` with `pair`.

matthias-ronge (changes requested):
Thanks for the solution, but using streams feels overcomplicated to me for this task. Why not just declare a

```java
private static final Comparator<Pair<View, LogicalDivision>> SORT_BY_IMAGE_ORDER = Comparator
        .comparing(pair -> pair.getLeft().getPhysicalDivision().getOrder());
```
and then do a
```java
Collections.sort(viewsToBeMoved, SORT_BY_IMAGE_ORDER);
```

solth:
> Thanks for the solution, but using streams feels overcomplicated to me for this task. Why not just declare a
> 
> ```java
> private static final Comparator<Pair<View, LogicalDivision>> SORT_BY_IMAGE_ORDER = Comparator
>         .comparing(pair -> pair.getLeft().getPhysicalDivision().getOrder());
> ```
> 
> and then do a
> 
> ```java
> Collections.sort(viewsToBeMoved, SORT_BY_IMAGE_ORDER);
> ```

Isn't that just as complicated, if not even more so? You are just moving the Comparator to a constant (which is only used once anyway, so I don't the reason to make a constant/variable in the first place)

matthias-ronge:
Yes, that's subjective, I find it easier to read than to create a stream especially for sorting. The advantage of the constant is that it is only created once when the class is initialized, while the argument, as you defined it, is created every time the method is called.  
But it's enough for me if you replace the `p` with a `pair`.

markusweigelt (approved):

solth:
> The advantage of the constant is that it is only created once when the class is initialized, while the argument, as you defined it, is created every time the method is called.

Ok, that's a good reason to move the comparator to a constant! Even though the method is still not called very often (and therefore the comparator wouldn't be instantiated very often, either), in the metadata editor I take every optimization I can!

matthias-ronge (approved):

oliver-stoehr (approved):

Issue #4777, opened 2021-10-21
Creating processes from process template using plus button


matthias-ronge:
Creating processes from process templates from inside the Create New Process page is implemented in a different concept than any other process creation:

![Screenshot](https://user-images.githubusercontent.com/3040657/138247410-a3e1b46f-edc0-4a0d-aa59-9935249874d3.png)

Any other process creation:

- Search for project / process template / superordinated process in the respective list → hit <kbd>+</kbd> button → Create New Process dialog shows → complete details → create process

But here

- Open Create New Process dialog → navigate to “process template” panel → select source process from list (no search available) → click apply → complete details → create process.

**Goal:** For processes, that are considered “template processes”, the <kbd>+</kbd> button will be enabled. Clicking the button opens the Create New Process dialog with the data filled in.

solth:
This was fixed by #4834, I think.

Issue #4322, opened 2021-04-06
Improvement for processcreation by configurable import wizards


subhhwendt:
The idea is to analyze for these different starting points the specific requirements and support by smart and configurable wizards instead of switching between tabs 

- catalogue import of monographs, multivolume Works, periodicals and newspapers for a chosen project
- adding periodical volumes from the process list
- adding newspaper issues by the calendar functionality connected to a chosen newspaper title
- adding processes by file upload
- adding processes by mass import (with extended possibilities for ingest of individual values for metadata fields like CurrentNoSorting or NumberProcessTitle independent from catalogue import)
 
It should be configurable
-  which fields are derived form the parent process in case of hierarchical structures (e.g. docLanguage, CatalogIDDigital, Collection)
- which fields are shown for expected input or for control
 
**To Do**

- [ ] development of an import wizard for **catalogue import** -  after review of all dependencies, metadata requirements and expected choices
- [ ] development of a wizard for processcreation for **periodical volumes** - after review of all dependencies, metadata requirements and expected choices
- [ ] development of a wizard for creation of **newspaper issues** - after review of all dependencies, metadata requirements and expected choices
- [ ] development of a wizard for processcreation by **fileupload** - after review of all dependencies, metadata requirements and expected choices
- [ ] devolopment of a wizard for processcreation by **massimport**
- [ ] Documentation of all configuration issues in the perspective of processcreation

**Estimated Cost and Complexity**

This will be a complex project for about 20 PT.

andre-hohmann:
The following issues might be related to this proposal. They could be closed, if they fit into the concept and are could be solved. 

- #3407
- #3408 (https://github.com/kitodo/kitodo-production/issues/3408#issuecomment-804003220)
- #3463
- #3561
- #3593
- #3970
- #3974

matthias-ronge:
Another way of process creation is mass process creation with a counter, which for example is used for stocks which are not yet cataloged at all. In this mode of operation, first processes are created, that only consist of a shelf code and a sequential number. Dockets are printed and the elements of the shelf are first digitized and later cataloged. Processes that are not used at the end of the shelf are deleted later.

matthias-ronge:
First of all, in the list of variants above, the “simple” creation of a process is missing, of course. That is: select DocType, fill in metadata fields, generate process title, done.

[...]

PR #4834, opened 2021-11-24
Import process design

### LINKED ISSUE #3529: Combination of "process data" and "process metadata" for creation of a new process
It could be helpful, if the data fields of "process data" and "process metadata" are combined, when a process is created. As both options have pros and cons, that should be discussed in the community. 

The issue was created due to the following reasons: 
- Even if it is only one click more, to change the tab: The more processes are created, the more uncomfortable it feels. 
- The more processes are created, the more often mandatory metadata is forgotten and it is necessary to open the tabs "process data" and "process metadata" to correct the data, if theprocess title has to be generated again. 

vorgangsdaten01

vorgangsdaten02

IkramMaalej:
Fixes #3529

Part of #4322

In this PR I have try to simplify the page design of creation of a new process from template by :
- Combining the first Tab 'Basic data' and the second tab 'Metadata' in one tab.

Bildschirmfoto 2021-11-25 um 10 06 16

- Convert 'process from template' tab to a dialog in order to make this way of  importing data looks like the other ways 'Catalogue search' and 'File upload'.

Bildschirmfoto 2021-11-25 um 10 09 24

- Grouping the buttons in the header into two menu buttons, one for "data transfer" and one for "saving" in order not to have a cluttered header space.

Bildschirmfoto 2021-11-25 um 10 08 32
Bildschirmfoto 2021-11-25 um 10 08 47


Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/dataEdit.xhtml
```diff
@@ -18,6 +18,9 @@
         xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
         xmlns:p="http://primefaces.org/ui">
 
+    <div style="font-weight:bold;padding:.75em 1.5em;">
```
solth, line 21: Please move this styling to a separate CSS class and use `padding: var(--default-half-size) var(--default-full-size);` instead of `em`.

Kitodo/src/main/webapp/pages/processFromTemplate.xhtml
```diff
@@ -36,40 +36,55 @@
...
+                        icon="fa fa-upload"
+                        iconPos="right"
+                        update="fileUploadForm">
+                <p:resetInput target="fileUploadForm"/>
+            </p:menuitem>
+            <p:menuitem id="searchEdit"
+                        process="@this"
+                        value="#{msgs.processTemplate}"
+                        oncomplete="PF('searchEditDialog').show();"
+                        icon="fa fa-download"
```
solth, line 82:
```suggestion
                        icon="fa fa-copy"
```

Kitodo/src/main/webapp/pages/processFromTemplate.xhtml
```diff
@@ -36,40 +36,55 @@
...
-                         process="@this"
-                         value="#{msgs['newProcess.fileUpload.heading']}"
-                         oncomplete="PF('fileUploadDialog').show();"
-                         icon="fa fa-upload"
-                         iconPos="right"
-                         update="fileUploadForm"
-                         styleClass="secondary">
-            <p:resetInput target="fileUploadForm"/>
-        </p:commandButton>
+        <p:splitButton id="saveContinue"
```
solth, line 39: I would prefer to _not_ combine both `Save` buttons into one split button but leave them as they are as two separate buttons that are always displayed, because they seem both equally relevant to me.

IkramMaalej: It will look then like this

Bildschirmfoto 2021-11-25 um 13 33 45

IkramMaalej:
> Resolves #3529, doesn't it?

Yes!

solth:
I believe this also fixed #4777.

PR #4846, opened 2021-11-29
Fix Multi-line single-select metadata is not saved

### LINKED ISSUE #4068: Always showing multi-select metadata is not saved
A multi-select-type metadata input is not saved if it is `alwaysShowing="true"`. (Extracted from #4061)

### LINKED ISSUE #4069: Multi-line single-select metadata is not saved
A multi-line single-selecd-type metadata (list of radio buttons) input is not saved. (Extracted from #4061)

IkramMaalej:
Fixes #4069
Fixes #4068

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
@@ -131,7 +131,7 @@
                                       layout="grid"
                                       columns="1">
                         <f:selectItems value="#{item.items}"/>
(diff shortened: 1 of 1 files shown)


matthias-ronge (approved):
According to the Coding Guidelines, I have no objection. I can't say whether that solves the problem, I can't test it right now, but I assume you've tried your solution.

solth (approved):
#4069 is indeed fixed by this change, but it has no effect on #4068, which is about `alwaysShowing` **checkboxes**. However, that problem seems to have been already fixed in the current master branch, so I guess it's ok to close that issue with thie pull request, too.

PR #4850, opened 2021-11-30
Remove lock to delete or rearrange process links

### LINKED ISSUE #3275: Process link cannot be deleted
A process link should be deletable by selecting the structural element which contains it, and then deleting it. This is not possible because the structural element of a link cannot be clicked on.

Cannot click the linked structural element

**Goal:** Allow clicking on linked structural elements again.

matthias-ronge:
Fixes #3275

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
@@ -20,7 +20,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -41,6 +40,7 @@ import org.kitodo.api.dataformat.LogicalDivision;
 import org.kitodo.api.dataformat.View;
 import org.kitodo.config.ConfigCore;
 import org.kitodo.config.enums.ParameterCore;
+import org.kitodo.data.database.beans.Process;
 import org.kitodo.exceptions.InvalidMetadataValueException;
 import org.kitodo.production.forms.createprocess.ProcessDetail;
 import org.kitodo.production.forms.createprocess.ProcessFieldedMetadata;
@@ -49,7 +49,6 @@ import org.kitodo.production.forms.dataeditor.StructureTreeNode;
 import org.kitodo.production.helper.Helper;
 import org.kitodo.serviceloader.KitodoServiceLoader;
 import org.primefaces.model.TreeNode;
-import org.kitodo.data.database.beans.Process;
 
 public class DataEditorService {
 
--- a/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
@@ -247,8 +247,15 @@ public class DataEditorService {
         }
         StructureTreeNode structureTreeNode = (StructureTreeNode) selectedLogicalNode.getData();
 
-        if (structureTreeNode.getDataObject() instanceof View) {
-            View view = (View) structureTreeNode.getDataObject();
+        Object dataObject = structureTreeNode.getDataObject();
+        // data object is null for structures inside parent processes
+        if (Objects.isNull(dataObject)) {
+            return null;
+        }
+
+        // default case
+        if (dataObject instanceof View) {
+            View view = (View) dataObject;
             if (Objects.isNull(view.getPhysicalDivision())) {
                 throw new IllegalStateException("View has no physical division assigned");
             }
@@ -258,13 +265,14 @@ public class DataEditorService {
                         dataEditor.getAcquisitionStage(), dataEditor.getPriorityList());
         }
 
-        if (structureTreeNode.getDataObject() instanceof Process) {
+        // data object is a Process for parent processes
+        if (dataObject instanceof Process) {
             return null;
         }
 
         throw new IllegalStateException("Data object is of bad type: "
-                + (Objects.isNull(structureTreeNode.getDataObject()) ? "null"
-                    : structureTreeNode.getDataObject().getClass().getName()));
+                + (Objects.isNull(dataObject) ? "null"
+                    : dataObject.getClass().getName()));
     }
 
     private static Collection<Metadata> getExistingMetadataRows(List<TreeNode> metadataTreeNodes)
--- a/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -48,6 +49,7 @@ import org.kitodo.production.forms.dataeditor.StructureTreeNode;
 import org.kitodo.production.helper.Helper;
 import org.kitodo.serviceloader.KitodoServiceLoader;
 import org.primefaces.model.TreeNode;
+import org.kitodo.data.database.beans.Process;
 
 public class DataEditorService {
 
@@ -170,6 +172,9 @@ public class DataEditorService {
         try {
             if (currentElement) {
                 structureView = getStructuralElementView(dataEditor);
+                if(Objects.isNull(structureView)) {
+                    return Collections.emptyList();
+                }
                 existingMetadata = getExistingMetadataRows(metadataNodes);
             } else {
                 structureView = dataEditor.getRulesetManagement()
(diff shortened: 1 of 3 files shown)


solth (changes requested):
These changes break the existing functionality and are insufficient to fix the linked issue. Just allowing the user to click linked processes in the metadata editor does not work without actually adding functionality to tell the system what to do with selected linked processes. 

Without that new functionality, with just these changes, selecting a linked process in the metadata editor leads to `IllegalStateExceptions` like the following: 
Bildschirmfoto 2021-12-02 um 10 44 37
@matthias-ronge it seems you didn't test your changes, did you?

matthias-ronge:
I only removed the block because we had a customer complaint that it was not possible to delete or move links. I had built in these functions two years ago and they worked at the time. I admit that I am amazed at the mistake. There should actually be no error. For example, you could give this element metadata—whether that is intentional or makes sense is another question. It's good that I'll look it again.

matthias-ronge:
this is the code where the `IllegalStateException` is thrown (the last line).
```java
/**
 * Determine and return StructureElementViewInterface corresponding to structure element currently selected or to be
 * added via AddDocStructTypeDialog.
 *
 * @param dataEditor DataEditorForm instance used to determine StructureElementViewInterface
 * @return StructureElementViewInterface corresponding to structure element currently selected or to be added
 */
public static StructuralElementViewInterface getStructuralElementView(DataEditorForm dataEditor) {
    Optional<LogicalDivision> selectedStructure = dataEditor.getSelectedStructure();
    if (selectedStructure.isPresent()) {
        return dataEditor.getRulesetManagement()
                .getStructuralElementView(
                        selectedStructure.get().getType(),
                        dataEditor.getAcquisitionStage(), dataEditor.getPriorityList());
    } else {
        TreeNode selectedLogicalNode = dataEditor.getStructurePanel().getSelectedLogicalNode();
        if (Objects.nonNull(selectedLogicalNode)
                && selectedLogicalNode.getData() instanceof StructureTreeNode) {
            StructureTreeNode structureTreeNode = (StructureTreeNode) selectedLogicalNode.getData();
            if (structureTreeNode.getDataObject() instanceof View) {
                View view = (View) structureTreeNode.getDataObject();
                if (Objects.nonNull(view.getPhysicalDivision())) {
                    return dataEditor.getRulesetManagement().getStructuralElementView(
                        view.getPhysicalDivision().getType(),
                            dataEditor.getAcquisitionStage(), dataEditor.getPriorityList());
                }
            }
        }
    }
    throw new IllegalStateException();
}
```
For me this is spontaneously confusing, because the selected structure should be given, right? Didn't you select anything when you selected the link? Or is the selection not considered selected by JSF? I am spontaneously at a loss.

matthias-ronge:
I am not experiencing the exception you reported. For me, this works as expected.

matthias-ronge:
I checked in the debugger: When I click the link, the function is called, but `selectedStructure.isPresent()` is true and so the if-condition applies, no error here. In your case it must be false for some reason which doesn’t make sense if you actually selected something, so this actually is an illegal state, but the question is why.

solth:
I think I found out why you couldn't reproduce the error. It only occurs for me when you click parent processes, but not for child processes:

https://user-images.githubusercontent.com/19183925/145376845-c34f8e2e-1d03-4fc5-9f4c-5963ebf51214.mov

solth:
Additionally, I think something is not updated correctly when removing a linked child process in the metadata editor structure tree using the "Remove element" context menu option. 
When expanding the row for the parrent process in the process list afterwards, it still displays all previously linked child processes, including the one I just removed in the editor (I didn't forget to save in the editor, the removed child process is not shown again in the editor when opening the parent process again!)

matthias-ronge:
Thanks for the hint. I'll take a closer look at that. If deleting doesn't work anymore, that's a bug.

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
```diff
@@ -227,25 +231,40 @@ public static StructuralElementViewInterface getStructuralElementView(DataEditor
...
+                    view.getPhysicalDivision().getType(),
+                        dataEditor.getAcquisitionStage(), dataEditor.getPriorityList());
         }
-        throw new IllegalStateException();
+
+        if (structureTreeNode.getDataObject() instanceof Process) {
+            return null;
+        }
+
+        throw new IllegalStateException("Data object is of bad type: "
```
solth, line 265: It seems to me this exception is always thrown! I get this exception when selecting the _parent_ process of the current process in the metadata editor (same as before, no problem with linked child processes):

https://user-images.githubusercontent.com/19183925/149781042-9ad6e51f-481e-41ea-9b75-33e31fc37abe.mov

solth (changes requested):
I still get the same error

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
matthias-ronge: There is literally nothing happening in the video.

The exception happened because the `dataObject` is of type `Process` for parent processes, and the _master_ expects it to be a `View`. But this is now checked for (line 261). So I don’t understand why the exception still happens for you.

The exception should print the "bad" object type now. What does it print for you?

matthias-ronge: @solth Thank you for catching! Looks like I didn’t get the error yesterday, probably because the process directory of my parent process is missing on disk… 🙄

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
```diff
@@ -227,25 +231,48 @@ public static StructuralElementViewInterface getStructuralElementView(DataEditor
...
-                        return dataEditor.getRulesetManagement().getStructuralElementView(
-                            view.getPhysicalDivision().getType(),
-                                dataEditor.getAcquisitionStage(), dataEditor.getPriorityList());
-                    }
-                }
+        }
+
+        TreeNode selectedLogicalNode = dataEditor.getStructurePanel().getSelectedLogicalNode();
+        if (Objects.isNull(selectedLogicalNode)) {
+            throw new IllegalStateException("No logical node selected");
```
solth, line 240: Please add message keys in English and German

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
```diff
@@ -227,25 +231,48 @@ public static StructuralElementViewInterface getStructuralElementView(DataEditor
...
+        }
+
+        if (!(selectedLogicalNode.getData() instanceof StructureTreeNode)) {
+            throw new IllegalStateException("Selected logical node data is of bad type: "
```
solth, line 244: Please add message keys in English and German

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
```diff
@@ -227,25 +231,48 @@ public static StructuralElementViewInterface getStructuralElementView(DataEditor
...
+        // data object is null for structures inside parent processes
+        if (Objects.isNull(dataObject)) {
+            return null;
+        }
+
+        // default case
+        if (dataObject instanceof View) {
+            View view = (View) dataObject;
+            if (Objects.isNull(view.getPhysicalDivision())) {
+                throw new IllegalStateException("View has no physical division assigned");
```
solth, line 260: Please add message keys in English and German

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
```diff
@@ -227,25 +231,48 @@ public static StructuralElementViewInterface getStructuralElementView(DataEditor
...
+                        dataEditor.getAcquisitionStage(), dataEditor.getPriorityList());
         }
-        throw new IllegalStateException();
+
+        // data object is a Process for parent processes
+        if (dataObject instanceof Process) {
+            return null;
+        }
+
+        throw new IllegalStateException("Data object is of bad type: "
```
solth, line 273: Please add message keys in English and German

solth (changes requested):
The exception is now handled correctly and selecting linked processes does not result in an `IlllegalStateException` anymore. 👍 

But there are still some problems that need to be resolved:
- rearraning linked child processes via drag'n'drop results in wrong error messages (see first attached video below)
- removing a linked _parent_ process does not work (see second attached video below); if this is not _supposed_ to work, the corresponding option in the context menu should be deactivated/hidden when selecting parent processes
- context menu options `Element hinzufügen`, `Typ des Strukturelements ändern` and `Medien hochladen` do not work for linked processes and should be deactivated or hidden (see screenshots below)

https://user-images.githubusercontent.com/19183925/149907417-99a7a14f-7f2a-402a-95e7-66e53d83e718.mov

https://user-images.githubusercontent.com/19183925/149907471-bde93979-58b2-4948-986f-61dbdaf44e87.mov

Bildschirmfoto 2022-01-18 um 10 12 43

matthias-ronge:
A rebase is no longer possible because the diff is no longer understandable. Development is terminated.

Issue #4276, opened 2021-03-15
Script of automatic task is not called if task is first workflow step


solth:
The script of an automatic task is not called when the task is the first in a workflow.

solth:
Duplicate of #3672

matthias-ronge:
Closed as duplicate

Issue #3672, opened 2020-05-20
Script task as first task is not started


matthias-ronge:
Extracted from [comment on issue #3279](https://github.com/kitodo/kitodo-production/issues/3279#issuecomment-626210008) by @mnscholz:
> If the automatic script task is the first task of a process, it seems that the script doesn't get executed at all. The task displays as locked (strike-thru bell) forever.

BartChris:
@matthias-ronge 
Hello,

While implementing a naive fix for enabling automatic first workflow tasks (https://github.com/kitodo/kitodo-production/pull/4853), i began to wonder how this  has to be seen in the context of the different process creation modes (https://github.com/kitodo/kitodo-production/issues/4322). 
If all those modes are using a the same template (workflow) the question arises how automatic starting workflow task should be handled. In theory this is a constellation where whenever a process creation event occurs (triggered by whatever procedure), it would have to be checked if this process has an automatic open task.  And the event has to be triggered after everything related to process creation (creation of METS-file) is finished since the task might operate on the METS file.

Automatic first workflow steps are probably a prerequisite of fully automated workflows (https://github.com/kitodo/kitodo-production/issues/3333). On the other hand i do not know what the effect would be if a mass import would trigger a lot of CPU intensive tasks (e.g. generation of image derivatives) by automatically stating a task. 
My naive fix above worked for a simple use case, but i i suppose i am not sure how and where in the code base this could be best handled.

matthias-ronge:
Our workflow engine is a very old legacy from 10 years of code history. And, we've never did large changes to it, just fixed immediate problems. Even a programmer who authored this a long time ago, should have said, "I'll never touch the workflow engine again". At the moment, the thread that closes the task and opens the next task does the basic work for the task (like creating or removing symbolic links), only a script is started asynchronously in a `TaskScriptThread` (you can sometimes see such in the Taskmanager). We have to think carefully, for example, if thousands of processes are automatically created for newspaper processes (this is done by a thread already), should each initialize a `TaskScriptThread` (they would stack up and possibly give out of memory errors) or participate in the same thread (that producing newspaper processes would then be done very, very slowly. What if the thread crashes or the server needs a reboot meanwhile?)

[...]

PR #4853, opened 2021-11-30
Enable first step as automatic task

BartChris:
Hello,
Right now it is not possible to define the first task in the workflow as an automatic task since the necessary checks seem only to be performed after another task is closed.

https://github.com/kitodo/kitodo-production/blob/51fbdd742fc272a8244964e8a8e4d06acb277bbd/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java#L439-L443

In order to enable an _automatic_ first task i added a check after the process is created and defined the necessary logic to start the automatic task. The solution works in my environment but please check if i am missing something.

Fixes https://github.com/kitodo/kitodo-production/issues/4276 and https://github.com/kitodo/kitodo-production/issues/3672

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
@@ -16,7 +16,6 @@ import java.io.OutputStream;
 import java.net.URI;
 import java.text.MessageFormat;
 import java.util.Collections;
-import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
@@ -36,9 +35,7 @@ import org.kitodo.api.dataformat.Workpiece;
 import org.kitodo.data.database.beans.Process;
 import org.kitodo.data.database.beans.Project;
 import org.kitodo.data.database.beans.Ruleset;
-import org.kitodo.data.database.beans.Task;
 import org.kitodo.data.database.beans.Template;
-import org.kitodo.data.database.enums.TaskStatus;
 import org.kitodo.data.database.exceptions.DAOException;
 import org.kitodo.data.exceptions.DataException;
 import org.kitodo.exceptions.CommandException;
@@ -51,7 +48,6 @@ import org.kitodo.production.enums.ObjectType;
 import org.kitodo.production.forms.BaseForm;
 import org.kitodo.production.helper.Helper;
 import org.kitodo.production.helper.TempProcess;
-import org.kitodo.production.helper.tasks.TaskManager;
 import org.kitodo.production.interfaces.RulesetSetupInterface;
 import org.kitodo.production.metadata.MetadataEditor;
 import org.kitodo.production.process.ProcessGenerator;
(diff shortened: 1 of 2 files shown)


markusweigelt:
@BartChris Please fix checkstyle problems. Thx.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -320,6 +321,18 @@ public String createNewProcess() {
         return this.stayOnCurrentPage;
     }
 
+    private void checkForAutomaticTask() {
```
markusweigelt, line 324: Suggest to rename something like "runAutomaticTask" because in method you trigger more than a check. With this method name, I expect to return a boolean if something is valid.

markusweigelt (changes requested):
- Combine nested if to one
- Rename function with more suitable name

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -436,6 +441,15 @@ private void createProcessHierarchy()
             }
         }
         ServiceManager.getProcessService().save(getMainProcess(), true);
+        Task currentTask = ServiceManager.getProcessService().getCurrentTask(getMainProcess());
+        if (Objects.nonNull(currentTask)) {
+            if (currentTask.isTypeAutomatic()) {
```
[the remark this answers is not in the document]
BartChris: thx, i changed that

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
BartChris: thx, i renamed the method

BartChris:
This requires more, open questions:
- what about other scenarios of creating a processes? 
- should the same behavior be implemented everywhere a process is created?

matthias-ronge:
At the moment, the thread that closes the task and opens the next task does the basic work for the task (like creating or removing symbolic links), only a script is started asynchronously in a `TaskScriptThread` (you can sometimes see such in the Taskmanager). You have to think carefully, for example, if thousands of processes are automatically created for newspaper processes (this is done by a thread already), should each initialize a `TaskScriptThread` (they would stack up and possibly give out of memory errors) or participate in the same thread (that producing newspaper processes would then be done very, very slowly. What if the thread crashes?)

I think we would need a completely different design in the background to solve this in a clean way. I think there should be a separate task engine where each automatic task is actually represented with a thread (or a few, a fixed number though) that handles those tasks. Actually, I would even see this as a separate program, so that you can install multiple task engines on different servers and they work together. But for that, we would have to reprogram the whole task handling.  Or maybe, in the future, make use of a powerful workflow engine and dispose of the self-made one.

matthias-ronge:
Sorry, I thought this was an issue discussion. My comment was general reflection, not feedback on the pull request.

BartChris:
Closed, needs more thoughts on side effects- See related issue.

PR #4874, opened 2021-12-10
Update log4j2 to fix CVE-2021-44228

henning-gerhardt:
More information about CVE-2021-44228:
- https://www.lunasec.io/docs/blog/log4j-zero-day/
- https://github.com/jas502n/Log4j2-CVE-2021-44228


stweil (approved):
This might be urgent for installations which are accessible from the Internet.

stweil:
Do we need updates for old releases, too?

PR #4884, opened 2021-12-15
Hierarchie file import

### LINKED ISSUE #4817: File upload multivolume: Missing link to existing superordinate processes
### Problem
If a process of a multivolume is created via file upload, it can be observed that:  
- the superordinate and the subordinate process are created correctly 
- the superordinate and the subordinate processes are linked to each other in the database and thus in the process list

If a second volume of the same superordinate process is created via file upload, it can be observed that: 
- the subordinate process is created correctly 
- a new superordinate process is created - instead of using the existing superordinate process

### Solution
For the import of multivolume file via file upload, it should be examined, if superordinate processes exists: 
- If a superordinate processes exists -> the new process should be linked to it
- If a superordinate processes does not exist -> a superordinate and a subordinate process should be created

Kathrin-Huber:
check for existing parents (in DB) on fileUpload
Fixes #4817

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/FileUploadDialog.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/FileUploadDialog.java
@@ -76,7 +76,6 @@ public class FileUploadDialog extends MetadataImportDialog {
 
             String parentID = importService.getParentID(internalDocument, this.createProcessForm.getRulesetManagement().getFunctionalKeys(
                     FunctionalMetadata.HIGHERLEVEL_IDENTIFIER).toArray()[0].toString());
-            if (Objects.nonNull(parentID)) {
                 importService.checkForParent(parentID,
                     createProcessForm.getTemplate().getRuleset().getId(), createProcessForm.getProject().getId());
                 if (Objects.isNull(importService.getParentTempProcess())) {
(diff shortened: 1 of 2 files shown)


andre-hohmann:
Fixes #4817

I have created four processes of an existing superordinate process. All four processes are created correctly and have been linked to the superordinate process. 
Thus, from the users' perspective, the adjustments in the pull request are correct.

andre-hohmann:
Unfortunately, there is a mistake regarding importing files of processes without hierarchies. 

If a file for a monograph, ... without hierarchy is imported, the "Titelsatzverknüpfung" is shown by default with the error message: 

`Sie können den Vorgang nicht mit diesem Titelsatz verknüpfen, da der ausgewählte DocType an keiner Stelle als Kind erlaubt ist.`

It is then necessary to de-select the processtitle in "Vorgang auswählen" by choosing "Bitte auswählen". This is cumbersome. 

Proposal: 
The tab with "Titelsatzverknüpfung" should only be shown, when the imported file contains hierarchies.

Kathrin-Huber:
Is this a general problem?
So does this also occure, when importing process from cataloge with existing parents and afterwards processes without parents at all?

if so, please open a bug ticket becuase it is not linked to this PR

andre-hohmann:
It seems to be a general problem, although it is more obvious during the file import: 
- #4898

solth (approved):

solth:
> Fixes #4817

@Kathrin-Huber must link the issue in the original description for it to be closed automatically when this PR is merged.

PR #4885, opened 2021-12-15
Add options 'type-metadatakey' to 'switch title' drop-down menu i…

IkramMaalej:
…n structural panel


Kathrin-Huber:
Why is this still a draft?

IkramMaalej:
@Kathrin-Huber  
It is not yet clear what should be the options of this drop-down menu and how to be configured.
See : https://github.com/kitodo/kitodo-production/discussions/4925

PR #4886, opened 2021-12-15
Rename hasParent() to match EL property rules

matthias-ronge:
Access to object properties in JSP should be done via El expressions for properties, and that defines that an _is_ property is to be used for `boolean`, not a _has_ property. With the renaming, that will be clarified.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/dto/ProcessDTO.java
+++ b/Kitodo/src/main/java/org/kitodo/production/dto/ProcessDTO.java
@@ -32,7 +32,7 @@ public class ProcessDTO extends BaseTemplateDTO {
     private String processBaseUri;
     private String batchID;
     private Integer parentID;
-    private boolean hasChildren;
+    private boolean parenthood;
     private Integer sortHelperArticles;
     private Integer sortHelperDocstructs;
     private Integer sortHelperImages;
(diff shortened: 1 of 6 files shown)


solth:
IMO `isParenthood` sounds unnatural and should be replaced with just `isParent`.

matthias-ronge:
I've given this some thought, but calling it `isParent()` would mean the property name is `.parent`, and then the reader would probably think it's accessing the parent as an object. That's exactly what I wanted to avoid.

solth:
One question: why do you think bean methods returning booleans should always start with `is`? Yes, EL can evaluate such methods in a way that they can be called from Facelets without the `is` prefix, but that is hardly a reason against `has` methods if those are more appropriate in a specific case. 
All things considered I think the existing method name `hasChildren` is a better fit than `isParenthood` and shouldn't be changed.

matthias-ronge:
I don't think so. It's a lingual bridge.  "get" is like "Give me your …", "is" is like "Do you have …".

What I wanted, is that the expression language values can be accessed through _properties_, not through _functions_. So, to access values of a class that start with "get", you omit the "get" and the parentheses as well. Same with "is". But with "has" you have to write the "has" and you have to remember that you have to put the parentheses here. And if you write `boolean hasParent()` and `process.hasParent`, it doesn't work either. Normally, the property name is equal to the name of the class field, but here it is not. You must use parentheses for this property, but for other properties you don’t have to.

So there are additional inconsistencies that one can always trip over. This change is intended to be able to remove an additional memory requirement from the developer brain. (Eclipse always shows me that as bad style, too.) That’s my vote for this change.

If you don’t like it, so be it.

Kathrin-Huber:
I really don't see the problem in `hasChildren()`

If you are really annoyed by the parentheses, you can rename the funktion to `isHasChildren()`

Sounds crazy, but we are actually doing this, for Example here:
https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/dto/TemplateDTO.java#L82

But I would prefere to just keep it as it is.

PR #4888, opened 2021-12-15
Export completed children on parent export

### LINKED ISSUE #4806: Enable hierarchical top-down export
Currently, the export of hierarchies is initiated from the lower level. Superordinate processes are also exported and, during the export of the superordinate processes, the links mentioned below for processes that have not yet been exported are removed from these. An export of the parent would only export these with the removed links.

**Goal:** When exporting a parent process, all subordinate processes whose workflow has ended but they have not yet been exported [a scenario where the workflow has no export task] are exported first. Then, the superordinate order is exported (once is sufficient) with all the links contained.

matthias-ronge:
Fixes #4806

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/export/ExportDms.java
+++ b/Kitodo/src/main/java/org/kitodo/export/ExportDms.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collection;
+import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -90,16 +96,21 @@ public boolean startExport(Process process) throws DataException {
         boolean wasNotAlreadyExported = !process.isExported();
         if (wasNotAlreadyExported) {
             process.setExported(true);
-            ServiceManager.getProcessService().save(process);
+            processService.save(process);
+        }
+
+        if (!exportCompletedChildren(process.getChildren())) {
```
Kathrin-Huber, line 102: If an error occures on child export, the process is still saved as "exported"

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
```diff
@@ -112,15 +112,8 @@ private HSSFWorkbook getWorkbook() {
             title.createCell(i).setCellValue("");
         }
 
-        HSSFRow rowHeader = sheet.createRow(1);
```
Kathrin-Huber, line 115: what does this change have to do with the issue?

Kathrin-Huber (changes requested):

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
matthias-ronge: Good catch. Needs fixing

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -46,16 +47,21 @@
...
 import org.kitodo.production.services.ServiceManager;
+import org.kitodo.production.services.data.ProcessService;
 import org.kitodo.production.services.file.FileService;
 import org.kitodo.production.services.workflow.WorkflowControllerService;
 
 public class ExportDms extends ExportMets {
     private static final Logger logger = LogManager.getLogger(ExportDms.class);
-    private boolean exportWithImages = true;
-    private final FileService fileService = ServiceManager.getFileService();
+    private static final String COMPLETED = "100000000000";
```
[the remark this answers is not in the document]
henning-gerhardt: Ignore it. As the state of task is calculated on the fly (which could result in bad / slow performance) instead of using the stored value inside the process (value is updated every time a state of task is changed) the here used value is correct.

PR #4889, opened 2021-12-16
#4838 add development profile to kitodo which puts the jsf project st…

### LINKED ISSUE #4838: ViewExpiredException when rendering error.xhtml
I encounter ViewExpiredException when opening a process in the metadata editor. Stack trace is:

```
	at org.kitodo.data.database.beans.User.<init>(User.java:122)
	at org.kitodo.production.services.data.UserService.getCurrentUser(UserService.java:225)
	at org.kitodo.production.controller.SessionClientController.getAvailableClientsOfCurrentUser(SessionClientController.java:148)
	at org.kitodo.production.controller.SessionClientController.userHasOnlyOneClient(SessionClientController.java:61)
	at org.kitodo.production.controller.SessionClientController.getCurrentSessionClientName(SessionClientController.java:47)
	at jdk.internal.reflect.GeneratedMethodAccessor509.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at javax.el.BeanELResolver.getValue(BeanELResolver.java:88)
```
I could find out that this happens when resolving the following expression: `#{SessionClientController.currentSessionClientName}` for which I found only one occurrence, in `navigation.xml`, line 109.

In `UserService.getCurrentUser(UserService.java:225)`:

    public User getCurrentUser() {
        return new User(getAuthenticatedUser());
    }

`getAuthenticatedUser()` returns `null` from `SecurityAccess.getCurrentAuthentication(SecurityAccess.java:139)`:

    protected Authentication getCurrentAuthentication() {
        return SecurityContextHolder.getContext().getAuthentication();
    }

because the security context it picks up doesn’t contain an authentication. The field `authentication` inside it is `null` in the debugger, which is returned.

This happens when I am logged in, and there are subsequent calls to these functions when the page is built, which correctly return the authenticated user.

markusweigelt:
…age on development

When maven profile development is used the JSF PROJECT_STAGE the project stage is set to "Development".

Maven WAR Plugin is updated too btw.

Fixes #4838

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -33,6 +33,7 @@
         <maxAllowedViolations>15</maxAllowedViolations>
         <selenium.version>3.141.59</selenium.version>
         <cargo.plugin.version>1.9.7</cargo.plugin.version>
+        <jsf.projectStage>Production</jsf.projectStage>
     </properties>
 
     <dependencies>
@@ -537,8 +538,9 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-war-plugin</artifactId>
-                <version>3.2.3</version>
+                <version>${maven-war-plugin.version}</version>
                 <configuration>
+                    <filteringDeploymentDescriptors>true</filteringDeploymentDescriptors>
                     <archive>
                         <manifest>
                             <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
@@ -815,6 +817,12 @@
                     </plugin>
                 </plugins>
             </build>
+        </profile>
+        <profile>
+            <id>development</id>
+            <properties>
+                <jsf.projectStage>Development</jsf.projectStage>
+            </properties>
         </profile>
     </profiles>
 </project>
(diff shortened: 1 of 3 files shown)


henning-gerhardt (approved):
Changes are okay. 

Change of `FACELETS_BUFFER_SIZE` to 5 MB should be okay too but must be maybe changed if this value is to big. Default value is 1024.

solth:
Works like a charm! 👍

matthias-ronge (commented):
I have to abstain here. I don’t know how Eclipse internally handles the debug mode. So I cannot say, if I test this, if the changes have any effect under Eclipse. I’m sorry.

markusweigelt:
@matthias-ronge Behaviour of this improvement is independent from your IDE.  Maven uses the "Development" Profile by default. For example when you run "mvn clean install" then the web.xml is adjusted during generation and the project stage is set to "Development". This causes the error messages from Java Server Faces to be displayed.

PR #4897, opened 2022-01-04
allow editing of aktive workflows

Kathrin-Huber:
As discussed in the "Anwendertreffen" in November, active workflows should be editable.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/projects/workflowList.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/projects/workflowList.xhtml
@@ -59,7 +59,7 @@
                             outcome="workflowEdit"
                             styleClass="action"
                             title="#{msgs['view']}"
-                            rendered="#{(item.status ne 'draft') and ((SecurityAccessController.hasAuthorityToViewWorkflow() and not SecurityAccessController.hasAuthorityToEditWorkflow()) or (SecurityAccessController.hasAuthorityToEditWorkflow()))}">
+                            rendered="#{(SecurityAccessController.hasAuthorityToViewWorkflow() and not SecurityAccessController.hasAuthorityToEditWorkflow())}">
                         <h:outputText><i class="fa fa-eye fa-lg"/></h:outputText>
                         <f:param name="id" value="#{item.id}"/>
                     </h:link>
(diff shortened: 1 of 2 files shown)


solth:
Perhaps that's only an error in the description as I am not sure about what exactly was decided at the Anwendertreffen (i thought it was about editing _templates_, not _workflows_, but could be wrong), but workflow **drafts** where always editable, weren't they?

Kitodo/src/main/webapp/WEB-INF/templates/includes/projects/workflowList.xhtml
```diff
@@ -59,7 +59,7 @@
                             outcome="workflowEdit"
                             styleClass="action"
                             title="#{msgs['view']}"
-                            rendered="#{(item.status ne 'draft') and ((SecurityAccessController.hasAuthorityToViewWorkflow() and not SecurityAccessController.hasAuthorityToEditWorkflow()) or (SecurityAccessController.hasAuthorityToEditWorkflow()))}">
+                            rendered="#{(SecurityAccessController.hasAuthorityToViewWorkflow() and not SecurityAccessController.hasAuthorityToEditWorkflow())}">
                         <h:outputText><i class="fa fa-eye fa-lg"/></h:outputText>
```
markusweigelt, line 63: Is this change correct? Previously users with the authority of editing workflow could view workflow too.

markusweigelt (changes requested):

Kathrin-Huber:
> Perhaps that's only an error in the description as I am not sure about what exactly was decided at the Anwendertreffen (i thought it was about editing _templates_, not _workflows_, but could be wrong), but workflow **drafts** where always editable, weren't they?

you are right, I meant "active workflows should be editable" I changed the PR description.

And we just discussed the editing of workflows, but of course, to make them used in the template, the templates should be editable too. Should I add this change?

solth:
> Should I add this change?

No, that can be done in a separate PR.

solth (commented):
@markusweigelt I think the situation is slightly different. _Processes_ (and their tasks) that have been created using a specific template (e.g. blueprint) and therefore a specific workflow should not be changed when editing that (active) workflow. IIRC that is what the community decided (that would also fit your description of already "started" tasks - because only process tasks _can_ be started, but not template tasks!). 
For templates, the situation is slightly different: changing workflow steps means the tasks of templates using this workflow _need_ to be updated accordingly, otherwise changing a workflow would not make much sense, would it? The user would still need to create a new template to make use of the adjusted workflow, so adjusting active workflow would not be simplified.

@Kathrin-Huber So I think just displaying the "Save" button is insufficient to solve the issue raised by the community and that allowing the user to edit active workflows has more ramifications that need to be adressed and that are not covered by the few changes in this pull request.

PR #4901, opened 2022-01-05
Mass import

IkramMaalej:
Redesign massImport 
  - Combining the two tabs 'file MassImport' and 'Text MassImport' in one tab.
  - Add metadataTable in order to offer the possibility to add more Metadata to the imported metadata.
 
 Part of #4322


Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessBooleanMetadata.java
```diff
@@ -72,15 +72,20 @@ public String getInput() {
...
+    }
+
+    @Override
+    public Collection<Metadata> getMetadata(boolean skipEmpty) throws InvalidMetadataValueException {
         if (!isValid()) {
             throw new InvalidMetadataValueException(label, settings.convertBoolean(active).orElse(""));
         }
         Optional<String> value = settings.convertBoolean(active);
-        if (value.isPresent()) {
+        if (!skipEmpty || value.isPresent()) {
```
markusweigelt, line 84: Suggest to return early here so you do not need the else.

```
if (skipEmpty || !value.isPresent()) {
}

MetadataEntry entry = new MetadataEntry();
entry.setKey(settings.getId());
entry.setDomain(DOMAIN_TO_MDSEC.get(settings.getDomain().orElse(Domain.DESCRIPTION)));
entry.setValue(value.get());
entry.setValue(value.isPresent() ? value.get() : "");
return Collections.singletonList(entry);
```

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -584,4 +596,56 @@ public int getProgress() {
...
+     *
+     * @param metadataNode TreeNode for which the check is performed
+     * @return whether given TreeNode contains ProcessFieldedMetadata and if any further metadata can be added to it
+     */
+    public boolean metadataAddableToGroup(TreeNode metadataNode) {
+        if (metadataNode.getData() instanceof ProcessFieldedMetadata) {
+            return !(DataEditorService.getAddableMetadataForGroup(getMainProcess().getRuleset(), metadataNode).isEmpty());
+        } else {
+            return false;
+        }
```
markusweigelt, line 641: Directly return false without the else.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessBooleanMetadata.java
```diff
@@ -72,15 +72,20 @@ public String getInput() {
 
     @Override
     public Collection<Metadata> getMetadata() throws InvalidMetadataValueException {
+        return getMetadata(true);
```
markusweigelt, line 75: name of method doesn't quite hit it "getMetadataWithoutEmpty" or something else would be better

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -592,6 +592,7 @@ messages=Messages
 metadata=Metadata
 metadataCorrupt=The metadata file is faulty, please inform the administrator.
 metadataDelete=Delete metadata
+metadataDeleteNotAllowed=Metadata is set as'Required' oder "alwaysShowing" in the ruleset
 metadataDoesNotEndWith=not end with the value
```
markusweigelt, line 596: From the user's point of view, a little more context may be required, for example "The deletion of the metadata is not possible because metadata is set as 'Required' oder "alwaysShowing" in the ruleset.

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -580,6 +580,7 @@ messages=Nachrichten
 metadata=Metadaten
 metadataCorrupt=Die Metadatendatei ist fehlerhaft, bitte informieren Sie ihren Administrator.
 metadataDelete=Metadaten l\u00F6schen
+metadataDeleteNotAllowed=Metadaten ist in Regelsatz als 'Required' oder "alwaysShowing" konfiguriert
 metadataDoesNotEndWith=endet nicht mit dem Wert
```
markusweigelt, line 584: Suggest change with more context see en properties

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessBooleanMetadata.java
```diff
@@ -71,16 +71,21 @@ public String getInput() {
...
         Optional<String> value = settings.convertBoolean(active);
-        if (value.isPresent()) {
+        if (!skipEmpty || value.isPresent()) {
             MetadataEntry entry = new MetadataEntry();
             entry.setKey(settings.getId());
             entry.setDomain(DOMAIN_TO_MDSEC.get(settings.getDomain().orElse(Domain.DESCRIPTION)));
-            entry.setValue(value.get());
+            entry.setValue(value.isPresent() ? value.get() : "");
             return Collections.singletonList(entry);
         } else {
```
markusweigelt, line 90: "else" is not needed here either. Directly return Collections.emptyList();

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -842,19 +844,32 @@ public String getReferringView() {
...
+        } else if (!Objects.equals(metadataPanel.getSelectedMetadataTreeNode(), treeNode.getParent())
+                || Objects.isNull(addMetadataDialog.getAddableMetadata())) {
+            prepareAddableMetadataForGroup(treeNode.getParent());
         }
-        else {
-            return true;
+        if (Objects.nonNull(addMetadataDialog.getAddableMetadata())) {
+            return addMetadataDialog.getAddableMetadata().stream()
+                    .map(SelectItem::getValue).collect(Collectors.toList()).contains(((ProcessDetail) treeNode.getData()).getMetadataID());
+        } else {
```
markusweigelt, line 862: No need of else ;)

markusweigelt (changes requested):
From my point of view, code is save except for the small changes and Codacy. However, I still have to test it in the application.

Kitodo/src/main/webapp/pages/massImport.xhtml
```diff
@@ -24,27 +24,136 @@
...
+                                               itemLabel="#{step.localizedLabel}"
+                                               itemValue="#{step.title}"/>
+                                <p:ajax update="@form"/>
+                            </p:selectOneMenu>
+                        </div>
+                        <h:panelGroup
+                                rendered="#{MassImportForm.importMode.equals('file') and MassImportForm.selectedCatalog ne null}">
+                            <div>
+                                <p:fileUpload value="#{MassImportForm.file}"
+                                              fileUploadListener="#{MassImportForm.handleFileUpload}"
```
solth, line 94: I think this needs to be updated to the Primefaces 8 syntax (see #4909)

solth (changes requested):

solth:
metadata-button-design

This vertical offset between input fields and command buttons could probably be fixed with something combination of CSS rules `display: inline-block` and `vertical-align: text-bottom` for the buttons.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessSelectMetadata.java
```diff
@@ -112,11 +112,18 @@ public String getInput() {
...
+    }
+
+    @Override
+    public Collection<Metadata> getMetadata(boolean skipEmpty) throws InvalidMetadataValueException {
         int items = selectedItems.size();
         Collection<Metadata> metadata = new HashSet<>((int) Math.ceil(items / .75));
         String key = settings.getId();
         MdSec domain = DOMAIN_TO_MDSEC.get(settings.getDomain().orElse(Domain.DESCRIPTION));
+        selectedItems.removeAll(Collections.singletonList(""));
+        selectedItems.removeAll(Collections.singletonList(null));
```
markusweigelt, line 126: removeAll should be called only once and with collection containing empty string and null for example or use with filter of stream.

BartChris:
I created two other issues regarding mass import today. I do not know if you want to adress them in this pull request as well, but maybe you want to tackle those problems as well while you are at it.
I leave theme here for reference

https://github.com/kitodo/kitodo-production/issues/4988
https://github.com/kitodo/kitodo-production/issues/4989

solth:
> I created two other issues regarding mass import today. I do not know if you want to adress them in this pull request as well, but maybe you want to tackle those problems as well while you are at it.
> I leave theme here for reference

I would opt to solve those issues in separate pull requests because this one is already huge enough as it is and shouldn't be delayed further.

The release management also prefers many small pull requests over few large ones, AFAIK.

Issue #4461, opened 2021-06-08
TSL/ATS calulation does not work properly


stefanCCS:
According to 'expectation' and source code here
https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/process/TitleGenerator.java#L103
I would expect, that TSL/ATS is build as combination of Author and Title.
My document (PPN 1022427806) has both, as you can see here
**Haupttitel:**
![grafik](https://user-images.githubusercontent.com/61045115/121173598-4928fe80-c859-11eb-9d70-faace447c45c.png)
**Verfasser:**
![grafik](https://user-images.githubusercontent.com/61045115/121173684-6231af80-c859-11eb-9d21-f445e23a38cc.png)

==> But, TSL/ATS is created out of Title only (after pressing "Zahnrad":
![grafik](https://user-images.githubusercontent.com/61045115/121173747-75447f80-c859-11eb-97b5-68aa833f1b1f.png)

==> Please check, if there is something wrong (in my assumptions or in source code)

matthias-ronge:
Maybe related: #3600

Kathrin-Huber:
@stefanCCS can this be closed?

andre-hohmann:
@henning-gerhardt : 

I think, here is the problem, that not the "old" author fields (without metadata groups) can be used for the generation of ATS (author title schlüssel). 
Due to the implementation of the metadata groups for personal and cooperation, the new fields must be assessed to create the ATS correctly according to the former times. That could be difficult, because the fields can be defined very flexible. 

It could discussed, if a division in TSL and ATS is still necessary, or if the creation of TSL without information of the author is sufficient.

henning-gerhardt:
@andre-hohmann : In 2.x the following  behavoiur is implemented: if there is no author given then the tls/ats is only build by title information. If there is an author a part of the tls/ats is builded with this author information. There is no requirement from where the title nor author is comming. But in this issue and maybe in the related: no tls/ats is build at all! So its looks for me that this method is not called or called with an empty title and author.

andre-hohmann:
@henning-gerhardt: 
I think the TSL is generated, but without author-information: 

> But, TSL/ATS is created **out of Title only** (after pressing "Zahnrad"

The following issues must be examined seperately, escpecially, because existing functionality have been los.  
- #5207 
- #5208

stefanCCS:
> @stefanCCS can this be closed?

[...]

PR #4905, opened 2022-01-07
Generate automaticallly 'process title', 'tiff header' and 'TSL/ATS' …

### LINKED ISSUE #3600: TSL/ATS is formed too late
When you import a process and then click _save_, (given a corresponding ruleset), TSL/ATS is blamed as missing metadata entry, and missing process title is reported, too. If you then click _generate process title_ button, TSL/ATS is secretly generated and added to the metadata list. You can click _save_ without errors. This is confusing behavior.

Screenshot

**Goal:** Apparently, the application is able to form TSL/ATS automatically. It should form and use it during the import, so that the first attempt to save would only complain about missing process title.

IkramMaalej:
…after importing a process

Part of #4322
Fixes #3600

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/MetadataImportDialog.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/MetadataImportDialog.java
@@ -118,6 +118,7 @@ public abstract class MetadataImportDialog {
                     && firstProcess.getMetadataNodes().getLength() > 0) {
                 createProcessForm.getProcessMetadata().initializeProcessDetails(
                         firstProcess.getWorkpiece().getLogicalStructure());
+                this.createProcessForm.getProcessDataTab().generateProcessTitleAndTiffHeader();
             }
         }
     }


andre-hohmann:
- #3600

I think, the problem is solved: The fields "Process title" and "Tiff header document name" are filled immediately after the metadata import and the process title does not need to be generated manually anymore. 

- #4461

I think the problem is not solved. The value of the field author is not regarded in the generation of the TSL/ATS. Probably the reason is the application of metadata groups instead of one author field. 
However, i am not sure, what the goal should be. Should the code be adjusted that only the field title is regarded, or should the code be adjusted, that also the fields of metadata groups for persons, are regarded for the generation of TSL/ATS?

I think the metadata groups should be regarded. But when the fields of the metadata groups are not standardized, it is difficult to name the fields. 

Maybe the author of the issue can describe the expected result.

matthias-ronge:
There is the `use` attribute in the ruleset to define the fields necessary for ATS/TSL creation. The title should be a `<key id="…" use="title">`, the author’s last name should be a `<key id="…" use="authorLastName">` (and I did expected this to be possible in a group like `Person`), and the result should be written to the `<key id="…" use="processTitle">`. This is part of the ruleset definition ([XSD file](/kitodo/kitodo-production/blob/master/Kitodo/rulesets/ruleset.xsd#L421-L436)), but at the moment, the ruleset is not yet taken fully into acount, but some fields are hard-coded.

IkramMaalej:
I have removed this one  [#4461 ](url) because it is not yet fixed. 
For now only this ruleset configuration `<key id="…" use="processTitle">` has been considered in the code base to generate the process Title and  ATS/TSL. 
In addition to that, I found out that if the `Role="Author"` is assigned to a metadata group `person`, then this latter will be also considered for the generation of ATS/TSL.

PR #4906, opened 2022-01-12
Add functional metadata for 'docType' metadata

### LINKED ISSUE #3389: Misleading error if import returns no docType
If import returns a field `docType`, but the ruleset does not know any field `docType`, then the field is displayed with yellow marking and no error message appears. (So it is correct.) However, if *import does not return* a field `docType`, then an error message says that there is no field docType *in the rule set.*

**Goal:** Clarify error message that error is in import XSLT, not tin the ruleset.

### LINKED ISSUE #4863: Hard-coded metadata docType is used
When creating a process, a hard-coded metadata `docType` is searched for. If it is not found, an error message will appear:

Screenshot 2021-12-01 103953

Once the aim was to stop using hard-coded metadata. For metadata to which a function is linked, there is the option of assigning `use="…"`-attributes in the ruleset.

**Goal:** This should be done for `docType`. If the `use="docType"` attribute was not assigned in the ruleset, no error message should appear stating that it was not found. It should be documented how exactly the metadata with the `use="docType"` is interpreted and is to be populated.

solth:
Fixes #4863 

**Note**: if this pull request is merged, mapping files and rulesets must be updated to include a metadata with `use="docType"` in order for import to continue to work!

_Edit_: also fixes #3389, I think


Kitodo/src/main/java/org/kitodo/production/helper/TempProcess.java
```diff
@@ -111,17 +113,23 @@ public NodeList getMetadataNodes() {
...
-                    this.getWorkpiece().getLogicalStructure().setType(docType);
+    public void verifyDocType() throws IOException, ProcessGenerationException {
+        if (Objects.nonNull(process.getRuleset())) {
+            List<String> doctypeMetadata = ImportService.getDocTypeMetadata(process.getRuleset());
+            if (doctypeMetadata.isEmpty()) {
+                throw new ProcessGenerationException("No doc type metadata defined in ruleset!");
+            }
+            if (Objects.nonNull(this.getWorkpiece().getLogicalStructure().getMetadata())) {
+                Optional<Metadata> docTypeMetadata = this.getWorkpiece().getLogicalStructure().getMetadata()
+                        .stream().filter(m -> m.getKey().equals(doctypeMetadata.get(0))).findFirst();
```
matthias-ronge, line 126: Please rename `m` variable to a full name.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -1206,4 +1211,19 @@ public Process importProcess(String ppn, int projectId, int templateId, String s
...
     }
+
+    /**
+     * Load doc type metadata keys from provided ruleset.
+     * @param ruleset Ruleset from which doc type metadata keys are loaded and returned
+     * @return list of Strings containing the IDs of the doc type metadata defined in the provided ruleset.
+     * @throws IOException thrown if ruleset file cannot be loaded
+     */
+    public static List<String> getDocTypeMetadata(Ruleset ruleset) throws IOException {
+        RulesetManagementInterface rulesetManagementInterface
```
matthias-ronge, line 1222: Please don’t use `Interface` suffix on a variable name. I.e., `RulesetManagementInterface rulesetManagement = `…

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -1206,4 +1211,19 @@ public Process importProcess(String ppn, int projectId, int templateId, String s
...
+                = ServiceManager.getRulesetManagementService().getRulesetManagement();
+        String rulesetDir = ConfigCore.getParameter(ParameterCore.DIR_RULESETS);
+        String rulesetPath = Paths.get(rulesetDir, ruleset.getFile()).toString();
+        rulesetManagementInterface.load(new File(rulesetPath));
+        return new ArrayList<>(rulesetManagementInterface.getFunctionalKeys(FunctionalMetadata.DOC_TYPE));
```
matthias-ronge, line 1227: There is no need to copy the result to another list.

    return rulesetManagementInterface.getFunctionalKeys(FunctionalMetadata.DOC_TYPE);

To use along with `contains()` as I mentioned above.

matthias-ronge (changes requested):
Please also update the example rulesets in `Kitodo/rulesets` directory. As I understand your comment, this is necessary. Or isn’t it?

solth:
@matthias-ronge thank you for the review. I implemented most of your requested changes, apart from the `<p>` tag in the java doc, because this was specifically how Codacy accepted it.

matthias-ronge (approved):

PR #4908, opened 2022-01-12
use delete_symLink only if it's neccessary

### LINKED ISSUE #4622: Delete symlink script is called on closing tasks
The script `script_deleteSymLink.sh` get called every time a task is closed. This is only necessary if the task property `write images` is set as then on taking over this task the script `script_createSymLink.sh` is called and the symbolic link between the images directory of the process into the user home directory is done

Kathrin-Huber:
fixes #4622

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -2419,15 +2419,12 @@ public class ProcessService extends ProjectSearchService<Process, ProcessDTO, Pr
      * @param task Task for which symlinks are removed
      */
     public static void deleteSymlinksFromUserHomes(Task task) {
-        WebDav webDav = new WebDav();
-
-        for (Role role : task.getRoles()) {
-            for (User user : role.getUsers()) {
-                try {
-                    webDav.uploadFromHome(user, task.getProcess());
-                } catch (RuntimeException e) {
-                    Helper.setErrorMessage(e.getLocalizedMessage(), logger, e);
-                }
+        if (task.isTypeImagesWrite()) {
+            WebDav webDav = new WebDav();
+            try {
+                webDav.uploadFromHome(task.getProcessingUser(), task.getProcess());
+            } catch (RuntimeException e) {
+                Helper.setErrorMessage(e.getLocalizedMessage(), logger, e);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
```diff
@@ -211,15 +211,13 @@ public void closeTaskByUser(Task task) throws DataException, IOException, DAOExc
                 if (!mih.checkIfImagesValid(task.getProcess().getTitle(), imageFolder)) {
                     throw new DataException("Error on image validation!");
                 }
+                this.webDav.uploadFromHome(task.getProcess());
```
henning-gerhardt, line 214: This move is wrong as the method is only called when the task should be verified on closing and the images could be modified.

I would recommend to 
- not move the call to this place
-  on original place wrap the call with a test to `task.isTypeImagesRead() || task.isTypeImagesWrite()`

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -2419,15 +2419,12 @@ private static void deleteMetadataDirectory(Process process) {
...
-        WebDav webDav = new WebDav();
-
-        for (Role role : task.getRoles()) {
-            for (User user : role.getUsers()) {
-                try {
-                    webDav.uploadFromHome(user, task.getProcess());
-                } catch (RuntimeException e) {
-                    Helper.setErrorMessage(e.getLocalizedMessage(), logger, e);
-                }
+        if (task.isTypeImagesWrite()) {
```
henning-gerhardt, line 2422: The test must be extended to a check if `task.isTypeImagesRead()` as even there a symlink is created but only a read-only symlink.

henning-gerhardt (changes requested):
There must be at least one more (or maybe more) calls to `webDav.uploadFromHome()` which are executed independent if the task has isTypeImagesRead or isTypeImagesWrite.

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
matthias-ronge: By the way, @Kathrin-Huber, could you maybe rename the method? Nothing is uploaded there. No idea how, but that name doesn't make any sense.

henning-gerhardt: Not only the method even the whole class WebDav did not make this what the name is suggesting. Renaming should be done after this pull request is merged.

Issue #4738, opened 2021-10-12
Enable corrections of volume-processes in the metadata editor


andre-hohmann:
### Problem
In the metadata editor the processes of volumes cannot be identified (only the label of the structure element is shown) and the order of the processes cannot be corrected. 
Theoretically, the order should be correct, when the processes are created correctly. Practically, mistakes will happen and it  should be possible to correct the mistakes. Otherwise, the incorrect process has to be deleted and created again. 

### Solution

<del>1. In the metadata editor the processes of volumes should be labeled with more information about the order. For example according to `use="displaySummary"` for the list in "Title record link" - see #4736. </del>
2022-09-01 (@andre-hohmann): Now, the process titles are shown - see comment below. 

2. In the metadata editor, it should be possible to correct the order processes of volumes manually. 

It is not the goal of this issue to enable the editing of the metadata of the volume processes via the parent process! It is just about the opportunity to <del> identify and </del>  order the volume processes. 

### Example
![BändeSortierung01](https://user-images.githubusercontent.com/14835770/136900992-7a00a6d9-e783-4bf0-9ea3-ecaf2962ff45.jpg)

andre-hohmann:
This could be solved with: 
- #4850

andre-hohmann:
Example, how it looks like, when the process title is shown: 

![exampleProcesstitle](https://user-images.githubusercontent.com/14835770/187910115-0e3578f5-ef9d-4be3-81e2-49be7115a757.jpg)

PR #4914, opened 2022-01-18
Show id and title of linked processes in metadata editor

solth:
Currently only the document type of linked processes is displayed in the metadata editor. That can become quite confusing and does not help a lot when multiple child processes of the same type are linked to one parent process, because it is not obvious which child node in the structure tree represents which child process, and thus, the order of those child processes (and potential pages in the parent process) cannot be validated/checked:

Bildschirmfoto 2022-01-18 um 13 12 21

The change in this pull request displays additional information about the linked processes, namely it's ID and process title, similar to what is displayed in when linking title records during metadata import. 
This helps identifying specific child processes and structuring linked processes and pages within a parent process:

Bildschirmfoto 2022-01-18 um 13 06 53

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -147,7 +147,7 @@ public class StructurePanel implements Serializable {
     }
 
     void deleteSelectedStructure() {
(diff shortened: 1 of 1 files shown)


andre-hohmann:
Thanks a lot, that is a really good improvement!

Maybe it could also be regarded, that the position of an issue might need be corrected, if the order is not correct.  
- #4738

andre-hohmann:
Good news - thanks a lot!

PR #4917, opened 2022-01-19
Add buttons to clear shortcuts

### LINKED ISSUE #4767: Shortcuts cannot be removed
When a shortcut is set, it is not possible to remove this shortcut.
Using the `delete` key to clear the input field results in a new shortcut containing the `delete` key.
Each input field should have a small button next to it to clear the input.

oliver-stoehr:
Adds a button for every shortcut to clear the corresponding input field.
Resolves #4767.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/shortcuts.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/shortcuts.js
@@ -21,6 +21,7 @@ let shortcuts = {
             keyCombination += event.originalEvent.code;
             event.originalEvent.target.value = keyCombination;
             event.preventDefault();
+            $(event.originalEvent.target).change();
         });
     }
 };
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/webapp/WEB-INF/resources/js/shortcuts.js
```diff
@@ -21,6 +21,7 @@ let shortcuts = {
             keyCombination += event.originalEvent.code;
             event.originalEvent.target.value = keyCombination;
             event.preventDefault();
+            $(event.originalEvent.target).change();
```
solth, line 24: What does this line do?

solth (approved):

Kitodo/src/main/webapp/WEB-INF/resources/js/shortcuts.js
oliver-stoehr: It dispatches a `change` event for the input field. This triggers the primefaces ajax update.

PR #4919, opened 2022-01-20
Bring back support for Active Directory

### LINKED ISSUE #4807: Can't authenticate via Ldap to a Windows Active Directory
By using firstname and lastname in the DN (LDAP-Gruppe->User DN-> cn={firstname} {lastname},ou=Benutzer,ou=Kitodo,...) 
it will be impossible to authenticate against a Windows Active Directory.

It seem's that the authentication using not the classes
org/kitodo/production/services/data/LdapServerService.java
but the "old" ldap classes 
org/kitodo/production/security/DynamicAuthenticationProvider.java
which not using replacing for firstname, lastname !?

One question is, is there a solution to configure the User DN, so that AD will accept them ?
If there is no trick to configure, why will not be used the function authentication in LdapServerService ?
Why exist the authentication (likewise code) twice ?

If there is no way to authenticate against the AD how about this solution to include the missing replacements in the 
function  authenticate(Authentication authentication) of DynamicAuthenticationProvider.java ?
##
95c95,98
<             configureAuthenticationProvider(ldapGroup.getLdapServer().getUrl(), ldapGroup.getUserDN());
---
>           String userDN = ldapGroup.getUserDN();
>             userDN = userDN.replaceAll("\\{lastname\\}", user.getSurname());
>             userDN = userDN.replaceAll("\\{firstname\\}", user.getName());
>             configureAuthenticationProvider(ldapGroup.getLdapServer().getUrl(), userDN);
##

matthias-ronge:
Fixes #4807

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
+++ b/Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
@@ -92,6 +92,9 @@ public class DynamicAuthenticationProvider implements AuthenticationProvider {
                 throw new AuthenticationServiceException("No LDAP server specified on user's LDAP group");
             }
             configureAuthenticationProvider(ldapGroup.getLdapServer().getUrl(), ldapGroup.getUserDN());
+            if (ldapGroup.getUserDN().contains("{ldaplogin}")) {
+                authentication = new UsernamePasswordAuthenticationToken(user.getLdapLogin(), authentication.getCredentials());
+            }
             return ldapAuthenticationProvider.authenticate(authentication);
         } else {
             return daoAuthenticationProvider.authenticate(authentication);
@@ -152,7 +155,7 @@ public class DynamicAuthenticationProvider implements AuthenticationProvider {
     }
 
     private String[] convertUserDn(String userDn) {
-        return new String[] {userDn.replaceFirst("\\{login}", "{0}") };
+        return new String[] {userDn.replaceFirst("\\{(?:ldap)?login}", "{0}") };
     }
 
     /**


Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
```diff
@@ -92,6 +92,9 @@ public Authentication authenticate(Authentication authentication) {
                 throw new AuthenticationServiceException("No LDAP server specified on user's LDAP group");
             }
             configureAuthenticationProvider(ldapGroup.getLdapServer().getUrl(), ldapGroup.getUserDN());
+            if (ldapGroup.getUserDN().contains("{ldaplogin}")) {
+                authentication = new UsernamePasswordAuthenticationToken(user.getLdapLogin(), authentication.getCredentials());
+            }
```
henning-gerhardt, line 97: Why is this change needed? I tested this with a modified user which contains the same values in `login` and `ldaplogin`, the user dn contain the `ldaplogin` placeholder and this user could be authenticated without this change.

henning-gerhardt (changes requested):

Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
matthias-ronge: Yes, of course, if the `ldaplogin` and the `login` contain the same value, it doesn't matter. However, the functionality of the `ldaplogin` is precisely that you can specify a _different_ name here, which is sent to the LDAP _instead of the login_. In the case of a login with a Microsoft Active Directory, this mus usually be “‹Firstname› ‹Lastname›”.

henning-gerhardt: Okay. But then I can not verify is this change correct or not nor I can not say if this change is working for the given use case.

matthias-ronge: You can, if you put a fake username (like "foobar") in the `login`, and your login name in the `ldapLogin`. You will see that you can log in with `foobar` and your password.

BTW, it fixed the problem at our customer’s site, who use it productively.

PR #4924, opened 2022-01-23
Improve newspaper creation

### LINKED ISSUE #3305: Make calendar editor change year on 'enter' press
In the calendar editor, if you type another year in the year box above the calendar and press <kbd>Enter</kbd>, the calendar editor wants to start creating processes.

**Goal:** If the cursor is in this box and you press <kbd>Enter</kbd>, the view should navigate to the year you entered. No processes should be created yet.

### LINKED ISSUE #3593: Selection list missing by adding metadata for newspaper issues
### Problem
If metadata is captured by creating the processes for newspaper issues, the configured selection lists of the ruleset are missing, which are offered by the creation of processes for monographs, manuscripts, ...
The correct capture of the metadata of the selection list is important, because otherwise further processing of it is not possible, as for example the correct assignment of the digital collection or the rights information. 
Especially, when several hundreds of processes are created, the correction of the metadata is quite extensive. 

### Solution
The configured selection lists need to be offered for the creation of processes for newspaper issues. 

For example by adding a step with the regular template (process metadata) of creating processes before the calender view is shown. There, the configured metadata can be selected. 

### Example
The expected collection name "_Abschlussarbeiten des Instituts für Literatur ‚Johannes R. Becher’ Leipzig_" has to be retrieved and copied. Furthermore, it is difficult to check it, because it is not shown completely. Errors will definitely occur.

newspapermetadata01

### LINKED ISSUE #3864: Calendar editor missing confirmation when deleting blocks
If a block is deleted in the calendar editor, no query appears.

**Goal:** Question box with message `calendar.block.remove.query` (_Do you want to delete the episode?_) should be shown.

### LINKED ISSUE #3974: Creation of processes for newspaper issues: Depiction of button "Add metadata"
### Problem
If metadata is captured for the newspaper issues, the button "Add metadata" is hidden after the first metadata was captured. It has to be scrolled down, until the button is shown again. 

### Solution
The button "Add metadata" should be depicted next to the button "close", which is always shown. 

### Examples
The button is not shown
issue01

After scrolling down, the button is shown
issue02

The button should always be shown
issue04

### LINKED ISSUE #4759: Creating newspaper processes - text fields are covered when browser windows is zoomed
### Problem
The text fields to capture the issue information are covered, when the browser window is zoomed. 

### Solution
The text fields to capture the issue information should always be shown. 

### Examples
ZeitungenÜberdeckungAusgabenanzeige

### LINKED ISSUE #4761: Creating newspaper processes - wrong color-labeling of issues
### Problem
When more then one newspaper issue per day is created, an own color is assigned to each issue-block. 
As soon as the text-label of the issues per day and the amount of issues per each day are the same, the color-label of one issue changes - and only one issue is recognized. In the end, only one issue block is created. 

This behaviour is difficult to describe - see the example to figure it out easier. 
This approach should not be applied in order to avoid duplicates. Maybe this is a kind of duplicate check. 
In case that this behaviour influences other behaviour, it is described in this issue. 

### Solution
The color-label of the issues should not be changed automatically. To avoid duplicates, an error message should be shown - #4712. Otherwise the user cannot interpret it. 

### Example 
ZeitungenAusgabenanzeigeHinzufügen

### LINKED ISSUE #4900: Duplicates are created during creation of newspaper-processes - 2
### Problem
When newspaper-processes are created with the calendar-plugin, the created processes are not examined regarding existing processes with the same processtitle. 

With the issue #4712 it no duplicates are created, when two issues of the same name are created in the same calendar. 
However, it is still not checked, if a process with the same processtitle already exists. See example below. 

### Solution
When newspaper-processes are created, all processtitles - also the existing ones - should be checked for duplicates, like during the creation of processes for other document types (monograph, volume). 

If one processtitle exists, the creation of the processes should be aborted. If possible, the labels of the duplicate processes should be shown. See also comments in the issue #4712. 

### Examples
duplicate01

IkramMaalej:
Part of #4322
Fixes #3305
Fixes #3593 
Fixes #3864
Fixes #3974
Fixes #4761 
Fixes #4759
Fixes #4900


Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/metadata/CountableMetadata.java
```diff
@@ -259,9 +269,29 @@ public void setStepSize(Granularity stepSize) {
...
+    }
+
+    /**
+     * Sets metadataDetail.
+     *
+     * @param metadataDetail value of metadataDetail
+     */
+    public void setMetadataDetail(ProcessDetail metadataDetail) {
+        this.metadataDetail = metadataDetail;
+        this.metadataType = this.metadataDetail.getLabel();
```
solth, line 292: Doesn't `getLabel` return a value translated to the language that is currently selected? So that wouldn't work as a "type", would it?

solth (commented):
Fixes #3305 :white_check_mark:
Fixes #3593 :white_check_mark:
Fixes #3864 :white_check_mark:
Fixes #3974 ❌ 
If I am not mistaken, the "Add metadata" buttons of the individual issues are still not fixed and leave the current view when scrolling through the metadata, don't they?

Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/metadata/CountableMetadata.java
IkramMaalej: Yes, I correct it. It should be `this.metadataType = this.metadataDetail.getMetadataID();`

IkramMaalej:
@andre-hohmann : Can you please test this issue #4900  again to see if it has been fixed in meantime

solth:
> Fixes #4761

✅ 

> Fixes #4759

✅ 

@andre-hohmann I added these changes to the current preview system. Could you check whether they fix #3974 in your opinion or not?

andre-hohmann:
@IkramMaalej : #4900 has not been fixed. It is still possible to create newspaper processes with existing process titles: 

DublettenZeitungen02

DublettenZeitungen

andre-hohmann:
@solth : #3974 is not fixed. It is necessary to scroll down to the new metadata element and then back up to the button "Metadatum hinzufügen" to create the next metadata element. 
The button "Metadatum hinzufügen" should be shown always. For example next to the button "Schließen" or next to "Mo. 2010-01-04 (Block 1)" . See the gif below. 
As the metadata element is created for all issues, there is no need for to buttons "Metadatum hinzufügen". 

Is is not possible to show the metadata in a list as in the regular creation of processes? The button "Als letzte Ausgabe festlegen" could be added to each metadata field. It would then not be necessary to add all needed metadata elements manually and to chose the fields and values, ...
It would also be helpful, if the window was bigger. 

Sorry, if i am only complaining, ... It is however great, that the standard-values of the ruleset are implemented. 

ZeitungMetadaten

solth:
> As the metadata element is created for all issues, there is no need for to buttons "Metadatum hinzufügen".

I am not sure, but I thought it's possible to add metadata to individual issues, hence the need for multiple buttons. 
See @oliver-stoehr's comment from the linked issue:
> The button was placed inside the scrollable area by design, because there can be multiple "add metadata" buttons.
When multiple issues occur on the same date, every issue has its own button to add metadata to this specific issue.
(The buttons are placed at the end bottom of the corresponding issue's area.)

oliver-stoehr:
That's right. The metadata is added to all issues _between the two specified dates and issues_ (e.g. "2010-01-04 2-Spätausgabe" and "2010-03-04 2-Spätausgabe"). If this specified beginning is for example the second issue on this date, the metadata will not be added to the first issue "Frühausgabe" (being outside the specified range).

andre-hohmann:
@solth , @oliver-stoehr : I do not really understand the current state. When i use the button "Metadatum hinzufügen" as shown in the gif above, the element is added to both issues. 
In my opinion, this is very good, because otherwise each metadata element has to be twice (if two issues will be created). That would be very cumbersome. 

My question is now: Is my observation incorrect? Is it not possible to move the button "Metadatum hinzufügen" to another position?

andre-hohmann:
@solth : Here two general hints: 
1. The Button "Ausgabe hinzufügen" is not striking enough. I click often "Block löschen" and "Block kopieren", because they are more striking due to the color. I use "Ausgabe hinzufügen" more often then "Block kopieren", ...
2. The field "Name" in the Issue-Blocks is mandatory and a text has to be inserted. When only one issue per day exists, we do not add a title. Therefore the field "Name" should not be mandatory - it must be possible to create newspaper processes without title.

matthias-ronge:
Maybe the dialog needs a more concise design:

dialog

( the blue arrow should be a pencil )

andre-hohmann:
@IkramMaalej: Thanks a lot for the improvements! 

### Summary
- #3305 -> is fixed
- #3593 -> is fixed
- #3864 -> is fixed
- #3974 -> is fixed
- #4761 -> is fixed
- #4759 -> is fixed
- #4900 -> is fixed

### Review
1. _Add metadata to all issues_: The German label needs a correction: "hinzufü</h4>gen" -> "hinzufügen". 
2. _Add metadata to all issues_: The values of text-fields are added to all issues, when it is entered in the first issue. The values of selection lists or multiselect-fields are not added to all issues, when it is entered in the first issue. 
The goal should be that all kinds fields (text, selection-list, multi-select, ...) and its values should be added to all issues, when the button "Add metadata to all issues" is applied. 
3. _Duplicate-check_: If two issues are created with the same label, the message is shown: `Duplicate issue designations were found. This may produce duplicate process titles!: (Block 1)`. If two issues are created without label, no message is shown and the duplicate processes are not created. The last option should not occur, but the behavior of duplicate check should be the same in all use cases. 
The goal should be that the error message is always shown when the label of issues are the same - even if they are empty.

andre-hohmann (approved):

PR #4927, opened 2022-01-25
Fix problems with duplicate LDAP logins not being reported

### LINKED ISSUE #4926: Misleading error message on duplicate existence of 'LDAP login's
If there is more than one user record in the database, that have the same value in the 'LDAP login' field (like, a user was deleted and then recreated), the error message is showing: _Login … not found!_ The original error—_Login … was found more than once!_—is hidden, and doesn’t appear to the log file either. The reason is that the error message is thrown as a `UserNotFoundException` and the catching code misleadingly assumed that there were 0 entries, not several, then tries to find the 'LDAP login' on the 'login' field and doesn't find it there.

**Workaround:** Issue database command: `UPDATE user SET ldapLogin = NULL WHERE deleted = 1;`

**Goal:**

- Do not use exceptions as part of the regular control flow
- Search for both 'LDAP login' and 'login' and report duplicates correctly
- When deleting users, set 'LDAP login' to `null`

matthias-ronge:
Fixes #4926

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/User.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/User.java
@@ -442,6 +442,7 @@ public class User extends BaseBean {
     public void selfDestruct() {
         this.deleted = true;
         this.login = null;
+        this.ldapLogin = null;
         this.active = false;
         this.name = null;
         this.surname = null;
--- a/Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
+++ b/Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
@@ -85,7 +85,7 @@ public class DynamicAuthenticationProvider implements AuthenticationProvider {
     @Override
     public Authentication authenticate(Authentication authentication) {
         try {
-            User user = ServiceManager.getUserService().getByLdapLoginWithFallback(authentication.getName());
+            User user = ServiceManager.getUserService().getByLdapLoginOrLogin(authentication.getName());
             if (!user.isActive()) {
                 throw new DisabledException(SpringSecurityMessageSource.getAccessor().getMessage(
                     "AbstractUserDetailsAuthenticationProvider.disabled",
--- a/Kitodo/src/main/java/org/kitodo/production/security/LdapUserDetailsContextMapper.java
+++ b/Kitodo/src/main/java/org/kitodo/production/security/LdapUserDetailsContextMapper.java
@@ -28,7 +28,7 @@ public class LdapUserDetailsContextMapper extends LdapUserDetailsMapper implemen
 
     @Override
     public UserDetails mapUserFromContext(DirContextOperations ctx, String username, Collection<? extends GrantedAuthority> authorities) {
-        User user = ServiceManager.getUserService().getByLdapLoginWithFallback(username);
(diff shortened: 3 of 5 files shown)


henning-gerhardt (changes requested):
If I understand your change correct then your change would create an exception if there are unique login information with overlapping ldapLogins like in this example:

Person A (ldapLogin provided by external LDAP service)
- name: Franziska
- surname: Mueller
- login: muellerF
- ldapLogin: mueller

Person B (ldapLogin provided by external LDAP service)
- name: Boris Friedrich
- surname: Mueller
- login: muellerB
- ldapLogin: muellerF

matthias-ronge:
No, the problem occurred when there was a user, which was was “deleted”, but the field _LDAP login_ was not set to `null`, but remained in the database. Later, the same user was recreated, so the value in the field _LDAP login_ existed **twice**.

The problem is: The user enters a “username” and Production tries to figure out whether this is the _LDAP login_, or the _login_. Both is possible.

Assume:

```
id | firstName | lastName | login  | ldapLogin      | deleted
---+-----------+----------+--------+----------------+--------
…
17 | null      | null     | null   | Matthias Ronge | 1
…
42 | Matthias  | Ronge    | mronge | Matthias Ronge | 0
```

Now I try to log in with “Matthias Ronge” and my password. The code should find out whether this is the _LDAP login_ or the _login_.

In this case, the original code found more than one entry and threw a `UsernameNotFoundException("Login " + login + " was found more than once")`, but the `catch` clause saw that there is a `UsernameNotFoundException` and assumed there is **no** database record with that string being the _LDAP login_,  therefore continued with the login using the user’s login name. A login with the value of the field _LDAP login_ didn’t happen, so the user couldn’t log in.

What the code did (simplified):

```java
// String login = "Matthias Ronge";
User user;
try {
    // is String "login" the LDAP login?
    user = UserService.getByLdapLogin(login); // throws UsernameNotFoundException “Login
                                              // Matthias Ronge was found more than once”
} catch (UsernameNotFoundException) { // not found. It is not the LDAP login
    user = UserService.getByLogin(login);
}
```

henning-gerhardt:
But your change did create the conflict which I mentioned above and which is not existing with your change.

matthias-ronge:
Now I understand what you mean. Yes, that’s true, but that's the behaviour from Production v. 2 on already. In your case, Franziska can only log in with the username "mueller", and Boris can log in with the username "muellerF" and "muellerB".

Please understand that this case is rather theoretical. As far as I know, the field _LDAP login_ is not used at all, except to authenticate against Microsoft Active Directory. The `login` is typically always and ever a String without white spaces (at least, I don’t know any other example), where the _LDAP login_ for Microsoft Active Directory must always be in the form of “‹Firstname› ‹Lastname›”, because the query interface of AD doesn’t accept anything else. So, in that case, you would have

Person A (ldapLogin provided by external LDAP service)

- name: Franziska
- surname: Mueller
- login: muellerF
- ldapLogin: Franziska Mueller

Person B (ldapLogin provided by external LDAP service)
 
- name: Boris Friedrich
- surname: Mueller
- login: muellerB
- ldapLogin: Boris Friedrich Mueller

Franziska can then log in to Production with login “muellerf” and “Franziska Mueller”, and Boris can log in with login “muellerb” and “Boris Friedrich Mueller”.

matthias-ronge:
The change re-enables authentication against MS AD in the staging system of one of our customers. So I can confirm it solves the problem described in the issue. I didn’t notice it does break something in another environment where there is no MS AD.

PR #4929, opened 2022-01-25
Use typed converter classes

solth:
Prevent untyped converter classes to improve type safety.

"Raw types bypass generic type checks, deferring the catch of unsafe code to runtime. Therefore, you should avoid using raw types."
(https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html)


Kathrin-Huber:
Why is this a draft? Looks good to me.
Is it because of the failing tests?

solth:
I think more adjustments are required, not just in the failing tests.

PR #4949, opened 2022-01-31
Make mandatory metadata required

### LINKED ISSUE #3403: Incomplete validation by creation of new processes
### Problem
If a new process is created, the metadata is not validated for all mandatory metadata, according to the ruleset. The following metadata, however, is validated: 
- TSL/ATS (all documenttyps)
- BandNr-Sortierung (multivolume)

### Solution
Kitodo.Production 2.x validated all shown metadata. That should be possible in Kitodo.Production 3.x, too. Probably, this is a question of configuration.

### LINKED ISSUE #4377: Process creation - no check, if values are available for mandatory meta data (for "option" meta data)
**Also, in official Release 3.3!**

Lets assume a ruleset which defines as mandatory meta data for something like this
grafik
With "monograph" being "processtitle".
As you can see, all three meta data are set to "minOccurs=1".
During process creation (=press save button) this is checked well for MODSrecordIdentifier and MODStitle, but not for MODSlanguageTerm. The difference is, that MODSlanguageTerm is a key having option, as the other two are just normal "text" fields.
--> in my opionion, it looks like, that there is no check made in the KITODO-Software for this kind of meta data.
Please clarify.

IkramMaalej:
Part of #4322 
Fixes #4377
Fixes #3403

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
@@ -97,6 +97,7 @@
                                       readonly="#{not item.editable}"
                                       styleClass="#{not item.editable or readOnly ? 'read-only disabled' : ''}"
                                       disabled="#{not item.editable or readOnly}"
+                                      required="#{item.required and (not empty param['editForm:save'] or not empty param['editForm:saveContinue'])}"
                                       showCheckbox="true">
                         <f:selectItems value="#{item.items}"/>
                         <p:ajax event="change" oncomplete="preserveMetadata(); #{request.requestURI.contains('metadataEditor') ? 'updateTitleMetadata();' : ''}"/>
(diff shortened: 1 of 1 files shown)


andre-hohmann:
@IkramMaalej : It seems as there is still no check for lists with checkboxes, of which at least one must be selected. In my tests, it is the list of "Digital Collections". See example below. 
I can create a process without selecting a digital collection, but when i validate the process in the metadata editor, i get a warning, that no digital collection is selected. 

I cannot assess the behaviour with drop-down-fields, because in these mandatory fields (script, rights information, ...), one value has to be chosen. It is not possible to select an "empty" value. This might be described in:
- #4377

Other mandatory text-fields, as for example TSL/ATS, CatalogIDDigital, ... are validated correctly. 

### Example
validationdigitalcollection

andre-hohmann:
@IkramMaalej: 
The validation works now. Processes cannot be created, when mandatory checkbox lists are not selected. 

However, there are error messages, which cannot be interpreted - see example below. Either, the text of the messages must be adjusted, or the messages must be hidden. 

In my opinion, the error messages can be hidden, because of the two following arguments: 
1. There are no error messages, when mandatory text fields are empty. 
2. The affected fields are highlighted by red lines and can be identified. 

Auswahl_Kollektion

PR #4952, opened 2022-02-02
Add logging for indexing process

### LINKED ISSUE #4939: Reindexing: no information
I tried to re-index all tasks. Nothing happens. The progress bar does not reset from 100% to anything. I do not get any visible feedback, except of the comment in the last line.

start indexing no progress

In the log file, there is only one line.

    [DEBUG] 2022-01-28 09:14:28.949 [ajp-nio-127.0.0.1-8009-exec-7] IndexingService - Starting indexing of type TASK

No idea, if indexing has started, is going on or has finished.

**Goal:** Show visual feedback

matthias-ronge:
Improves logging for Indexing


Kathrin-Huber:
I'm pretty sure, there is a way without writing an own class which is instantiated every time the index is called.

PR #4955, opened 2022-02-03
mark the index status of base beans within a transaction as done with…

markusweigelt:
…out using the base type derivation

Fixes that the indexing hangs when the entity is marked as indexed.


Kathrin-Huber (approved):
please rebase

PR #4966, opened 2022-02-08
Database subset by config parameter

markusweigelt:
I don't currently know if it's a feature that belongs in the master. It helps me in development to work more efficiently on large databases.


matthias-ronge:
What exactly does your filter apply to, why is it helpful, and how should it be configured.

solth (approved):
Changes work as intended and only processes/tasks configured by `database.subset.dates` are indexed.

I also don't know if this belongs in the production code or should only be used to aid development, but as long as it doesn't do any harm to the actual running system, I don't see any reason not to include this option. 

One question to this end: does the "identity" clause in the SQL query (" 1=1 ") for the date filter (if `database.subset.dates` is not configured) have any impact on database query performance? I suspect it doesn't just wanted to ask to be sure.

PR #4968, opened 2022-02-09
Support output of metadata in docket creation

### LINKED ISSUE #4963: Missing exmple/documentation for file kitodo_exportXml.xml
Class `ExportXmlLog` references an undocumented file `kitodo_exportXml.xml`. It is not clear what this file should look like and what it does. There is no example file in the repository either.

### LINKED ISSUE #4964: ExportXmlLog still searching for meta_anchor.xml
The class `ExportXmlLog` is still searching for a file `meta_anchor.xml` which isn’t existent in Production version 3 anymore.

**Goal:** The code can be cleaned up.

### LINKED ISSUE #4984: Save same XML log output to file, as is used for docket creation
The output of the XML log via the corresponding export button, and the generation of the XML log as input for the XSLT process for creating dockets, is implemented twice: once in the core, and once in the Docket module. Both implementations differ significantly: Two different XML processing frameworks are used, and different XPATH parsers are used as well. For the ruleset, one implementations outputs the title, the other outputs the file name. One class allows access to metadata using a configuration file, the other does not care about this configuration file. In addition, different logging frameworks are used (Log4j vs. SLF4j).

**Goal:** Both places where the XML log is generated are to be equalized. Functional equality ᴍᴜsᴛ be established. The same frameworks sʜᴏᴜʟᴅ be used. At best, the code ᴄᴀɴ be deduplicated.

matthias-ronge:
Fixes #4964, fixes #4963, fixes #4984


Kathrin-Huber:
please fix checkstyle

Kathrin-Huber (changes requested):
Please add tests (with and without parent metadata)

matthias-ronge:
This pull request can no longer be rebased because the resulting diffs no longer make sense. It needs to be recreated.

Kitodo-Docket/src/main/java/org/kitodo/docket/ExportXmlLog.java
```diff
@@ -282,7 +305,119 @@ private List<Element> prepareProperties(List<Property> properties, Namespace xml
...
+                for (int i = 0; i <= count; i++) {
+                    String name = config.getString("namespace(" + i + ")[@name]");
+                    String value = config.getString("namespace(" + i + ")[@value]");
+                    nss.put(name, value);
+                }
+            }
+        } catch (ConfigurationException | RuntimeException e) {
+            logger.debug(e.getMessage(), e);
+            nss = new HashMap<>();
+        }
```
henning-gerhardt, line 349: This whole code block is dublicated in "new" method `getMetsFieldsFromConfig` ... maybe move this code to its own method and reuse it?

Kitodo-Docket/src/main/java/org/kitodo/docket/ExportXmlLog.java
```diff
@@ -282,7 +305,119 @@ private List<Element> prepareProperties(List<Property> properties, Namespace xml
...
+                for (int i = 0; i <= count; i++) {
+                    String name = config.getString(xmlpath + "(" + i + ")[@name]");
+                    String value = config.getString(xmlpath + "(" + i + ")[@value]");
+                    fields.put(name, value);
+                }
+            }
+        } catch (ConfigurationException | RuntimeException e) {
+            logger.debug(e.getMessage(), e);
+            fields = new HashMap<>();
+        }
```
henning-gerhardt, line 399: Dublicated code block - see comment above.

henning-gerhardt (changes requested):

solth:
Please resolve conflicts

Kitodo-Docket/src/main/java/org/kitodo/docket/ExportXmlLog.java
matthias-ronge: Thank you for noticing. The code was deduplicated now

Kitodo-Docket/src/test/java/org/kitodo/docket/ExportXmlLogTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class ExportXmlLogTest extends ExportXmlLog {
+
+    @Test
+    public void shouldExportXmlLogWithMetadata() throws IOException {
+        DocketData data = new DocketData();
+        data.setMetadataFile("src/test/resources/meta.xml");
+        ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+        super.startExport(data, ((OutputStream) buffer));
+        Assert.assertTrue("Output should not be empty", buffer.size() > 0);
```
henning-gerhardt, line 30: If the output of the export is some kind of garbage or random generated bytes then this test will not detect this as you only checking of the amount of returned data.

Kitodo-Docket/pom.xml
```diff
@@ -70,6 +70,12 @@
             <groupId>org.apache.pdfbox</groupId>
             <artifactId>pdfbox</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>commons-collections</groupId>
+            <artifactId>commons-collections</artifactId>
```
henning-gerhardt, line 76: - used version of commons-collections is very, very old and should be replaced by commons-collections4
- in the new created unit test there is no usage of this library or I did not see the usage

Kitodo-API/src/main/java/org/kitodo/api/docket/DocketData.java
```diff
@@ -17,30 +17,82 @@
...
+        return metadataFile;
+    }
+
+    /**
+     * Sets the metadataFile.
+     * 
+     * @param metadataFile
+     *            The metadata file.
+     */
+    public void setMetadataFile(String metadataFile) {
```
henning-gerhardt, line 68: Should we not use URI class for the metadata file as the file must not necessary residence on disc?

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -2017,10 +2016,15 @@ private List<DocketData> getDocketData(List<Process> processes) {
      *            The process to create the docket data for.
      * @return The DocketData for the process.
      */
-    private DocketData getDocketData(Process process) {
+    private static DocketData getDocketData(Process process) throws IOException {
         DocketData docketdata = new DocketData();
 
         docketdata.setCreationDate(process.getCreationDate().toString());
+        URI metadataFilePath = fileService.getMetadataFilePath(process);
+        docketdata.setMetadataFile(fileService.getFile(metadataFilePath).toString());
```
henning-gerhardt, line 2024: In my opinion the cast `toString` should not be used to get the path to a file - there are better and more reliable methods to get this information.

henning-gerhardt (changes requested):
This is only a code review, no function review and beside my code remarks I have one more:

Changing nearly all methods to `static` of the moved class and some of the existing methods in other classes only to get ride out of the new operator for the class instance is wrong as the static methods are violating against the poly morph principle of object operating programming and is a sign of lazy programming in my opinion.

solth:

Kitodo-Docket/src/test/java/org/kitodo/docket/ExportXmlLogTest.java
matthias-ronge: Test was changed

Kitodo-API/src/main/java/org/kitodo/api/docket/DocketData.java
matthias-ronge: Changed

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
matthias-ronge: Changed

Kitodo-Docket/pom.xml
henning-gerhardt: Is then the used scope `test` right? Should then the scope not changed to the default value so this dependency is visible and not hidden solved?

matthias-ronge: It is only needed for the test to work. Don’t ask me why

henning-gerhardt: The reason is, that this dependency exists somewhere else too and is resolved there and as the application provide all needed dependencies the dependency  here is solved in a "hidden" (or magic) way.

matthias-ronge: I removed `test` scope

henning-gerhardt (commented):
Again only a code review as we did not use this feature:

Looks okay but I don't see any requirements to made a lot of methods static. This may be possible but I don't likes static methods except for only a few tiny reasons.

matthias-ronge:
A function **is** `static` if it does not use member variables, and does not use other functions that use member variables. In my opinion, any function to which this applies should then also be labeled as such. Then, as a programmer reading the source code, you know that these characteristics are met for this function. This is just like how we mark constants as `final`. You don't have to do this, it would work without it, but it says something to the programmer reading the code. I like to do that, also for myself, when I read my own code again years later. I find it clearer. Why does it bother you?

henning-gerhardt:
Independent of the "static method" changes I'm unsure if all this changes should belong to the "Kitodo-Docket" module as this module in my understanding is only for creating the dockets and not for general generating pdf files depending on xslt, xml and meta data of the process.

matthias-ronge:
The sole purpose of generating the XML log is to create the docket from it. Previouslyz there were two places where the XML log was generated (duplicate code), one copy in the core for saving it to the user folder, and another copy in the docket module for creating the docket. These two have now been deduplicated and the code now only exists once.

In addition, due to incomplete copy-paste, the saved XML log was not exactly identical to what was generated in the Docket module, so if you developed an XSLT for a docket for a customer and it ran with the saved XML log, it still didn't work reliably with the docket module. This was very nerve-wracking when the customer wants many adjustments to the docket.xslt.

I think this change makes sense for both reasons, code deduplication and better configurability.

henning-gerhardt:
This may be true (I did not validate your statement in any little detail) but and in mind of the current module concept the base (generating the XML log) should then be in a separate module and the current docket module should depend on this new module or if there are only some common logic: the common logic should be in one module and the outputs in two depending modules of this common module. 

But I don't know if the current module concept is still valid or not. Maybe this should be discussed but there is no discussion ongoing as no one is participating as everyone has only his/her stuff to do. So @solth should be decide if this changes should all go into the current docket module or not.

matthias-ronge:
Hi @henning-gerhardt: I removed the `static` modifiers from the static functions. From the point of view of the code review, is the source code consistent with the coding guidelines? If so, I would ask you to mark the pull request as approved.

I think the questions of dividing the functions into modules, and how to proceed at all with the modularization of the source code (there are quite a few more open questions, I think), are important, and you are welcome to open a GitHub discussion for this.

henning-gerhardt:
I can review this maybe next week and I'm discussing some in deep details only in my native language.

henning-gerhardt (approved):

PR #4969, opened 2022-02-09
Support includes in ruleset

### LINKED ISSUE #4596: Support includes in ruleset
The mapping of all project-specific information in the ruleset leads to an expansion of the rulesets, so that in the worst case there is one ruleset per project. In many ways, these rulesets are nevertheless the same. However, to adapt the same change now for all projects requires a change in a large number of files.

**Goal:** Rulesets should allow includes to be defined. So that the included rulesets are read first, and then the configurations from this ruleset are added.

_Example:_ This ruleset first loads a general ruleset (`bib_shared.xml`), then a ruleset in which only the language field is defined (because its declaration is so extensive, `bib_shared_language-field.xml`). Then, only the collections are specified specifically for the current project. Everything else is already defined in the general ruleset for all projects.

```xml
<ruleset xmlns="http://names.kitodo.org/ruleset/v2">
    <include>bib_shared.xml</include>
    <include>bib_shared_language-field.xml</include>
    <declaration>
        <key id="singleDigCollection">
            <label>Digital collection</label>
            <label lang="de">Digitale Kollektion</label>
            <option value="eins">
                <label>Collection 1</label>
            </option>
            <option value="zwei">
                <label>Collection 2</label>
            </option>
            <option value="drei">
                <label>Collection 3</label>
            </option>
            <preset>eins</preset>
        </key>
    </declaration>
</ruleset>
```

Notes:
* Includes should work recursively.
* It is clear that this causes dependencies between the different files. Using the includes is optional, but if they are used, this must be dealt with.

matthias-ronge:
Fixes #4596

### DIFF
--- a/Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
+++ b/Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
@@ -218,7 +218,7 @@ public class RulesetManagement implements RulesetManagementInterface {
      */
     @Override
     public void load(File rulesetFile) throws IOException {
-        this.ruleset = readRecursive(rulesetFile);
+        this.ruleset = read(rulesetFile);
         initializeNamespaces(ruleset.getKeys(), rulesetFile.getParentFile());
     }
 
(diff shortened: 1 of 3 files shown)


Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
```diff
@@ -262,6 +262,18 @@ private void initializeNamespaces(List<Key> keys, File home) throws IOException
         }
     }
 
+    private static Ruleset readRecursive(File rulesetFile) throws IOException {
```
Kathrin-Huber, line 265: remove "recursive" here

Kathrin-Huber (changes requested):

PR #4998, opened 2022-02-17
add resource expiration in development environment

markusweigelt:
When changes are made to css or javascript, for example, they are not reloaded even though the browser cache is disabled. Only an explicit refresh of the resource url can fix this.

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -821,8 +821,8 @@
             </activation>
             <properties>
                 <jsf.projectStage>Development</jsf.projectStage>
-                <!-- resource expiration after two minutes -->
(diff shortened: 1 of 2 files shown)


Kathrin-Huber:
Don't you think two minutes is too short?
When opening the Metadata Editor it can easily take two minutes to load the page, are the resources then loaded multiple times?

matthias-ronge:
When loading one page, a browser typically doesn’t load a referenced file several times, even if it is referenced several times in that page. But I would even assume that the script resources are only referenced once from the generated HTML (but that’s something that’s easy to check).

markusweigelt:
Expiring is changed to four minutes.

PR #5024, opened 2022-03-02
change from CurrentNo to CurrentNoSorting

### LINKED ISSUE #4737: Migrated processes of volumes should be ordered by logical order
### Problem
Processes for periodicals or multivolumes must be assigned to the correct position. The correct position follows the logical order, as for example the order number of volumes or the date of issuance. 

For the migrated processes in Kitodo.Production 3.x, the order of the processes follows the Kitodo-ID and thus the date of the creation of the processes. That can be random due to several reasons, as for example: missing volumes are added belated from other institutions, ...

This leads also to problems during the creation of new processes #4736. It is in some cases impossible to find a correct logical position. 

### Solution
**If the order in Kitodo.Production influences the order in Kitodo.Presentation**, the order in Kitodo.Production must be adjusted during the migration. I am aware of the difficulties in finding criteria that meets the demands of all institutions. 

If the order in Kitodo.Production affects only Kitodo.Production, this issue might not be regarded.

Kathrin-Huber:
fixes #4737

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
@@ -347,7 +347,7 @@ public class HierarchyMigrationTask extends EmptyTask {
      */
     private static Integer getCurrentNo(LogicalDivision includedStructualElement) {
         Integer currentNo = includedStructualElement.getMetadata().parallelStream()
-                .filter(metadata -> metadata.getKey().equals("CurrentNo")).filter(MetadataEntry.class::isInstance)
+                .filter(metadata -> metadata.getKey().equals("CurrentNoSorting")).filter(MetadataEntry.class::isInstance)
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
```diff
@@ -347,7 +347,7 @@ private static Integer convertChildMetsFile(URI metadataFilePath) throws IOExcep
      */
     private static Integer getCurrentNo(LogicalDivision includedStructualElement) {
```
matthias-ronge, line 348: You could rename it, like
```
private static Integer getCurrentNoSorting(LogicalDivision logicalDivision)
                                   ^^^^^^^                 ^^^^^^^^^^^^^^^
```

matthias-ronge (approved):

Kitodo/src/main/java/org/kitodo/production/helper/tasks/HierarchyMigrationTask.java
Kathrin-Huber: You are right, it's renamed :-)

PR #5026, opened 2022-03-04
Allow editing metadata of all processes in hierarchy during import

### LINKED ISSUE #4595: Values not visible for hierarchy import
When importing a volume, and the import automatically imports the superior multi-volume work as well, the metadata fields of the multi-volume work are not visible and cannot be edited in the import mask.

**Goal:** Show and allow to edit metadata from all hierarchically imported records.

Screenshot

solth:
This pull request enables the user to edit the metadata of all imported processes during hiearchichal import (i.e. importing parent or children processes from accordingly configured search interfaces) and thereby fixes #4595.

The imported process hiearchy is displayed in the upper section of the slightly redesigned import mask if more than one process is imported at once. The user can than click on the individual processes and edit the process basic and metadata in the lower part of the import mask. The process hierarchy can also be collapsed to have more room for editing data in the lower half).

https://user-images.githubusercontent.com/19183925/156761343-7c38e03e-0054-4846-bd55-fe54905cf6dd.mov

Part of #4322


IkramMaalej:
- All of a person's subfields should be as wide as the role subfield.

Bildschirmfoto 2022-03-21 um 10 01 58

- If I add  a new person to a process and edit it and then I switch to another process and I add a new Person to it, then I get a new pre-filled person with data from what  I eneterd in the first process.

- If I add a new metadata to one process and then switch to another. When I return to the first edited process, the newly added metadata disappears.

- The newly added metadata will not be saved and not displayed in ME (but the edited ones will be)

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -651,4 +623,74 @@ public boolean metadataAddableToGroup(TreeNode metadataNode) {
...
+        return this.childProcesses;
+    }
+
+    /**
+     * Get ID metadata.
+     *
+     * @param tempProcess TempProcess whose ID metadata is returned
+     * @return ID metadata
+     */
+    public String getID(TempProcess tempProcess) {
```
Kathrin-Huber, line 651: I would expect to get an database id here (I know, its a tempProcess, but still... ) Can you be more specific in the methodname?

Kathrin-Huber (approved):
please rebase

Kathrin-Huber:
Please fix ST test

Kathrin-Huber:
please resolve conflicts and fix ST test

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
solth: This method returns the catalog ID which is used to query individual records from the search interface. It is configured in the `kitodo_opac.xml` using the `identifierMetadata` element. This metadata is displayed in the import mask on the buttons in the process hierarchy to identify the individual records imported.

henning-gerhardt: Is the name `getCatalogueId` not a better name then `getMetadataId`?

PR #5036, opened 2022-03-18
Change sorting in excel from title to id

Kathrin-Huber:
Sorting in Excel should be done by id, not by title

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
@@ -69,7 +69,8 @@ public class SearchResultGeneration {
     private List<ProcessDTO> getResultsWithFilter() {
         List<ProcessDTO> processDTOS = new ArrayList<>();
         try {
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
```diff
@@ -70,7 +70,7 @@ public HSSFWorkbook getResult() {
         List<ProcessDTO> processDTOS = new ArrayList<>();
         try {
             processDTOS = ServiceManager.getProcessService().findByQuery(getQueryForFilter(ObjectType.PROCESS),
-                ServiceManager.getProcessService().sortByTitle(SortOrder.ASC), false);
+                    ServiceManager.getProcessService().sortById(SortOrder.ASC),false);
```
henning-gerhardt, line 73: `false` needs a white space in front and maybe this must be changed to `true` as we / I got time out issues with ` false`

Kathrin-Huber: right, but this is a different issue ;-)
This PR is only about ordering.
I will add the whitespace

solth (approved):

PR #5039, opened 2022-03-21
Import metadata from mass import CSV files

### LINKED ISSUE #4989: No automatic check for existing parents during mass import
It appears as if there is no check for existing parents during a mass import. A logic similar to the one implemented for file uploads ( https://github.com/kitodo/kitodo-production/pull/4884 ) is probably needed for mass imports as well in order to establish a link to the parent process.

solth:
Allow importing metadata from CSV files.
For this to work, the first line in the CSV file has to contain the metadata keys to be added. The consequent lines then hold the corresponding metadata values to be added for the individual imported records.
Additionally, the user can set a CSV separator character which is used in the uploaded file as column separator.

Also fixes #4989


BartChris:
@solth This looks useful. I have some questions: Is this feature in general always intended to be used with a catalogue backend? So that the data from the csv file is always used as additional metadata for the imported recorded? If this is the case: Are there plans to also add a CSV import which takes all the data from the csv?

solth:
@BartChris currently this is just intended as an improvement of the existing mass import from catalog search interfaces. It shouldn't be too difficult to expand the functionality to create new processes just from the data in the CSV files, without OPAC search, but that is outside the scope of this pull request.

PR #5080, opened 2022-04-26
Allow parent process's process title as prefix in child title generation

matthias-ronge:
If the schema for the child process title starts with a plus, Production will prepends the process title of the parent process. This pattern already exists for newspaper processes, and it is now made available for common child processes as well.


Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessDataTab.java
```diff
@@ -209,10 +209,14 @@ public void generateProcessTitleAndTiffHeader() {
                     new Object[] {getDocTypeLabel(docType), process.getRuleset().getTitle() });
             }
 
+            Process parentProcess = createProcessForm.getTitleRecordLinkTab().getTitleRecordProcess();
+            if (processTitle.startsWith("+") && Objects.nonNull(parentProcess)) {
+                processTitle = '\'' + parentProcess.getTitle() + '\'' + processTitle;
```
Kathrin-Huber, line 214: Why the backslashes?

matthias-ronge: Since the apostrophe is a character constant, and such are enclosed in apostrophes, it must be escaped. It’s the same reason why a double quote in a string literal needs to be preceded by a backslash.

```
char    String   value
-----   ------   -----
'f'     "f"      f
'"'     "\""     "
'\''    "'"      '
```

matthias-ronge: The text between the apostrophes is interpreted as text by the title generator, text without apostrophes is understood as a metadata key. So a `processTitle="+'_'+IssueNo` becomes string `"'ParentProcess_1234567X'+'_'+IssueNo"`, and then the generator turns it into `"ParentProcess_1234567X_123"`. Without the apostrophes, the title generator would look for a metadata entry with key `ParentProcess_1234567X`, which does not exist. I didn't invent the syntax for the title generator at this point, this is something [we inherited from the past](/kitodo/kitodo-production/blob/2.x/Goobi/config/goobi_projects.xml#L65-L67).

Kathrin-Huber:
Waiting for a test here

Kitodo/src/test/java/org/kitodo/production/forms/createprocess/ProcessDataTabTest.java
```diff
@@ -0,0 +1,125 @@
+package org.kitodo.production.forms.createprocess;
```
henning-gerhardt, line 1: Missing license header?

matthias-ronge: Fixed

PR #5081, opened 2022-04-26
Allow automatic adding the counting for a child process

### LINKED ISSUE #4993: Offer automatic child counting when creating processes
When creating child processes (multi-volume works), the volume number must always be specified manually, if it isn’t imported from a catalog. This is annoying, if you’re going to do them sequentially anyway.

**Goal is:** A metadata key can be defined with a `use="childCount"`, which is pre-assigned with the next volume number (the number of existing child processes plus one) when a child process is created using the plus button on the parent process.

matthias-ronge:
Resolves #4993.


Kathrin-Huber:
Also a test is making sense here

Issue #4783, opened 2021-10-22
Performance problem: Process list


andre-hohmann:
### Problem
If the process list shows 100 processes, it takes several seconds until the list is shown.
Probably, the reason is that too many icons have to loaded.

### Solution
The list of processes should be shown quicker, also when 100 processes are shown.

Issue #4195, opened 2021-02-18
Metadata editor takes a long time initializing


matthias-ronge:
If a process with a large number of images opens the metadata editor, it takes a very long time. With 1200 images, `FileService.searchForMedia()` took 28 seconds to process.

**Goal:** This should be investigated to speed up.

markusweigelt:
The main problem here is the gallery panel. All chapters incl. the subchapters are loaded as a structure element below each other. In addition, an image gallery is created for each structure element.  The more chapters there are, the longer it takes. With about 160 chapters the loading time is about 1 minute.

We should reduce the number of structural elements here by pagination within the gallery or by another solution.

matthias-ronge:
This is where JSF's lazy loading functionality would make sense, not at the image level though, but at the gallery strip level. Only those gallery strips that are in the current viewport will be loaded, the other strips will not be requested until you scroll to them.

markusweigelt:
Do you mean the lazy loading function at the componente "p:dataList" (attribute "lazy") or is there a general implementation? From my observations this is only useful in connection with a for example pagination. If all structure elements are loaded on one page then this option is not useful.

matthias-ronge:
It is [`<p:dataScroller>`](http://primefaces.org/showcase/ui/data/datascroller/basic.xhtml)

markusweigelt:
The component looks quite good at first sight. We just need to figure out how to jump to non-existent structural elements.

markusweigelt:
I think the problem bigger than thought. The initial loading can probably be fixed by lazy loading but there are several problems here. Solving this problem will also solve this ticket. #5012

Issue #3533, opened 2020-04-24
Improve reaction speed of pages with a table


matthias-ronge:
The user interface takes an incredibly long time to build pages with tables, especially if you keep in mind that the tables (should) be shown from a search engine index. This mainly affects the list of processes and the dashboard. The rotary key symbol appears often and for a long time. It is particularly offending that, after search queries, a table with incorrect values is slowly built up, then it flashes briefly, and the correct data appears in a split second.

![Screenshot 1](https://user-images.githubusercontent.com/3040657/80218012-e8346c00-8640-11ea-9d72-befb52f4fba3.jpg)

![Screenshot 2](https://user-images.githubusercontent.com/3040657/80218400-6c86ef00-8641-11ea-85e8-158675c4ec7f.jpg)

**Goals:**
- [ ] In any case, the relevant data should be displayed immediately, not then via an update.
- [ ] Tables should be displayed 100% from the search engine index. This must be reviewed and corrected where necessary.
- [ ] If a display from the index is technically not possible, the possible components from the index should be displayed and the remaining components reloaded asynchronously in order to quickly get a reactive user interface again.

PR #5096, opened 2022-04-27
Performance Caching Current Locale

thomaslow:
Contributes to:
- https://github.com/kitodo/kitodo-production/issues/3533
- https://github.com/kitodo/kitodo-production/issues/4195
- https://github.com/kitodo/kitodo-production/issues/4783

On many occasions, e.g., when translating messages, the method [LocaleHelper.getCurrentLocale](https://github.com/kitodo/kitodo-production/blob/6b5bb5e2e5fc4631e28ea33b753bfa8e68105c0d/Kitodo/src/main/java/org/kitodo/production/helper/LocaleHelper.java#L57) is called. If a user is logged in, this request will trigger a database query for the current language setting of the user, see [Line 84](https://github.com/kitodo/kitodo-production/blob/6b5bb5e2e5fc4631e28ea33b753bfa8e68105c0d/Kitodo/src/main/java/org/kitodo/production/helper/LocaleHelper.java#L84). Naturally, issuing a database query for each `#{msgs.something}` will reduce performance significantly.

In theory, this database query could be cached by Hibernate. Unfortunately, Hibernate caching is currently not working in Kitodo:
- the Hibernate first-level cache is not used, because Hibernate sessions do not persist for longer than a single database query, see most methods of [BaseDAO](https://github.com/kitodo/kitodo-production/blob/6ce1ce5c60cb6d70d5e6b2d8b9b31ea645e58bb7/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java), and the first-level cache is bound to the current Hibernate session
- the Hibernate second-level cache is enabled globally, but requires the annotation `@org.hibernate.annotations.Cache` for each Bean, which are missing, see [Documentation](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#caching-config): 
```
By default, entities are not part of the second level cache and we recommend you to stick to this setting.
[...]
Use the @org.hibernate.annotations.Cache annotation for this purpose.
```

Even with database caching enabled, one could argue that each `#{msgs.something}` should still not require a full re-evaluation of the Hibernate query framework. In my opinion, the current user locale should be cached in memory for each request.

The proposed pull request caches `LocaleHelper.getCurrentLocale` via a simple request scoped cache implementation using `FacesContext`, meaning the cache is cleared after each HTTP request. Caching improves performance significantly (together with #5095):

- Dashboard from 500ms to 350ms (#5095) to 250ms (both pull requests)
- List of 100 processes from 3100ms to 1400ms (#5095) to 800ms (both pull requests)
- Metadataeditor backend from ~4000ms to ~2600ms (#5095) to ~1900ms (both pull requests)

### Drawbacks

Caching via `FacesContext` could be argued to not be a very clean way of implementing this. Ideally, such information should be cached via beans that are annotated with `@RequestScope`. However, I'm not aware of any other way to introduce request scope to low-level static service methods like this. I suppose this is a JSF architectural design issue. Any suggestions are welcome.


Kitodo/src/main/java/org/kitodo/production/helper/cache/RequestScopeCacheHelper.java
```diff
@@ -0,0 +1,89 @@
...
+ * <p>If this cache is used outside of a JSF request context, it will simply re-evaluate the supplier 
+ * function upon every request.</p>
+ * 
+ */
+public class RequestScopeCacheHelper {
+
+    private static final String ATTRIBUTE_NAME = "RequestScopeCacheHelper";
+
+    @SuppressWarnings("unused")
+    private static final Logger logger = LogManager.getLogger(RequestScopeCacheHelper.class);
```
markusweigelt, line 47: Do we need this logger? If not then I suggest to remove it.

markusweigelt (changes requested):
Great idea to cache within the RequestScrope. From my side the code and documentation fits except of the small comment.

Kitodo/src/main/java/org/kitodo/production/helper/cache/RequestScopeCacheHelper.java
thomaslow: It's not needed. Removed

markusweigelt (approved):

Kitodo/src/main/java/org/kitodo/production/helper/cache/RequestScopeCacheHelper.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+import javax.faces.context.FacesContext;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
```
henning-gerhardt, line 22: As there is no Logger anymore this imports are not needed anymore.

markusweigelt: I had overlooked thx

PR #5100, opened 2022-04-27
Allow adding new folders to existing projects

### LINKED ISSUE #5097: Sub-folders cannot be added to projects
### Problem
Since fixing #4833, no sub-folders can be added to the projects (see example below). 
- the field "Folder" is blocked for the cursor
- the field "File type" is grey and the value cannot be changed

### Solution
It is necessary to avoid changing the paths of existing folders, but it should be possible to
1. add sub-folders to new projects
2. add sub-folders to existing projects (for example to add the folders for the preview- and mediaview folders)

### Example
subfolder01

solth:
Fixes #5097


andre-hohmann:
@solth : Thanks for the quick solution! @henning-gerhardt applied in our system and it works. I can now add new subfolders, but cannot change the paths of existing subfolders.

PR #5109, opened 2022-04-28
Explicitly load Saxon, without incurring any side impacts on the JVM

### LINKED ISSUE #5108: XMLSchemaConverter breaks foreign web applications
XMLSchemaConverter [redefines the JVM system variable `javax.xml.transform.TransformerFactory`](/kitodo/kitodo-production/blob/0c5c76bbb4c2afa9f6b4e2ed9093c4d82854f586/Kitodo-XML-SchemaConverter/src/main/java/org/kitodo/xmlschemaconverter/XMLSchemaConverter.java#L71). This has implications for other web applications running in the same JVM / Tomcat instance, when they want to process XML, they suddenly try to load `net.sf.saxon.TransformerFactoryImpl`, which they don't even have in their classpath, because it is not intended by the developer, and fail with `ClassNotFoundException`.

Workaround: Copy `Saxon-HE-9.9.1-5.jar` from `webapps/kitodo/WEB-INF/lib/` to the `WEB-INF/lib/` directory of the broken web app, then restart Tomcat.

**Goal:** Instead of redefining `javax.xml.transform.TransformerFactory` and using `TransformerFactory.newInstance()`, the code should use `new net.sf.saxon.TransformerFactoryImpl()` to load Saxon explicitly.

matthias-ronge:
Fixes #5108

### DIFF
--- a/Kitodo-XML-SchemaConverter/pom.xml
+++ b/Kitodo-XML-SchemaConverter/pom.xml
@@ -43,6 +43,7 @@
         <dependency>
             <groupId>net.sf.saxon</groupId>
             <artifactId>Saxon-HE</artifactId>
+            <version>${saxon.version}</version>
         </dependency>
     </dependencies>
 
--- a/Kitodo-XML-SchemaConverter/pom.xml
+++ b/Kitodo-XML-SchemaConverter/pom.xml
@@ -43,7 +43,6 @@
         <dependency>
             <groupId>net.sf.saxon</groupId>
             <artifactId>Saxon-HE</artifactId>
-            <version>${saxon.version}</version>
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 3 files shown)


solth (changes requested):
Please rebase / resolve conflicts!

Kathrin-Huber:
still waiting for a rebase here

matthias-ronge:
Doesn’t work anymore

Kitodo-XML-SchemaConverter/pom.xml
```diff
@@ -43,6 +43,7 @@
         <dependency>
             <groupId>net.sf.saxon</groupId>
             <artifactId>Saxon-HE</artifactId>
+            <version>${saxon.version}</version>
```
henning-gerhardt, line 46: This should not necessary as the used version of a package is controlled in the global pom file.

solth (approved):
Tested changes with catalog import and it worked as before.

PR #5115, opened 2022-04-29
Performance Enable Gzip Compression

thomaslow:
Related Issues:
- #4195 

When loading a jsf page, the jsf instructions are sent as raw text to the browser. For large pages, e.g., a meta data editor with 1000 thumbnails, this results in a huge amount of data (~10 MB) that is transferred without compression. With compression enabled, only 3.4MB are transferred. On slow network connections, compression can improve loading times significantly.

This pull request enables gzip compression of all resources application-wide. 

Together with the following other pull requests
- #5113
- #5114

the loading time of the meta data editor with 1000 pages is reduced from 5.79 seconds to 2.27 seconds (desktop workstation) or 20.81 seconds to 6.01 seconds (laptop).

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/web.xml
+++ b/Kitodo/src/main/webapp/WEB-INF/web.xml
@@ -111,6 +111,17 @@
         <session-timeout>120</session-timeout>
     </session-config>
 
+    <!-- gzip filter compressing resources -->
+    <filter>
+        <filter-name>gzipResponseFilter</filter-name>
+        <filter-class>org.omnifaces.filter.GzipResponseFilter</filter-class>
+    </filter>
+    <filter-mapping>
+        <filter-name>gzipResponseFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+        <dispatcher>REQUEST</dispatcher>
+    </filter-mapping>
+
     <!-- Welcome files -->
     <welcome-file-list>
         <welcome-file>index.jsp</welcome-file>


henning-gerhardt:
The reduced data is only true if you access the Tomcat instance direct by your browser. If you have an application proxy (like Apache HTTP or nginx) in front of a Tomcat instance and in this application proxy is already a gzip compression active (Debian 11 Apache HTTP 2.4 by default) you did note a difference between.

thomaslow:
You are right, this is true for Apache. I quickly checked the same for Nginx and Tomcat in Ubuntu 20.04. Apparently, it is not the case for Nginx (only compresses `text/html` by default) nor Tomcat (no compression by default). 

Instead of enabling compression as suggested by this pull request, maybe we can add some instructions to the "Installationsanleitung" instead? To my knowledge, there is no section about how to set up Tomcat for production, or proxy Kitodo with either Apache or Nginx.

I realize that a complete guide on how to set up and secure a web-facing web server like Apache or Nginx is out of the scope of the "Installationsanleitung", but an example configuration or some notes might help less experienced system admins to deploy Kitodo in a more secure and more efficient way, e.g., by pointing out things like:

> ### Notes on Web Server Configuration 
>
> 1. Make sure to enable compression for at least the resource types `text/html`, `text/css`, `text/xml`, `application/javascript`, `application/xml`, see ([Apache](https://httpd.apache.org/docs/2.4/mod/mod_deflate.html), [Nginx](https://docs.nginx.com/nginx/admin-guide/web-server/compression/), [Tomcat](https://tomcat.apache.org/tomcat-9.0-doc/config/http.html#:~:text=compressibleMimeType))
> 2. If Apache or Nginx is used as a proxy, make sure Tomcat is not accessible by blocking its port with your firewall
> 3. Make sure to enable SSL/TLS since otherwise passwords might be exposed when logging in, see (Apache, Nginx, Tomcat)
> 4. Set up fail2ban to prevent malicious login attempts
> 5. Set up rate limiting to slow down malicious crawler
> 6. Allowing the user to upload large files might require to modify settings X and Y
> 7. ...

henning-gerhardt:
As I did not use nginx I did not know if there is gzip compression is active by default or not. So your pull request enable compression in any case and if you run an application proxy in front of it you must configure the application proxy correct.

Issue #5099, opened 2022-04-27
NPE on GalleryPanel on non defined media directory


henning-gerhardt:
A null pointer exception is thrown if no media or preview folder is defined in a project and a process of this project is opened in the meta data editor:

```
java.lang.NullPointerException
        at org.kitodo.production.forms.dataeditor.GalleryPanel.getMediaViewMimeType(GalleryPanel.java:163)
        at org.kitodo.production.forms.dataeditor.GalleryMediaContent.getMediaViewData(GalleryMediaContent.java:98)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
```

The null pointer exception is created on accessing the folder information and is set explicit to `null` if there is no media folder defined [GalleryPanel.java#L403](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java#L403) and [GalleryPanel.java#L405](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java#L405) but there is no check if this the case later on working with the value ([GalleryPanel.java#L187](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java#L187) and [GalleryPanel.java#L163](https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java#L163)).

This could maybe be solved if there is not an other issue #5097 which is preventing adding new folder information to existing and new projects. 

But the application should even work if there are no media folder(s) defined inside the project settings and not show a null pointer exception to the user. The application may should show in this case a nice warning message that there are no media folders in the project are defined and no pictures will be shown.

matthias-ronge:
The problem here is that the `<p:graphicImage>` for the media view is not `rendered` conditionally, but shall be shown even if there is no data.

matthias-ronge:
[...]

PR #5117, opened 2022-04-29
Prevent possible NPE if no media or preview folder is selected

henning-gerhardt:
Possible fix for #5099


matthias-ronge (changes requested):
This solution is wrong, as media variants must correspond to the METS file groups. The value of `use` is the value of the USE attribute of the `<mets:fileGrp>`.

matthias-ronge:
Your approach reveals that the code flow is not ideal here. If present, media files do always have a media variant. This could be an improvement (but won’t fix the NPE): https://github.com/matthias-ronge/kitodo-production/pull/new/nju

henning-gerhardt:
@matthias-ronge : Open a pull request with your changes and I will close my pull request as you have more knowledge about this stuff as I.

PR #5121, opened 2022-05-02
Performance loading Mets workpiece

thomaslow:
Related issues:
- #4195 

Loading a mets file ("meta.xml") via `MetsService.loadWorkpiece` is slow in case there are many images referenced in the xml. An xml that contains 2000 images needs ~1.5 seconds to load (desktop workstation). 

The problem is that the xml file is traversed multiple times to figure out the "use" parameter for each image file (e.g. THUMB, MAX, LOCAL). By pre-loading this parameter for each file and using that as a cache while generating `PhysicalDivision`, the loading time improves drastically to around 0.1 seconds.

This affects multiple parts of kitodo-production (wherever `MetsService.loadWorkpiece` is [called](https://github.com/kitodo/kitodo-production/search?q=loadWorkpiece&type=code)), e.g.:
- when opening the meta data editor
- when calculating process statistics via the "number of metadata" button at the bottom of a process list
- ...

### DIFF
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/FileXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/FileXmlElementAccess.java
@@ -20,6 +20,8 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.kitodo.api.MdSec;
 import org.kitodo.api.dataformat.MediaVariant;
 import org.kitodo.api.dataformat.PhysicalDivision;
@@ -31,9 +33,11 @@ import org.kitodo.dataformat.metskitodo.FileType;
 import org.kitodo.dataformat.metskitodo.MdSecType;
 import org.kitodo.dataformat.metskitodo.Mets;
 import org.kitodo.dataformat.metskitodo.MetsType;
-import org.kitodo.dataformat.metskitodo.MetsType.FileSec.FileGrp;
 
 public class FileXmlElementAccess {
+
+    @SuppressWarnings("unused")
+    private static final Logger logger = LogManager.getLogger(FileXmlElementAccess.class);
 
     /**
      * The data object of this file XML element access.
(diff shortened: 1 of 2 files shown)


Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/FileXmlElementAccess.java
```diff
@@ -31,10 +33,12 @@
 import org.kitodo.dataformat.metskitodo.MdSecType;
 import org.kitodo.dataformat.metskitodo.Mets;
 import org.kitodo.dataformat.metskitodo.MetsType;
-import org.kitodo.dataformat.metskitodo.MetsType.FileSec.FileGrp;
 
 public class FileXmlElementAccess {
 
+    @SuppressWarnings("unused")
+    private static final Logger logger = LogManager.getLogger(FileXmlElementAccess.class);
```
henning-gerhardt, line 40: No used logger - remove it including the imports for Logging ;)

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
```diff
@@ -135,11 +139,16 @@ private MetsXmlElementAccess(Mets mets) {
         Map<String, FileXmlElementAccess> divIDsToPhysicalDivisions = new HashMap<>();
         if (optionalPhysicalStructMap.isPresent()) {
             DivType div = optionalPhysicalStructMap.get().getDiv();
-            FileXmlElementAccess fileXmlElementAccess = new FileXmlElementAccess(div, mets, useXmlAttributeAccess);
+            Map<FileType, String> fileUseByFileCach = createFileUseByFileCache(mets);
```
henning-gerhardt, line 142: Typo? Instead of `fileUseByFileCach` maybe `fileUseByFileCache` is better?

thomaslow: Yes. Thanks. Will fix that. Something bigger is wrong, though. Integration tests are failing and its not clear to me why yet.

PR #5132, opened 2022-05-10
Allow non-LDAP user to change his own name and location

### LINKED ISSUE #4652: User cannot change his own name, location or login
Currently, users without special permissions cannot change their own name, location or login. Instead, he requires the `editUser` authority, which in fact is intended for editing _other_ users. Instead, every user should be allowed to change this personal information without the need for some specific permissions.

Bildschirmfoto 2021-09-08 um 10 26 13

solth:
Fixes #4652


henning-gerhardt (approved):
I was able to use the one and only database user to change is first name and surname and his location. LDAP based users (except administrators) could not change first name nor surname nor location with and without your changes.

PR #5156, opened 2022-05-16
Use the same label pattern "image number : order label" everywhere in the MetaDataEditor

### LINKED ISSUE #5031: Inconsistent application of image number
### Problem
In the structure tree of the metadata editor, the image number is shown besides the pagination information. According to issue #4038, this is correct. 
In the pagination frame and in the window to add a structure element, the internal order number is shown. 

The different image numbers are very confusing and leads to wrong pagination or structuring. See the GIF in Example below. 

### Solution
The image numbers should be applied consistently according to the ones in the structure tree. 

A possible solution might be described in https://github.com/kitodo/kitodo-production/issues/4038#issuecomment-940097340

### Example
In the GIF, the image number in the structure tree, in the pagination frame and in the window "New docstruct" must be compared. 

PaginierungAbweichung_2

thomaslow:
Fixes #5031

The following pull request changes how labels of the gallery thumbnails, pagination list items and add-doc-struct dialog are generated. Instead of using `<order> : <order label>`, all labels are now generated the same way as in the logical structure tree, meaning using `<short image file number> : <order label>`.

Before:

Screenshot from 2022-05-16 17-54-00

After:

Screenshot from 2022-05-16 17-51-19


thomaslow:
The pull request was updated to also show the correct label in the header of the preview panel.

bitmap

In case there is no page currently selected, the header now also shows a message "no selection":

Screenshot from 2022-05-18 09-52-39

PR #5165, opened 2022-05-19
Usability MetaDataEditor: Logical Structure Options Menu

### LINKED ISSUE #4599: Enable folding of image-information in the structure tree
### Problem
In processes with many images and complex structures, it is very difficult to keep the overview of the structure elements. Especially, when all images are assigned and only the metadata or the position of the structure elements are examined for typos, missing elements, .... It is possible to fold each structure element manually, but this is in some cases very cumbersome. 

### Solution
There should be a button, which hides all image information in the structure tree in the respective structure elements. 

### Example
FoldingElements

thomaslow:
Fixes #4599 

This pull request proposes to add a small options menu to the logical structure panel, which allows to customize the logical structure tree. Two options are available at the moment:
- "show first and last media label" will append the label of the first and last media for all logical structure nodes
- "show media" allows to hide media nodes similar to the "separate structure" workflow option, but without actually showing the physical structure tree

Since I'm not super familiar with Kitodo's terminology and style guidelines, please feel free to suggest how these options should be labeled or whether the User Interface styling should be improved in some way.

Screenshot:
Screenshot from 2022-05-19 11-14-35

This pull depends on and includes the pull request:
- #5147 

For reviewers: The commit c7c9e9200f6f3a0e79376da82fde44a58975bcfd and b7a8a4d21bff061fa5dae43b37d231ce3fa0daa3 contains changes relevant to this pull request.


andre-hohmann:
@thomaslow : Thanks for the great proposal. 

Regarding the labels, i have the following suggestions: 
- "show first and last media label" 
-> "_Zeige zugewiesene Seiten_" - "_Show assigned pages_"
- "show media" 
-> "_Zeige nur Strukturelemente_" - "_Show only structure elements_"

I hope, i understood the functions correctly. The labels are partly derived from the 2.x version and should be familiar to the users. 
I know that "Seiten" and "pages" are too narrow, as they do not fit to audio, video, ...,, but that should be discussed generally as there are other fields, labels, ... which are affected. 

Regarding the interface styling, i need some more time, but at the first glance it looks good. It is the right position next to the option to change label of the structure elements.

andre-hohmann:
I tried it now myself and from my point of view, the interface styling is good. Considering the minimal demands regarding the styling in the issue, it is difficult to implement it better. 

The function of the button could be better described, because it is difficult to find it intuitively. Otherwise, it is good, that it does not take much space. 
I have no better ideas and in my opinion, ideas should be discussed in the community and the proposals should be created by the users. 

I am glad about the functionality, which helps a lot to review the integrity of the structure elements: 
ReduzierungStrukturbaum

Issue #5123, opened 2022-05-09
Deleting user defined filter results in error entry


henning-gerhardt:
If you delete an existing user defined filter the following error message get be logged but not displayed to the user

```
[ERROR] 2022-05-09 08:57:52,806 [http-nio-127.0.0.1-8080-exec-42] org.kitodo.production.services.data.UserService - Cannot not remove filter from user with id 182
org.kitodo.data.exceptions.DataException: org.kitodo.data.database.exceptions.DAOException: javax.persistence.OptimisticLockException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [org.kitodo.data.database.beans.Filter#7564258]
        at org.kitodo.production.services.data.base.SearchService.remove(SearchService.java:365) ~[classes/:3.4.3-SNAPSHOT]
        at org.kitodo.production.services.data.UserService.removeFilterFromUser(UserService.java:461) ~[classes/:3.4.3-SNAPSHOT]
        at org.kitodo.production.services.data.UserService.removeFilter(UserService.java:400) ~[classes/:3.4.3-SNAPSHOT]
        at org.kitodo.production.forms.BaseForm.removeFilterFromUser(BaseForm.java:189) ~[classes/:3.4.3-SNAPSHOT]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
        at org.apache.el.parser.AstValue.invoke(AstValue.java:247) ~[tomcat9-jasper-el-9.0.43.jar:9.0.43]
        at org.apache.el.MethodExpressionImpl.invoke(MethodExpressionImpl.java:267) ~[tomcat9-jasper-el-9.0.43.jar:9.0.43]
        at org.jboss.weld.util.el.ForwardingMethodExpression.invoke(ForwardingMethodExpression.java:40) ~[weld-servlet-2.4.8.Final.jar:2.4.8.Final]
        at org.jboss.weld.el.WeldMethodExpression.invoke(WeldMethodExpression.java:50) ~[weld-servlet-2.4.8.Final.jar:2.4.8.Final]
[...]

PR #5173, opened 2022-05-24
Update user after adding or removing a filter

### LINKED ISSUE #5124: Deleting user defined filter is visible until next log in
If you delete an user defined filter, the filter is still usable and visible until the user logged out and logged in back.

**Goal:** The deleted filter should vanish after deleting.

### LINKED ISSUE #5125: Adding user defined filter is invisible until next log in
If you create a new user defined filter the filter is not usable nor visible until the user logged out and logged in back

**Goal:** New created user defined filter should be usable and visible after creating.

solth:
Fixes #5124 
Fixes #5125 
I was unable to reproduce #5123 so I cannot say if its fixed or not.


henning-gerhardt (approved):
Changes are okay and adding and / or removing a user filter is working as expected. All 3 issues #5123 , #5124 and #5125 are fixed with this changes in this pull request.

PR #5195, opened 2022-06-07
Fix action to remove symbolic links from home directory for selected processes

### LINKED ISSUE #4554: Remove link from home directory as "possible action" does not work
### Problem
In the menu "possible actions" it is possible to chose "Remove link from home directory". If it is applied to several processes in a list, an error message and a confirmation message is shown: 

- Error loading 'Process' with ID 0
- Removed all folder from home directory: fertig/

However, the links are not removed. 

It is possible to remove the link for each process by the action button on the right side. 

### Solution
It should be possible to remove the links from the home directory with the action. 

### Examples
homeverzeichnis02

homeverzeichnis01

thomaslow:
Closes #4554 

Issue #4554 reports that the action `Remove link from home directory` is not working as expected. Currently, the action triggers code that only removes symbolic links in the `fertig/` subdirectory (configuration property `doneDirectory`) of a users home directory. Symbolic links of processes created via the action `Link within home directory` are not removed.

This pull request changes the behavior of this action to remove symbolic links for all selected process from the home directory of the current user.

This means that the previous behavior (removing symbolic links from the `fertig/` directory) can no longer be triggered by a user. If no longer needed, the corresponding code could be removed (which is not part of this pull request).


henning-gerhardt:
> The error should be gone now.

Thank you. I will test this as soon as possible.

> There was an old jsf statement that did not make sense, probably from 2.x times.

This could be possible.

henning-gerhardt:
Your latest change to this pull request did fix the appearing error message. Works for me now without an issue.

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -350,6 +350,7 @@ digitalCollections=Digitale Kollektionen
 directory=Verzeichnis
 directoryRemoved=Verzeichnis aus Benutzerhome entfernt
 directoryRemovedAll=Alle Verzeichnisse aus Benutzerhome entfernt
+directoryRemovedSelected=Alle ausgewählten Verzeichnisse aus Benutzerhome entfernt
```
Kathrin-Huber, line 353:
```suggestion
directoryRemovedSelected=Alle ausgew\u00E4hlten Verzeichnisse aus Benutzerhome entfernt
```

please encode

Kathrin-Huber (changes requested):

PR #5218, opened 2022-07-05
Fix checkstyle violations and lower number of allowed violations

henning-gerhardt:
Fix a lot of current allowed checkstyle violations by extracting code into its own methods.


henning-gerhardt:
It would be nice that after merging this pull request all open pull requests are rebased against master branch to show which open pull requests have checkstyle violations.

Issue #4080, opened 2020-11-03
Too long texts in ReadTheDocs


matthias-ronge:
Our menu texts are too long, so the control line breaks into two lines and covers the upper part of the help page:

![Screenshot of Kitodo.Production help](https://user-images.githubusercontent.com/3040657/97959611-05948080-1db0-11eb-8591-c0db988f5d8e.png)

With short texts, it looks right:

![Screenshot of correct example](https://user-images.githubusercontent.com/3040657/97959617-07f6da80-1db0-11eb-8760-796702dcb5bd.png)

Erikmitk:
My preferred solution would be to change/adjust the theme. https://www.mkdocs.org/user-guide/styling-your-docs/

Erikmitk:
Actually… it looks like the ReadTheDocs (RTD) infrastructure is at fault. The long values aren't even supposed to be there. README has no place in the navigation. Something fishy in the build process is going on.

My fork builds cleanly in RTD. The difference is that I have the Python Interpreter for building set to 3.x. When I try to change the official documentation to the same settings it fails to build. I suspect an outdated mkdocs dependency due to the error message. I tried [wiping the build environment](https://docs.readthedocs.io/en/stable/guides/wipe-environment.html) but that didn't change a thing. 😞

Erikmitk:
https://docs.readthedocs.io/en/stable/intro/getting-started-with-mkdocs.html
> We strongly recommend to pin the MkDocs version used for your project to build the docs to avoid potential future incompatibilities.

We might have to add a configuration.

(Although it's super weird that my fork builds the documentation clean out of the box.)

bertsky:
> My fork builds cleanly in RTD.

@Erikmitk, what do you mean by that? Do you have a readthedocs.io deployment for your fork (where?), or is it just your local `mkdocs serve -t readthedocs` (or just `mkdocs serve`) that looks correct?

IMO the issue here is actually much broader: currently, in the RTD deployment, none of the images in the user documentation resolve correctly. I attempted #5221 to solve that (by prefixing all references with `/docs`), only to find out the image paths were correct, but the deployment was not.

So this is quite urgent, actually. (Also, running mkdocs should be **documented** and probably put into the CI...)

Erikmitk:
@bertsky This is quite an old issue but I just checked if it still persists. I have access to two projects on RTD:

[...]

PR #5221, opened 2022-07-06
docs: fix all img src paths

bertsky:
None of the images in the generated user documentation resolve correctly (neither on Github or in the ~~Github Pages~~ ReadTheDocs deployment). I found no existing issue on that. This just auto-replaces all such paths.


bertsky:
Wait. When generating mkdocs locally, the previous image paths (without leading `/docs`) do work.

So the question becomes: why is the current deployment on readthedocs.io broken? Seems to be #4080 actually...

PR #5222, opened 2022-07-07
Fix new checkstyle violation and lower amount of allowed violations

henning-gerhardt:
Fixes introduced checkstyle error:

```
src/main/java/org/kitodo/export/ExportDms.java:[137] (coding) OverloadMethodsDeclarationOrder: Overload methods should not be split. Previous overloaded method located at line '94'.
```


henning-gerhardt:
Try to fix snyk integration.

PR #5225, opened 2022-07-07
Docs fix img src paths

bertsky:
Follow-up to #5221, see https://github.com/kitodo/kitodo-production/issues/4080 for discussion.

This adds a docs/requirements.txt so that the readthedocs.org deployment will install a compatible version of `mkdocs`.

Moreover, to fix the image references in the user guide (which are broken in readthedocs.org deployments as well as Github pages, but somehow magically work when just building locally with `mkdocs serve`), making them resolve in all versions, I prefixed the absolute paths with `../../..`. Not sure if this is correct – perhaps it _is_ an mkdocs / RTD / GH pages configuration issue, after all.


Kathrin-Huber:
We are not using read the docs anymore, don't we @solth?

solth:
> We are not using read the docs anymore, don't we @solth?

I am sorry, I don't know.

matthiaskissler:
Hi, the RTD documentation is obsolete and will be moved to Github Wiki in the near future; the user documentation part has already found a new home there, I still have to move the developer chapters to a fitting place within the Wiki pages.

bertsky:
@matthiaskissler understood, thanks for clarifying!

I presume we can therefore close?

matthiaskissler:
Yes, this issue can be closed, thank you

PR #5227, opened 2022-07-08
Update bug_report.md

solth:
Remove "Smartphone" section and add "Release" section

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -23,15 +23,12 @@ A clear and concise description of what you expected to happen.
 **Screenshots**
 If applicable, add screenshots to help explain your problem.
 
+**Release**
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -23,16 +23,13 @@ A clear and concise description of what you expected to happen.
...
 
+**Release**
+Please specify the version of Kitodo.Production in which the bug was encountered [e.g. 3.3.1, 3.4.3 etc]
+
 **Desktop (please complete the following information):**
  - OS: [e.g. iOS]
  - Browser [e.g. chrome, safari]
  - Version [e.g. 22]
 
-**Smartphone (please complete the following information):**
```
henning-gerhardt, line 31: The smartphone section must not be removed but maybe only renamed to "non-desktop (tablet, mobile phone, ..)" or merge both sections to one (f.e. device) and add a field for entering the used device (desktop, tablet, mobile phone, ...) so it is maybe more useful in future if other devices are present.

Issue #5219, opened 2022-07-06
Problems with the new import configuration


BartChris:
I have some problems transferring the current xml based configuration to the new import configuration (https://github.com/kitodo/kitodo-production/pull/5038). Right now i cannot make some catalogs work which worked before. It seems that some of the features got lost which originally ( https://github.com/kitodo/kitodo-production/pull/3374) allowed for the usage of arbitrary search interfaces. It would be good if those features could be ported to the new interface as well to make use of catalogs which worked before

I have some questions in this context:

-  The xml based configuration had a way to specfiy custom url parameters. This allowed to add parameters to the URL in additon to those which are given by the user query. How can something like that be done in the new interface?
```
<urlParameters>
    <param name="format" value="oai_mods" />
</urlParameters>
```
- What does Query delimiter actually stand for and what has to go there?

![grafik](https://user-images.githubusercontent.com/7516499/177597311-b2213e73-9358-4ecc-8783-3dbfad249700.png)

- Is it ensured that the mapping files are actually applied in the right order if i want to chain multiple transformations?

![grafik](https://user-images.githubusercontent.com/7516499/177595143-f89f2f33-c39d-4404-9187-b3fe9f476f00.png)

- How can a prefix for the identifier be specfied:

`<identifierParameter prefix="prefix:" value="id" />`

solth:
> * What does Query delimiter actually stand for and what has to go there?
> 
> 
> ![grafik](https://user-images.githubusercontent.com/7516499/177597311-b2213e73-9358-4ecc-8783-3dbfad249700.png)

The query delimiter is an optional character in which the query part of the URL can be enclosed. This was necessary for some SRU interfaces like the "LfULG - DiGAS" where the query would need to be enclosed in `"` characters in order for the interface to process the query successfully.

solth:
> * Is it ensured that the mapping files are actually applied in the right order if i want to chain multiple transformations?
> 
> 
> ![grafik](https://user-images.githubusercontent.com/7516499/177595143-f89f2f33-c39d-4404-9187-b3fe9f476f00.png)

Yes, the mapping files are saved in a list internally, preserving the order in which they have been assigned to the import configuration and ensuring they are applied in the correct order to the imported metadata file. In fact, an exception will be thrown if you try to save a an ImportConfiguration where 
- the input metadata format of the _first_ of a sequence of mapping files does not correspond to the metadata format of the ImportConfiguration, or
[...]

PR #5038, opened 2022-03-18
Import Configurations

solth:
This pull request introduces a new `ImportConfiguration` class that holds information about specific ways to import metadata in Kitodo.Production, thus replacing the current configuration file `kitodo_opac.xml`.

Each `ImportConfiguration` can have one of the following types:
- `OPAC_SEARCH`: corresponds to `catalogue` entries in the `kitodo_opac.xml` with `<fileUpload>false</fileUpload>` (or no `<fileUpload>` setting at all) and offers the same settings relevant for this type of configuration, like URL, search interface type (SRU, OAI etc.), search fields, default import depth etc.
- `FILE_UPLOAD`: corresponds to `catalogue` entries in the `kitodo_opac.xml` with `<fileUpload>true</fileUpload>` and offers the settings relevant for this type of configuration, like file format, metadata format and mapping file for included parent record
- `TEMPLATE_PROCESS`: represents a configuration where a specific template process is preselected as the source to copy metadata when creating a new process

ImportConfigurations can be created and edited via the GUI instead of having to edit an XML file on the server. This allows Kitodo system admins with appropriate permissions (also added in this pull request) to configure everything necessary via the frontend without depending on file system access on the server.
Bildschirmfoto 2022-03-18 um 12 15 25

Moving the import configurations from the `kitodo_opac.xml` configuration file to the database and allowing the user to edit it via the GUI brings the following advantages:
- prevent configuration errors by only allowing specific options for certain settings like "interface type" or "metadata format" (when using a configuration file, all kinds of wrong values can be entered that lead to unpredictable errors in the system)
- directly assign default import configurations to individual projects in order to always open the appropriate dialog and default settings when creating a process for that project (fixes #4252, fixes #3970, fixes #4236)
- reduce number of configuration files that need to be maintained (one of the goals during the development of Kitodo.Production 3, which so far has only been partly achieved)

Additionally, xslt mappings are now also saved as `MappingFile` entites to the database and mapped to import configurations. Mapping file entites now also save information about which input metadata format is transformed into which output metadata format. This way, the system can validate that the metadata format of the configured search interface or upload file fits the configured metadata format and is eventually transformed into the internal Kitodo metadata format.
Bildschirmfoto 2022-03-18 um 13 21 54

The configuration edit mask also shows tooltips for all import configuration input fields to support the user when editing configurations.
Bildschirmfoto 2022-03-18 um 12 15 59

**Known issues**:
[...]

PR #5231, opened 2022-07-08
Add input field for identifier prefix

### LINKED ISSUE #5228: ImportConfiguration is missing identifier parameter prefix configuration
**Bug description**
As reported by @BartChris in #5219 the new `ImportConfiguration` class is missing the `prefix` attribute for identifier parameters that were supported in the former `kitodo_opac.xml` catalogue configurations.

**Expected behavior**
The new `ImportConfiguration` class should support the `prefix` parameter for identifier parameters in the same way the old `kitodo_opac.xml` catalogue configurations did.

solth:
Fixes #5228 

(As it turned out, the database schema for import configurations already contained a field "idPrefix"; just the form to create or edit import configurations was missing a corresponding input field)

### DIFF
--- a/Kitodo/src/main/resources/messages/messages_de.properties
+++ b/Kitodo/src/main/resources/messages/messages_de.properties
@@ -571,6 +571,7 @@ importConfig.searchField=Suchfeld
 importConfig.searchFields=Suchfelder
 importConfig.searchFields.defaultField=Default-Suchfeld
 importConfig.searchFields.idSearchField=ID-Parameter
+importConfig.searchFields.idPrefix=ID-Prefix
 importConfig.searchFields.parentIdField=ID-Parameter der \u00DCberordnung
 importConfig.searchFields.selectDefaultField=Bitte Default-Suchfeld w\u00E4hlen
 importConfig.searchFields.selectIdSearchField=Bitte Suchfeld f\u00FCr Suche nach Identifier w\u00E4hlen
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
@@ -80,12 +80,28 @@
                                   noSelectionOption="true"/>
                     <f:selectItems value="#{importConfigurationEditView.searchFields}"/>
                     <p:ajax event="change"
+                            update="editForm:importConfigurationTabView:idParameterPrefix"
                             oncomplete="toggleSave();"/>
                 </p:selectOneMenu>
                 <p:commandButton id="idSearchFieldHelp" type="button"
                                  styleClass="help-button" icon="fa fa-lg fa-question-circle-o"/>
                 <p:tooltip for="idSearchFieldHelp"
                            value="URL parameter for unique identifiers. This parameter is used for importing individual records from the search interface."/>
+            </div>
+            <div>
+                <p:outputLabel for="idParameterPrefix"
+                               value="#{msgs['importConfig.searchFields.idPrefix']}"/>
+                <p:inputText id="idParameterPrefix"
+                             styleClass="input-with-button"
+                             value="#{importConfigurationEditView.importConfiguration.idPrefix}"
+                             disabled="#{importConfigurationEditView.idSearchField eq null}">
+                    <p:ajax event="change"
+                            oncomplete="toggleSave();"/>
+                </p:inputText>
+                <p:commandButton id="idParameterPrefixHelp" type="button"
+                                 styleClass="help-button" icon="fa fa-lg fa-question-circle-o"/>
+                <p:tooltip for="idParameterPrefixHelp"
+                           value="This optional string will be prepended to each ID when querying datasets by ID. This can be used to simplify querying OAI interfaces for example, where each identifier must be preceedded with a valid set identification."/>
             </div>
             <div>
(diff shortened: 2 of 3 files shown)


Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
```diff
@@ -80,13 +80,29 @@
...
+                             styleClass="input-with-button"
+                             value="#{importConfigurationEditView.importConfiguration.idPrefix}"
+                             disabled="#{importConfigurationEditView.idSearchField eq null}">
+                    <p:ajax event="change"
+                            oncomplete="toggleSave();"/>
+                </p:inputText>
+                <p:commandButton id="idParameterPrefixHelp" type="button"
+                                 styleClass="help-button" icon="fa fa-lg fa-question-circle-o"/>
+                <p:tooltip for="idParameterPrefixHelp"
+                           value="This optional string will be prepended to each ID when querying datasets by ID. This can be used to simplify querying OAI interfaces for example, where each identifier must be preceedded with a valid set identification."/>
```
Kathrin-Huber, line 104: please use message file here

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -571,6 +571,7 @@ importConfig.searchField=Suchfeld
 importConfig.searchFields=Suchfelder
 importConfig.searchFields.defaultField=Default-Suchfeld
 importConfig.searchFields.idSearchField=ID-Parameter
+importConfig.searchFields.idPrefix=ID-Prefix
```
henning-gerhardt, line 574: "Prefix" -> "Präfix" ... looks like you used a lot of English words one to one in German like "Default", "ID-Parameter" and maybe more.

Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
```diff
@@ -80,13 +80,29 @@
                                   noSelectionOption="true"/>
                     <f:selectItems value="#{importConfigurationEditView.searchFields}"/>
                     <p:ajax event="change"
+                            update="editForm:importConfigurationTabView:idParameterPrefix"
                             oncomplete="toggleSave();"/>
                 </p:selectOneMenu>
                 <p:commandButton id="idSearchFieldHelp" type="button"
                                  styleClass="help-button" icon="fa fa-lg fa-question-circle-o"/>
                 <p:tooltip for="idSearchFieldHelp"
                            value="URL parameter for unique identifiers. This parameter is used for importing individual records from the search interface."/>
```
henning-gerhardt, line 89: Translatable string would be here nice too.

henning-gerhardt (changes requested):
Translation is not to write English words in in German even they have a similar meaning ;)

Kitodo/src/main/resources/messages/messages_de.properties
solth: Thanks for pointing out the "ID-Prefix", I will change this. But what's wront about "ID-Parameter"? Parameter is a german word, too, I don't see the problem here. English terms like "Default" are also very establised in the German language, so I thought it would be appropriate here. Or would you prefer something like "Standardsuchfeld"?

Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
solth: Yes, I know. I haven't translated the tooltips for all those `ImportConfiguration` edit form fields, yet. See my description in #5038, I was planning to add the translations for all those tooltips in a separate pull request, probably in a new "tooltip" bundle.

Kitodo/src/main/resources/messages/messages_de.properties
henning-gerhardt: ID Parameter looks for me wrong but I did not have for now a better wording. Maybe ask a librarian or other real user which wording is better.

Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
solth: Please see  #5235

Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
solth: See #5235

henning-gerhardt: Should be okay if resolving the mentioned issue is done in near future - at least before next release.

henning-gerhardt (approved):
Changes looks good but I did not have a catalogue system which is needing this. Maybe @BartChris can verify this as he has such a system.

BartChris:
Thanks, the identifier is correctly added to the constructed URL, so this change should be good. I can however only test it with an proxy interface because i need additional custom parameters for my tested custom interface to query it directly (see: https://github.com/kitodo/kitodo-production/issues/5219#issuecomment-1177479957)

PR #5232, opened 2022-07-08
Save mapping file order for import configurations

### LINKED ISSUE #5229: Order of mapping files in import configuration is not saved
**Bug description**
As reported by @BartChris in #5219 the order of `MappingFiles` is not saved to the table `importconfiguration_x_mappingfile`. Therefore it is not guaranteed that the files are applied to the imported XML in the correct order.

**To Reproduce**
1. Create a new `ImportConfiguration`
2. Assign multiple mapping files to the new `ImportConfiguration`
3. Save the `ImportConfiguration`
4. Check the order of the mapping files in the database table `importconfiguration_x_mappingfile`

**Expected behavior**
The order of the `MappingFiles` should be explicitely saved to the one configured by the user in the `Edit import configuration` form.

solth:
Fixes #5229 

**Note** @BartChris : since the order of mapping files in import configurations wasn't saved explicitly to the database yet, I had to initialize the new `sorting` column statically with `0`. This means that for import configurations with multiple mapping files that you created before this pull request, you need to edit the configuration again and re-assign the mapping files in the correct order. Then they will be saved correctly to the database.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ImportConfiguration.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ImportConfiguration.java
@@ -25,6 +25,7 @@ import javax.persistence.ManyToMany;
 import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
 import javax.persistence.OneToOne;
+import javax.persistence.OrderColumn;
 import javax.persistence.Table;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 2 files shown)


henning-gerhardt:
@solth : I did already a review (which was good) and suggested to review this changes by BartChris. Why should I now review this pull request again as no changes was made since my last review?

solth:
> Why should I now review this pull request again as no changes was made since my last review?

Because you didn't approve the pull request in your first review and @BartChris indeed confirmed that the changes work as intended.

henning-gerhardt (approved):

PR #5233, opened 2022-07-08
Add amount of meta data to index to improve performance

henning-gerhardt:
Improve performance for creating the excel file inside the process list with thousand of processes. For archive this the only needed data from the meta data file is moved into the index as every other data for the excel file is already taken from the index. Instead of reading thousand of meta data files which consumes a lot of time the needed data is available after search.

This is a co-work from @Kathrin-Huber  (general logic) and @henning-gerhardt  (test, cleanup, ...) and commit history is squashed as this changes needs a lot of round tripping.

After merging this pull request the index for processes must be rebuild.


solth (changes requested):
Since I lack the amount of processes to verify the performance improvement when generating Excel files I can only confirm that the functionality still works like before.

The code changes also look sound, apart from a few unused imports that should be removed.

solth (approved):

PR #5247, opened 2022-07-15
Fix error with file upload import configurations

solth:
Added missing converter to fileupload dialogs import configuration select menu to prevent error.


henning-gerhardt:
Then both pull requests depends on each other but this is not mentioned :-( I will apply the other pull request too.

henning-gerhardt:
We tested this and the already mentioned pull request #5252 together and both issues are gone. It would be nice if both pull requests had some automatic tests to prevent such kind of issues in future as manual testing consumes more time then writing tests.

oliver-stoehr (approved):

PR #5252, opened 2022-07-18
Enable file format select menu and restrict available formats to XML

### LINKED ISSUE #5251: Return format not saved for import configurations of type "file upload"
**Describe the bug**
When creating a new import configuration of type "File Upload", the "Return file format" is not saved because the corresponding form input field is deactivated (deactivated fields are skipped during form submission). The field is deactivated because Kitodo currently only supports XML as return format anyway.
**Note**: this does not apply to import configurations of type 'Catalog import' and 'Process template', because the 'Return format' input field is not deactivated for those import configuration types!

**To Reproduce**
1. Go to page 'Projects'
2. Click on 'New', then 'Import Configuration'
3. Select 'File upload' as configuration type
4. Fill out all mandatory fields and save the configuration
5. Check the database: column 'return_format' in table 'importconfiguration' will be `null` for new import configuration object!
6. Trying to use the new import configuration will lead to an `NullPointerException`!

**Expected behavior**
The file format should always be saved in the import configuration and the `NullPointerException` upon usage should be prevented!

**Screenshots**
Bildschirmfoto 2022-07-18 um 09 36 07

**Release**
3.4.4-SNAPSHOT

 - OS: MacOS
 - Browser Firefox
 - Version 102

solth:
Fixes #5251

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ImportConfigurationEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ImportConfigurationEditView.java
@@ -126,12 +126,12 @@ public class ImportConfigurationEditView extends BaseForm {
     }
 
     /**
-     * Get file formats.
+     * Get file formats. Currently only XML is supported.
      *
      * @return file formats
      */
     public List<FileFormat> getFileFormats() {
-        return List.of(FileFormat.values());
+        return Collections.singletonList(FileFormat.XML);
     }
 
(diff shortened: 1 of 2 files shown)


henning-gerhardt:
See my comment in https://github.com/kitodo/kitodo-production/pull/5247#issuecomment-1189863550

oliver-stoehr (approved):

PR #5253, opened 2022-07-18
Add edit button for search fields of ImportConfigurations

### LINKED ISSUE #5237: ImportConfiguration: "Edit" button for "SearchFields"
When creating `ImportConfiguration` objects of type "Catalogue search", the user has to add at least one `SearchField` to the configuration. `SearchField`s are used to query the configured search interface.

Currently, existing `SearchField`s cannot be edited, but just deleted:
Bildschirmfoto 2022-07-11 um 20 23 08
So in order to fix an error in an existing `SearchField` it first needs to be deleted and than re-added again. Instead, it should be possible to edit existing `SearchFields` to fix small errors easier.

solth:
Resolves #5237


Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/updateSearchFieldDialog.xhtml
```diff
@@ -0,0 +1,91 @@
...
+                                         value="#{updateSearchFieldDialogView.searchField.label}"/>
+                            <p:commandButton id="searchFieldLabelHelp" type="button"
+                                             styleClass="help-button" icon="fa fa-lg fa-question-circle-o"/>
+                            <p:tooltip for="searchFieldLabelHelp"
+                                       value="Label displayed to the user when importing processes using the current OPAC import configuration"/>
+                        </div>
+                    </p:row>
+                    <p:row>
+                        <div>
+                            <p:outputLabel for="searchFieldValue" value="Search field value"/>
```
IkramMaalej, line 45: Translation is missing here too

IkramMaalej (approved):

PR #5254, opened 2022-07-18
Show only regular files in list of available xslts for mapping file

### LINKED ISSUE #5240: Strange entry "xslt" on adding / updating mapping file entries
**Describe the bug**
If you add a new mapping file entry at the end of the file list the entry "xslt" appears. This entry even appears if you edit an existing mapping entry. This entry appears even if there is no file in the XSLT directory, with only .xsl files, with only .xslt files or mixed .xsl and .xslt files.

**To Reproduce**
1. Go to "Projects" -> Mapping files
2. Edit an existing entry
3. Scroll down the file list

or
1. Go to "Projects" 
2. Click on "New" -> "New mapping file"
3. Scroll down the file list

**Expected behavior**
Maybe remove this entry as this entry is not needed.

**Screenshots**
Screenshot_2022-07-12_12-58-09

 - OS: Devuan 4
 - Browser: Firefox
 - Version: 102

solth:
Fixes #5240


henning-gerhardt (approved):
Changes are okay (would be perfect with a test) and ignore directories with a xsl or xslt ending but include symbolic links which link to real files.

PR #5276, opened 2022-08-03
Fix update thumbnails-overlay after paginating

### LINKED ISSUE #5275: Thumbnails overlay in gallery panel do not update after paginating the pages
**Describe the bug**
After paginating the pages, the thumbnails overlay in gallery panel do not update and keep the old order labels.

**To Reproduce**

Aug-03-2022 12-56-56

**Expected behavior**
Thumbnails in Gallery panel should have  the right order label after the pagination is done.

**Release**
3.4.4-SNAPSHOT

IkramMaalej:
Fixes #5275

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -3028,7 +3028,7 @@ Column content
     margin-right: 3px;
 }
 
-.thumbnail-overlay {
+.ui-widget .thumbnail-overlay {
     position: absolute;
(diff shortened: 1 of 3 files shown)


oliver-stoehr (changes requested):
It would be better to only update the components that actually change, instead of updating the whole form. You should only update `@(.thumbnail-overlay)` to update the labels in the overlay with the new pagination. 

This is currently not possible with the overlay being rendered by `h:panelGroup` elements. Bu you could use `p:outputPanel` instead.

oliver-stoehr (approved):

PR #5287, opened 2022-08-12
Use search delimiter even on hierachy search

henning-gerhardt:
While try to fix the access to an other catalogue I found out that the defined search delimiter was not used on hierarchy search. This pull request is fixing this.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
@@ -640,7 +640,8 @@ public class ImportService {
         // ################ IMPORT #################
         importModule = initializeImportModule();
         DataRecord dataRecord = importModule.getFullRecordById(
-                createDataImportFromImportConfiguration(importConfiguration), identifier);
+                createDataImportFromImportConfiguration(importConfiguration),
+                getSearchTermWithDelimiter(identifier, importConfiguration));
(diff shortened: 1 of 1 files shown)


solth (approved):
This does work for hierarchical import, but I think the query delimiter needs to be incorporated for automatic child process import as well, even if we do not have a search interface that supports both, child process import _and_ query delimiters, yet.

henning-gerhardt:
Can you give me a hint where to find this? As my catalogue with the need of search delimiter did not support importing of child elements and my catalogue with child import support did not work as I can not switch this button on did not need the search delimiter and there are no tests which need one or the other thing it is difficult to find the right place for this. Or is your comment to interpret to open a new issue for this?

solth:
> Can you give me a hint where to find this? As my catalogue with the need of search delimiter did not support importing of child elements and my catalogue with child import support did not work as I can not switch this button on and did not need the search delimiter and even there are no tests which need one or the other thing it is difficult to find the right place for this. Or is your comment to interpret to open a new issue for this?

I think this needs to be incorporated in https://github.com/kitodo/kitodo-production/blob/037a345dfafb932edda0dec6f02902c5e5cdb114/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java#L124
I opened a new issue to keep track of this problem: #5319

Issue #5286, opened 2022-08-11
Multiple xlink:to elements for one xlink:from element in mets:structLink


andre-hohmann:
**Describe the bug**
In Kitodo.Production 3.x multiple `xlink:to` elements are assigned to one `xlink:from` element in `<mets:structLink>`. 
This happens for the element that are assigned to several structure elements. 
In the DFG-Viewer, this leads to the multiple depiction of the metadata of the selected elements. See example in "Additional context". 

**To Reproduce**
1. Examination of the exported METS file of a process with several structure elements on different levels. 

**Expected behavior**
One `xlink:to` element should be assigned to one `xlink:from` element only once

**Release**
- 3.4.4-SNAPSHOT

**Desktop (please complete the following information):**
- OS: Windows 10
- Chrome 103.0.5060.134 (Offizieller Build) (64-Bit)

**Additional context**

1 Example in DFG-Viewer (SUB): 
- https://dfg-viewer.de/show?id=9&tx_dlf%5Bid%5D=http%3A%2F%2Fmets.sub.uni-hamburg.de%2Fkitodo%2FPPN1773384848&tx_dlf%5Bpage%5D=471

The metadata fields "Titel", "Autor", "Einrichtung", ... are doubled, when structure elements on lower level are selected - in this case: Image 471. 

2 Comparison of the `<mets:structLink>` of another process (SLUB)
- a `<mets:structLink>` in Kitodo.Production 3.x - internal METS-file

<details>

``` xml
    <mets:structLink>
        <mets:smLink xlink:to="PHYS_0001" xlink:from="LOG_0001"/>
        <mets:smLink xlink:to="PHYS_0002" xlink:from="LOG_0001"/>
        <mets:smLink xlink:to="PHYS_0003" xlink:from="LOG_0001"/>
        <mets:smLink xlink:to="PHYS_0004" xlink:from="LOG_0001"/>
        <mets:smLink xlink:to="PHYS_0005" xlink:from="LOG_0004"/>
        <mets:smLink xlink:to="PHYS_0006" xlink:from="LOG_0005"/>
        <mets:smLink xlink:to="PHYS_0007" xlink:from="LOG_0005"/>
        <mets:smLink xlink:to="PHYS_0008" xlink:from="LOG_0007"/>
    </mets:structLink>
```
</details>

- b `<mets:structLink>` in Kitodo.Production 3.x - external METS-file

<details>

``` xml
   <mets:structLink>
[...]

PR #5288, opened 2022-08-15
Avoid duplicated links in the exported METS by using a LinkedHashSet

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/5286


matthias-ronge (approved):
Looks correct

PR #5290, opened 2022-08-17
Add 'prestructured import' flag to MappingFile class

### LINKED ISSUE #5273: Add 'prestructured import' as flag to MappingFiles
The option "prestructured import" allows adopting existing structures from imported documents. This is achieved by delegating the creation of METS structures from internal Kitodo functions to the configured `MappingFile`s. These can then contain rules to either create _new_ METS structures or to copy existing structure maps from imported METS documents.

Currently the flag "prestructured import" is configured for `ImportConfiguration`s. This flag should also be added to the `MappingFile` class to mark whether a specific XSLT file creates METS structures or not. This would also improve configuration stability by verifiying that the flag value of the last `MappingFile` of a `ImportConfiguration` corresponds to the configured flag in the `ImportConfiguration` itself.

solth:
Fixes #5273


Kitodo/src/main/java/org/kitodo/production/forms/validators/ImportConfigurationMappingValidator.java
```diff
@@ -49,41 +51,49 @@ public void validate(FacesContext facesContext, UIComponent uiComponent, DualLis
...
+
+    private void validateLastMappingFile(MappingFile mappingFile, Boolean prestructuredImport) {
+        // last mapping file must map to KITODO internal metadata format
+        if (!Objects.equals(mappingFile.getOutputMetadataFormat(), MetadataFormat.KITODO.name())) {
+            throw new ValidatorException(new FacesMessage(FacesMessage.SEVERITY_ERROR, Helper.getTranslation(
+                    "importConfig.validation.error.lastMappingFormat", mappingFile.getTitle(),
+                    mappingFile.getOutputMetadataFormat()), null));
+        }
+        // 'prestructured import' flag of mapping file must be equal to corresponding flag in import configuration
+        if (!Objects.equals(mappingFile.getPrestructuredImport(), prestructuredImport)) {
```
markusweigelt, line 94: Is this a convention to use Object.equals for comparing Booleans? If not I would suggest to check them without Object.equals.

solth: Since `Boolean` objects can be `null` - in contrast to the primitive type `boolean` - I prefer using `Objects.equals`, because it includes a null check to prevent null pointer exceptions.

markusweigelt: Yes, you are right, it is not a primitive type. Take back my comments here. Did not say anything. ;)

markusweigelt (approved):

PR #5291, opened 2022-08-17
Add message keys for tooltips

### LINKED ISSUE #5235: Add translations for import configuration tooltips
**Problem description**
#5038 added a form to create and edit `ImportConfiguration`s. The input fields of this form have tooltips that are displayed when hovering over the "?" icons next to the fields. These tooltip are currently part of the template files and only exist in English.

The tooltips should be translated and added to the message files.

solth:
Resolves #5235


henning-gerhardt (changes requested):
I did only a read review without looking if the message appear on the right position.

henning-gerhardt (approved):

PR #5292, opened 2022-08-17
Delete existing data editor settings for tasks of a changed workflow

BartChris:
When changing an already activated workflow (see: https://github.com/kitodo/kitodo-production/pull/5206) the template  tasks constructed based on those workflow are deleted and then recreated. This however does not work if there are existing data editor setting for a specific task. (Mysql table `dataeditor_setting`)

In this case the change of the template tasks fails because of a foreign key constraint. 

```sql
SqlExceptionHelper - Cannot delete or update a parent row: a foreign key constraint fails (`kitodo`.`dataeditor_setting`, CONSTRAINT `FK_dataeditorsetting_task_id` FOREIGN KEY (`task_id`) REFERENCES `task` (`id`))
```

The user interface reports the following error:

grafik

The pull requested tries to adress this by deleting all data editor settings related to template tasks which are changed during a workflow update.
The disadvantage is that all editor settings for this task get lost, but since one cannot say what exactly has changed, it probably serves no purpose to preserve those settings.

**Edit**: An alternative approach would be  to handle this on the database level by specifying "ON DELETE CASCADE" for this foreign key as `dataeditor_setting` does not hold critical data. The same setting could be applied for the foreign key pointing to the `user` table.


solth:
I would prefer not to lose the DataEditorSettings and think we should strive for fixing the mentioned problem without losing any data, if possible. Shouldn't it be possible to first deatch the DataEditorSettings from the old task object before it is deleted and then attech them to the new, re-created task?

BartChris:
That would be good, i agree. The problem is that it would be very hard to identify the same task if there were other tasks added in between or if a name of a task was changed. In the end this would mean some mechanism to make sure that the new task is attached to the right old settings.

solth:
You are right. If the user completely changes the sequence of tasks it could be impossible to re-assign the data editor settings to the original tasks. Therefore I take back my first comment about trying to keep the data editor settings.

solth:
On second thought I am not convinced that the current proposal is optimal, since users will loose there DataEditorSettings for no apparent reason. Even though these settings aren't critical for the system to work, it could easily be conceived as a bug in the system when previously saved settings are suddenly lost.

Perhaps a popup dialog could be shown when changing an active workflow in cases where DataEditorSettings exist that informs the workflow engineer that saving the workflow will reset user settings and that the users should be informed about this. The engineer can than decide whether he wants to abort changing the workflow or to change the workflow nonetheless and accept the loss of user data.

Kitodo/src/main/java/org/kitodo/production/services/data/DataEditorSettingService.java
```diff
@@ -68,13 +68,34 @@ public Long countResults(Map filters) throws DAOException {
...
+    }
+    
     private List<DataEditorSetting> getByUserAndTask(int userId, int taskId) {
         Map<String, Object> parameterMap = new HashMap<>();
         parameterMap.put("userId", userId);
         parameterMap.put("taskId", taskId);
         return getByQuery("FROM DataEditorSetting WHERE user_id = :userId AND task_id = :taskId ORDER BY id ASC", parameterMap);
     }
 
+
```
henning-gerhardt, line 97: Please remove this non useful empty lines.

henning-gerhardt (changes requested):
Your changes looks good and are working for me (in a simple data world with not many settings) - so please provide tests which shows that your changes do exactly this what they want to do and nothing else as removing in general and removing settings from users are dangerous.

solth:
@BartChris please add a test as requested by @henning-gerhardt

BartChris:
Sorry for the delay, i will add those tests next weeks. 

On a general note on the problem of preserving old state (editor settings) attached to a task:
While thinking about the problem of workflow changes in general (https://github.com/kitodo/kitodo-production/discussions/5351) i was also thinking again about the possibility to compare older with newer workflow tasks as this would be necessary to eventually preserve the editor settings. In general one could compare the tasks either by their name or by their workflowId. (If you you do not want to compare every single property of a task). But i am also not very convinced, if this (see also discussion thread) gives any  reliable indication if the editor settings should be preserved or not.

henning-gerhardt (commented):
Thank you for adding a test. But can you add a second test or extend your existing test for the scenario that a user has data settings for 2 workflows / templates and only one is changed so that one get deleted and the other stay remain?

BartChris:
I tried to make it more clear what is happening in the test.  The test should do, what you want it to do. 
The first template has a predefined workflow, tasks and dataeditor settings. The workflow gets replaced by a new workflow thereby updating all the tasks of this template.
The second template does not have tasks defined initially so i generate a task for that template and assign data editor settings to this task. Those settings are preserved.

henning-gerhardt (approved):

solth:
@BartChris One little Codacy problem still remains. Can you fix that issue? Then I will merge this pull request.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowForm.java
```diff
@@ -183,6 +189,26 @@ private void updateTemplateTasks() throws DAOException, IOException, WorkflowExc
         }
     }
 
+    /**
+     * Check if there are data editor setting for the tasks of the current workflow.
+     *
+     * @return true if one of the tasks has data editor settings defined
+     */
+    public boolean hasWorkflowDataEditorSettingsDefined() {
```
solth, line 197: Sorry for having another change request but I just noticed that you added this new function in the `WorkflowForm` class contains a whole bunch of loops and condition checks. Although we have many places where form classes/beans still contain a of such business logic, we are trying to avoid adding _new_ such cases to the beans. Could you instead just add the function to the `DataEditorSettingService` and save the return value to a flag in the form class? Than this function could simply return the determined value and it wouldn't have to be determined again each time the value is requested.

BartChris: I think it is not possible to store this value (session wide?). The value depends on the workflow which is currently opened in the Workflow editor (form). But I moved the logic to the service. In order to hopefully speed up the query i rewrote the check to HSQL.

solth: Thanks for the refactoring. 

Couldn't this new property (e.g. `boolean dataEditorSettingsDefined`) just be set once in `WorkflowForm.load` whenever the workflow editor is opened with a specific workflow? Thus the value would be evaluated _once_ every time a workflow is opened and the evaluated value would just be returned?

So something like this. First define new property of `WorkflowForm`: 
```
boolean dataEditorSettingsDefined;
```
then evaluate its value in the `load` method:
```
public void load(int id) {
...
dataEditorSettingsDefined = this.dataEditorSettingService.areDataEditorSettingsDefinedForWorkflow(this.workflow);
...
}
```
and then in the "getter" just return the evaluated value
```
public boolean hasWorkflowDataEditorSettingsDefined() {
   return dataEditorSettingsDefined;
}
```
Would that work?

BartChris: Thank you for the suggestion. I implemented that and it works.

BartChris:
Right now the check for existing data editor settings is also evaluated to true for compeltely new workflows, i have to inspect, why this happens.

Issue #5300, opened 2022-08-18
Close task on accept doesn’t work


matthias-ronge:
**Describe the bug**
Workflow tasks can be configured `typeAcceptClose`. In this case, clicking on the bell to take over a task should have the effect of taking over the task and finishing it in one single user interaction. This doesn’t work. The task behaves as if the option wasn’t chosen and opens conventionally.

**To Reproduce**
1. Create a workflow with a task of type “close on accept”
2. Create a process of it.
3. Progress to the task, then click on the bell
4. The task page opens what it shouldn’t do

**Expected behavior**
When clicking the bell, this should have the effect of clicking “finish task”.

Issue #4693, opened 2021-09-22
Misleading meaning of task option "Finalise task after accepting"


m-kotzyba:
The thrown exception by using the option "Finalise task after accepting" was solved: #4568 
However, the meaning of the option is misleading:

Having a Workflow with tasks 'a' --> 'b' --> 'c', where only task 'b' has the option "Finalise task after accepting", the following was observed:
- Accepting and finish task 'a' opens and closes 'b' automatically and leaves 'c' open. (‘b’ was never accepted manually, only ‘a’ was finished)
- That is, 'b' is not finalized after accepting it but finalized by finishing the task 'a' before.
- This is in accordance with our use case, where the first task 'A' of a workflow has always the status "closed" and hence, the second task 'B' with option "Finalise task after accepting" is shown to the user. After accepting 'B' (click on the bell), the task 'B' is not finalized but the user has to finish it manually. That is, in out use case task 'B' is not finalized automatically, because the task 'A' before has just status "closed" but was not finished and hence cannot trigger the automatic finalization of 'B'.

If this is intended, the naming of the option in not appropriate. If this is a bug, please have a look on that.

matthias-ronge:
The described behaiviour is a bug.

- Finishing task 'a' should leave 'b' open (unless is is an _automatic_ step; in which case the the “close after accepting” is meaningless)
- After click on the bell for task 'b', it should be finished, and 'c' should be open. The user should not need to finish the task manually

PR #5301, opened 2022-08-18
Enable automatic closing of task with type typeAcceptClose

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/5300
fixes https://github.com/kitodo/kitodo-production/issues/4693 

I am not quite sure where to redirect if a task is closed after acceptance. I setup a redirect to the desktop for this case.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
@@ -173,9 +173,9 @@ public class WorkflowControllerService {
         URI metadataFileUri = ServiceManager.getProcessService().getMetadataFileUri(task.getProcess());
         Workpiece workpiece = ServiceManager.getMetsService().loadWorkpiece(metadataFileUri);
         RulesetManagementInterface ruleset = ServiceManager.getRulesetManagementService().getRulesetManagement();
-        ruleset.load(new File(Paths.get(
-                ConfigCore.getParameter(ParameterCore.DIR_RULESETS),
-                task.getProcess().getRuleset().getFile()).toString()));
+        ruleset.load(new File(
+                Paths.get(ConfigCore.getParameter(ParameterCore.DIR_RULESETS), task.getProcess().getRuleset().getFile())
+                        .toString()));
         ValidationResult validationResult = ServiceManager.getMetadataValidationService().validate(workpiece, ruleset);
         if (State.ERROR.equals(validationResult.getState())) {
             Helper.setErrorMessage(Helper.getTranslation("dataEditor.validation.state.error"));
@@ -248,7 +248,9 @@ public class WorkflowControllerService {
 
     /**
      * Checks if all children of a process are closed.
-     * @param process the process to check
+     * 
+     * @param process
+     *            the process to check
      * @return true if all children are closed
      */
     public static boolean allChildrenClosed(Process process) {
@@ -339,7 +341,8 @@ public class WorkflowControllerService {
     /**
      * Unified method for report problem .
      *
-     * @param comment as Comment object
+     * @param comment
+     *            as Comment object
      */
     public void reportProblem(Comment comment) throws DataException {
         Task currentTask = comment.getCurrentTask();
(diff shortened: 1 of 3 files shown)


solth:
@BartChris thank you for the pull request. I think it would be more appropriate to redirect to the task list instead of the desktop after a task of type "acceptClose" has been closed, but I am not a user of the system. What is your opinion about this, @subhhwendt and @andre-hohmann?

henning-gerhardt (changes requested):
- This change fixes only #5300 but not #4693. If the first task is a normal task and the following task a task of "accept and close" then this task is closed too!
- Returning to desktop after a task is closed in this way feels wrong as I assume that the user is expecting to return back to the task list
- Please provide tests which shows that your change is correct or it is easily happen that this kind of function get broken or missing again.

BartChris:
I implemented a fix for https://github.com/kitodo/kitodo-production/issues/4693 and added a test. The build succeeds, but Codacy fails because a missing Javadoc annotation for the test. I do not know, what i have to change here in order for Codacy to succeed. 

grafik

Kitodo/src/test/java/org/kitodo/production/forms/CurrenTaskFormIT.java
```diff
@@ -0,0 +1,95 @@
...
+        Task taskTypeAcceptClose = createAndSaveTask(TaskStatus.OPEN, 1, process, null, true);
+        Task followingTask = createAndSaveTask(TaskStatus.LOCKED, 2, process, null, true);
+        currentTaskForm.setTaskById(taskTypeAcceptClose.getId());
+        currentTaskForm.takeOverTask();
+        Task taskTypeAcceptCloseUpdated = taskService.getById(taskTypeAcceptClose.getId());
+        Task followingTaskUpdated = taskService.getById(followingTask.getId());
+
+        assertEquals("Task of type typeAcceptClose was closed!", TaskStatus.DONE,
+            taskTypeAcceptCloseUpdated.getProcessingStatus());
+        assertEquals("Folowing task is open!", TaskStatus.OPEN, followingTaskUpdated.getProcessingStatus());
```
henning-gerhardt, line 78: Grammar: `Folowing` -> `Following`

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
```diff
@@ -652,14 +647,12 @@ private boolean runXPathCondition(Process process, String xpath) throws IOExcept
         return ServiceManager.getProcessService().getNodeListFromMetadataFile(process, xpath).getLength() > 0;
     }
 
-    private void verifyTask(Task task) {
+    private void checkForAutomaticTask(Task task) {
```
henning-gerhardt, line 650: Maybe there is a better method naming possible like `perfomActionsForAutomaticTask` or something similar which is describing what this method is doing.

henning-gerhardt (changes requested):
Changes looks good, but maybe it is possible to rename the mentioned method and fix the grammar error.

henning-gerhardt:
> The build succeeds, but Codacy fails due to a missing Javadoc annotation for the test. I do not know, what i have to change here in order for Codacy to succeed.

I don't know how this was handled in the past and how @solth is handling this. Easiest fix would be to provide a JavaDoc comment above the test method.

solth:
> Easiest fix would be to provide a JavaDoc comment above the test method.

I agree. Even though we have many tests without JavaDoc in the existing code, we should at least try to have a short JavaDoc for new tests.

Kitodo/src/test/java/org/kitodo/production/forms/CurrenTaskFormIT.java
BartChris: fixed!

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
BartChris: renamed the method

PR #5302, opened 2022-08-18
Unused 'parentElementXpath' field

### LINKED ISSUE #5299: Catalogue import: Unused parent element - XPath
**Describe the bug**
If you define inside a catalogue import  configuration for a catalogue import a xpath expression for the parent element then is this expression not used.

**To Reproduce**
1. Define a xpath expression for parent element inside the catalogue import configuration for a catalogue
2. Define or not the other needed options for the catalogue

**Expected behavior**
Defined xpath expression is used and information about the parent is retrieved.

**Release**
3.4.4-SNAPSHOT

solth:
Fixes #5299 

As it turns, the `parentElementXPath` field hasn't been used even before the introduction of the new `ImportConfiguration` class. Instead, catalog IDs of parent records are extracted from documents that are already transformed into the internal Kitodo.Production format, using metadata flagged as `higherlevelidentifier`.


henning-gerhardt (approved):
Changes looks correct.

PR #5304, opened 2022-08-18
Fix bug when importing child records

### LINKED ISSUE #5303: Child records cannot be imported
**Describe the bug**
Kitodo.Production offers the option to automatically import child records of an directly imported "main" record. For this to work the queried search interface must support queries by parent record ID (for example Kalliope, with query parameter `context.ead.id`) and the corresponding ImportConfiguration needs to be configured with a search field that is used as "ID parameter of parent record". This functionality is currently bugged. Even if the ImportConfiguration is configured correctly and a record with existing child records is imported, those child records are omitted during import.

**To Reproduce**
1. Create a new `ImportConfiguration` or edit an existing one of type "Catalog import" that supports querying child records by parent record ID
2. Configure a search field that is used as "ID parameter of parent record"
3. Import a record that has child records in the remote search interface and activate the "Import children" switch
4. The child records will not be imported

**Expected behavior**
Given a correctly configured ImportConfiguration of type "Catalog import" that queries a search interface with support for search queries by parent record ID, existing child processes should be imported when the "Import children" switch is activated during import

**Release**
3.4.4-SNAPSHOT

solth:
Fixes #5303

### DIFF
--- a/Kitodo/src/main/resources/messages/messages_de.properties
+++ b/Kitodo/src/main/resources/messages/messages_de.properties
@@ -763,7 +763,7 @@ newProcess.catalogueSearch.loadingChildrenSuccessfulSummary=Kindvorg\u00E4nge er
 newProcess.catalogueSearch.loadingChildrenSuccessfulDetail={0} Kindvorg\u00E4nge geladen
 newProcess.catalogueSearch.manyChildrenWarning=Es wurden {0} Kindvorg\u00E4nge gefunden.
 newProcess.catalogueSearch.parentIDMetadataMissing=Es wurde kein Metadatum zum Speichern von IDs \u00fcbergeordneter Vorg\u00E4nge ("higherlevelIdentifier") im Regelsatz gefunden. Importtiefe zur\u00fcckgesetzt auf 1!
-newProcess.catalogueSearch.parentIDParameterMissing=Es wurde kein Parameter zum Suchen von Kindelementen eines bestimmten Elterndokuments f\u00fcr den ausgew\u00E4hlten Katalog in der kitodo_opac.xml konfiguriert (fehlender Konfigurationsparameter "parentElement")!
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -763,7 +763,7 @@ newProcess.catalogueSearch.loadingChildrenSuccessfulSummary=Kindvorg\u00E4nge er
 newProcess.catalogueSearch.loadingChildrenSuccessfulDetail={0} Kindvorg\u00E4nge geladen
 newProcess.catalogueSearch.manyChildrenWarning=Es wurden {0} Kindvorg\u00E4nge gefunden.
 newProcess.catalogueSearch.parentIDMetadataMissing=Es wurde kein Metadatum zum Speichern von IDs \u00fcbergeordneter Vorg\u00E4nge ("higherlevelIdentifier") im Regelsatz gefunden. Importtiefe zur\u00fcckgesetzt auf 1!
-newProcess.catalogueSearch.parentIDParameterMissing=Es wurde kein Parameter zum Suchen von Kindelementen eines bestimmten Elterndokuments f\u00fcr den ausgew\u00E4hlten Katalog in der kitodo_opac.xml konfiguriert (fehlender Konfigurationsparameter "parentElement")!
+newProcess.catalogueSearch.parentIDParameterMissing=Kindvorg\u00E4nge k\u00F6nnen nicht importiert werden, da der hierfür ben\u00F6tigte "ID-Parameter der \u00DCberordnung" in der Import-Konfiguration nicht eingestellt wurde.
```
henning-gerhardt, line 766: German umlaut ü should be escaped in 'hierfür'

markusweigelt (approved):
Importing child processes works for me. After importing the process with child processes into the form and before importing the process without child processes, the child processes are correctly resetted.

henning-gerhardt (approved):

PR #5313, opened 2022-08-21
Catalog config importer

### LINKED ISSUE #5224: Importer for kitodo_opac.xml
### Problem
After the introduction of the new `ImportConfiguration` class in #5038 catalog configurations previously configured in the file `kitodo_opac.xml` do not work anymore.

### Solution
A new `ImportConfigurationImporter` should be implemented that converts `<catalog>` configurations from existing `kitodo_opac.xml` files to `ImportConfiguration` instances with the correspondig values. `SearchField` and `MappingFile` instances have to be created, too. This import or conversion has to be a guided process, because `MappingFile` instances now  store the input and output metadata formats which the user needs to select from the list of available formats.

### LINKED ISSUE #5262: Remove catalog configurations from "kitodo_opac.xml"
Since catalog configurations are now maintained using the new class `ImportConfiguration` introduced in #5038, the corresponding `<catalog>` entries in the main OPAC config file [kitodo_opac.xml](https://github.com/kitodo/kitodo-production/blob/156006e7cedcf145e20171546800468b19ad20e1/Kitodo/src/main/resources/kitodo_opac.xml) should be removed to reflect this change and to avoid confusion. 

**Note:** the catalog configurations from the corresponding _test_ configuration file [kitodo_opac.xml](https://github.com/kitodo/kitodo-production/blob/156006e7cedcf145e20171546800468b19ad20e1/Kitodo/src/test/resources/kitodo_opac.xml) should **not** be removed because it will be used for testing the upcoming `CatalogConfigurationImporter` (see #5224)

solth:
This adds an catalog configuration importer that transforms opac configurations from existing `kitodo_opac.xml` files into new `ImportConfiguration` objects. Also creates corresponding `MappingFile` objects where needed.

Fixes #5224 
Fixes #5262


markusweigelt (changes requested):
I just went through the code with these changes. Now follows the test in Kitodo.Production

markusweigelt:
Import works well except for the description which is not imported. Otherwise, the imported catalogs seem to have all the search fields and settings as in the XML.

In general, I think it would be great to have a general documentation with a hint to migrate the XML or to extend or adapt the existing one. 

In perspective the documentation should be updated in a PR. But for that we first need a useful and complete user documentation. :-/

markusweigelt (approved):

PR #5334, opened 2022-08-30
Add Kitodo Script to set labels automatically

### LINKED ISSUE #5016: Automatically set outline headings
In version 2, the headings of the outline (which are stored in `<div LABEL="…">`) were taken from metadata `TitleDocMain`. This relationship was hardwired. In version 3, this hard-wired takeover was removed, so that outline headings can—but also must—be treated independently of other metadata. Although the editing user has more freedom in designing the outline of the digital copy, in practice it means more work, as the outline headings and the title (for example, of a chapter) are usually the same.

**Goal:** It needs an automatic takeover. To do this, the command could accept a list of simple metadata keys, and if the heading is empty, it is filled with the first metadata value found for a key in the sequence.

matthias-ronge:
Resolves #5016

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
@@ -28,10 +28,10 @@ import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.text.StrTokenizer;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.util.Strings;
 import org.kitodo.api.MetadataEntry;
 import org.kitodo.api.dataformat.Division;
 import org.kitodo.api.dataformat.Workpiece;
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -22,11 +23,17 @@
...
 import java.util.Objects;
+import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.text.StrTokenizer;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.util.Strings;
```
henning-gerhardt, line 34: I would prefer the `isEmpty` method from `StringUtils` class of the `apache-commons`.`lang3` package (or the old lang package if the old package should be used future and replacement is not planed in near future)  instead a string checking method from Log4J.

matthias-ronge:
Rebased

solth:

henning-gerhardt (changes requested):
Please add tests to prove your changes are correct implemented.

matthias-ronge:
Can be solved by XSLT.

PR #5353, opened 2022-09-15
Add switch to hide/show inactive templates

### LINKED ISSUE #3311: Remove switch to create unreachable production templates

solth:
Fixes #3311 

Instead of removing the switch to create unreachable templates, as was proposed in the linked issue, this pull requests restores the - presumably - intended behaviour of filtering the list of templates available to create a new process for a project by the templates `active` values. In addition, a switch in the template list on the project page can be used to toggle whether inactive templates are shown in that list or not.


Kitodo/src/main/java/org/kitodo/production/dto/ProjectDTO.java
```diff
@@ -223,12 +224,12 @@ public void setClient(ClientDTO client) {
...
     /**
-     * Get list of templates.
+     * Get list of active templates.
      *
-     * @return list of templates as TemplateDTO
+     * @return list of active templates as TemplateDTO
      */
     public List<TemplateDTO> getTemplates() {
-        return templates;
+        return templates.stream().filter(TemplateDTO::isActive).collect(Collectors.toList());
```
thomaslow, line 232: I think it would be better not to add logic to a DTO bean. If inactive templates are not needed when querying a project for its templates, they should probably not be indexed in the first place?

Kitodo/src/test/java/org/kitodo/selenium/ListingST.java
```diff
@@ -234,4 +235,27 @@ public void listUsersTest() throws Exception {
...
+        projectsPage.goToTemplateTab();
+        assertEquals("Wrong number of templates before hiding first template", 2,
+                projectsPage.getTemplateTitles().size());
+        TemplateEditPage editTemplatePage = projectsPage.editTemplate();
+        editTemplatePage.hideTemplate();
+        editTemplatePage.save();
+        assertEquals("Wrong number of templates after hiding first template", 1,
+                projectsPage.getTemplateTitles().size());
+        projectsPage.goToTemplateTab();
+        Thread.sleep(500);
```
thomaslow, line 256: Considering that the CI pipeline already needs 20+ minutes to run, maybe it is possible to avoid generic sleep statements (I know there are a lot of others). There are some examples in other Selenium tests on how to wait for certain conditions like html-elements being available.

thomaslow (commented):
The implemented toggle UI component looks a bit out of character:
image

In other views there is a menu at the bottom of the list, which allows to show or hide certain elements, e.g. the task list: 

image

Maybe such a menu would be more consistent with the existing design language of Kitodo?

Kitodo/src/main/java/org/kitodo/production/dto/ProjectDTO.java
solth: I agree with you in principle about avoiding business logic in DTOs, but I think we still need to add inactive templates to the index. Most list views are filled from the index and we wouldn't be able to edit and re-activate inactive templates if they were _not_ in the index and consequently would not appear in the list, would we?

Issue #5245, opened 2022-07-14
Rel 3.4.3: Process Creation: Meta Data Window in not updated, when docType is changed


stefanCCS:
If I want to create a new process, and press "cancel" in catalogue dialog (because I want to input the meta data manually), then I can choose the docType on the left and I would expect that the meta data on the right would will change accordingly (more or less or different required meta data depending on docType).
Unfortunately, this does not happen (or in more detail, it does not happen, if I try out it this way I have described, but I have seen this happening in one case, where I have worked on different window (here in GitHub) and have come back to KITODO-Window).
I am using Chrome.
Pressing F5 does not change this behavior.

Looks like to be very strange update behavior - please check.

![image](https://user-images.githubusercontent.com/61045115/178993317-61734f31-a16c-4e7c-bb5f-858bf8e89458.png)

stefanCCS:
Additional information provide by SUBHH:

I go via the project list and add an operation via the plus sign - leaving out the catalog import. I get "Band" as pre-selection - the metadata selection matches our rule set.
![image](https://user-images.githubusercontent.com/61045115/180416249-761e2943-ffa7-4490-8da1-f3fe3308f49c.png)

I select a couple of other document types in this mask and then return to the "Band" selection - and get a different selection of metadata (in this case from periodical) - as you can hopefully see well.
![image](https://user-images.githubusercontent.com/61045115/180416347-dfa376ce-5b4d-4c88-8079-6e34385168af.png)

==> So, it is even worse: It is not just not updating but it can be even wrong!

stefanCCS:
Checked in Rel 3.5.0-beta - works fine.

Issue #3664, opened 2020-05-19
Don't forget metadata when changing DocType


matthias-ronge:
If a new process is created without using the catalog, and the doctype is changed **after** metadata has already been entered manually, the metadata is forgotten. In addition, if the metadata input mask only contains digital collections, it is not updated, and it appears that the metadata has been recorded. However, an empty metadata file is saved.

**Goal:** The metadata will be retained when changing the DocType.

solth:
@matthias-ronge it seems this has been fixed by #5355 . Can you confirm this?

matthias-ronge:
This doesn’t happen any more

PR #5355, opened 2022-09-20
Fix update of metadata table after doctype is changed

markusweigelt:
Fixes: #5245


solth:
Does this also fix #3664 ?

markusweigelt:
Maybe it sounds different for me but @matthias-ronge should be able to tell.

PR #5356, opened 2022-09-20
Default record xpaths in import configurations

### LINKED ISSUE #5317: Default xpath expressions for record identifier and record title
While testing pull request #5293 I missed the default values for xpath record identifier and record title expressions for new catalogue import configurations. As this default expressions are added while running the database migration there are no entries for a new catalogue import configuration. You must copy them or if you running a new empty system retrieving this needed information from somewhere.

If you add a new catalogue configuration it should be possible to select a default identifier and title xpath expression or enter your own xpath expression if you like.

solth:
Provide default xpaths for ID and title when editing import configurations.

Fixes #5317


henning-gerhardt (approved):
Changes look okay, except that (maybe selenium based) tests are missing.

First I expected that if I choose a specific metadata format than corresponding entries get pre selected but this was not the requirement in the feature request. So choosing the right one or use them as a base is okay.

Issue #3792, opened 2020-06-29
Sorting in lists


stefanCCS:
The sorting in any list kind of view (e.g. tasks) in supported only in a limited way (e.g. in Task-List it can be sorted for Task-Name only).
Is there anything to configure this sorting?
If not, in my opionion sorting of all columns should be standard in all list views --> Feature Request.

andre-hohmann:
I absolutely agree.

andre-hohmann:
Maybe related to: #4245.

solth:
Since all lists are now sortable by all columns (the exceptions are the multivalued fields `roles`, `clients` and `projects` in the user list, which have been excluded from this issue), this issue can now be closed.

stefanCCS:
@solth :
It looks like in the "All processes" view the sorting of the two columns "Processing begin/end of last task" is not possible.
Is there a special reason for it, why this is not possible?

thomaslow:
@stefanCCS: Sorting of those two columns should work. Can you elaborate on what happens in your case? Are processes sorted incorrectly, or not at all?  There should be an icon in the column header visualizing when the table is sorted by that column:

![image](https://user-images.githubusercontent.com/6214043/203787788-fa557990-b3c6-477f-b623-0e9d7e7ad770.png)

stefanCCS:
Thanks for your comment - indeed it works:
I just have not tried it out, because this sorting symbol 
![image](https://user-images.githubusercontent.com/61045115/203816063-1d002dc6-faa0-493c-9fd6-d1bfd10d1e80.png)
is not visible in my UI.
![image](https://user-images.githubusercontent.com/61045115/203816026-a2586016-a761-4d48-aec0-5b29d47091dc.png)
But I can use to Mouse to do the sorting as expected.

--> this means, this is more a UI related Issue.

PR #5360, opened 2022-09-23
Make Process Table Columns Sortable

### LINKED ISSUE #4245: Process list not sortable by progress
In Production 2, the process list can be sorted by the progress of the processes, if you click on the table header column. This is missing in Production 3.

**Goal:** Make processes list sortable by progress

thomaslow:
Related Issues:
- Contributes to #3792 
- Fixes #4245 

This pull request makes all columns of the process table sortable. In order to achieve this, the data for each column needs to be available as a property in the elastic search index. Some columns were previously generated on-the-fly, such that sorting them was not possible. 

These column values are now calculated at indexing time. As a consequence, indexing speed may be slightly reduced, and UI performance (of the process table) should have slightly improved. 

This pull request requires a full re-indexing (including an update of the elastic search mapping).

Demo of Process Table:

https://user-images.githubusercontent.com/6214043/191960063-ba6391c4-1716-4779-9143-44ae15b535d2.mp4


Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/index/converter/ProcessConverter.java
```diff
@@ -84,18 +84,21 @@ private static Task getLastProcessedTask(Process process) {
      * <p>For process that have children, their repective tasks are also included.</p>
      * 
      * @param process the process
+     * @param considerChildren whether to include tasks of children processes
      * @return the list of tasks of the process (and potentially its children)
      */
-    private static List<Task> getListOfTasksForProgressCalculation(Process process) {
+    private static List<Task> getListOfTasksForProgressCalculation(Process process, Boolean considerChildren) {
```
henning-gerhardt, line 90: Why are you using nullable `Boolean` class instead of primitive `boolean` as the values should only be `true` or `false` ?

thomaslow: Hi Henning, I didn't really think about it. Fixed it below.

thomaslow:
@solth In order for the process status to be sortable, it needs to be kept up-to-date in the ElasticSearch index whenever a task status (OPEN, DONE, INWORK, LOCKED) is changed. This requires that the accompanying process is re-indexed whenever a task status is changed. This conflicts with a pull request from last year, see https://github.com/kitodo/kitodo-production/pull/4543, which presumably improved performance when saving tasks. 

How do we deal with this?
- Re-index processes whenever tasks are saved, potentially reducing performance
- Don't index process progress (and thus, make it not sortable)

thomaslow:
The performance problem mentioned in the previous comment is fixed now by calculating the process status via a custom SQL query. This SQL query allows to efficiently count how many tasks have a certain state (open, locked, inwork, done) even for parent processes that may have hundreds of child tasks (e.g. newspaper processes). Previously, this counting would take more than 1 second for parent processes with many children.

The SQL query uses an SQL statement (`with recursive`) that is not supported by Hibernate but works with most databases. I checked support for MySQL 8+, MariaDB 10.2.2+, H2Database 1.4+, but other databases should work too, e.g. PostgreSQL, Oracle. The syntax is the same for all databases I know of.

In case the SQL query does not work, there is a chance that the resulting exception is intercepted and the previous bean-based status calculation is used as a fallback. However, this depends on how the query fails. 

I also added a selenium test to check that sorting processes by state works.

thomaslow:
Merged with master so github only shows code changes to current master.

solth:
> The performance problem mentioned in the previous comment is fixed now by calculating the process status via a custom SQL query.

So that means re-indexing processes whenever a task is saved does not impose any performance problems anymore? Can you elaborate a little more how using a recursive SQL query helps in this case? If I am not mistaken, #4543 improved performance by reducing read/write operations on the index, not on the database. How does an optimised SQL query have an effect on that?

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -203,7 +204,7 @@ private boolean startExport(Process process, LegacyMetsModsDigitalDocumentHelper
 
     private boolean exportCompletedChildren(List<Process> children) throws DataException {
         for (Process child:children) {
-            if (processService.getProgress(child.getTasks(), null).equals(COMPLETED) && !child.isExported()) {
+            if (ProcessConverter.getCombinedProgressAsString(child, false).equals(COMPLETED) && !child.isExported()) {
```
solth, line 207: Wouldn't calling `ProcessConverter.getCombinedProgressAsString` with second parameter `considerChildren = false` with an parent process without own tasks always result in "000000000100", e.g. 100% locked (as oposed to the the expected "100000000000", e.g. 100% completed?

solth (changes requested):
Thank you for this pull request. I tested it and it works very well with the small list of processes I have on my local development system. I cannot say whether the changes concerning indexing have any negative impact on the performance on larger systems with many processes, though.

On the code side I found only a handful of minor issues like a few typos and unused imports. I am unsure about how progress for processes without own tasks is calculated. Perhaps you could just comment on my question (see below) and elaborate a little more on that specific part.

thomaslow:
> So that means re-indexing processes whenever a task is saved does not impose any performance problems anymore?

No, there are still performance problems related to the general save-strategy in Kitodo as outlined in #5368. The pull request #5371 slightly improves on this issue a bit for processes with parents. Since this pull request requires that processes are re-indexed when their task status changes, it doesn't make things easier, but there is no other way at the moment.

> Can you elaborate a little more how using a recursive SQL query helps in this case?

My original assesment of the problem in this [comment](https://github.com/kitodo/kitodo-production/pull/5360#issuecomment-1257820266) was a bit incomplete. The process of a task was already re-indexed when its task status changed (except in one case, see line 94 of the [WorkflowControllerService.java](https://github.com/kitodo/kitodo-production/pull/5360/files#diff-20932f6779d0304fe9420f36a3cc458dc0bfd8c5584e11859c0094cad3bfb21d)). So, the conflict with #4543 seems to be resolved without reverting it. Still, without always re-indexing a process whenenver a task changes, I can not guarantee that there might be some code somewhere, which modifies a task outside of the logic of `WorkflowControllerService.java` without saving and re-indexing its process (such that the process-index will not contain the most up-to-date information).

Besides that, the performance problem of this pull request was caused by two issues:

1. Changing a task status triggers a save-operation for its process, which leads to the performance issue described in #5368, which caused many re-indexing operations for its parent processes. This still is a problem, but can not be fixed by this pull request.

2. Calculating the task status of a parent process took a very long time when retrieving all related tasks for a parent process via bean getter methods (each triggering a database query for the tasks of a child process). This issue was multiplied by problem 1, because parent processes were saved many times (re-calculating the process status multiple times). The custom SQL query allows to efficiently calculate the process status for parent processes (which previously was only partially implemented by Kathrin via the `sortHelperStatus` property). It is still (unnecessarily) calculated multiple times, but much faster now.

I hope this clears things up a bit.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/TaskDAO.java
```diff
@@ -14,8 +14,17 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
+import javax.persistence.PersistenceException;
```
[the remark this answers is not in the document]
thomaslow: Yes, sorry, we should probably add a check for unused imports to the linter.

Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/index/converter/ProcessConverter.java
```diff
@@ -0,0 +1,296 @@
...
+     * 
+     * @param process the process
+     * @param considerChildren whether to also count tasks of children processes
+     * @return a map providing the percentage of tasks having a certain status (done, open, locked, inwork)
+     */
+    public static Map<TaskStatus, Double> getTaskProgressPercentageOfProcess(Process process, boolean considerChildren) {
+        Map<TaskStatus, Integer> counts = countTaskStatusOfProcess(process, considerChildren);
+        Integer total = counts.values().stream().mapToInt(Integer::intValue).sum();
+        
+        // report processes without any tasks as if they had a single locked task
```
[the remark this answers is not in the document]
thomaslow: I'm not sure why this was implemented this way. This was implemented the same before in [ProcessService](https://github.com/kitodo/kitodo-production/blob/b2f2837032369c64f0c9c7f56d5e5f95b8245cc7/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java#L1494). 

> Does this mean the task progress of a process without own tasks is always encoded as "100% locked tasks"? 

Yes. Unless `considerChildren=true` and the process has children processes with tasks that are not locked.

> Shouldn't it instead be illegal (e.g. throw an exception) to try to determine the "task progress" of a process without workflow/tasks (when the parameter considerChildrenis false)?

Maybe.

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
thomaslow: If this list of children contains processes that have no tasks, they are not catched by this if statement, correct. I don't know what this export code is doing, which is why I didn't change the implementation of how the progress status is calculated.

thomaslow:
@solth I commited all of your suggested changes. Thank you for your review.

solth (commented):
Thanks for implementing the change requests. I still want to make a few tests, so I haven't merged this pull request, yet.

Since some of your other PRs have been merged in the meantime, a small code conflict arose. Could you resolve this conflict while I condict my tests?

solth (approved):

PR #5362, opened 2022-09-26
Fix missing user column in task table

thomaslow:
The optional column "last editing user" is not shown in the task list when selecting it from the column configuration drop-down menu. This pull request fixes this problem. The column "last editing user" is now available.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/tasks/taskList.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/tasks/taskList.xhtml
@@ -86,7 +86,7 @@
 
             <p:column id="processingUserColumn"
                       headerText="#{msgs.lastEditingUser}"
-                      rendered="#{CurrentTaskForm.showColumn('task.lastProcessingUser')}">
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
Your change looks strange for me: every around your code change is referring to `processingUser` and not to `lastEditingUser` except the column heading. If the `lastEditingUser` should be displayed then should not everything (user information, column id name, ...) used from `lastEditingUser` or named by this?

solth:
@henning-gerhardt this is due to an unfortunate difference between the task property `processingUser` and the database entry `task.lastEditingUser` in the `listcolumn` table, which controlls whether the `processingUser` of the task should be displayed or not.

Thus this pull request indeed fixes the described problem.
An alternative solution would be to change the `title` value of the database entry from `task.lastEditingUser` to `task.processingUser` using a migration file in order to adjust bean property and list column flag.

henning-gerhardt:
Thank you @solth for the clarification but reading the code without this information is confusing and I will maybe not the last user which gets confused.

solth (approved):

Issue #4916, opened 2022-01-18
File pointer in exported METS get duplicated if they have the same filepath


BartChris:
I am not sure wether this is intended behaviour, but consider the following constellation:

I have defined three filegroups: ORIGINAL, DEFAULT, THUMBS. For none of those filegroups a METS-PATH is defined in the project settings, see for example the settings for the ORIGINAL filegroup.

![grafik](https://user-images.githubusercontent.com/7516499/149984415-90733e67-24f9-4c62-97be-0c79ff037c44.png)

In the internal meta.xml everything looks fine. But when the process gets exported, the physical structure section is duplicating the UUID of the thumbnail.

![grafik](https://user-images.githubusercontent.com/7516499/149985252-3242dc34-ab04-45b6-ba08-566a95f1f130.png)

If i then define individual METS-paths for each filegroup, e.g. /images/tif/ , /images/jpg/ , /images/thumbs all of the filegroups are referenced in the phyical structure:
 
![grafik](https://user-images.githubusercontent.com/7516499/149984978-7e9e8a3d-3955-437c-b282-2f9cc0344ee1.png)

There seems to be a hard requirement hat the images in the file groups can be differentiated via their mediatype or the METS path. Is this correct?

matthias-ronge:
I don't understand the point here. The exported METS file is supposed to reference two filesets (I mean `DEFAULT` and `THUMBS`) with two _different_ image versions. If you don't specify a path here, it will refer to the _same_ file (`00000001.jpg`) twice! Do you also use the DEFAULT image as a thumbnail?

It is not yet defined what should happen if a fileset without a specified _METS pointer path_ is to be exported. In order for the DFG Viewer to work, the file group must describe the absolute path to the images, as they can subsequently be retrieved from a server by the DFG Viewer (i.e. in the external view). The requirement here is that the thumbnail must be 150 pixels in size and the DEFAULT file must have the original resolution, so two different files must be referenced here.

If you don't want to use the DFG viewer, you're free here, but what information do you want to encode here?

I agree that this is probably a bug nonetheless. Apparently the UUID is looked up using the file name.

BartChris:
Thank your for the detailed answer. You are of course correct that in this form the Filegroups do not make any sense. So let me explain how i got there. I had the problem that i could not use the interface in Kitodo

![grafik](https://user-images.githubusercontent.com/7516499/151372616-87dbf96c-18b5-4f46-a292-4db02467c1e5.png)

[...]

PR #5369, opened 2022-09-29
Filename variables

pontus-osterdahl:
This pull request adds some of the functionality requested in [#4916 ](https://github.com/kitodo/kitodo-production/issues/4916) by introducing two placeholders for use in METS-paths in folder configurations. 

- "filename": the referenced filename (without path).
- "basename": the basename of the file (without path). 

The METS-path "http://www.example.com/content/(basename)/default/(filename)" should now in the exported METS be resolved as "http://www.example.com/content/Image00001/default/Image00001.jpg".

The METS-path specifies the URL by which an image can be retrieved by the DFG-Viewer. Currently, the filename is added to the end of the URL, but information such as the filename may be necessary at other positions when external image servers are used. The addition of placeholders will hopefully reduce the need for post-processing.

Notes:
   - The filename-placeholders are only used during export and will not be accessible e.g. in script-paths.
   - If a file-placeholder is replaced, the filename is not by default added to the end of the METS-path. The reason for this is to improve flexibility. The filename must in such cases, as in the example above, be added via a placeholder.
   - If the METS-path does not contain "(filename)" or "(basename)", the filename is by default added to the end of the METS-path.


Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
```diff
@@ -368,4 +396,121 @@ private String determineReplacementForFirstchild(Matcher variableFinder, String
...
+     *         invalid, the string found is returned and no replacement takes
+     *         place. If the variable cannot be replaced because the requested
+     *         data cannot be accessed, the empty string is returned.
+     */
+    private String determineReplacementForFilePlaceholder(Matcher variableFinder, String filename) {
+        switch (variableFinder.group(6)) {
+    	    case "filename":
+                return determineReplacementForFilename(variableFinder, filename);
+    	    case "filename_without_extension":
+                return determineReplacementForFilenameWithoutExtension(variableFinder, filename);
```
matthias-ronge, line 468: You can simplify this a lot (and get rid of two more functions) by using `org.apache.commons.io.FilenameUtils`:
```java
switch (variableFinder.group(6)) {
    case "filename":
        return variableFinder.group(1) + FilenameUtils.getName(filename);
    case "basename":
        return variableFinder.group(1) + FilenameUtils.getBaseName(filename);
    default:
        ...
```

matthias-ronge (changes requested):
In summary, a good and previously missing functionality! But the code can be improved in some places.

- `(filename_without_extension)` is long. I suggest `(basename)` instead.
- Make sure your `.java` files do not contain tabs, only spaces. (Configure your editor or code formatter settings.)
- Whenever possible, please do not duplicate (copy-paste) code. You can refactor the code if necessary (moving existing code into another method, etc.).
- Javadoc: The method description (the text before the first `@param` or the `@return` label) needs to contain a period (`.`). This is because everything before the first period is copied into the summary table on top of the Javadoc page. This sentence should be a verbal phrase (a phrase starting with a verb, mentally completing the sentence “This function …”) Good: “[This function] Replaces filename variables with the file name.” Not: “[This function] If a filename is to be determined, it is determined here.” The `@param` and `@return` descriptions do not need a period at the end. They never go into the summary table. BTW, you don’t need to write Javadoc for `private` functions as they do not show in the Javadoc at all. If there is something really important to explain about a `private` function, you can point this out using a non-Javadoc comment (starting the comment with `/*` instead `/**`).

See also my detail comments below.

pontus-osterdahl:
Thank you @matthias-ronge for your thorough review and helpful suggestions! They should be included now.

matthias-ronge:
When you think you've incorporated all changes, you can click this button …

the button

… to request a new review. The pull request then reappears for me in [the list of assigned pull request (you can bookmark this link for own list)](https://github.com/kitodo/kitodo-production/pulls/review-requested/@me).

matthias-ronge (approved):
That looks good! (I haven't downloaded and tested it, I assume it works.)

PR #5370, opened 2022-09-29
Allow to combine filters with alternatives and negations

### LINKED ISSUE #4981: Extended Search: Combine filter queries with operators such as AND, OR and NOT
It would be very helpful, to combine queries with an operator such as OR.

Scenario: Find all Tasks that are in Step “Export” independent of the status but not “closed”;
Possible filter query: "stepopen: Export" OR "stepinwork: Export" (… maybe combined with a project-condition)

thomaslow:
Related Issues:
- Fixes #4981

This pull request adds the ability to specify alternatives for conditions in a single filter (and to negate conditions). In part, this was already possible for some properties, e.g. the id property using the following filter (selecting processes for multiple ids):

> "id:81279 81278"

Another example is the filtering for tasks by their order range (1=Scanning, 2=QC, 3=Structure, 4=Export):

> "stepinwork:1-2"

As far a I know, the most recent documentation for the current filter syntax can be found in the "[Anwenderdokumentation Kitodo 1.11](https://github.com/kitodo/kitodo-production/wiki/Suche-und-Filtern)".

However, it was not possible to combine alternatives for arbitrary conditions. This pull request allows to add alternative conditions by separating them using a vertical line character "|" in a filter. For example, in order to filter tasks that are either currently in progress in the step "Scanning" or already done in step "QC", you can now use the following syntax:

> "stepinwork:Scanning | steplocked:QC"

Additionally, negations can now be applied to every condition universally by adding a "-" (minus) as a prefix. Again, for some properties this was already implemented. For example, it was possible to write the following filter before:

> "-stepinwork:Scanning"

However, the result of that filter was not the exact inverse of the non-negated filter `stepinwork:Scanning`. The previous implementation applied the negation to both parts of the filter, meaning the fact that the task is a `Scanning` task and the fact that it is a task that is currently in progress. As a result, the list of tasks would not include any `Scanning`-tasks and also not include any tasks currently in progress (independent of their name). The exact inverse (and in my opinion more intuitive interpretation of the negation) would be to just exclude `Scanning`-tasks that are currently in progress. This pull request changes the semantics of this filter. Therefore, users should be notified about this change when upgrading.

This also affects the Integration Tests, which have not yet been updated (TODO).

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/SearchForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/SearchForm.java
@@ -311,14 +311,16 @@ public class SearchForm {
         }
         if (StringUtils.isNotBlank(this.processPropertyValue)) {
             if (StringUtils.isNotBlank(this.processPropertyTitle)) {
-                search += "\"" + FilterString.PROPERTY.getFilterEnglish() + this.processPropertyTitle + ":"
-                        + this.processPropertyValue + "\" ";
+                search += "\"" + this.processPropertyOperand + FilterString.PROPERTY.getFilterEnglish() 
+                        + this.processPropertyTitle + ":" + this.processPropertyValue + "\" ";
             } else {
-                search += "\"" + FilterString.PROPERTY.getFilterEnglish() + "*:" + this.processPropertyValue + "\" ";
+                search += "\"" + this.processPropertyOperand + FilterString.PROPERTY.getFilterEnglish() 
+                        + "*:" + this.processPropertyValue + "\" ";
             }
         } else {
             if (StringUtils.isNotBlank(this.processPropertyTitle)) {
-                search += "\"" + FilterString.PROPERTY.getFilterEnglish() + this.processPropertyTitle + ":*\" ";
+                search += "\"" + this.processPropertyOperand + FilterString.PROPERTY.getFilterEnglish() 
+                        + this.processPropertyTitle + ":*\" ";
             }
         }
         return search;
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
@@ -21,7 +21,9 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.StringTokenizer;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


henning-gerhardt:
Please keep in mind that a task title can contain white spaces like `Erfassen der Meta- und Strukturdaten" and this is the need for the `"` as surrounding of the expression like `"stepopen:Erfassen der Meta- und Strukturdaten".

thomaslow:
Yes, I considered this. The following filter works just fine:

> "stepinwork:Structure and Metadata | stepinwork:Scanning"

Unless there is a vertical line "|" somewhere in a title, it should work. If you think this could be a problem, I could adjust the syntax such that you could optionally add parentheses around a condition in order to mask the "|", e.g.:

> "(stepopen:Task with | in Title) | (stepopen:Long Text)"

henning-gerhardt:
If this is working (and maybe proofed by a test) then it is fine for me.

thomaslow:
@solth Is there a good place to document the new filter syntax? Here are some options I found:
- as a pull request for a new markdown file in the [docs/user/introduction](https://github.com/kitodo/kitodo-production/tree/master/docs/user/introduction) directory
- by modifying the wiki page [Anwenderdokumentation Kitodo.Production 3.X / Suche und Filter](https://github.com/kitodo/kitodo-production/wiki/Suche-und-Filter)

solth:
> @solth Is there a good place to document the new filter syntax? Here are some options I found:
> 
>     * as a pull request for a new markdown file in the [docs/user/introduction](https://github.com/kitodo/kitodo-production/tree/master/docs/user/introduction) directory
> 
>     * by modifying the wiki page [Anwenderdokumentation Kitodo.Production 3.X / Suche und Filter](https://github.com/kitodo/kitodo-production/wiki/Suche-und-Filter)

I think we are trying to gather such information primarily in the Github Wiki, now. 

@matthiaskissler Is that correct?

matthiaskissler:
Yes, that is correct; both developer and user documentation can be found in the wiki now (here is an overview page for the 3.x documentation sites: https://github.com/kitodo/kitodo-production/wiki/Informationen-zu-Kitodo.Production-3.x)

thomaslow:
I added some information about the general search as well as a description of the new filter syntax to the Wiki page [Suche und Filter](https://github.com/kitodo/kitodo-production/wiki/Suche-und-Filter).

oliver-stoehr (approved):

PR #5371, opened 2022-09-30
Performance when saving processes with a parent

### LINKED ISSUE #5368: Improve performance when saving and indexing processes, tasks, etc.
> tltr; I propose to use a queue to collect save operations within business logic and trigger saving multiple objects in one go at the end.

When saving things in Kitodo (especially processes and tasks), often a number of related entities are needed to be saved and indexed again. This can quickly get out of hand. Let's say there are 3 processes:

1. Grand Parent process
2. Parent process
3. Child process

Updating a task in the child process also affects the parent and grand parent process. This is already handled in the [save](https://github.com/kitodo/kitodo-production/blob/2e80926c57f17dfac8ed47d818130a4c7950a59a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java#L281) method of the ProcessService, which uses a recursive call to save parent processes. In addition, parent processes are saved twice (before saving the child and again after the child was saved). In total there are 9 actual database and indexing operations triggered in the above scenario. The grand parent process is saved 6 times, the parent process twice, the child once.

In addition, the business logic code is not very structured in this regard. Some methods call the BaseDAO.save methods \~whenever they like\~. This makes sense when they are called independently, but not when they are called as part of bigger batch updates. For example, increasing the current task status (which involves re-assigning a new user to the next task, and potentially closing a prior task, etc.) triggers multiple calls to the save method of the same task. If this method would trigger a process save (as required for pull request #5360), you can get a coffee before anything is saved.

What can we do about it?

In this ticket I would like to propose the idea to adapt the general save-strategy in Kitodo in a way that allows to defer or enqueue save operations until the respective business logic is done, and afterwards trigger that all enqueued objects are saved and indexed in one go. In my opinion, this should greatly improve the performance when saving things and at the same time keep the convenience of "sprinkling" save calls into the business logic code wherever appropriate.

For each object type (process, task, etc.) there should be a queue containing "save"-intentions that are deduplicated (by object id). So, a recursive save of a process could easily generate 9 save-intentions for a grand parent process, but due to the deduplication, the process is actually only saved once when a final "save-and-index-all-queued-objects" call is triggered. 

What do you think?

thomaslow:
Related Issues:
- #5368

This pull request improves the performance when saving processes that have a parent process. It doesn't fix #5368 but removes a lot of unnecessary save operations previously triggered due to a double recursion in the `ProcessService.save` method.

This improves the performance for actions like "set status of process up/down". For example. for 20 selected processes (with parents) the time improves from 15.1s to 4.5s (status up) and 10.7s to 6.0s (status down).


solth:
#5360 has been merged. Please resolve resulting conflicts.

IkramMaalej:
In my local environment (OS: macOS, Browser Firefox v 106.0.2) :
- for 20 selected processes (with parents) the time improves from 43s to 19.5s when setting the processes status down  and from 46s to 11.5s when setting the processes status up.
- When creating of 54 processes (with children) the time is also reduced from 1min05s to 53s.

Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
```diff
@@ -611,11 +613,22 @@ public void testFindAllIDs() throws DataException {
...
+        // to be returned. It could be any order ElasticSearch chooses. It's weird that this doesn't fail randomly.
+        // Assert.assertTrue("id's contain wrong entries", allIDs.containsAll(Arrays.asList(5, 3, 1, 2, 6)));
 
         allIDs = ServiceManager.getProcessService().findAllIDs(5L, 10);
         Assert.assertEquals("Wrong amount of id's in index", 2, allIDs.size());
-        Assert.assertTrue("id's contain wrong entries", allIDs.containsAll(Arrays.asList(7, 4)));
+
+        // This test is very strange. The method findAllIDs doesn't specify an order in which processes are supposed 
+        // to be returned. It could be any order ElasticSearch chooses. It's weird that this doesn't fail randomly.
+        // Assert.assertTrue("id's contain wrong entries", allIDs.containsAll(Arrays.asList(7, 4)));
```
IkramMaalej, line 631: Same here.

Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
```diff
@@ -611,11 +613,22 @@ public void testFindAllIDs() throws DataException {
...
-        Assert.assertTrue("id's contain wrong entries", allIDs.containsAll(Arrays.asList(5, 3, 1, 2, 6)));
+        Assert.assertEquals("Duplicate ids in index", allIDs.size(), new HashSet<Integer>(allIDs).size());
+        Integer maxId = allIDs.stream().mapToInt(Integer::intValue).max().getAsInt();
+        Integer minId = allIDs.stream().mapToInt(Integer::intValue).min().getAsInt();
+        Assert.assertTrue("Ids should all be smaller than 8", maxId < 8);
+        Assert.assertTrue("Ids should all be larger than 0", minId > 0);
+        
+        // This test is very strange. The method findAllIDs doesn't specify an order in which processes are supposed 
+        // to be returned. It could be any order ElasticSearch chooses. It's weird that this doesn't fail randomly.
+        // Assert.assertTrue("id's contain wrong entries", allIDs.containsAll(Arrays.asList(5, 3, 1, 2, 6)));
```
IkramMaalej, line 624: Right, I would also suggest to delete this test because the test in `line:612` checks if allIDs contains wrong entries or not.
`Assert.assertTrue("id's contain wrong entries", allIDs.containsAll(Arrays.asList(5, 2, 6, 4, 1, 7, 3)));`

IkramMaalej (approved):

Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
thomaslow: Removed comment and test.

Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
thomaslow: Thank you for your feedback. I deleted both tests.

PR #5373, opened 2022-10-05
Make authorities table sortable

thomaslow:
Related Issues:
- Contribues to #3792 

This pull request adds the ability to sort authorities by their translated title and type. Authorities are now sorted by title by default. 

Since the translated titles are not available in the SQL database, the implementation is currently restricted to a maximum of 1000 authorities. The implementation loads all authorities (up to 1000) into memory, sorts them by their translated title, and renders the requested page.

Screenshot for authorities sorted in German:
image


solth (approved):
The changes work as you described and the code looks good to me.

PR #5374, opened 2022-10-05
Upgrade selenium chrome driver to 106

thomaslow:
@solth Please merge and trigger rebuild of recently failed CI pipelines, thank you!


solth:
@thomaslow thanks for the pull request. Just re-triggering failed builds won't help, though, because the code they are based on still contains Chrome driver Version 104. Please rebase your corresponding branches against the updated master and then force-push your rebased branches instead. This will trigger a new build with the up-to-date Chrome driver.

PR #5387, opened 2022-10-11
Add link to filter for parent process to row expansion

### LINKED ISSUE #4982: Button to show superordinate process in process list
In the process list, displaying further information of a superordinate process (via the arrow down button) allow to show all subordinate processes in a (new, filtered) process list (see image 1 and 2)

image 1:

image 2:

However, to show (all) superordinate processes of a subordinate process in a (new, filtered) process list is not possible with one click. The available options to open the “Edit”-Options or the Metadata Editor of a process’s superordinate processes (cf. image 2) is sometimes not sufficient, e.g. if a docket or METS file have to be generated. Instead the user has to filter “id:<Id-of-superordinate processes>” (cf. image 3) but even the ID is not directly available (cf. image 2).
 
image 3:

oliver-stoehr:
Resolves #4982

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
@@ -1101,17 +1101,28 @@ tr.ui-widget-content,
     display: inline-block;
 }
 
-.row-expansion-wrapper > table > tbody {
-    width: 70%;
-}
-
 .row-expansion-wrapper > table > tbody > tr {
     width: 100%;
 }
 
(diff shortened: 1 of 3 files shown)


thomaslow:
Hi Oliver, there is no white space between the process title and the "show" link:

image

If possible, it would probably be a good idea to add the same link to the search result table (which uses the exact same drop down view for parent and children processes). This link shoud (in my opinion) redirect the user to the processes list and apply the id filter.

oliver-stoehr:
Thank you for pointing out! I added the link to the search result table and fixed the missing space.

thomaslow (approved):
Alright, looks good to me.

PR #5388, opened 2022-10-12
Fix import of opac configurations without mapping files

### LINKED ISSUE #5382: Catalog configurations without `mappingFiles` parameter are not converted
**Describe the bug**
#5293 removed metadata format specific implementations of the `SchemaConverterInterface`. That left `XMLSchemaConverter.java` as the _only_  schema converter currently available. The class contains a check for "supported source metadata formats". That check was originally meant to enable catalog configurations in `kitodo_opac.xml` for specific metadata formats like MODS or MARCXML without explicitly configuring special mapping files. These configurations would instead use default mapping files for those formats when no `mappingFiles` element was found in a catalog configuration in `kitodo_opac.xml`.

#5038 replaced the configurations in `kitodo_opac.xml` with the new `ImportConfiguration` class. Mapping files with specific metadata input and output formats are now mandatory with this class, so the case described above cannot occur anymore. The check for supported metadata formats is already done when saving the import configuration itself and falsely prevents catalog configurations without `mappingFiles` element in `kitodo_opac.xml` to be converted to new `ImportConfiguration` instances with the error message `Missing mandatory XML element "mappingFiles"`

Instead, the check for the `mappingFiles` element during the conversion of `kitodo_opac.xml` entries to new `ImportConfiguration` instances should add default mapping files instead of leading to an conversion error.

**To Reproduce**
1. Convert a `kitodo_opac.xml` configuration file that contains a catalog configuration without `mappingFiles` element
2. The conversion will fail with the error message `Mandatory configuration "mappingFiles" is missing`

**Expected behavior**
The conversion should add default mapping files to catalog configurations in `kitodo_opac.xml` that do not contain a `mappingFiles` element instead of 

**Release**
3.4.4-SNAPSHOT (Commit b65bd96)

### LINKED ISSUE #5383: XMLSchemaConverter incorrectly checks for supported metadata formats
**Describe the bug**
In addition to the problem described in #5382, the (now obsolete) check for supported file formats also prevents other input metadata formats like EAD from being processed. These can now be configured as `OTHER` in the import configuration setup since #5313  but will lead to the error `UnsupportedFormatException: No SchemaConverter found that supports OTHER and XML!` during import.

Bildschirmfoto 2022-10-10 um 12 23 42

**To Reproduce**
1. Configure an `ImportConfiguration` with metadata format `OTHER`
2. Try to import a record using the new `ImportConfiguration`

**Expected behavior**
Using import configurations with metadata format `OTHER` should not lead to an error.

**Release**
3.4.4-SNAPSHOT (Commit b65bd96)

solth:
OPAC configurations in `kitodo_opac.xml` without `<mappingFiles>` element where falsely skipped during conversion to new `ImportConfiguration` objects. These configurations should instead use standard mapping files where available (MODS, MARC and PICA). That change is implemented in this pull request and fixes #5382 

Additionally the changes in this pull request skip the check for `supportedSourceMetadataFormats` during metadata conversion (because that is now controlled by the configured mapping files themselves) and therefor also fixes #5383


Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -310,9 +310,7 @@ private SchemaConverterInterface getSchemaConverter(DataRecord record) throws Un
         KitodoServiceLoader<SchemaConverterInterface> loader =
                 new KitodoServiceLoader<>(SchemaConverterInterface.class);
         List<SchemaConverterInterface> converterModules = loader.loadModules().stream()
-                .filter(c -> c.supportsSourceMetadataFormat(record.getMetadataFormat())
-                        && c.supportsSourceFileFormat(record.getFileFormat())
-                        && c.supportsTargetMetadataFormat(MetadataFormat.KITODO)
+                .filter(c -> c.supportsSourceFileFormat(record.getFileFormat())
```
markusweigelt, line 313: Please changing "c" to something to more meaningful variable name e.g. "converter".

Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
```diff
@@ -11,22 +11,39 @@
...
+    private final String source;
+    private final MetadataFormat targetFormat;
 
     /**
      * Constructor setting filename and source URI for the XSL transformation file.
      * @param filename Filename including suffix without any path.
      * @param source Remote source where the file can be retrieved from if it is not yet available in Kitodo.
      */
-    MetadataFormatConversion(String filename, String source) {
+    MetadataFormatConversion(String filename, String source, MetadataFormat targetFormat) {
```
markusweigelt, line 43: `@param targetFormat` doc is missing.

Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
```diff
@@ -11,22 +11,39 @@
...
+    static {
+        DEFAULT_CONVERSIONS.put(MetadataFormat.MODS, Collections.singletonList(MODS_2_KITODO));
+        DEFAULT_CONVERSIONS.put(MetadataFormat.PICA, Collections.singletonList(PICA_2_KITODO));
+        DEFAULT_CONVERSIONS.put(MetadataFormat.MARC, Arrays.asList(MARC_2_MODS, MODS_2_KITODO));
+    }
 
-    private String fileName;
-    private String source;
+    private final String fileName;
+    private final String source;
```
markusweigelt, line 35: I think it is more clear when `source` is directly of type URL.

markusweigelt (changes requested):
Reviewed your code changes. Test of the functionality in the application follows.

markusweigelt:
Works very well for me! Tested the import with standard for input metadata format "PICA" and "MODS" when no mapping file was specified for catalog in kitodo_opac.xml.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
solth: Done

Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
solth: Done

PR #5390, opened 2022-10-12
Prevent NPE when querying catalogs with invalid IDs

### LINKED ISSUE #5389: Importing unknown IDs results in `NullPointerException`
**Describe the bug**
When querying search interfaces for records by ID, Kitodo bypasses the creation of hitlists and directly tries to import the record identified by the provided ID. It assumes that the search interface will return exactly one hit. If the provided ID is invalid and does not correspond to record in the search interface, though, Kitodo will trigger a `NullPointerException` when trying to retrieve data from the empty result. 

This is caused by a missing check for the number of hits when querying by ID, e.g. method `XMLResponseHandler.extractNumberOfRecords` is only called when hitlists are generated by searches for non-ID fields. This error was caused by the introduction of the `CatalogImportDialog.skipHitList` method in #5313 

**To Reproduce**
1. Create an `ImportConfiguration` of type `CatalogImport`
2. Search for an invalid ID with that `ImportConfiguration`
3. Observe error

**Expected behavior**
When searching for an invalid ID, the import mask should not result in a `NullPointerException` but instead display a corresponding message that no data record was found.

**Screenshots**
Bildschirmfoto 2022-10-12 um 12 49 11

**Release**
3.4.4-SNAPSHOT (commit 7ca97bd)

solth:
Fixes #5389

### DIFF
--- a/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
+++ b/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
@@ -48,6 +48,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.net.ftp.FTPClient;
 import org.apache.commons.net.ftp.FTPFile;
 import org.apache.commons.net.ftp.FTPFileFilter;
+import org.apache.http.HttpEntity;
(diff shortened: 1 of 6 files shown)


markusweigelt (approved):
Works for me! Unlike before, now the message that the ID was not found.

Issue #5271, opened 2022-08-02
Arbitrary SearchFields in ImportConfigurations for OAI interfaces


solth:
**Describe the bug**
Currently, all `ImportConfiguration`s of type `Catalog search` allow configuring arbitrary `SearchField`s for search queries. This makes sense for SRU interfaces, but not for OAI interfaces which are designed not for searching but for harvesting. OAI interfaces in Kitodo.Production only use the predefined OAI verb **GetRecord** to retrieve individual records and therefore cannot generate search result _lists_, so defining custom `SearchField`s does not make sense for OAI configurations in Kitodo.Production.

**To Reproduce**
1. Go to 'Projects'
2. Click on 'New' and then 'ImportConfiguration'
3. Select 'Catalog search'  in the 'Configuration type' menu
4. Select 'OAI' in the 'Search interface type' menu
5. Scroll down to 'Search fields'. Here, custom fields can be added, which should not be allowed for OAI

**Expected behavior**
Adding custom `SearchField`s should not be possible for OAI `ImportConfiguration`s.

**Release**
3.4.4-SNAPSHOT

BartChris:
You are probably correct. One problem i see is, that there might be a custom interface, which responds in an OAI-conformant way but also supports other search fields. I used custom "searchfields" (ID) to return an OAI conformant response. So this would reduce flexibility which is on the other hand re-introduced by e.g. https://github.com/kitodo/kitodo-production/issues/5257

The question is if there is a real disadvantage in allowing for more flexibility.

solth:
I guess the best solution would be to add another search interface type "Custom", which does allow freely adding search fields but does not impose the restrictions that real OAI or SRU interfaces have (OAI interfaces need a configured `metadataPrefix`, but do not allow for custom search fields, for example, while SRU interfaces _always_ need a `recordSchema` and also support - or even require! - configuring arbitrary search fields). 

Such a "Custom" search interface could also be used for other APIs like Solr or ElasticSearch servers or direct access to databases. In addition, it should probably also be the option where the user can add the custom URL parameters described in #5257.

One of the motivations behind replacing the opac XML configurations with the new `ImportConfiguration` class was to reduce the risk of faulty configurations by having a tighter control over which type of configuration can contain which type of setting. I would prefer not to weaken this newly gained "stability" by allowing parameters for interface types that normally do not support them. 

The proposed "Custom" interface type seems like the best way to allow for more "unconventional" search interfaces.

[...]

PR #5397, opened 2022-10-14
Add arbitrary URL parameters to CUSTOM interfaces

### LINKED ISSUE #5257: Reinstate option to add arbitrary URL parameters for ImportConfigurations
**Describe the bug**
Even though it wasn't intentional, catalog configurations in `kitodo_opac.xml` allowed adding arbitrary URL parameters. These custom URL parameters have been used in some catalog configurations. #5038 replaced the catalog configurations in `kitodo_opac.xml` with a new `ImportConfiguration` class. This new class only supports interface type specific URL parameters like `recordSchema` for SRU interfaces or `metadataPrefix` for OAI interfaces, thus rendering those catalog configurations useless that depended on custom URL parameters.

**Expected behavior**
`ImportConfiguration`s of type `Catalog search` should allow adding arbitrary URL parameters.

**Release**
3.4.4-SNAPSHOT

**Additional context**
This issue was initially reported by @BartChris in #5219

solth:
This pull request adds the option to add arbitrary URL parameters to `ImportConfigurations` with search interface type `CUSTOM`. It also limits the number of search fields in `CUSTOM` search interfaces to one, because the current implemention for `CUSTOM` interfaces does not support hitlists, but only queries by ID (which should be sufficient in most cases anyway.) Fixes #5257 

@BartChris could you check whether or not this re-enables you to import from the simulated OAI interface you mentioned in #5271 ? If so, I will add a test to this pull request and mark it ready for review.


BartChris:
@solth Thanks a lot for your efforts. This works for my simulated OAI interface.

solth:
@BartChris would you mind adding a review for this pull request?

solth:
Unfortunately I don't know what's wrong with the Selenium test. Locally it always runs successful, but on GitHub it fails every time.

BartChris:
@solth yes, i will check it

Kitodo/src/main/java/org/kitodo/production/forms/ImportConfigurationEditView.java
```diff
@@ -225,6 +226,43 @@ public void removeSearchField(SearchField searchField) {
...
 
+    /**
+     * Add given UrlParameter to importConfiguration.
+     *
+     * @param urlParameter UrlParameter to add
+     */
+    public void addUrlParameter(UrlParameter urlParameter) {
+        if (importConfiguration.getUrlParameters().stream().map(UrlParameter::getParameterKey)
+                .collect(Collectors.toList()).contains(urlParameter.getParameterKey())) {
+            Helper.setErrorMessage("URL parameter keys must be unique ('" + urlParameter.getParameterKey()
```
BartChris, line 237: The error message is shown, but only showing for the fraction of a second.

Kitodo/src/test/java/org/kitodo/production/services/data/ImportServiceIT.java
```diff
@@ -83,31 +91,73 @@ public static void cleanDatabase() throws Exception {
...
+        Process importedProcess = importProcess(RECORD_ID, MockDatabase.getK10PlusImportConfiguration());
         Assert.assertEquals("WrongProcessTitle", "Kitodo_" + RECORD_ID, importedProcess.getTitle());
         Assert.assertEquals("Wrong project used", 1, (long) importedProcess.getProject().getId());
         Assert.assertEquals("Wrong template used", 1, (long) importedProcess.getTemplate().getId());
         Assert.assertEquals("Not the correct amount of processes found", 8, (long) processService.count());
     }
 
-    private static void setupServer(String serverResponse) {
+    @Test
+    public void shouldCreateUrlWithCustomParameters() throws DAOException, ImportException, IOException {
```
BartChris, line 102: The naming does not fully describe what is happeing in the test as the creation of url parameter is happening in the MockDatabase class and not in this test. Maybe this should be renamed to reflect what is happening in the test (`shouldCreateProcessWithCustomUrlParameters.`)  and also check if the imported processes is the expected one and not just "not null"

BartChris:
@solth Functionality works good for me. I added some comments.

BartChris:
> Unfortunately I don't know what's wrong with the Selenium test. Locally it always runs successful, but on GitHub it fails every time.

The selenium test is also failing locally for me

<pre>
[ERROR] addCustomImportconfigurationWithUrlParameters  Time elapsed: 17.862 s  <<< FAILURE!
org.opentest4j.MultipleFailuresError: 
Multiple Failures (2 failures)
        org.awaitility.core.ConditionTimeoutException: Condition with lambda expression in org.kitodo.selenium.testframework.pages.ImportConfigurationEditPage was not fulfilled within 5 seconds.
        org.openqa.selenium.UnhandledAlertException: unexpected alert open: {Alert text : }
  (Session info: chrome=106.0.5249.119): {Alert text : 
</pre>

Kitodo/src/test/java/org/kitodo/production/services/data/ImportServiceIT.java
solth: Actually, the MockDatabase only prepares the ImportConfiguration with custom URL parameters. This test does not check the creation of the _parameters_, though, but instead the creation of the _URL_ itself, using the parameters of the ImportConfiguration. For that reason I think the method is named correctly. 

The creation of an ImportConfiguration with custom URL parameters on the other hand is tested in the Selenium Test `AddingST.addCustomImportconfigurationWithUrlParameters`.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/UrlParameterDAO.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param offset
+     *            result
+     * @param size
+     *            amount of results
+     * @return constrained list of persisted beans
+     * @throws DAOException if a HibernateException is thrown
+     */
+    @Override
+    public List<UrlParameter> getAll(int offset, int size) throws DAOException {
+        return retrieveObjects("FROM urlparameter ORDER BY ID ASC", offset, size);
```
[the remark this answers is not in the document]
solth: The name of the database table is `urlparameter` (all lowercase characters), not `UrlParameter` (camal case). Why do you suggest to change the parameter in the query to camel case?

solth:
> Edit: The test seems to use the wrong ID for the button. I changed
> 
> https://github.com/kitodo/kitodo-production/blob/128bdb2097d0a65cf275db739f89877238787a45/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/ImportConfigurationEditPage.java#L48-L49

Great! Thanks alot for looking into the failing test!

Kitodo/src/main/webapp/WEB-INF/templates/includes/importConfigurationEdit/rows/searchFieldRows.xhtml
```diff
@@ -72,8 +72,11 @@
                 <p:tooltip for="searchFieldHelp"
                            value="#{msgs['tooltip.importConfig.searchFieldHelp']}"/>
                 <p:commandButton icon="fa fa-plus"
+                                 id="addSearchField"
+                                 title="#{(importConfigurationEditView.importConfiguration.interfaceType eq 'CUSTOM' and importConfigurationEditView.importConfiguration.searchFields.size() > 0) ? msgs['importConfig.custom.idSearchOnly'] : msgs['importConfig.addSearchField']}"
```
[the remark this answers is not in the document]
solth: Yes, we had the same issue with some other deactivated command components in the past. Fixed it by wrapping the button in a separate container element and moving the tooltip from the button to the wrapper. Thanks for pointing this out.

solth:
> ```java
> ```java
> 
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> After that the tests finish succesful. Is using a static ID here the only possibility and will this break again when other changes are introduced to the form?
> ```

I think the best way is to explicitely set the ID to a fixed value and use that ID in the Selenium test because then the ID won't change with new form elements being added later on.

Kitodo/src/main/java/org/kitodo/production/forms/ImportConfigurationEditView.java
solth: This seems to have been caused by the `saveButtonToggler`. This hidden button is used to activate the `Save` button and is called whenever the properties of an object being edited - in this case the `ImportConfiguration` - change. And since the `error-messages` component was set to `autoUpdate`, the previously displayed "URL parameter keys must be unique" message would be immediately replaced with an empty message array when the `saveButtonToggler` was triggered. 

To resolve the issue, I replaced the `error-messages` component with `autoUpdate` with an identical `error-messages` component _without_ the `autoUpdate` in the `importConfigurationEdit` view.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/UrlParameterDAO.java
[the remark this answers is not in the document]
BartChris: I have not tested it, but i think in HQL you have to use the name of the Entity, not the table name. This is also done for the getAll-method in other DAO-classes.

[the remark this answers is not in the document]
solth: You are right! I just tested the method and it did indeed throw the following exception:
```org.hibernate.hql.internal.ast.QuerySyntaxException: urlparameter is not mapped [FROM urlparameter ORDER BY ID ASC]```
Implementing your change request does resolve the problem!

**Comment by solth** [comment]
@BartChris thanks a lot for the review. I implemented your change requests.

**Comment by solth** [comment]
@BartChris would you mind reviewing this PR?

File: Kitodo/src/main/java/org/kitodo/production/forms/ImportConfigurationEditView.java [hunk 5]
```diff
@@ -225,6 +226,43 @@ public void removeSearchField(SearchField searchField) {
...
 
+    /**
+     * Add given UrlParameter to importConfiguration.
+     *
+     * @param urlParameter UrlParameter to add
+     */
+    public void addUrlParameter(UrlParameter urlParameter) {
+        if (importConfiguration.getUrlParameters().stream().map(UrlParameter::getParameterKey)
+                .collect(Collectors.toList()).contains(urlParameter.getParameterKey())) {
+            Helper.setErrorMessage("URL parameter keys must be unique (\"" + urlParameter.getParameterKey()
```
BartChris, line 237: error message should probably be translated.

BartChris, line 237: is it possible to translate this error message as well?

BartChris (approved):

BartChris:
I think i discovered another bug: When i try to change a url parameter, all of the defined parameters are changed and not just the one i want to change.

BartChris (changes requested):
Updating url params is broken

solth:
> I think i discovered another bug: When i try to change an url parameter, all of the defined parameters (if i have multiple defined) are changed and not just the one i want to change.

Thanks a lot for discovering this nasty bug. The same seems to be true for _unsaved_ search fields.

solth:
@BartChris I added message keys for the error message you mentioned and fixed the bug when updating URL parameters. Could you test it once more? The similar bug with the search fields seems to be caused by something different and isn't fixed yet, but I would suggest to deal with that in a separate pull request because it isn't really connected to the topic of this one.

BartChris (approved):
Looks good to me!

solth:
Thanks for the review. I added a new issue for bug when updating unsaved search fields: #5420

PR #5398, opened 2022-10-14
Update used actions to new version

henning-gerhardt:
Updating used GitHub actions to new versions to get solve the warning

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-java, actions/setup-java, actions/checkout
```

on the actions overview after a build process is done.


thomaslow (approved):
I'm no expert with GitHub actions. According to the documentation, there are no migration steps necessary when upgrading from v2 to v3. The maven build log seems to be the same as before. So, it should be fine.

Issue #5217, opened 2022-07-05
NPE during import when import mapping and ruleset do not match


solth:
I got a NullPointerException when trying to import some data using a mapping file that creates metadata groups and metadata entries within the group that are both unknown to the ruleset:
<img width="1424" alt="Bildschirmfoto 2022-07-05 um 09 52 00" src="https://user-images.githubusercontent.com/19183925/177278709-ca194bde-da85-450e-8018-7dcfa289e035.png">

markusweigelt:
Same problem last week but i did not report as issue. :sweat:

Problem seams the IllegalStateException in ProcessFieldedMetadata line 346. In my case this exception is more meaningful. "Got complex metadata entry with key "ContributorCorporation" which isn't declared as substructured key in the rule set." So we should log this exception, improve or reconsider error handling. Cause after this exception is catched a fallback mechanism handle missing field using metadata.createUndefinedMetadataTable and null is passed as parameter which then leads to the npe.

```java.lang.NullPointerException
	at org.kitodo.production.forms.createprocess.ProcessTextMetadata.addLeadingZeros(ProcessTextMetadata.java:48)
	at org.kitodo.production.forms.createprocess.ProcessTextMetadata.<init>(ProcessTextMetadata.java:39)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createMetadataEntryEdit(ProcessFieldedMetadata.java:340)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createUndefinedMetadataTable(ProcessFieldedMetadata.java:223)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createMetadataEntryEdit(ProcessFieldedMetadata.java:350)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createMetadataTable(ProcessFieldedMetadata.java:198)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.<init>(ProcessFieldedMetadata.java:123)
	at org.kitodo.production.services.data.ImportService.initializeProcessDetails(ImportService.java:1049)
	at org.kitodo.production.services.data.ImportService.transformToProcessDetails(ImportService.java:997)
	at org.kitodo.production.services.data.ImportService.addTitleAndTiffHeaderDataToTempProcess(ImportService.java:437)
	at org.kitodo.production.services.data.ImportService.createTempProcessFromDocument(ImportService.java:427)
[...]

PR #5399, opened 2022-10-14
Catch NPE when unknown metadata and metadata groups are created during import

solth:
This PR catches the `NullPointerException` that can occur during import that is decribed in #5217 and shows a corresponding error message. It does *not* fix the underlying issue, but just mitigates the problem.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/ProcessHelper.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/ProcessHelper.java
@@ -30,6 +30,7 @@ import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.StructuralElementViewInterface;
 import org.kitodo.api.dataformat.LogicalDivision;
 import org.kitodo.data.database.beans.Process;
+import org.kitodo.exceptions.CatalogException;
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/ProcessHelper.java
```diff
@@ -99,7 +100,11 @@ public static ProcessFieldedMetadata initializeProcessDetails(LogicalDivision st
             RulesetManagementInterface managementInterface, String stage, List<Locale.LanguageRange> priorityList) {
         StructuralElementViewInterface divisionView = managementInterface.getStructuralElementView(structure.getType(),
                 stage, priorityList);
-        return new ProcessFieldedMetadata(structure, divisionView);
+        try {
+            return new ProcessFieldedMetadata(structure, divisionView);
+        } catch (NullPointerException e) {
```
markusweigelt, line 105: I think when we catch NPE than we should be more direct on the problem e.g. in function `ProcessFieldedMetadata.createMetadataEntryEdit`. This function should throw `ProcessException` or a new  `ProcessFieldException`. Otherwise it leads to confusion, misinterpretation or hides other problems

markusweigelt (changes requested):

solth:
As it turns out, handling for the null pointer exception mentioned in #5217 is already covered by the changes in #5390, so this pr can be closed.

PR #5402, opened 2022-10-17
Sort ImportConfigurations alphabetically in 'Create new process' form

solth:
ImportConfigurations in the popup menu on the "Create new process" form were sorted in the order they were created. This was very confusing whenever more than just a few ImportConfigurations were available.

This pull request orders the ImportConfigurations in the mentioned form alphabetically and moves the business logic of loading and filtering them to the corresponding service class.


markusweigelt (changes requested):
Tested and works for me. Tiny suggestion to change error message of assert.

markusweigelt (approved):

PR #5403, opened 2022-10-17
Improve "correction K" tooltip

### LINKED ISSUE #4994: Show corrections messages and comments on hovering over the ‘K’
Currently, hovering over the correction ‘K’ only show the corrections messages. (see image)

However, it would be very helpful to show also the comments, e.g. in case a correction turned out to be not necessary (cf. #4935) and hence, the tasks can be just finished instead of opening eh Metadata Editor and click on the comments.

oliver-stoehr:
The tooltip is now rendered using `p:tooltip` and also lists all normal comments in addition to the correction comments.
Correction comments in the tooltip are marked with a `K`/`C` in red (not yet resolved) or grey (already resolved).
Screenshot of improved tooltip

The column is renamed from "Correction message" to "Comments".
It can now display three different symbols:
- a grey speech bubble for normal comments
- a red `K`/`C` for an unresolved correction comment
- a grey `K`/`C` for at least one resolved correction comment
Bildschirmfoto 2022-10-18 um 14 26 24

Resolves #4994.


thomaslow:
This will create substantial merge conflicts with #5360, which stores the comment status in ElasticSearch to be sortable.

solth:
#5360 has been merged. Please resolve resulting conflicts.

solth:

thomaslow:
@oliver-stoehr 
I tested your pull request. Functionally, it seems to work just fine and resolves the conflicts mentioned earlier. From this perspective, the pull request is acceptable. 

However, I do have some performance concerns. When rendering a process list with 100 processes, the performance of the backend call that retrieves the DataTable contents (inspected via the F12 browser network tab) degrades from ~0.8 seconds (current master) to ~1.3 seconds. It is not a huge deal, but noticable. Still, in my database I only have in total ~8 comments. I could imagine, if there are more comments in your database, performance might be affected even more.

I suspect the performance degradation is caused by multiple calls to `ProcessForm.getComments(process)` in `processList.xhtml`, which each will trigger additional 100 database queries to find all comments of all listed processes. Unfortunately, theses database queries are also not cached in Kitodo. So, two calls to `ProcessForm.getComments(process)` will trigger 200 database queries. Maybe it is possible to improve the code to save one or two calls or even write a custom database query that retrieves all comments with a single database query? For example, you could create a RequestScoped bean that issues the custom database query and caches the comments during the lifetime of the request (and potentially multiple calls to the bean getter method when rendering the jsf).

Ideally, since the comments are not even shown to the user without hovering over the comment symbol, the tooltip contents could be loaded only when hovering occurs, instead of pre-loading all comments of all listed processes. Unfortunately, the Primefaces Tooltip component doesn't seem to support this.

oliver-stoehr:
Thank you for pointing out! I moved the `getComments` methods to a separate bean which caches the comments per request. So there should now be only one database query per process.  
The comment icon's `rendered` attribute now relies on index information and does not query the database. Same applies to processes without comments.

solth:
@thomaslow could you finish your review for this pull request and check whether your prior remarks have been incorporated sufficiently?

thomaslow:
Sorry. As I said in November, I really do not have any more time left for Kitodo. The review above was already made as a favor to the community.

solth:
@thomaslow sure, I understand no problem. Thanks for the initial review, though!

Kitodo/src/main/java/org/kitodo/production/forms/CurrentTaskForm.java
```diff
@@ -15,6 +15,7 @@
 import java.net.URI;
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Collections;
```
solth, line 18:
```suggestion
```
It seems this import is no longer required

Kitodo/src/main/java/org/kitodo/production/forms/CurrentTaskForm.java
```diff
@@ -29,6 +30,7 @@
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.kitodo.data.database.beans.Batch;
+import org.kitodo.data.database.beans.Comment;
```
solth, line 33:
```suggestion
```
It seems this import is no longer required

Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
```diff
@@ -13,6 +13,7 @@
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
```
solth, line 16:
```suggestion
```
It seems this import is no longer required

Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
```diff
@@ -23,6 +24,7 @@
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.ElasticsearchStatusException;
+import org.kitodo.data.database.beans.Comment;
```
solth, line 27:
```suggestion
```
It seems this import is no longer required

Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -17,6 +17,7 @@
 import java.net.URI;
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Collections;
```
solth, line 20:
```suggestion
```
It seems this import is no longer required

Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -34,6 +35,7 @@
 import org.apache.logging.log4j.Logger;
 import org.kitodo.config.ConfigCore;
 import org.kitodo.config.enums.ParameterCore;
+import org.kitodo.data.database.beans.Comment;
```
solth, line 38:
```suggestion
```
It seems this import is no longer required

solth (changes requested):
Please rename the SQL migration file because "V2_117..." already exists.

solth (approved):

PR #5406, opened 2022-10-18
Display confirmation dialog for image deletion

BartChris:
When a user has the right to delete images in the metadata editor no confirmation dialog is shown for the deletion of images. I think it should be necessary to confirm the deletion.

### DIFF
--- a/Kitodo/src/main/resources/messages/messages_de.properties
+++ b/Kitodo/src/main/resources/messages/messages_de.properties
@@ -205,6 +205,7 @@ confirmClose=Schlie\u00DFen best\u00E4tigen
 confirmDelete=L\u00F6schen best\u00E4tigen
 confirmDeleteClient=Mandant "{0}" l\u00F6schen?
 confirmDeleteDocket=Laufzettel "{0}" l\u00F6schen?
+confirmDeleteImage=Bild wirklich l\u00F6schen?
 confirmDeleteImportConfiguration=Importkonfiguration "{0}" l\u00F6schen?
 confirmDeleteLdapgroup=LDAP-Gruppe "{0}" l\u00F6schen?
 confirmDeleteMappingFile=Abbildungsdatei "{0}" l\u00F6schen?
--- a/Kitodo/src/main/resources/messages/messages_en.properties
+++ b/Kitodo/src/main/resources/messages/messages_en.properties
@@ -209,6 +209,7 @@ confirmClose=Confirm close
 confirmDelete=Confirm delete
 confirmDeleteClient=Delete client "{0}"?
 confirmDeleteDocket=Delete docket "{0}"?
+confirmDeleteImage=Really delete the image?
 confirmDeleteImportConfiguration=Delete import configuration "{0}"?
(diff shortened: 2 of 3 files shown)


solth:
Thanks for the pull request. If I recall correctly, images removed via the contexed menu in the metadata editor are not really remove from the file system until the user clicks "Save", though. Is that correct, @IkramMaalej ?

Therefor I am not sure if we really need a confirmation dialog here, since the risk of losing data is quite low at that point.

BartChris:
I think you are correct here, a additional save is necessary. However: The deletion of an image might happen accidentally without the user being aware. If the user the performs some other actions and saves those the image is lost as well. I would prefer to have this safeguard.

IkramMaalej:
> Thanks for the pull request. If I recall correctly, images removed via the context menu in the metadata editor are not really removed from the file system until the user clicks "Save", though. Is that correct, @IkramMaalej ?
> 
> Therefor I am not sure if we really need a confirmation dialog here, since the risk of losing data is quite low at that point.

Hi @BartChris Thanks for your contrinution.

Correct. The deleted media will be permanantly removed from the file system, only after clicking the "Save" or the "Save & exit" button.

Maybe it would be better if we could show a single warning dialog for all deleted media just after clicking the "Save" or "Save & exit" buttons, in order to avoid showing this dialog (many times) after each single delete operation.

In addition, I think if we decide to go with this solution, then it might be more consistent to  apply  the same behaviour also when deleting a structural  element or deleting a metadata.

BartChris:
Thank you for your thoughts. I agree that it would be more consistent to always request a delete dialog (also for other elements). I introduced it for images because it has more consequences if a image gets deleted  (maybe you need a new scan then) then it has to delete a structure element for example. 
Regarding your idea of showing a dialog before saving: Would that mean that you show the name of every image that is going to be deleted? I would probably be fine with that.

Maybe @m-kotzyba or @andre-hohmann can comment on the idea of having confirmation dialogs for delete operations (of all kind) in the metadata editor.

m-kotzyba:
> Maybe @m-kotzyba or @andre-hohmann can comment on the idea of having confirmation dialogs for delete operations (of all kind) in the metadata editor.

My (Kitodo) colleagues mentioned several times that the less necessary clicks the better (to save time). So, to confirm each deletion, I think, makes the editing cumbersome. However, to get a reminder at the end (after using the save button), what elements will be deleted/edited, sounds attractive. Maybe it is possible to provide this "reminder" feature as option in the settings?

solth:
Currently it seems the consensus is to show just one warning dialog upon saving in the editor when images were removed prior to clicking the "Save" button. Perhaps this behavior could even be made configurable with an additional switch/flag in "User settings" -> "Metadata Editor" to accomodate more user wishes? (e.g. "Show confirmation dialog when saving after images were added or deleted")

@BartChris could you adjust your implementation accordingly? Otherwise I would suggest to close this pull request until the solution described above is available. Alternatively we could first start a discussion to determine most desirable implementation that covers most user wishes. What do you think?

BartChris:
@solth I agree. I will close this for now, and we wait for an implementation following your recommdation - either by me or by somebody else.

PR #5430, opened 2022-11-07
Autofocus search field in OPAC import dialog popup

solth:
This pull request adds two small QOL improvements to the catalog import popup dialog:
- remove the `no selection option` of the `Search field` selection, so that a search field is always selected when the dialog is opened; thus even catalog search configurations of interface types `OAI` (which do not allow for custom search fields and therefor cannot define a default search field) now always come with the OAI `Identifier` search field pre-selected, saving the user two clicks during each import
- autofocus the `Search term` field when the popup dialog is opened when default import configuration and default search field are configured, saving the user another click


markusweigelt:
@solth For my case it works if an import configuration was already selected and popup dialog is opened again while creating process. 

When I start to create new process the import configuration must always be selected. After selecting import configuration the search field is preselected but the focus is not in the suitable value field.

markusweigelt (approved):

PR #5431, opened 2022-11-07
Simplify XML reading and get off a dependency

henning-gerhardt:
Replace XML file reading by external dependency (commons.configuration.XMLConfiguration) with Java internal mechanic.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/editor/XMLEditor.java
+++ b/Kitodo/src/main/java/org/kitodo/production/editor/XMLEditor.java
@@ -20,6 +20,7 @@ import java.io.StringReader;
 import java.nio.file.Files;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.annotation.PostConstruct;
 import javax.faces.application.FacesMessage;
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/editor/XMLEditor.java
```diff
@@ -114,12 +114,10 @@ public String getCurrentConfigurationFile() {
...
      */
     public void loadXMLConfiguration(String configurationFile) {
-        try (StringWriter stringWriter = new StringWriter()) {
+        try {
             currentConfigurationFile = configurationFile;
             this.configurationFile = KitodoConfigFile.getByName(configurationFile);
-            XMLConfiguration currentConfiguration = new XMLConfiguration(this.configurationFile.getAbsolutePath());
-            currentConfiguration.save(stringWriter);
-            this.xmlConfigurationString = stringWriter.toString();
+            this.xmlConfigurationString = Files.lines(this.configurationFile.getFile().toPath()).collect(Collectors.joining("\n"));
```
solth, line 120: `Files.lines` returns a `Stream<String>`, which isn't closed here in case of an `IOException`, if I am not mistaken. Would it be possible to use a `try-with-resources` statement here to make sure the stream is always closed?

solth (approved):

PR #5433, opened 2022-11-07
Replace ConfigurationException with FileNotFoundException

henning-gerhardt:
I think that throwing a FileNotFoundException is in this cases the better option as throwing an external dependency exception.


solth (approved):

henning-gerhardt:
I will adjust this pull request after #5431 is merged as the catch block XMLEditor class must be changed.

solth:

PR #5434, opened 2022-11-08
Joergleh 2022 08 11

joergleh:
Aligned the English and German messages line by line, deleted superfluous keys and added two keys missing in the English messages file


solth:

PR #5439, opened 2022-11-14
Add settings to show comments and pagination panel by default

### LINKED ISSUE #4978: Metadata Editor: Save state of open/closed comments or pagination window
Similar to the state of "Structured view" or "Detail view", what is saved in the background and used each time if the Metadata editor is (re-)opened, it would be helpful to save also the state of open/closed comments or pagination window.

Alternatively, the option to set a default option on open/closed comments or pagination window in the User data & settings would solve this issue.

solth:
This PR adds user settings to save the default state of comments and pagination panel and thereby fixes #4978 (second option)

Also removes unused "Unstructured view" from metadata editor gallery.


oliver-stoehr (changes requested):
I just found two typos. Apart from that, it looks good to me and works fine.

solth:
@ m-kotzyba do you mind testing this solution from a user side and check whether it resolves the issue in your opinion?

m-kotzyba:
@solth I had a look on the solution. If the default values are chosen in the user settings, e.g., both windows open:



in the Meta Data Editor, as expected, they are open:



Same with closed windows as default and the same with the gallery default view.
Thanks for the implementation. :)

oliver-stoehr (approved):

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -324,6 +324,7 @@ private void init() {
         structurePanel.getSelectedPhysicalNode().setSelected(true);
         metadataPanel.showLogical(getSelectedStructure());
         metadataPanel.showPhysical(getSelectedPhysicalDivision());
+        galleryPanel.setGalleryViewMode(GalleryViewMode.getByMessageKey(user.getDefaultGalleryViewMode()).name());
```
markusweigelt, line 327: In my opinion, we should not use message keys for behavioral testing in logic, because it can change and does not describe a state. I think here user.getDefaultGalleryViewMode() should directly return a GalleryViewMode or only the name of the enum should be used. Conceivable would be here also a number. 0 for the list 1 for the preview etc.

markusweigelt (changes requested):

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
solth: Good point. Implemented.

markusweigelt (approved):
Works for me. Why is this configuration actually part of the user and not the project or the application? Probably this will be useful for all users of a project or of the application. But that's certainly a different issue. ;)

solth:
> Works for me. Why is this configuration actually part of the user and not the project or the application? Probably this will be useful for all users of a project or of the application. But that's certainly a different issue. ;)

Thank you for the review. 

The linked issue describes the feature request for individual users. There is, however, already another issue which describes a similar feature request for a system wide setting (#5493) which - as you said - should be implemented separately.

PR #5451, opened 2022-11-23
Add auto-generated METS and Kitodo source files to repository

solth:
This pull request adds the Java files generated from schema files `mets.xsd` and `kitodo.xsd` via JAXB to the repository and removes the dependency on online resources from the library of congress.

Thus resolves #5449 and fixes the broken builds of recent pull requests.

This PR is a draft until any open source licence related uncertainties have been solved.

**Edit**: auto generated Java files are excluded from checkstyle checks because JAXB does not adhere to configured rules when creating those files

PR #5461, opened 2022-11-27
Add required LoC resources to repository

### LINKED ISSUE #5449: Github builds failing because requests to LoC resources are blocked
**Describe the bug**
The files `mets.xsd`, `xlink.xsd` and `marc21slim2mods.xsl` are downloaded from the website of the Library of Congress (loc.gov) during builds of Kitodo.Production and are used for test cases and to automatically create Java classes for METS concepts like `StructLink`, `MdSec`etc. Since about two weeks, these downloads fail when initiated by automatic Github builds that are triggered by each new commit, with http status code **403** ("forbidden") and the following error message in the build log:

```
Could not get content
org.apache.maven.wagon.authorization.AuthorizationException: authorization failed for https://www.loc.gov/standards/mets/version111/mets.xsd, status: 403 Forbidden
```

This also means builds triggered by pull requests will fail and automatic tests and other checks are not performed. This hinders further development and must be fixed before any new pull requests can be merged.

A potential fix could be to add all auto generated Java files to the repository and thus make builds independent of the availability of resources from loc.gov. A draft version for this fix has been implmented in 441169ff7da92d0c10cafda50d95f303ac5f28c5 and indeed seems to resolve the issue (the corresponding build was successful, marked by the green check mark) but before this fix can be merged to the master branch it has to be verified that adding those auto generated files without the Kitodo license header is unproblematic or not.

**To Reproduce**
1. Commit changes and push to repository or fork of repository

**Expected behavior**
Builds should not fail

**Additional context**
The requests for the loc.gov resources succeed and the files are downloaded when triggered by builds on my local development environment. This presumably means only requests from Github to loc.gov resources are blocked specifically.

solth:
Alternative solution to #5451 that also fixes #5449 

In contrast to #5451 this pull request does not add the Java files auto generated by the `mets.xsd` and `kitodo.xsd` schema definitions to the repository, but instead adds the METS and XLINK schema files and themselves.

Also adds the files `marc21slim2mods.xsl` and `MARC21slimUtils.xsl` to the repository which are required by a test case.

Both approaches - this pull request as well as #5451 - resolve the need to download resources from loc.gov during builds and therefor fix the linked issue.

### DIFF
--- a/Kitodo-XML-SchemaConverter/src/main/resources/xslt/marc21slim2mods.xsl
+++ b/Kitodo-XML-SchemaConverter/src/main/resources/xslt/marc21slim2mods.xsl
@@ -1,0 +1,5508 @@
+<xsl:stylesheet xmlns="http://www.loc.gov/mods/v3" xmlns:marc="http://www.loc.gov/MARC21/slim"
+	xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+	exclude-result-prefixes="xlink marc" version="1.0">
+	<xsl:include href="./src/main/resources/xslt/MARC21slimUtils.xsl"/>
+	<xsl:output encoding="UTF-8" indent="yes" method="xml"/>
+	<xsl:strip-space elements="*"/>
+
+	<!-- Maintenance note: For each revision, change the content of <recordInfo><recordOrigin> to reflect the new revision number.
+	MARC21slim2MODS3-4 (Revision 1.95) 20141219
+
+Revision 1.95 - Added a xsl:when to deal with '#' and ' ' in $leader19 and $controlField008-18 - ws 2014/12/19
+Revision 1.94 - Leader 07 b mapping changed from "continuing" to "serial" tmee 2014/02/21
+Revision 1.93 - Fixed personal name transform for ind1=0 tmee 2014/01/31
+Revision 1.92 - Removed duplicate code for 856 1.51 tmee tmee 2014/01/31
+Revision 1.91 - Fixed createnameFrom720 duplication tmee tmee 2014/01/31
+Revision 1.90 - Fixed 520 displayLabel tmee tmee 2014/01/31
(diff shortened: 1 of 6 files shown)


Kitodo-XML-SchemaConverter/src/main/resources/xslt/marc21slim2mods.xsl
```diff
@@ -0,0 +1,5508 @@
+<xsl:stylesheet xmlns="http://www.loc.gov/mods/v3" xmlns:marc="http://www.loc.gov/MARC21/slim"
+	xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+	exclude-result-prefixes="xlink marc" version="1.0">
+	<xsl:include href="./src/main/resources/xslt/MARC21slimUtils.xsl"/>
```
henning-gerhardt, line 4: Should the reference in `href` not contain only a relative path from the current file as both xsd files are in the same directory?

henning-gerhardt (changes requested):
If it is possible use a relative path. If it is not possible then the changes are approved.

Kitodo-XML-SchemaConverter/src/main/resources/xslt/marc21slim2mods.xsl
solth: I tried different relative path variations (`MARC21slimUtils.xsl`, `./MARC21slimUtils`, `.MARC21slimUtils` ...) but always got an error akin to
```Caused by: java.io.FileNotFoundException: /Users/arved/Projects/Kitodo/kitodo-production/Kitodo-XML-SchemaConverter/MARC21slimUtils.xsl (No such file or directory)```
The path variant I ended up with is the only one which worked and did not cause an error.

henning-gerhardt: I tried it in the last hour too but I did not even get this working. Maybe we found a better way in future.

PR #5465, opened 2022-11-28
Improve id column for larger ids

oliver-stoehr:
The id column is now a little bit wider to display ids with up to seven digits on the desktop and the processes list.
In addition the column is now resizable on the processes list.
  
Id column in processes list


solth:

andre-hohmann (approved):
@oliver-stoehr : Thanks! From the user perspective it is OK.

Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/processesWidget.xhtml
```diff
@@ -22,7 +22,7 @@
                  styleClass="default-layout"
                  var="process">
         <p:column headerText="#{msgs.id}"
-                  width="50"
+                  width="70"
```
henning-gerhardt, line 25: I don't know if its possible to move the `width` declarations here and on the other place  to a style sheet property so the adjustment must be done only once and can maybe defined more granular.

henning-gerhardt (commented):
Looks good so far as I can tell this :)

henning-gerhardt (approved):

Issue #5455, opened 2022-11-24
Metadata cannot be selected when creating several divisions


pontus-osterdahl:
**Describe the bug**
If you create several divisions at once in the metadata editor, it is possible to automatically assign metadata keys and values. The metadata key to be added is selected in the dropdown-list "Metadata". However, the list is always empty.

**To Reproduce**
1. Open Metadata Editor.
2. Right-click on a division and select "+Add element".
3. In the resulting dialog set count to more than one 1 and blur the field.
4. An empty dropdown-list appears.

**Expected behavior**
Dropdown-list should contain the allowed metadata keys. 

**Release**
3.4.3, 3.5.0-beta-1

**Additional context**
Metadata keys can be selected in 3.4.2.

PR #5466, opened 2022-11-29
Enable select metadatakey when creating several divisions 

pontus-osterdahl:
Fixes [#5455](https://github.com/kitodo/kitodo-production/issues/5455) by preparing the addable metadata.


IkramMaalej:
The list of addable metadata  is no longer empty, but the suggested metadata entries are incorrect and do not always match the selected metadata type.

IkramMaalej:
Another small note, it is not possible to add more than one docstruct without adding metadata, there is no 'No selection' entry in the dropdown menu.

IkramMaalej:
The Metadata drop-down list shows the correct suggestions only in the second execution. 
In the first execution, the metadata's suggestions are always coming from the previous 'add element' operation, regardless in which process this operation has been executed.

https://user-images.githubusercontent.com/34158824/211517204-3070194e-e94f-458d-a323-bb263974a88e.mov

IkramMaalej (commented):
The Metadata drop-down list shows the correct suggestions only in the second execution.
In the first execution, the metadata's suggestions are always coming from the previous 'add element' operation, regardless in which process this operation has been executed.


https://user-images.githubusercontent.com/34158824/211519828-821fdf25-fc4b-4eae-9b86-d15b2956c08a.mov

pontus-osterdahl:
Many thanks for the comments! The structure type was not reset correctly. Now it should be working.

pontus-osterdahl:
Thanks again! The metadata-selection is now disabled if no docstruct type is selected.

IkramMaalej (approved):

solth:
@pontus-osterdahl please rebase against current master as your PR is the next merge candidate

PR #5467, opened 2022-11-29
`ViewScoped` metadata editor

### LINKED ISSUE #3860: MetadataLock not updated when user leaves metadata editor via menu
The class `MetadataLock` manages a list of processes that are currently opened in the metadata editor by a user. Processes are "unlocked" when the user exits the metadata editor via the `Exit` or `Save & Exit` buttons, but not when the user leaves the metadata editor via some other link, for example the Kitodo logo, the breadcrumbs or the navigation menu. 

This leads to many processes still being marked as "in work by another user" even though they are not openend in the metadata editor anymore.

### LINKED ISSUE #4182: SessionScoped DataEditorForm prevents opening multiple processes in multiple tabs
Since the `DataEditorForm` currently is a `SessionScoped` bean, the user can always just store information to one single process in it at any given time. Once a different process is opened in the medata editor, the data of the previous process is replaced. 

This leads to the problem that if the user opens two processes in the metadata editor in two separate browser tabs, the data of the first process loaded in the first tab is replaced by the second process, resulting in both tabs showing data of the same process. 

This behaviour may be explained by historical reasons (since Kitodo.Production 2 used JSF 1, which did not support `ViewScope`) but nonetheless poses some very confusing and user-unfriendly behaviour for a modern web application and frankly should be considered a nasty bug.

### LINKED ISSUE #4868: Metadata lock is not removed if metadata editor is closed without using the close buttons
I just encountered the problem that when the metadata editor is closed without using the buttons ("Schließen", "Speichern und schließen"), e.g. by jumping directly to the desktop, the metadata lock is not removed.  And it seems like that even an admin cannot easily remove that lock. 
I am not sure if this is intended since somebody might lock the editing of the metadata without being aware of it. It is probably not an easy fix but closing the metadata editor should probably always remove the lock - even if the user did not use the Buttons. Another option would be to give an indication to the user that the editor is locked for others and that he or she should should remove the lock, if the work is done.

solth:
This pull request changes the scope of the `DataEditorForm` bean from `SessionScoped` to `ViewScoped`. As suggested by @BartChris in https://github.com/kitodo/kitodo-production/issues/4868#issuecomment-985368385 it makes use of the OmniFaces variant of the `@ViewScoped` annotation. This particular version of `ViewScoped` provides a way to call a method (in this case the editors `close` method that - among others - releases the `MetadataLock` of a process) whenever the user leaves the metadata editor - even if he clicks on a separate link apart from `Close`, manually enteres a different URL or closes the browser tab/window altogether. 

To achieve this, the `@PreDestroy` annotated method of the `ViewScoped` bean utilizes a combination of `onload` event handlers and the `Navigator.sendBeacon` Javascript method (see https://showcase.omnifaces.org/cdi/ViewScoped for details). 

Because the current check for changes in the metadata editor also uses the `unload` event to display a warning message to the user, it has been removed. Since this check was bugged, however, this is a reasonable change in my opinion. (currently, the "Unsaved changes" warning often occurs even when the user did not change any date at all in the editor)

As a side effect, the metadata editor is now opened via URLs containing the ID of a process, making individual processes in the editor bookmarkable.

Fixes #3860 
Fixes #4182 
Fixes #4868 

**Edit**: the separate `ApplicationScoped` `ImageProvider` bean is required (and recommended) because `<p:graphicImage>` and `StreamedContent` (which we use for loading the images of a process in the gallery) do not work with `ViewScoped` beans (see https://stackoverflow.com/questions/8207325/display-dynamic-image-from-database-or-remote-source-with-pgraphicimage-and-str/12452144#12452144 for details)


markusweigelt (approved):
The source code looks good. Could not test the function locally yet. If it is not urgent I will do it next week. Otherwise here is the "Approve" on the source code.

thomaslow:
I quickly tested your changes with my artificial example database in the following scenarios:
- drag & drop in structure tree and gallery
- loading large process with 2000 images
- loading newspaper process with multiple parents
- loading the editor in separate structure tree mode

There were no obvious problems. Everything seemed to work fine. However, I didn't test every possible action.

The only downside I can see is that now a page refresh (F5) will immediately revert any ongoing changes. This was not the case before and I believe some users were frequently using F5 in case something was not displayed correctly in the editor. This could potentially irritate some users if there is a minor problem in the editor (e.g. clicking something causes a Javascript exception, or some other problem) and any ongoing changes can not be recovered via F5. 

In general, I agree with the solution. Maybe we should make sure to notify users that - starting from now - it is a good idea to frequently save even small changes. Or maybe we could add an old-school "are you sure to leave the editor without saving"-message?

solth:
@markusweigelt & @thomaslow thank you for the feedback!

@markusweigelt merging this PR is not immediately necessary, so I would be very glad if you'd find some time to check the functionality, too!

@thomaslow thanks for the comprehensive tests! I agree that the loss of the confirmation dialog when leaving (or reloading) the editor with unsaved changes is not optimal. The problem is that the OmniFaces `@ViewScoped` annotations `@Predestroy` method utilizes exactly the same `unload` event to achieve the functionality we need as the `Do you want to leave the page?` dialog does. 

Optimally, I would like to _first_ process the `unload` event manually to display the confirmation dialog and only if the user confirms the question in the dialog, bubble the event to the system so the `@Predestroy` method of the editor is triggered to release the metadata lock. Unfortunately I have not yet found a way to control the order of JavaScript event handlers in such a way.

Since the `Do you want to leave?` confirmation dialog is actually bugged right now (see PR description, it is often displayed after just collapsing or expanding the structure tree, even without changing any data), I think it is justifiable to remove it for the time being and inform the users accordingly in the release notes like you suggested until we find a way to re-implement the functionality so that is works correctly and in tandem with the `ViewScoped` editor bean. What do you think?

markusweigelt (approved):
Works for me!

For the implementation of `Do you want to leave?` dialog you can use the `window.onbeforeunload`. 

> This way the @ViewScoped unload can detect it and take it into account and continue to work properly.

https://javadoc.io/static/org.omnifaces/omnifaces/2.7.18/org/omnifaces/cdi/ViewScoped.html -> "Using window.onbeforeunload"

solth:
> For the implementation of `Do you want to leave?` dialog you can use the `window.onbeforeunload`.
> 
> > This way the @ViewScoped unload can detect it and take it into account and continue to work properly.
> 
> https://javadoc.io/static/org.omnifaces/omnifaces/2.7.18/org/omnifaces/cdi/ViewScoped.html -> "Using window.onbeforeunload"

Thanks for this hint! I have tried to only trigger the destruction of the bean when the user confirms the "unload" event but haven't been able to get it to work, yet. But at least the documention you quoted and linked does seem to suggest that it should be possible. I will try to incorporate that change before merging this pull request.

solth:
@markusweigelt thanks to your hints I was now able to incorporate our custom `beforeunload` event handler script in combination with the `@preDestroy` annotated `close` method in `DataEditorForm` correctly so that the bean will only be destroyed when the user confirms leaving the page despite unsaved changes. 

Would you mind reviewing this last part, too?

markusweigelt:
Now a warning message appears when data has been changed and it works very well.

Atm if you click the close button and data has been change the warn message alert does not appear.

After closing the metadata editor the metadata editor button is disabled until the next browserrefresh.
Peek 2023-02-14 15-06

solth:
> Does this occur with you as well? Maybe it is not part of this PR and occurs in the master. I'll test that right away.

Yes, you are right, that occurs for me, too. That issue will be limited to this branch because the metadata editors icon is never deactivated in the master branch, currently.

solth:
Should be fixed now

markusweigelt (changes requested):
After saving the changed data in the editor without closing it directly and without changing any further data, the warning continues to appear when you navigate away.

solth:
> After saving the changed data in the editor without closing it directly and without changing any further data, the warning continues to appear when you navigate away.

Thanks for pointing this out. Should also be fixed now.

markusweigelt (approved):
Works for me!

BartChris:
@solth It seems that the usage of the omnifaces library broke something for me when i start Kitodo from IntelliJ; might be a Java Version issue, i will investigate, what is the exact issue here:

grafik

BartChris:
Casting the httpServletRequest to FacesContext seems to fix the error for me.

```java
   if (httpServletRequest.getRequestURI().contains("pages/workflowEdit")
                        && securityAccessService.hasAuthorityToViewWorkflow()) {
                    chain.doFilter(request, response);
                } else if (ViewScopeManager.isUnloadRequest((FacesContext) httpServletRequest)) {
                    chain.doFilter(request, response);
                } else {
                    denyAccess(httpServletRequest, httpServletResponse);
                }

```

henning-gerhardt:
@BartChris: Is there a way to reproduce this? I can not confirm your mentioned issue with this change. But I got a lot of null pointer exceptions in the catalina.out file like

```
15-Feb-2023 11:10:47.112 SEVERE [http-nio-127.0.0.1-8080-exec-7] org.apache.catalina.core.StandardWrapperValve.invoke Servlet.service() for servlet [Faces Servlet] in context with path [/kitodo] threw exception
		at org.primefaces.model.DefaultStreamedContent.getStream(DefaultStreamedContent.java:104)
		at org.primefaces.application.resource.StreamedContentHandler.handle(StreamedContentHandler.java:115)
		at org.primefaces.application.resource.PrimeResourceHandler.handleResourceRequest(PrimeResourceHandler.java:108)
		at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:72)
		at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:72)
		at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:72)
		at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:72)
		at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:72)
		at javax.faces.webapp.FacesServlet.service(FacesServlet.java:196)
		at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:227)
		at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
		at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
		at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:189)
		at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
		at org.kitodo.production.servletfilter.EncodingFilter.doFilter(EncodingFilter.java:68)
		at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:189)
		at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
		at org.omnifaces.filter.GzipResponseFilter.doFilter(GzipResponseFilter.java:183)
		at org.omnifaces.filter.HttpFilter.doFilter(HttpFilter.java:108)
		at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:189)
		at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
		at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:337)
		at org.kitodo.production.security.SecurityObjectAccessFilter.doFilter(SecurityObjectAccessFilter.java:89)
		at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:346)
....
```

solth:
> @solth It seems that the usage of the omnifaces library broke something for me when i start Kitodo from IntelliJ; might be a Java Version issue, i will investigate, what the exact issue here is:

@BartChris this might be a Java version related issue. Which version are you using? Currently, Kitodo.Production only supports JDK 11, although that is not officially documented, yet.

henning-gerhardt:
@solth : Regarding the Null Pointer Exception posted above: Removing this changes did not trigger the Null Pointer Exception anymore. Step to reproduce: open a process in the meta data editor which has no image files.

solth:
> Casting the httpServletRequest to FacesContext seems to fix the error for me.
> 
> https://github.com/effective-webwork/kitodo-production/blob/b9868a93eae985ed4ae7f7202a27304e7b865413/Kitodo/src/main/java/org/kitodo/production/security/SecurityObjectAccessFilter.java#L58-L59
> 
> ```java
>    if (httpServletRequest.getRequestURI().contains("pages/workflowEdit")
>                         && securityAccessService.hasAuthorityToViewWorkflow()) {
>          chain.doFilter(request, response);
>     } else if (ViewScopeManager.isUnloadRequest((FacesContext) httpServletRequest)) {
>          chain.doFilter(request, response);
>     } else {
>          denyAccess(httpServletRequest, httpServletResponse);
>     }
> ```

@BartChris this is not caused your Java version, but by the OmniFaces library which has been updated in this pull request. You should be able to fix your error when you reload your maven dependencies before building the software.

solth:
@henning-gerhardt I cannot reproduce the error you reported. Given the NPE in the stack trace you provided occurs in the `doFilter` methods of Kitodo and OmniFaces itself during the `beforeunload` event, I would hazard a guess that it is caused by the same outdated Maven dependency.

PR #5475, opened 2022-12-06
Bump commons-net version to 3.9.0

henning-gerhardt:
From dependabot security bot:

> Prior to Apache Commons Net 3.9.0, Net's FTP client trusts the host from PASV response by default. A malicious server can redirect the Commons Net code to use a different host, but the user has to connect to the malicious server in the first place. This may lead to leakage of information about services running on the private network of the client.
The default in version 3.9.0 is now false to ignore such hosts, as cURL does. See https://issues.apache.org/jira/browse/NET-711.

Fixes CVE CVE-2021-37533 / GHSA GHSA-cgp8-4m63-fhh5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
         <commons-configuration.version>1.10</commons-configuration.version>
         <commons-io.version>2.11.0</commons-io.version>
         <commons-lang3.version>3.12.0</commons-lang3.version>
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
I did not notice that for this vulnerability pull request #5474 is already open.

Issue #5481, opened 2022-12-08
Asynchronous DMS export finishes task although export fails


BartChris:
**Describe the bug**

When exporting a process to DMS using the asynchronous mode the current task is always closed even if the export fails. The wrong behavior seems to be in these lines:

https://github.com/kitodo/kitodo-production/blob/0e71e58149b66dccef33105aafdd87e78de4a28b/Kitodo/src/main/java/org/kitodo/export/ExportDms.java#L128-L136

When a export is scheduled for background processing the method `startExport(Process process, URI unused)` always returns `true`.
At this point however the export could fail for multiple reasons, so returning true here is probably not correct.
This is problematic, because the returned boolean decides wether the export is succesful or not. This value is then also used to decide wether the current task in the workflow should be closed or not.

https://github.com/kitodo/kitodo-production/blob/0e71e58149b66dccef33105aafdd87e78de4a28b/Kitodo/src/main/java/org/kitodo/export/ExportDms.java#L83-L87

What happens is that a export does not happen but the export step is still closed.

**To Reproduce**
1. Rename the standard xsl file so that the xsl transformation of the exported Kitodo METS will fail
2. Define a task in the workflow as automatic and check the option "Export DMS"
3. Close the step before the automatic export
4. The export will fail
5. The task will be closed

**Expected behavior**
A failure in the export must set the export task to open. The workflow must not continue. The closing of the workflow step after the asynchronous task has finished succesfuly should probably happen here:

https://github.com/kitodo/kitodo-production/blob/0e71e58149b66dccef33105aafdd87e78de4a28b/Kitodo/src/main/java/org/kitodo/production/helper/tasks/ExportDmsTask.java#L63-L68

**Release**
Kitodo 3.5-beta

Issue #5008, opened 2022-02-21
Automatic export step is closed even if XSLT transformation throws errors or process data is not valid


BartChris:
When the export is configured to run automatic 

![grafik](https://user-images.githubusercontent.com/7516499/154996393-ebbc3e31-ef2f-49e8-9c38-8f939b2675f1.png)

the export step is finished even if the XSL process throws an error like:

`Writing Mets file failed!: A sequence of more than one item is not allowed as the second argument of fn:concat() (<kitodo:metadata>, <kitodo:metadata>) `

The other problem is that if the process metadata is invalid according to the ruleset nothing is exported but the export step is closed as succesful nevertheless. The export step should in both cases not be closed since no valid METS file is produced,

PR #5482, opened 2022-12-09
Fix async export

BartChris:
resolves https://github.com/kitodo/kitodo-production/issues/5481
resolves https://github.com/kitodo/kitodo-production/issues/5008

This pull request tries to adress the problems outlined in the above issue by
- not immediatly indicating a succesful export in case of an asynchronous export and prevent closing a workflow step  after a failed automatic export
- ensuring that the export status is correctly propagated through all the method calls involved
- removing the static "export finished" message from the Desktop and Process list form to ensure that a success message is only displayed in case of an actual succesful export


oliver-stoehr (approved):

solth:
@BartChris please rebase against current master since your PR is the next merge candidate

PR #5486, opened 2022-12-13
Activate 'Search' button when search field and term are not empty

### LINKED ISSUE #5485: Release 3.5.0-beta - catalogue search needs to be opened twice
### Discussed in https://github.com/kitodo/kitodo-production/discussions/5463

<div type='discussions-op-text'>

<sup>Originally posted by **aetherfaerber** November 28, 2022</sup>
After switching to the 3.5.0-beta release we are experiencing the following oddity:

1. Open a new process/„Vorgang“.
2. Click the data transfer („Datenübernahme“) button, open catalogue search.
3. The pop-up menu allows entries in the search field etc. but the Search OPAC (maybe this label should be changed?) button stays greyed out and cannot be clicked even with all values set/changed.
4. Choose Cancel.
5. Open catalogue search again.
6. Values are still set and button is now active and can be clicked.

We have set up an import configuration with a search interface of type CUSTOM and metadata format OTHER. We did not use the configuration importer to source our old kitodo_opac.xml but created a fresh configuration through the webinterface.

So far I have no idea if this i a bug or caused by a configuration on our side.

</div>

solth:
Fixes #5485 

This was caused by removing the 'no selection option' in #5430 without preselcting a valid 'Search field' option for import configurations without default search field configured!

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
@@ -242,6 +242,8 @@ public class ImportService {
             return Helper.getTranslation("recordId");
         } else if (Objects.nonNull(importConfiguration.getDefaultSearchField())) {
             return importConfiguration.getDefaultSearchField().getLabel();
+        } else if (!importConfiguration.getSearchFields().isEmpty()) {
+            return importConfiguration.getSearchFields().get(0).getLabel();
         }
         return "";
     }
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/dialogs/import.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/dialogs/import.xhtml
@@ -66,7 +66,8 @@
                                          required="#{not empty param['catalogSearchForm:performCatalogSearch']}"
                                          value="#{CreateProcessForm.catalogImportDialog.hitModel.selectedField}">
                             <f:selectItems value="#{CreateProcessForm.catalogImportDialog.searchFields}" var="field"/>
-                            <p:ajax update="catalogSearchForm:catalogSearchButton"/>
+                            <p:ajax update="catalogSearchForm:searchTerm
+                                            catalogSearchForm:catalogSearchButton"/>
                         </p:selectOneMenu>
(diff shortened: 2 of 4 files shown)


solth:
@aetherfaerber since you reported the initial issue, would you mind testing whether this branch fixes the problem for you or not?

henning-gerhardt (approved):
Code changes looks good but I tested this only with a real catalogue without a defined default search field which is not the scenario of the original issue poster.

aetherfaerber:
Again thank you @solth for fixing this issue. We have deployed this branch and can confirm that using a custom interface it does not work without this but does work with the mr applied.

aetherfaerber:
On a side note, we observed that regardless if a custom or OAI interface is used the `value` field requires a keystroke to unlock the search button. Choosing an item from the browsers autofill dropdown is not sufficient. I can't tell if this is intentional. It also might be a completely different issue.

henning-gerhardt:
> On a side note, we observed that regardless if a custom or OAI interface is used the `value` field requires a keystroke to unlock the search button. Choosing an item from the browsers autofill dropdown is not sufficient. I can't tell if this is intentional. It also might be a completely different issue.

I noted this too but this behaviour is even existing in the master branch so it is at least a new feature request and is not introduced by this pull request.

solth:
> > On a side note, we observed that regardless if a custom or OAI interface is used the `value` field requires a keystroke to unlock the search button. Choosing an item from the browsers autofill dropdown is not sufficient. I can't tell if this is intentional. It also might be a completely different issue.
> 
> I noted this too but this behaviour is even existing in the master branch so it is at least a new feature request and is not introduced by this pull request.

I added another `p:ajax` tag that should at least partially take care of this issue. It will update the "Search" button also on selecting a browser suggestion after clicking in the input text field, but only once the field looses focus, so when the user clicks somewhere outside the input text field. It's not perfect, but AFAIK there is no better JS event supported by PF that would achieve 100% what we want to do, so I hope this improvement suffices for now.

solth:
@henning-gerhardt and @aetherfaerber would you mind testing the updated branch if this sufficiently resolves the problem reported by you?

solth:
I will merge this PR now since you already approved it. If the need for another issue concerning the "Search" button arises, we can always open another issue.

PR #5495, opened 2022-12-28
Remove 'Identifier metadata' field from import configuration class

### LINKED ISSUE #5470: `identifier_metadata` not set to default value when converting catalog configurations
**Describe the bug**
When converting catalog configurations from the `kitodo_opac.xml` file to new `ImportConfiguration` instances using the catalog configuration converter, the field `identifier_metadata` should be set to the (historical) default value `CatalogIDDigital` if no other value has been configured for a catalog (using the XML element `<identifierMetadata>`) in `kitodo_opac.xml`. 

This does not happen, and since this field is mandatory in `ImportConfiguration`s, trying to edit and save a converted catalog configuration (that did not contain the mentioned `<identifierMetadata>` element) without explicitely setting the field "Metadata for catalog ID" (frontend label for `identifier_metadata`) results in an error.

**To Reproduce**
1. Make sure `kitodo_opac.xml` contains a catalog configuration without `<identifierMetadata>` element
2. Go to `Processes` -> `Import configurations`
3. Click button `Import catalogue configurations`
4. Selected the catalog mentioned at 1.
5. Start conversion of catalogue configurations
6. Added automatically created `ImportConfiguration` corresponding to mentioned catalog configuration
7. Check field "Metadata for catalog ID"

**Expected behavior**
The field `identifier_metadata` should be set to the default value `CatalogIDDigital` if no other metadata was specified in the catalogue configurion.

**Release**
3.5.0.beta-1

**Additional context**
Ideally, this metadata should not be configured at the import configuration but in the corresponding ruleset (using a new "use=" rule), because it always needs to denote an existing metadata key anyway.

solth:
Remove `identifier_metadata` property from `ImportConfiguration` class and retrieve corresponding metadata field as "Functional metadata" from ruleset instead, as originally intended.

_Note_: this makes it necessary to add the `use="recordIdentifier"` attribute to a metadata - in most cases probably `CatalogIDDigital` - in your ruleset.

Fixes #5470


solth:
@BartChris would you mind reviewing this pull request since you pointed out the best solution to me?

BartChris:
@solth Sure, i will do that!

solth:
Thanks a lot! I will assign you as reviewer.

BartChris (commented):
Your changes work for me (having tried the normal and the mass import) and i cannot detect any flaws in the logic. 
The consequence of the change is however that we have to make sure, that the functional key "recordIdentifier" is set. 

Otherwise you have the problem that:

- the system might throw a Nullpointer exception when the user searches for the title of a book in catalogue search and gets back a list of titles. The system does not now what to use to retrieve the metadata of that dataset.

solth:
Thank you for the review, @BartChris 

I think you are right and we should ensure some metadata is configured as `recordIdentifier` in the current ruleset to avoid the problems you mentioned. I would suggest to add a check to the regular and mass import pages testing exactly this condition and displaying a warning dialog with the corresponding information to the user if the condition is not met. Closing this dialog via the "Ok" button would take the user back to the desktop page.

Would that solve the problem in your opinion?

BartChris:
This is probably a good idea. The problem might be that we cannot prevent defect behaviour in all scenarios. We might integrate a check that there is at least on key defined as `recordIdentifier`. But there might be scenarios where somebody is using different metadata keys as `recordIdentifier`. E.g. for journals `CatalogIDDigitalJournal` and for multivolume works `CatalogIDDigital` .

Maybe this is unlikely but i would propose that this has to be adressed in multiple ways:

1. with a check if at least one value is set as  `recordIdentifier` as you have proposed. 
2. A note in the import dialog declaring that the user should ensure that for all doctypes imported via catalog a  `recordIdentifier` should be set. Also for parent doc types which might be imported at the same time.

Maybe in that dialog. 

grafik

3. The inclusion of the  `recordIdentifier` in all example rulesets for all(?) root doctypes in  case they are imported via catalog.
4. A explanation of that behaviour in the documentation, maybe with some "best practice" advice.

solth:
Although I am quite sure that most Kitodo.Production 3.x installations use the default fallback metadata `CatalogIDDigital` for _all_ doc types because the option to configure this metadata was not well documented or known, I agree with you that if we add such a check it should be sound. 

Additionally I think your remark is important that this functional metadata is only required when importing metadata from a remote search interface. 

Therefor I suggest to perform the check whether _all_ top level structure types (e.g. document types) in the current ruleset allow at least one functional metadata of type `recordIdentifier` right before the import dialog is displayed (the one of which you showed a screenshot in your last comment). If the configuration is _not_ complete in this regard, the dialog about missing mandatory functional metadata for catalog import is displayed instead. Clicking the "Ok" button on this dialog will close it but still allow the user to manually create a process.

As you suggested we should also add a corresponding paragraph to the documentation and adjust the example rulesets accordingly.

Would this approach be sufficient in your opinion?

BartChris:
@solth Yes, that sounds good to me, if it is possible for you to detect all the configured root types.

BartChris:
In my local setup i have a root document type which is not imported by catalog (periodical volumes), so i get a warning since no recordIdentifier is defined there. I am not sure how we can handle that. The easiest would probably be to always demand having a recordIdentifier defined even if no catalog import is used for this doc type. Otherwise we would need another mechanism to actually decide which divisions definiteley need a metadata element of type `recordIdentifier`, which complicates things a lot. 
I would therefor opt for making this recordIdentifier alwasy required for root types even if it is (right now!) not used in the application 
What do you think?

Another problematic aspect:
The detection of the root element types right now relies on them having declared a `processTitle` attribute, which is not enforced and defined a little bit unclear in the ruleset documentation (on top of the ruleset example in the repo)

<pre>
use="processTitle" -> For document types which are created with the calender without CatalogIDDigital, like newspaper issues. The process title can be transformed in the export XSL.
</pre> 

https://github.com/kitodo/kitodo-production/blob/61e5cc4bcfaac3160b2dd64995d092d56a6f83b9/Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Ruleset.java#L229-L230

We can therefor not really be sure if all root divisions can be identified.

solth:
@BartChris thanks again for the review and comments. 

I think we should apply the requirements for the existence of the functional metadata `recordIdentifier` to _all_ document types, which are defined as those divisions in the ruleset for which a `processTitle` rule exists and which are used as such in the "Create new process form", as you already noticed. 

Since adding this functional metadata to divisions like periodical volumes (that might not actually require it) does not  pose any real disadvantages I think it would be preferable to do so instead of further complicating the configuration of rulesets by unnecessarily distinguishing between imported and manually created document types.

We can of course always discuss to extend the current rules in the ruleset but that is out of scope of this pull request and the linked issue, IMO.

BartChris (approved):

BartChris:
I agree with your proposal. Changes work good for me.

PR #5497, opened 2022-12-28
Fix mapping other metadata formats and parent mapping file

### LINKED ISSUE #5458: Catalog configuration converter does not always assign correct mapping files
**Describe the bug**
Kitodo.Production provides a converter tool to transform catalog configurations from previously used `kitodo_opac.xml` files into new `ImportConfiguration` instances. This transformation also transforms any configured mapping files into new `MappingFile` objects which are then assigned to the new `ImportConfiguration`. 

Sometimes the converter assigns the wrong `MappingFiles` if there are multiple `MappingFiles` whose input metadata format corresponds to the metadata format of the `ImportConfiguration`.

**To Reproduce**
1. Add multiple catalog configurations with the same metadata format but different mapping files to `kitodo_opac.xml`
2. Use the catalog configuration converter to create `ImportConfiguration` instances from the `kitodo_opac.xml` file
3. At least one of the created `ImportConfiguration`s will have a `MappingFile` that does not correspond to the mapping files of the original catalog configuration

**Expected behavior**
The catalog configuration converter should always assign exactly those `MappingFile` instances to an `ImportConfiguration` object that were originally assigned to them in the `kitodo_opac.xml` file.

**Release**
3.5.0-beta1

### LINKED ISSUE #5494: CatalogConfiguration import fails with `UnknownFormatConversionException` for custom metadata formats
**Describe the bug**
The metadata format `OTHER` is supposed to cover all formats other than MARC and MODS that might be imported from a search interface. This does work for manually created `ImportConfiguration`s, but not when transforming catalog configurations from `kitodo_opac.xml` that configure `<metadataFormat>`s like for example `EAD` to new `ImportConfiguration` instances.

**To Reproduce**
1. Configure catalog configuration with metadata format different than MODS, MARC or PICA in `kitodo_opac.xml`
2. Try to import catalog configurations 
3. observe error

**Expected behavior**
The catalog configuration converter should set the metadata format to `OTHER` for unrecognized metadata formats in `kitodo_opac.xml` configurations.

**Screenshots**
Bildschirmfoto 2022-12-28 um 13 49 02

**Release**
3.5.0-beta1

### LINKED ISSUE #5496: `parentMappingFile` not converted for `OPAC search` type import configurations
**Describe the bug**
If a catalog configuration in the file `kitodo_opac.xml` contains a `<parentMappingFile>` element, it is not mapped to the newly created `ImportConfiguration` of type `OPAC SEARCH`. Instead, such parent mapping files are only automatically set during transformation for `ImportConfigurations` of type `FILE UPLOAD`. 

**To Reproduce**
1. Add a catalog configuration to `kitodo_opac.xml` with a `<parentMappingFile>` _and_ a search interface host configuration
2. Convert the catalog configuration using the "CatalogConfigurationConverter" of Kitodo
3. Check the settings of the resulting `ImportConfiguration`, which will not contain the configured "parentMappingFile"

**Expected behavior**
`<parentMappingFile>` settings should not only be added `ImportConfigurations` of type `FILE UPLOAD`, but also to those of type `OPAC SEARCH` automatically when importing catalog configurations from the file `kitodo_opac.xml` using the "CatalogConfigurationConverter".

**Release**
3.5.0-beta1

**Additional context**
@andre-hohmann I think this was relevant for you, wasn't it?

solth:
Fixes #5494 
Fixes #5496 
Fixes #5458

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/CatalogConfigurationImporter.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/CatalogConfigurationImporter.java
@@ -87,6 +87,7 @@ public class CatalogConfigurationImporter {
             importConfiguration.setConfigurationType(ImportConfigurationType.OPAC_SEARCH.toString());
             importConfiguration.setReturnFormat(OPACConfig.getReturnFormat(catalogName).toUpperCase());
             setMetadataFormat(importConfiguration, catalogName);
+            setParentMappingFile(importConfiguration);
             setSearchInterfaceType(importConfiguration, catalogName);
             setUrl(importConfiguration, catalogName);
             setItemFields(importConfiguration, catalogName);
@@ -188,21 +189,21 @@ public class CatalogConfigurationImporter {
         fileUploadConfiguration.setDescription(OPACConfig.getOPACDescription(catalogTitle));
         fileUploadConfiguration.setReturnFormat(OPACConfig.getReturnFormat(catalogTitle).toUpperCase());
         setMetadataFormat(fileUploadConfiguration, catalogTitle);
-        MappingFile parentMappingFile = null;
-        for (MappingFile mappingFile : ServiceManager.getMappingFileService().getAll()) {
-            if (mappingFile.getFile().equals(OPACConfig.getXsltMappingFileForParentInRecord(catalogTitle))) {
-                parentMappingFile = mappingFile;
-                break;
-            }
-        }
-        if (Objects.nonNull(parentMappingFile)) {
-            fileUploadConfiguration.setParentMappingFile(parentMappingFile);
-        }
+        setParentMappingFile(fileUploadConfiguration);
         fileUploadConfiguration.setMappingFiles(getMappingFiles(fileUploadConfiguration));
         fileUploadConfiguration.setIdentifierMetadata(OPACConfig.getIdentifierMetadata(catalogTitle));
         // update title to include "File upload" postfix! (original title is required until here to load mapping files!)
         fileUploadConfiguration.setTitle(fileUploadConfigurationTitle);
         ServiceManager.getImportConfigurationService().saveToDatabase(fileUploadConfiguration);
+    }
+
+    private void setParentMappingFile(ImportConfiguration config) throws DAOException {
+        for (MappingFile mappingFile : ServiceManager.getMappingFileService().getAll()) {
+            if (mappingFile.getFile().equals(OPACConfig.getXsltMappingFileForParentInRecord(config.getTitle()))) {
+                config.setParentMappingFile(mappingFile);
+                break;
+            }
+        }
     }
 
     private List<SearchField> getSearchFields(ImportConfiguration configuration) {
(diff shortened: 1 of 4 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/CatalogConfigurationImporter.java
```diff
@@ -222,11 +223,14 @@ private List<MappingFile> getMappingFiles(ImportConfiguration configuration) thr
...
         List<MappingFile> allMappingFiles = ServiceManager.getMappingFileService().getAll();
-        if (OPACConfig.getCatalog(configuration.getTitle()).containsKey(MAPPING_FILES)) {
+        try {
+            OPACConfig.getCatalog(configuration.getTitle()).configurationAt(MAPPING_FILES);
             for (String filename : OPACConfig.getXsltMappingFiles(configuration.getTitle())) {
                 mappingFiles.add(getConfiguredMappingFile(allMappingFiles, filename, configuration));
             }
-        } else {
+        } catch (IllegalArgumentException e) {
+            logger.debug("No 'mappingFiles' element found in catalog configuration '" + configuration.getTitle()
```
markusweigelt, line 232: I suggest to change the log level to "info". That would be a more suitable log level cause it is a relative important message out of any debug context.

markusweigelt (changes requested):

markusweigelt:
In my case the expected behaviour of #5494  is not reached. I get the error in a better way but the `metadataFormat` with value `EAD` is not assigned to metadataformat `OTHER`.
image

Kitodo/src/main/java/org/kitodo/production/helper/CatalogConfigurationImporter.java
solth: Done

solth:
@markusweigelt Thanks for the review!

> In my case the expected behaviour of #5494 is not reached. I get the error in a better way but the `metadataFormat` with value `EAD` is not assigned to metadataformat `OTHER`. For `OTHER` there exist a mapping file.

I just tried to verify whether the problem still occurs and wasn't able to reproduce it with the current state of the branch - catalog configurations with `<metadataFormat>EAD</metadataFormat>` from the `kitodo_config.properties` file are correctly converted to metadata format `OTHER` in my case.

Since the essential change to fix issue #5494 has been made in the API module this module has to be properly recreated for the fix to take effect. To avoid errors caused by outdated modules I normally delete the existing modules manually before recomiling (I have had quite a few occasions already where existing modules wouldn't get properly overwritten otherwise). Perhaps you can try the same to make sure you are testing the current state of the module?

Where you able to test whether the other two linked issues #5496 and #5458 are fixed or not?

markusweigelt (approved):

markusweigelt:
> Where you able to test whether the other two linked issues https://github.com/kitodo/kitodo-production/issues/5496 and https://github.com/kitodo/kitodo-production/issues/5458 are fixed or not?

I had tested the issues and they seemed to be fixed.

PR #5500, opened 2023-01-04
Bug import additional field twice

### LINKED ISSUE #5454: Rel 3.5.0-beta - Import of additional metadata - field was added, even if it was there before
I have tried out the feature of "Import of additional metadata" with the following data
1) Create process with meta data of K10plus, ppn=1022427806
2) Add metadata from HEBIS, ppn=058289550

==> this has created a 2nd entry of "Haupttitle (METS)" like this
image

Which IMO should not be the case (only new fields are filed, but no "overloading" is done).

Please be aware, that this is a METS-Field in the RuleSet-Definition (maybe this is the cause for this behavior):
```
		<key id="LABEL" domain="mets:div">
			<label>Main title ‹METS›</label>
			<label lang="de">Haupttitel ‹METS›</label>
		</key>

```

markusweigelt:
Fixes #5454

### DIFF
--- a/Kitodo/kitodo.iml
+++ b/Kitodo/kitodo.iml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module version="4">
-  <component name="CheckStyle-IDEA-Module" serialisationVersion="2">
-    <option name="activeLocationsIds" />
-  </component>
-</module>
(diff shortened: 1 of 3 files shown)


IkramMaalej (commented):
https://github.com/kitodo/kitodo-production/issues/5454#issuecomment-1377135145

IkramMaalej (approved):

PR #5501, opened 2023-01-06
Updated media references

### LINKED ISSUE #5281: Meta data editor: wrong validation result on wrong image count
**Describe the bug**
If you changed the amount of images on a correction workflow and use the validation button in the meta data editor you get no error message that there are not paginated images or previous paginated images are now missing. Validation tell you that everything is okay but this is not the case and the opposite to 2.x.

**To Reproduce**
1. Process a process like normal
2. Start a correction workflow in the meta data editor
3. Add / Remove some images outside of the meta data editor and close the correction workflow successful
4. Open the meta data editor for this process and click on the validation button

**Expected behavior**
Validation in meta data editor should even check that the amount of paginated images to the real amount of images inside the defined image directory like in 2.x and show error messages if there are images missing which are not paginated or paginated images which are missing or ...

**Release**
3.4.4-SNAPSHOT

### LINKED ISSUE #5285: Changes of image files are not written in METS file by "Read in pagination from images"
**Describe the bug**
In Kitodo.Production 2.x, the function "Read in pagination from images" is used to adjust the METS file (`<mets:fileSec>`, `<mets:structLink>`, ...) according to the current amount of image files in the respective image folder. This is necessary, for example, after the deletion or the addition of image files. 
In Kitodo.Production 3.x the button exists, too (See: [Beispiele](https://github.com/kitodo/kitodo-production/wiki/Paginierung#beispiele)), but without any functionality. Thus, it seems as if it is not possible to apply changes of images files to the METS file. 

**To Reproduce**
1. Delete some image files of a specific process
2. Generate the preview image files
3. Open the metadata editor
4. For the deleted image files, only empty frames are shown
5. Still, the same amount of image files are shown in the structure tree
6. It is not possible to adjust the METS file to the current amount of image files 

**Expected behavior**
If "Read in pagination from images" is clicked, the METS file is updated according to the current amount of image files. 

**Release**
- 3.4.4-SNAPSHOT

- OS: Windows 10
- Chrome 103.0.5060.134 (Offizieller Build) (64-Bit)

solth:
Fixes #5281 
Fixes #5285


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -251,7 +257,7 @@ public String open(String processID, String referringView) {
             String metadataLanguage = user.getMetadataLanguage();
             priorityList = LanguageRange.parse(metadataLanguage.isEmpty() ? "en" : metadataLanguage);
             ruleset = ServiceManager.getRulesetService().openRuleset(process.getRuleset());
-            openMetsFile();
+            mediaUpdated = openMetsFile();
```
matthias-ronge, line 260: Set `this.mediaUpdated` inside `openMetsFile()`

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -301,15 +307,15 @@ private void loadDataEditorSettings() {
      * @throws IOException
      *             if filesystem I/O fails
      */
-    private void openMetsFile() throws IOException, InvalidImagesException {
+    private boolean openMetsFile() throws IOException, InvalidImagesException {
```
matthias-ronge, line 310: A `boolean` return value from `openMetsFile()` is confusing. Please change this back to `void`, and set `this.mediaUpdated` inside the function itself.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
```diff
@@ -301,15 +307,15 @@ private void loadDataEditorSettings() {
...
         mainFileUri = ServiceManager.getProcessService().getMetadataFileUri(process);
         workpiece = ServiceManager.getMetsService().loadWorkpiece(mainFileUri);
         workpieceOriginalState = ServiceManager.getMetsService().loadWorkpiece(mainFileUri);
         if (Objects.isNull(workpiece.getId())) {
             logger.warn("Workpiece has no ID. Cannot verify workpiece ID. Setting workpiece ID.");
             workpiece.setId(process.getId().toString());
         }
-        ServiceManager.getFileService().searchForMedia(process, workpiece);
+        return ServiceManager.getFileService().searchForMedia(process, workpiece);
```
matthias-ronge, line 318: Set `mediaUpdated` here

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
```diff
@@ -347,7 +347,7 @@ private Map<String, URI> listDirectory(Pair<URI, Pattern> query, boolean absolut
         FilenameFilter filter = (dir, name) -> query.getRight().matcher(name).matches();
         try (Stream<URI> relativeURIs = fileService.getSubUris(filter, query.getLeft()).parallelStream()) {
             Stream<URI> resultURIs = absolute ? relativeURIs.map(
-                uri -> new File(FilenameUtils.concat(ConfigCore.getKitodoDataDirectory(), uri.getPath())).toURI())
+                uri -> new File(ConfigCore.getKitodoDataDirectory().concat(uri.getPath())).toURI())
```
matthias-ronge, line 350: Question: Why did you change this?

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1183,39 +1200,92 @@ private void addNewURIsToExistingPhysicalDivisions(Map<String, Map<Subfolder, UR
         }
     }
 
+    private void removeMissingMediaFromWorkpiece(List<String> mediaToRemove, Workpiece workpiece,
+                                                 Collection<Subfolder> subfolders) {
+        List<PhysicalDivision> pages = workpiece.getAllPhysicalDivisionChildrenFilteredByTypePageAndSorted();
+        for (String removal : mediaToRemove.stream().filter(Objects::nonNull).collect(Collectors.toList())) {
```
matthias-ronge, line 1206: Filtering `mediaToRemove` from `null` Strings looks like it is hiding a bug elsewhere. I assume, there shouldn’t be any `null` Strings in `mediaToRemove`. My suggestion would be not to check for it here at all.

If you absolutely **must** do this, don’t first make a list of the result again, for then only iterating over the list. Either:

    for (String removal : mediaToRemove) {
        if (Objects.isNull(removal)) {
            continue;
        }    

Or:

    mediaToRemove.stream().filter(Objects::nonNull).forEach(removal -> {

matthias-ronge (approved):
I did a code review but didn't try the functionality. I found two little things (and people can differ about that) and I had one question. Maybe you would like to take a look. But it doesn't get in the way of a merge for now, so I approve that because this pull request is a valuable improvement.

andre-hohmann (approved):
From the users perspective the functionality works correctly for the configuration and workflows in the SLUB Dresden. 

**_Test protocol_**
The following description is created due to transparency and traceability. Maybe other use cases which are not applied in the SLUB Dresden must be tested, too. 

_Workflow SLUB Dresden_
- After deleting or adding image files, the file-labels are relabeled according to the order of the original pages: 0000001, 0000002, 0000003, ...
- The jpg-files for preview and mediaview are generated externally. 

The tests were conducted with the following configuration:
- `metsEditor.useAutomaticPagination` = "true"
- `metsEditor.defaultPagination` = "uncounted"

_General observation_
1. When image files are deleted or added, the message "Geänderte Dateireferenzen festgestellt" appears. 
2. Only when a process is saved, the changes shown in the structure tree are written in the METS-divisions `<mets:structMap TYPE="PHYSICAL">` and `<mets:fileSec>`. 

_Use cases_
1. In a new process, the pages are labeled with " - "
2. If image files 13 and 14 from 14 image files are deleted, only image files 1-12 are shown in the structure tree and the values in `ORDERLABEL` are kept. 
3. If image files 13 and 14 are added to 12 image files, the image files 1-14 are shown in the structure tree and the values in `ORDERLABEL` of files 1-12 are kept. The `ORDERLABEL` for image files 13 and 14 is set to " - "
4. If image files 7 and 8 are added to 12 image files and the file-labels are labeled according to the order, the image files 1-14 are shown in the structure tree and the values in `ORDERLABEL` of files 1-12 are kept. The `ORDERLABEL` for image files 13 and 14 is set to " - "

These use cases are correct. 

**_Question_**
One use case cannot be evaluated, because the workflow is not applied in the SLUB Dresden as file names are re-labeled without gaps after corrections. 
1. Image files 13 and 14 from 16 are deleted
2. Image files 13 and 14 from 14 are added again
3. The image files 13 and 14 are shown at the end of the structure tree - expected was the order 12, 13, 14, 15, 16

METS-Aktualisierung10

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
solth: This was a tricky one. `FilenameUtils.concat` returns `null` for relative paths. And since the Kitodo data directory path is relative in the Selenium test environment, this led to `NullPointerException`s and made it impossible to retrieve the actual metadata directory of a specific process during Selenium tests. The `File.concat` function on the other hand can handle relative paths correctly, so I decided to use it here.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
henning-gerhardt: Why this in confusing? A "open" method could or should return a boolean value except there are rules to use exception for different cases. Setting a class global state is a possible solution, right, but there are a lot of cases too where this state is not changed even a change has been performed.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1183,39 +1200,92 @@ private void addNewURIsToExistingPhysicalDivisions(Map<String, Map<Subfolder, UR
...
+            for (URI mediaUri : mediaFiles.values()) {
+                if (mediaUri.getPath().startsWith(folder.getFolder().getRelativePath())) {
+                    String canonical = folder.getCanonical(mediaUri);
+                    if (Objects.nonNull(canonical)) {
+                        return canonical;
+                    }
+                }
+            }
+        }
+        return null;
```
henning-gerhardt, line 1244: According to the coding rules `null` return should be avoided. If I see the current usage right than an empty string (`""`) should be good enough too or?

henning-gerhardt (approved):
This is a code only review as the functional review was done by a team mate: A side of a maybe small change the code looks good and adding tests is very good thing.

solth:
@andre-hohmann Thank you for the thorough review!

> **_Question_** One use case cannot be evaluated, because the workflow is not applied in the SLUB Dresden as file names are re-labeled without gaps after corrections.
> 
>     1. Image files 13 and 14 from 16 are deleted
> 
>     2. Image files 13 and 14 from 14 are added again
> 
>     3. The image files 13 and 14 are shown at the end of the structure tree - expected was the order 12, 13, 14, 15, 16

If I am not totally mistaken this is something we discussed at length and agreed upon during our meeting but unfortunately missed to document in the issue description: when first opening a process that does not yet contain any links to media, the canonical file names will be used for initial sorting of the media. 

Once a workpiece already contains explicitely ordered media, however, new pages/files will not be sorted according to their filenames anymore, but instead will always have to be appended to the _end_ of the root element. This is because in Kitodo.Production 3 the `ORDER` attribute determines the order of media files and since media files can be re-ordered this `ORDER` attribute does not have to correspond with the filenames used during initial sorting of media in a process. Hence, filenames of pages added later are ignored once media with saved `ORDER` attribute exists in a process/workpiece. 

Another reason why filenames cannot be used for automatic sorting of files added in a later correction step is that at that point the process strucuture might already exist. If - in your example - page 12 belongs to one chapter and pages 15 and 16 belong to the next chapter (or a substructure of the first chapter) it would be ambiguous to determine where to add newly added pages 13 and 14 - as last pages of the first chapter or as first pages of the next chapter?

So these are the reasons new pages added at a later step are always added as _last_ media, independent of there filenames.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
solth: Thanks for pointing this out. I changed this but did not just check just for `null`, but opted for a "blank" check instead to also filter emptry Strings here - which might be returned from `getCanonical` after implementing the proposed change by @henning-gerhardt. I hope that works for you, too.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
solth: Thanks, changed!

henning-gerhardt: You added a check with the old stringutils class. Can you switch to stringutlils3 class instead or we need to add this old library as dependency again.

solth: > You added a check with the old stringutils class. Can you switch to stringutils3 class instead or we need to add this old library as dependency again.

:dizzy_face: 

thanks for pointing this out 😅

andre-hohmann:
@solth : Thanks for clarifying. I really forgot the discussion - probably, because it is not relevant for us due to the new labeling of our image files after changes. 
I know this is not part of the Issue and the PR, but it reminds me that we should think about a better documentation to prevent "unnecessary" discussions. 

From my point of view, the question is answered and thus the issue is approved from the users perspective.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
solth: I agree with @henning-gerhardt that returning a boolean to signal whether a function was executed successfully or not is neither confusing nor unheard of. Many often used methods like `File.mkdir` do something similar.

solth:
Added issue #5281 to list of fixed issed as discussed.

PR #5503, opened 2023-01-10
Set directory modules externaly

markusweigelt:
**Problem** 
The Properties Maven plugin reads the `kitodo_config.properties` and uses the property `directory.modules` as output directory to copy the module files using the Maven Resource plugin. This unfavorable behavior mixes maven properties with the configuration variables from the application. All works as far as possible when developing with native Tomcat. However, my development environment is completely based on Docker. By adjusting the `directory.modules` property in `kitodo_config.properties` and build and deploy the `WAR`, the modules directory is not found in the Docker container because the directory is only on the host. Conversely, you get the error `Cannot create resource output directory: /usr/local/kitodo/modules` if `kitodo_config.properties` is not customized. This will force you to modify the `kitodo_config.properties`. 

**Solution**

- Add `directory.modules` as maven property with Kitodo modules directory as default
- Skipping Properties Maven plugin to prevent overwriting of `directory.modules` using `-Dproperties-maven-plugin.phase=none`
- Set target directory for copy modules from maven cli using `-Dproperties-maven-plugin.phase=none -Ddirectory.modules=Target Directory`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,8 @@
         <main.basedir>${project.basedir}</main.basedir>
         <kitodo.version>3.5.0-beta-1</kitodo.version>
         <kitodo.modules>${main.basedir}/Kitodo/modules/</kitodo.modules>
+        <!-- Skip overwriting with values of kitodo_config.properties using -Dproperties-maven-plugin.phase=none as maven parameter -->
+        <directory.modules>${kitodo.modules}</directory.modules>
         <checkstyle.config.location>config/checkstyle.xml</checkstyle.config.location>
         <phase.prop>none</phase.prop>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
--- a/pom.xml
+++ b/pom.xml
@@ -30,9 +30,10 @@
     <url>https://github.com/kitodo/kitodo-production</url>
 
     <properties>
-        <kitodo.version>3.5.0-beta-1</kitodo.version>
-        <checkstyle.config.location>config/checkstyle.xml</checkstyle.config.location>
         <main.basedir>${project.basedir}</main.basedir>
+        <kitodo.version>3.5.0-beta-1</kitodo.version>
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
Mhh, I'm unsure about this fundamental change. Your indention is maybe right (splitting configuration out of the property files)  and maybe good for the docker usage but did Kitodo.Production support docker in an official way?

Should this kind of change not set inside a maven profile like all the other optional adjustments instead of set this in a global way?

If this change goes in I must adjust scripts and IDE settings for building and deploying the application with this now changed way and document the change for future use and understanding why all this parameters are needed. A lot of work for - in my opinion - only small improvement.

markusweigelt:
> Mhh, I'm unsure about this fundamental change. 

Why this is a fundamental change? Nothing has changed in the existing behavior. Only the target path of the copy for the modules was made overwritable/configurable in this PR.

The described problem occurs only in the Development Profile. Most developers use only the modules under config-local/modules for deploying and not the copy of the directory with Maven resource plugin. (Last year as I cleaned up the pom.xml, a bug with the `directory.modules` value came up and only one developer had a problem with it. For everyone else it worked since the changes were already a time in the master.)

> ... and maybe good for the docker usage but did Kitodo.Production support docker in an official way?

This is no docker specific change. Same problem may be occurs in all external/virtual environments when deploying WAR file to them. 

> If this change goes in I must adjust scripts and IDE settings for building and deploying the application with this now changed way and document the change for future use and understanding why all this parameters are needed. A lot of work for - in my opinion - only small improvement.

No there is no need to adjust something.

henning-gerhardt:
Okay, then I'm the only person which is working in a different way and usage. Good to know :-( I'm open to know how other developers are using the application from their IDE (IntelliJ IDEA) with this changes in mind.

markusweigelt:
I can overwrite `directory.modules` quite simple. 

`mvn clean install -Ddirectory.modules=TARGET MODULES`

Had just assumed that setting the property as a parameter is overridden by the Properties Maven plugin but it works.

PR #5507, opened 2023-01-16
Enable manipulation of child process links in metadata editor

### LINKED ISSUE #3275: Process link cannot be deleted

### LINKED ISSUE #4738: Enable corrections of volume-processes in the metadata editor

solth:
Fixes #3275 
Fixes #4738 

Based on #4850 by @matthias-ronge

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/MetadataPanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/MetadataPanel.java
@@ -147,7 +147,7 @@ public class MetadataPanel implements Serializable {
     }
 
     void showLogical(Optional<LogicalDivision> optionalStructure) {
-        if (optionalStructure.isPresent()) {
+        if (optionalStructure.isPresent() && Objects.isNull(optionalStructure.get().getLink())) {
             logicalMetadataTable = createProcessFieldedMetadata(optionalStructure.get());
             dataEditorForm.getAddMetadataDialog().prepareAddableMetadataForStructure(
                     getLogicalMetadataRows().getChildren());
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -1363,16 +1363,17 @@ public class StructurePanel implements Serializable {
                 dropStructure.getType(), dataEditor.getAcquisitionStage(), dataEditor.getPriorityList());
 
         LinkedList<LogicalDivision> dragParents;
-        if (divisionView.getAllowedSubstructuralElements().containsKey(dragStructure.getType())) {
+        if (divisionView.getAllowedSubstructuralElements().containsKey(dragStructure.getType())
+                || Objects.nonNull(dragStructure.getLink())) {
             dragParents = MetadataEditor.getAncestorsOfLogicalDivision(dragStructure,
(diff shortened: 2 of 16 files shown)


solth:
@BartChris thanks for the comments. 

Yes, it is already possible to (re-)establish links to subordinate processes directly from the metadata editor, and in fact it is possible from the very dialog that can be seen in your screenshot. You just have to check the checkbox "Verlinkung" and the GUI of the dialog will change to one that allows you to add process links. This means it is quite easy to create links between parent and child processes, even if a link was accidentally removed. 

Because of this I am not convinved that we need another, separate permission to control whether links between processes can be removed (and added, for that matter), but this is probably something that the users are better suited to decide than me.

Concerning the question of your second edit: that is how Kitodo 3 is designed: only the ruleset controls what structures can be contained in another structure or document. This is of course something which can be discussed, but it something which is currently at the core of Kitodo.Production.

BartChris:
@solth Thanks a lot for your explanation, i forgot about that. While checking that i encountered a bug in the search functionality, for which i made another issue. https://github.com/kitodo/kitodo-production/issues/5509

You are right that the restoration of links to subprocesses is quite easy (if the search function is repaired), but i still am afraid that the deletion of a process link might happen by accident and stays unnoticed. But maybe others have an different oppinion on that. We already had a similar discussion here  

The general feature of adding images to a superordinate process, is probably better discussed in a Github discussion.  https://github.com/kitodo/kitodo-production/pull/5406

andre-hohmann (approved):
I have tested the following scenario to modify subordinate process titles in the metadata editor view of the superordinate process. 

1 Change the order of subordinate process titles in the metadata editor 
- [The order of the subordinated process titles was changed in the metadata editor. 
- The order of the subordinated process titles was changed in the METS file of the superordinate process.]

2 Unlink subordinate process title in the metadata editor 
- An existing subordinate process title was unklinked in the metadata editor.
- An existing subordinate process title was unklinked in the METS file of the superordinate process. 

3 Add subordinate process title in the metadata editor
- A subordinate process title was added in the metadata editor. 
- A subordinate process title was added in the METS file of the superordinate process. 

The error message described in #5509 could not be reproduced.

Thus, from the user perspective the the PR is approved.

solth:
@andre-hohmann thank you for the review!

> The error message described in #5509 could not be reproduced.

@BartChris already fixed that issue with PR #5514 which has been merged into the master branch. This branch has been rebased against the updated master.

henning-gerhardt (approved):
Changes look good but the white space changes made is hard to read.

PR #5510, opened 2023-01-17
Add 'strict' mode to let metadata validation fail on warnings

solth:
The changes in this PR are based on adjustments made by @henning-gerhardt to allow validation to fail not just on `ERROR`s, but also on `WARNING`s. It also adds a new boolean parameter `validationFailOnWarning` to the file `kitodo_config.properties` to make this behaviour configurable, thus introducing a `strict` validation made.


henning-gerhardt (approved):
Changes look good and is working as expected.

PR #5513, opened 2023-01-17
Fix hierarchical mass import

### LINKED ISSUE #5512: Wrong process links created during hierarchical mass import
**Describe the bug**
When importing mulitple processes with hierarchical relations in mass import some incorrect process links can be created when importing IDs in a constellation like the following:
- [1st parent process ID]
- [ID of 1st child of 1st parent-process]
- [ID of 2nd child of 1st parent process]
- [2nd parent process ID]
- [...]
In such a situation the 2nd parent process will be incorrectly linked as another child of the 1st parent process, even though the record of the 2nd parent does not actually contain the ID of the first parent as it's own parent.

**To Reproduce**
1. Go to the 'Mass import' page to import multiple processes into a project at once
2. Upload a CSV file with IDs in the constellation described above
3. Start Mass Import

**Expected behavior**
No incorrect process links should be inserted during mass import

**Release**
3.5.0-beta1

solth:
Fixes #5512

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
@@ -1227,8 +1227,8 @@ public class ImportService {
 
     private void setParentProcess(String parentId, int projectId, Template template)
             throws DAOException, IOException, ProcessGenerationException {
+        parentTempProcess = null;
         if (StringUtils.isNotBlank(parentId)) {
-            parentTempProcess = null;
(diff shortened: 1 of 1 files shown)


solth:
@subhhwendt could you do a functional review on your test system?

subhhwendt:
> @subhhwendt could you do a functional review on your test system?

I've tested with different import files including hierachical structures - and everything was ok!

oliver-stoehr (approved):

Issue #5509, opened 2023-01-16
Searching for a process by title in linking dialogue search breaks for strings.


BartChris:
**Describe the bug**
If one searches for a string in the AddDocStructDialog

![grafik](https://user-images.githubusercontent.com/7516499/212746290-73aa881a-d02a-4d9c-a963-e9e1ab267ff5.png)

Elasticsearch throws an "search_phase_executuion_exception":

![grafik](https://user-images.githubusercontent.com/7516499/212746445-1ace07cd-4d42-452d-81bf-dbc70e328607.png)

**To Reproduce**
1. Go to the Dialog for adding a new structural element. Use the "linking" functionality.
2. Search for a string like "retro" 
3. You get an Elasticsearch exception

**Expected behavior**
The search should work. 

**Release**
Current Master

**Additional context**
It seems like both fields `id` and `title` are queries with a string. This might lead to an exception because `id` is defined of type `long` in the Elasticsearch mapping

![grafik](https://user-images.githubusercontent.com/7516499/212747554-985018ed-0541-4fd4-985e-02332334a95d.png)

I was able to fix that by defining setting the lenient attribute on the `id` field
https://javadoc.io/doc/org.elasticsearch/elasticsearch/7.5.0/org/elasticsearch/index/query/MatchQueryBuilder.html#lenient()

https://github.com/kitodo/kitodo-production/blob/964a09f5a4b6f5dd1bb19b0f480816487183befc/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java#L765-L768

```java
 BoolQueryBuilder query = new BoolQueryBuilder()
                .should(new MatchQueryBuilder(ProcessTypeField.ID.getKey(), searchInput).lenient(true))
                .should(new MatchQueryBuilder(ProcessTypeField.TITLE.getKey(), "*" + searchInput + "*"))
                .must(new MatchQueryBuilder(ProcessTypeField.RULESET.getKey(), rulesetId));
```

I am still investigating wether the returned results are actually correct.

solth:
@BartChris thanks a lot for reporting and analyzing this bug. I remember encountering it some time ago, too, but I didn't know it can be fixed this easily. Would you mind opening a pull request with your fix?

BartChris:
[...]

PR #5514, opened 2023-01-18
Fix the search for linkable children

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/5509

I implemented a wildcard query (* QUERYTERM *) on the title field to align the implementation with the general search in Kitodo.


BartChris:
@solth Oh, i was not aware of this error, i concentrated on fixing the search. I will look into that as well.

BartChris:
I fixed this error by removing those lines:

https://github.com/kitodo/kitodo-production/blob/78952540391ab3a6df2c551439688ac96ab86c09/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java#L697-707

after a process is selected we already have a handle `selectedProcess` so there is no need to search again.

PR #5517, opened 2023-01-20
Update to latest chrome version

markusweigelt:
This fix avoids the permanent adjustment of the Chrome version. The CI workflow always seems to use the latest version of the Chrome browser.

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
@@ -33,8 +33,6 @@ public class WebDriverProvider {
     private static final Logger logger = LogManager.getLogger(WebDriverProvider.class);
 
     // https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection
-    // Please don't rely on the LATEST_RELEASE file without a version suffix.
-    // It exists for backward compatibility only, and will be removed in the near future.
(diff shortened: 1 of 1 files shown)


solth:
Thanks for the pull request. This change would definitely make it simpler to keep a working Chrome driver for the Selenium tests without the frequent manual version updates.
I am a little hesitent though because of the warning just above the changes code:

>  Please don't rely on the LATEST_RELEASE file without a version suffix.
>  It exists for backward compatibility only, and will be removed in the near future.

I am not sure if it still holds true (the comment is from 2019 and the LATEST_RELEASE without suffix is still available).
@Erikmitk could you comment on this since you added the warning originally?

Erikmitk:
The comment was straight from the [documentation website](https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection).

Now it reads:

> In addition, the version of ChromeDriver for the current stable release of Chrome can be found at https://chromedriver.storage.googleapis.com/LATEST_RELEASE. **However, using on this file may be less reliable than methods described above.**

I guess we would only run into problems, if the build pipeline would use an ✌🏻outdated✌🏻 version of Chrome instead of fetching the latest release so Browser and Driver wouldn't match. But since we're lagging behind and have to adjust the URL, it would be fair to switch to use latest. @markusweigelt should remove the warning or at least only keep the link to the site.

markusweigelt:
Updated the comment: removed warning and left the link in.

PR #5534, opened 2023-02-07
Remove support for Travis use

henning-gerhardt:
CI based on Travis is not anymore the use case, so remove the needed and used code for this usage.


solth (commented):
Do you think it would be possible to adapt the implementation of `MailSender` to reuse it for Github CI instead of Travis instead of removing it? The files in question does seem to contain very little Travis specific code, so perhaps it would be possible to retain this useful functionality?

henning-gerhardt:
This could be possible but I don't know how to archive that. If somebody want to implant this functionality, feel free to re-use this code. I want to only remove unused code.

henning-gerhardt:
I don't know if this functionality was really working or was only working for specific users.

solth:
It only worked partially for me - I did get screenshots send when a selenium test failed, but those sceenshots always only contained an white, empty browser window. So I guess it didn't really work.

solth (approved):

henning-gerhardt:
Famous last words: we switched from Travis CI to GitHub actions a long time ago and no one was complaining about this missing functionality.

PR #5536, opened 2023-02-10
Fix check for unsaved changes

### LINKED ISSUE #4821: Close metadata editor without saving raises no warning?
We have observed that if you have been working in the metadata editor, then you close with the "Close" button, it goes away, but there is no warning of the form: "Do you want to exit without saving?" This is functionally correct, but we were surprised because we felt we were expecting a warning pop-up.

**Question:** Is that intended? Should we include a warning, or is it fine without?

solth:
Fixes #4821 

(see https://github.com/kitodo/kitodo-production/issues/4821#issuecomment-1426843178 for details)

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -1295,7 +1295,7 @@ public class FileService {
         }
         for (PhysicalDivision physicalDivision : physicalDivisions) {
             if (Objects.isNull(physicalDivision.getOrderlabel())) {
-                physicalDivision.setOrderlabel(" - ");
+                physicalDivision.setOrderlabel("-");
             }
         }
     }
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -610,9 +610,9 @@ public class ProcessFieldedMetadata extends ProcessDetail implements Serializabl
             throw new IllegalStateException("never happening exception");
         }
         if (skipEmpty) {
-            result.setGroup(metadata instanceof List ? metadata : new ArrayList<>(metadata));
+            result.setGroup(metadata instanceof List ? metadata : new HashSet<>(metadata));
         } else {
-            result.setGroup(new ArrayList<>(DataEditorService.getExistingMetadataRows(treeNode.getChildren())));
+            result.setGroup(new HashSet<>(DataEditorService.getExistingMetadataRows(treeNode.getChildren())));
         }
         return Collections.singletonList(result);
     }


solth:
An alternative (better?) fix could be to change the type of `MetadataGroup.group` from the abstract class `Collection` to a specific implementation like `HashSet` (which is used in the code anyway) to avoid class mismatches when comparing two `MetadataGroup` instances via its `equals` method where `group` has been set to different implementations of `Collection`. This way the risk of future code changes leading to problems like the current where two `MetadataGroup` objects are evaluated as different even though they are in fact the same could be avoided.

The other change could probably be improved upon by saving the default pagination value `-` to a constant which is than used in all relevant parts of the code to avoid using different variations of the same character in the future.

matthias-ronge:
The internal pagination value of uncounted pages should neither be “ - ” nor “-”, it is: `uncounted`, and only in the exported METS file it should be translated to “ - ” (**with** whitespace before and behind).

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1295,7 +1295,7 @@ private void repaginatePhysicalDivisions(Workpiece workpiece) {
         }
         for (PhysicalDivision physicalDivision : physicalDivisions) {
             if (Objects.isNull(physicalDivision.getOrderlabel())) {
-                physicalDivision.setOrderlabel(" - ");
+                physicalDivision.setOrderlabel("-");
```
matthias-ronge, line 1298: Should be:

    physicalDivision.setOrderlabel("uncounted");

See also: `org.kitodo.production.helper.metadata.pagination.PaginatorType.UNCOUNTED`

matthias-ronge:
> An alternative (better?) fix could be to change the type of `MetadataGroup.group` from the abstract class `Collection` to a specific implementation like `HashSet` (which is used in the code anyway) to avoid class mismatches when comparing two `MetadataGroup` instances via its `equals` method where `group` has been set to different implementations of `Collection`. This way the risk of future code changes leading to problems like the current where two `MetadataGroup` objects are evaluated as different even though they are in fact the same could be avoided.

I think you are right, and I think, this  `Collection` should be a `Set`, since metadata in MODS doesn’t have any order, and has no wish for duplicates. I made a separate issue for it:  #5543

BartChris:
> I think you are right, and I think, this `Collection` should be a `Set`, since metadata in MODS doesn’t have any order, and has no wish for duplicates. I made a separate issue for it: #5543

I am no expert in this regard, but just to ask: Do you mean with duplicates that there are no metadata values with the the same key AND the same value? A Person (=Metadata Group) can have multiple roles (all Metadata elements with key "row" inside a Group), those roles have different values usually. 
As far as i can see only the domain and the key are considered in the equals operation right now:
https://github.com/kitodo/kitodo-production/blob/22ef36fa6260a54d2b459fdba005abd5189b74a1/Kitodo-API/src/main/java/org/kitodo/api/Metadata.java#L73-L75

solth:
> The internal pagination value of uncounted pages should neither be “ - ” nor “-”, it is: `uncounted`, and only in the exported METS file it should be translated to “ - ” (**with** whitespace before and behind).

This seems to require some additional work. I will close this PR and instead open a new pull request to just fix the part about the `MetadataGroup.group` variable as described and discussed in #5543

Issue #5484, opened 2022-12-12
Add metadata group leads to two new metadata groups


andre-hohmann:
**Describe the bug**
If a metadata group (e.g. person or corporation) is added in the metadata editor to a process which already contains a metadata group the following occurs:
1. the existing metadata group is duplicated with all values in the metadata fields
2. another new and empty metadata group is added
3. Thus, instead of two expected ones, there are three metadata groups

**To Reproduce**
1. Open a process with a metadata group (e.g. PPN 1745885765) in the metadata editor
4. Click on '+' (add new metadata)
5. Choose a metadata group (e.g. person) and click 'apply'
6. Scroll down and count the amount of metadata groups

**Expected behavior**

- Only one empty metadata group should be added.
- The duplicate metadata group should not be created. 

**Release**
KITODO.PRODUCTION Version 3.5.0-beta-1

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome Version 108.0.5359.99

BartChris:
The problem seems to be in these lines:

https://github.com/kitodo/kitodo-production/blob/fb896b484c24147770a5e40a956de4142aaab20c/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java#L238-L245

both methods `addLabel` and DataEditorService.getExistingMetadataRows retrieve metadata elements, but `addLabel` constructs them from 'metadata' which holds Metadatagroups without empty value. DataEditorService.getExistingMetadataRows on the other hand returns MetadataGroups with empty values (this is explicitely set here: https://github.com/kitodo/kitodo-production/blob/ed3b6c6301630e0d045156e49d3171b5328ae3ac/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java#L277), which are added to the collection `entered` which is implemented as a Hashset. 

So although they is only one Metadata Group on the page Kitodo thinks there are two (one with empty values and the other without), so it renders two old groups and the new one.

I am not quite sure wether this problem is directly associated with the problem discussed here:
https://github.com/kitodo/kitodo-production/pull/5536

BartChris:
This problem also appears when any other field is added to the metadata tree, not only complex elements. It does not appear if all subvalues of group are filled out because then there is no difference in retrieving the elements with different settings for empty elements.

PR #5549, opened 2023-02-21
Avoid duplication of metadata groups while adding new metadata fields

BartChris:
Resolves [5484](https://github.com/kitodo/kitodo-production/issues/5484)

As outlined in the issue: existing metadata groups are added twice to the list of metadata elements when new elements are added to the list. This PR addresses this by explicitely removing the metadata groups which have the empty elements filtered out before the existing groups are reconstructed from the tree (including their empty values).


IkramMaalej (approved):

BartChris:
I have no idea why the Selenium tests are failing. They all pass in my local environment.

solth:
@BartChris please rebase against current master as your pull request is the next merge candidate.

Issue #4357, opened 2021-04-22
Metadata should be sorted to better support diff’ing


matthias-ronge:
The metadata is stored in Kitodo in a random order. This is correct, as metadata is out of order in MODS. However, when two file versions are compared by the programmer (with “diff”), there are many changes, and these are not changes, but they are caused due to the random order of entries.

**Goal:** is to save the metadata in sorted order, by key, and then by value. Since the order is meaningless, it can also be stored alphabetically. It makes no difference, but, for the implementer, that's great because “diff” works just fine.

PR #5556, opened 2023-02-23
Change MetadataGroup.group variable to HashSet

### LINKED ISSUE #5543: Improve MetadataGroup
@solth wrote in #5536:

> An alternative (better?) fix could be to change the type of `MetadataGroup.group` from the abstract class `Collection` to a specific implementation like `HashSet` (which is used in the code anyway) to avoid class mismatches when comparing two `MetadataGroup` instances via its `equals` method where `group` has been set to different implementations of `Collection`. This way the risk of future code changes leading to problems like the current where two `MetadataGroup` objects are evaluated as different even though they are in fact the same could be avoided.

I think that `Collection` should be a `Set`, since metadata in MODS doesn’t have any order, and has no wish for duplicates. We can change this separately

solth:
Fixes #5543 

This fixes one part of #4821

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/MetadataGroup.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/MetadataGroup.java
@@ -11,7 +11,6 @@
 
 package org.kitodo.api;
 
-import java.util.Collection;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.stream.Collectors;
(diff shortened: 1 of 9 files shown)


matthias-ronge (approved):

matthias-ronge:
Could you replace `HashSet` with `TreeSet`? This would then also fix #4357. Just a question. You can also first merge this

solth:
I considered using `TreeSet` instead of `HashSet` (as you already suggested during our conversation) but ultimately decided against it because of potential performance implications (see point 2.4 on https://www.baeldung.com/java-hashset-vs-treeset). 

Although I can see the advantage of always having the same order of metadata currently I think avoiding any impacts on performance especially in the metadata editor has to have higher priority.

PR #5585, opened 2023-03-08
Create correct metadatatype

### LINKED ISSUE #5583: Selecting a Metadagroup when creating several divisions results in exception.
**Describe the bug**
In the metadataeditor several structures can be added at once with optional metadata. According to the defintions in the ruleset, the keys of both simple metadataentries and/or metadatagroups can be selected. If a metadatagroup is selected, clicking on the generated structure element throws an exception. The reason is that the generated metadata is a simple metadata entry, although the ruleset specifies it to be a metadata group.

**To Reproduce**
1. Open metadataeditor.
2. Open "New Docstruct" dialog (e.g. by right-clicking a structure element and clicking "Add Element").
3. Select a Structure element and set number of elements to > 1
4. Select a metadatum type that is a metadatagroup in the relevant ruleset.
5. Click apply.
6. Click on the new structure.
7. Exception "java.lang.IllegalStateException: Got simple metadata entry with key "person" which is declared as substructured key in the rule set." is thrown.

**Expected behavior**
The system should be able to correctly generate either metadata groups and metadata entries from the given metadata key.

**Screenshots**

**Release**
Observed on 3.4.2 and current master branch (since #5466).

 - OS: [Windows 10]
 - Browser [Firefox]

**Additional context**

pontus-osterdahl:
This pull request fixes #5583 by ensuring that the correct subclass of `Metadata` is created from the metadatakey passed to the newly created division.


IkramMaalej:
Thanks for this Fix.
The problem with the "java.lang.IllegalStateException" no longer occurs.
When a complex metadata key is selected, I think that the "Value" input field in the dialog should not be rendered because its entered value will be lost, or we can also show the metadata subfields of the selected metadata group, wich is a more complicated solution.

IkramMaalej (commented):
Thanks for this Fix.
The problem with the "java.lang.IllegalStateException" no longer occurs.
When a complex metadata key is selected, I think that the "Value" input field in the dialog should not be rendered because its entered value will be lost, or we can also show the metadata subfields of the selected metadata group, wich is a more complicated solution.

pontus-osterdahl:
Thanks for your comment! I would tend to, at least for now, go with the first suggestion. To show the subfields would be great, but I am not sure that it is worth the effort since I do not know how often this functionality would actually be used.

IkramMaalej (approved):

PR #5592, opened 2023-03-11
Expire user session on http session timeout

### LINKED ISSUE #5591: "logged in" status in user list not updated on session expiration
**Describe the bug**
The user list does not reliably display the "logged in" status of users (with the green check mark) because the underlying Spring session is only updated when the user actively logs out of the system, not when the configured session timeout is reached. Even if a user is effectively logged out after his session expires, he still appears as "logged in" on the user page/user list.

**To Reproduce**
1. (optional) lower `session-timeout` value in `web.xml` to 1 and restart Tomcat to make (testing) life easier
2. Log into Kitodo with one account in one browser
3. Log into Kitodo with another account in a different browser
4. Observe user list in first browser
5. Refresh page (just in first browser) from time to time to refresh session and avoid timeout with the corresponding user
6. Once session timeout is reached for the _second_ user, he will be logged out (resulting in page reloads in second browser to be redirected to login page)
7. In the first browser, though, the second user will still appear with the green check mark signaling he is logged in, even after reloading the page/list

**Expected behavior**
The user list should reliably reflect whether a user is logged in or not and the green check mark should disappear - on page reload - for users that have been logged out by session expiration.

**Screenshots**
Bildschirm­foto 2023-03-10 um 22 52 07

**Release**
3.6.0-SNAPSHOT

**Additional context**
This issue is related to #3169 and #4129 but describes a separate problem.

solth:
Fixes #5591

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/version/KitodoVersionListener.java
+++ b/Kitodo/src/main/java/org/kitodo/production/version/KitodoVersionListener.java
@@ -25,6 +25,10 @@ import javax.servlet.http.HttpSessionBindingEvent;
 import javax.servlet.http.HttpSessionEvent;
 import javax.servlet.http.HttpSessionListener;
 
+import org.kitodo.production.security.SecurityUserDetails;
+import org.kitodo.production.services.ServiceManager;
+import org.springframework.security.core.context.SecurityContextImpl;
+
 /**
  * Listener to set up Kitodo versioning information from Manifest on application
  * startup.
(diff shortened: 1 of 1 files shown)


BartChris:
I succesfully tested your changes: After session expiration in browser 2 the user is no longer marked as logged in in browser 1. Functionality wise i approve the PR. I have no real knowledge on the session handling in general in Kitodo, so i cannot judge the changes technically.

Issue #5579, opened 2023-03-02
File references are gone on editing meta data entries of finished process


henning-gerhardt:
**Describe the bug**

Solving the issue #5285 changed the update of file references. A lot of use cases were discussed but this one was missed:

* Opening completed processes (all tasks green) without files (media files (LOCAL) and derivatives) in the metadata editor leads to the loss of the file references.
  * If either the media files inside the source directory (LOCAL) or the media files in the derivative directories are available then the file references are kept.
* There is no message, that the references are removed
* We discovered this use case because of [this comment](https://github.com/kitodo/kitodo-production/issues/4999#issuecomment-1447743880).

The use case is important for the SLUB Dresden due to the following reasons: 

1. The media files (LOCAL) are archived, when a process is completed. 
2. The derivatives are deleted, when a process is completed. 
3. In completed processes, metadata is corrected or completed, or structural data is supplemented. 

Thus, it is now impossible to correct metadata of completed processes and save the corrections without losing the references. 

**To Reproduce**

1. Open a process in meta data editor which contains file references but no the media files (original and derivatives)
2. Save the process in meta data editor
3. References to files are gone

**Expected behavior**

If processes with references but without any media files are modified in the metadata editor, the references should not be removed. 

At least, there should be a message that includes an explicit request for the user to delete the references. If this is not confirmed, the references remain. This message must be worded very clearly to avoid accidental wrong selection. 

**Release**

master branch from 2023-03-01 commit 414b1b4

BartChris:
> At least, there should be a message that includes an explicit request for the user to delete the references. If this is not confirmed, the references remain. This message must be worded very clearly to avoid accidental wrong selection.

I also think that this is absolutely essential. I want to add here that there might also be the scenario, that only the large Tiff files are moved to another system and the derivatives stay where they are. It is OK if Kitodo would not be fully functional at this point but the System should never automatically without any warning remove references to the originals.  I have not checked this: Was the concept of deletion of references originally not bound to the user expliciteley using a button? (https://github.com/kitodo/kitodo-production/issues/5285)

andre-hohmann:
@BartChris : 
You are right, as long as derivatives or the original data are available, the references are not removed. 
[...]

PR #5598, opened 2023-03-15
Skip automatic removal of missing media from workpiece, when no media…

markusweigelt:
… was found, Fix some potential npes

Fixes issue #5579

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -292,12 +292,18 @@ public class StructurePanel implements Serializable {
     }
 
     Optional<LogicalDivision> getSelectedStructure() {
+        if (Objects.isNull(selectedLogicalNode)) {
+            return null;
+        }
         StructureTreeNode structureTreeNode = (StructureTreeNode) selectedLogicalNode.getData();
         Object dataObject = structureTreeNode.getDataObject();
         return Optional.ofNullable(dataObject instanceof LogicalDivision ? (LogicalDivision) dataObject : null);
     }
 
     Optional<PhysicalDivision> getSelectedPhysicalDivision() {
+        if (Objects.isNull(selectedPhysicalNode)) {
+            return null;
+        }
         StructureTreeNode structureTreeNode = (StructureTreeNode) selectedPhysicalNode.getData();
         Object dataObject = structureTreeNode.getDataObject();
         return Optional.ofNullable(dataObject instanceof PhysicalDivision ? (PhysicalDivision) dataObject : null);
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -293,7 +293,7 @@ public class StructurePanel implements Serializable {
 
     Optional<LogicalDivision> getSelectedStructure() {
         if (Objects.isNull(selectedLogicalNode)) {
-            return null;
+            return Optional.empty();
         }
         StructureTreeNode structureTreeNode = (StructureTreeNode) selectedLogicalNode.getData();
         Object dataObject = structureTreeNode.getDataObject();
@@ -302,7 +302,7 @@ public class StructurePanel implements Serializable {
 
     Optional<PhysicalDivision> getSelectedPhysicalDivision() {
         if (Objects.isNull(selectedPhysicalNode)) {
-            return null;
+            return Optional.empty();
         }
         StructureTreeNode structureTreeNode = (StructureTreeNode) selectedPhysicalNode.getData();
         Object dataObject = structureTreeNode.getDataObject();
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
@@ -210,7 +210,7 @@ public class KitodoScriptService {
     }
 
     private boolean executeRemainingScript(List<Process> processes, String script)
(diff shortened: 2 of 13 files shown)


markusweigelt:
@henning-gerhardt 

> At least, there should be a message that includes an explicit request for the user to delete the references. If this is not confirmed, the references remain. This message must be worded very clearly to avoid accidental wrong selection.

With the current implementation of this pr, the references are not remove and only a warning message is shown what happend.
Is it necessary to show a dialog to select loading behaviour.  I think it can be very annoying and may also lead to user errors. What do you think? Does the problem occur not so often and the selection errors are negligible (cause you can reload metadataeditor if you select the wrong loading behaviour if the user notices it)?

andre-hohmann:
@markusweigelt : Thanks a lot!

A warning without a dialog function is absolutely sufficient. If someone needs a dialog function, this can be developed supplementary. 

Unfortunately, it is difficult to estimate how often certain cases occur in the respective facilities. The discussion following  https://github.com/kitodo/kitodo-production/issues/5579#issuecomment-1460308477 confirms in my opinion, that your suggestion is the best way to solve the basic problem (removal of needed references) with little effort. 

Especially https://github.com/kitodo/kitodo-production/issues/5579#issuecomment-1466264052 shows that there are probably no use cases in which the removal of references is needed.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -292,12 +292,18 @@ public void setSelectedPhysicalNode(TreeNode selectedPhysicalNode) {
     }
 
     Optional<LogicalDivision> getSelectedStructure() {
+        if (Objects.isNull(selectedLogicalNode)) {
+            return null;
```
henning-gerhardt, line 296: According to Coding Guidelines returning `null` value should be avoided. There are a few places in the source code where an additionl null check must be added and not only at the one place later in this pull request.

According to my IDE (IntelliJ IDEA) the return statement here should be written as `return Optional.empty()`.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -292,12 +292,18 @@ public void setSelectedPhysicalNode(TreeNode selectedPhysicalNode) {
...
+        }
         StructureTreeNode structureTreeNode = (StructureTreeNode) selectedLogicalNode.getData();
         Object dataObject = structureTreeNode.getDataObject();
         return Optional.ofNullable(dataObject instanceof LogicalDivision ? (LogicalDivision) dataObject : null);
     }
 
     Optional<PhysicalDivision> getSelectedPhysicalDivision() {
+        if (Objects.isNull(selectedPhysicalNode)) {
+            return null;
```
henning-gerhardt, line 305: See comment above.

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -203,13 +204,13 @@ private boolean executeOtherScript(List<Process> processes, String script)
                 copyDataToChildren(processes, script);
                 break;
             default:
-                return executeRemainingScript(processes, script);
+                return executeRemainingScript(processes);
```
henning-gerhardt, line 207: Change is not related to the issue. Please revert this and open a separate pull request for this change.

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -203,13 +204,13 @@ private boolean executeOtherScript(List<Process> processes, String script)
...
         }
         return true;
     }
 
-    private boolean executeRemainingScript(List<Process> processes, String script)
-            throws DataException, IOException, InvalidImagesException {
+    private boolean executeRemainingScript(List<Process> processes)
```
henning-gerhardt, line 212: See comment above. Removing of the unused parameter is a separate issue.

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
```diff
@@ -242,7 +242,7 @@ public static List<SelectItem> getAddableMetadataForMediaUnit(DataEditorForm dat
      */
     public static StructuralElementViewInterface getStructuralElementView(DataEditorForm dataEditor) {
         Optional<LogicalDivision> selectedStructure = dataEditor.getSelectedStructure();
-        if (selectedStructure.isPresent()) {
+        if (Objects.nonNull(selectedStructure) && selectedStructure.isPresent()) {
```
henning-gerhardt, line 245: The null check could be avoided if the `getSelectedStructure()` method is changed in the already mentioned way (`return null` -> `return Optional.empty()`.

henning-gerhardt (changes requested):
This is a code only review. Functionality check will be performed later. Maybe adding tests should help here and prevent changes in future in a bad way.

markusweigelt:
> Especially [#5579 (comment)](https://github.com/kitodo/kitodo-production/issues/5579#issuecomment-1466264052) shows that there are probably no use cases in which the removal of references is needed.

Maybe it makes more sense to take out the deletion directly than to introduce this distinction. Alternatively we could also add a project option to clean up the file references which is turned off by default or a button in metadataeditor to do it manually.

Who decides here if this is still needed?

andre-hohmann:
@markusweigelt : Thanks a lot for the PR. I had a look on it and it works!
The only thing I noticed is that when the language is set to English, the message in German language is displayed. 
It's not "bad" and maybe it's just because of our configuration. Nevertheless, I want to draw attention to it.

henning-gerhardt (approved):
Changes looks nice and clean and adding a test is very nice.

Why the German warn message appears but the interface is set in English is unclear and maybe not related to the changes itself.

PR #5603, opened 2023-03-27
Prevent NPE when creating unknown metadata group and metadata

### LINKED ISSUE #5217: NPE during import when import mapping and ruleset do not match

solth:
This PR fixes #5217 by preventing the NullPointerException that is currently being displayed to the user when the import mapping xslt file creates a metadata group that is undefined in the used ruleset.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessTextMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessTextMetadata.java
@@ -43,7 +43,7 @@ public class ProcessTextMetadata extends ProcessSimpleMetadata implements Serial
     }
 
     private String addLeadingZeros(String value) {
-        if (Objects.equals(super.settings.getInputType(), InputType.INTEGER)) {
+        if (Objects.nonNull(super.settings) && InputType.INTEGER.equals(super.settings.getInputType())) {
             int valueLength = value.length();
             int minDigits = super.settings.getMinDigits();
             return valueLength >= minDigits ? value : "0".repeat(minDigits - valueLength).concat(value);
--- a/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
@@ -139,13 +139,15 @@ public class DataEditorService {
         ProcessFieldedMetadata fieldedMetadata = ((ProcessFieldedMetadata) metadataNode.getData());
         ComplexMetadataViewInterface metadataView = fieldedMetadata.getMetadataView();
         List<SelectItem> addableMetadata = new ArrayList<>();
-        for (MetadataViewInterface keyView : metadataView.getAddableMetadata(fieldedMetadata.getChildMetadata(),
-                fieldedMetadata.getAdditionallySelectedFields())) {
-            addableMetadata.add(
-                    new SelectItem(keyView.getId(), keyView.getLabel(),
-                            keyView instanceof SimpleMetadataViewInterface
-                                    ? ((SimpleMetadataViewInterface) keyView).getInputType().toString()
-                                    : "dataTable"));
+        if (Objects.nonNull(metadataView)) {
+            for (MetadataViewInterface keyView : metadataView.getAddableMetadata(fieldedMetadata.getChildMetadata(),
+                    fieldedMetadata.getAdditionallySelectedFields())) {
+                addableMetadata.add(
+                        new SelectItem(keyView.getId(), keyView.getLabel(),
+                                keyView instanceof SimpleMetadataViewInterface
+                                        ? ((SimpleMetadataViewInterface) keyView).getInputType().toString()
+                                        : "dataTable"));
+            }
         }
         return sortMetadataList(addableMetadata, ruleset);
     }
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -578,16 +578,18 @@ public class ProcessFieldedMetadata extends ProcessDetail implements Serializabl
         assert division == null;
         MetadataGroup result = new MetadataGroup();
         result.setKey(metadataKey);
-        result.setDomain(DOMAIN_TO_MDSEC.get(metadataView.getDomain().orElse(Domain.DESCRIPTION)));
-        try {
-            this.preserve();
-        } catch (NoSuchMetadataFieldException e) {
-            throw new IllegalStateException("never happening exception");
-        }
-        if (skipEmpty) {
-            result.setMetadata(metadata instanceof List ? metadata : new HashSet<>(metadata));
-        } else {
-            result.setMetadata(new HashSet<>(DataEditorService.getExistingMetadataRows(treeNode.getChildren())));
+        if (Objects.nonNull(metadataView)) {
+            result.setDomain(DOMAIN_TO_MDSEC.get(metadataView.getDomain().orElse(Domain.DESCRIPTION)));
+            try {
+                this.preserve();
+            } catch (NoSuchMetadataFieldException e) {
+                throw new IllegalStateException("never happening exception");
+            }
+            if (skipEmpty) {
+                result.setMetadata(metadata instanceof List ? metadata : new HashSet<>(metadata));
+            } else {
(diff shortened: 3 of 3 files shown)


BartChris:
This might be about a special scenario but nevertheless: I tried the changes with the scenario described in the linked issue (importing an entry from K10 plus without having all keys defined in the ruleset). The first NPE does not happen anymore. I still get an NPE in this specific scenario because "PERSON" was not defined in my ruleset and the construction of the ATS does fail, since it identifies a PERSON key in the data, but not the other values which are required here.

	at org.kitodo.production.forms.createprocess.ProcessTextMetadata.getMetadataID(ProcessTextMetadata.java:62)
	at org.kitodo.production.services.data.ImportService.getListOfCreators(ImportService.java:827)

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessTextMetadata.java
```diff
@@ -43,7 +43,7 @@ public ProcessTextMetadata(ProcessTextMetadata template) {
     }
 
     private String addLeadingZeros(String value) {
-        if (Objects.equals(super.settings.getInputType(), InputType.INTEGER)) {
+        if (Objects.nonNull(super.settings) && InputType.INTEGER.equals(super.settings.getInputType())) {
```
matthias-ronge, line 46: I would like better:

    Objects.nonNull(super.settings) && Objects.equals(InputType.INTEGER, super.settings.getInputType())
Not a requirement, but I think it is better readable. And, Objects.equals does `null`-checks.

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
```diff
@@ -136,13 +136,15 @@ public static List<SelectItem> getAddableMetadataForGroup(Ruleset ruleset, TreeN
...
         ComplexMetadataViewInterface metadataView = fieldedMetadata.getMetadataView();
         List<SelectItem> addableMetadata = new ArrayList<>();
-        for (MetadataViewInterface keyView : metadataView.getAddableMetadata(fieldedMetadata.getChildMetadata(),
-                fieldedMetadata.getAdditionallySelectedFields())) {
-            addableMetadata.add(
-                    new SelectItem(keyView.getId(), keyView.getLabel(),
-                            keyView instanceof SimpleMetadataViewInterface
-                                    ? ((SimpleMetadataViewInterface) keyView).getInputType().toString()
-                                    : "dataTable"));
+        if (Objects.nonNull(metadataView)) {
```
matthias-ronge, line 139: Wondering why `metadataView` can be `null`. This looks like a bug elsewhere.

matthias-ronge (commented):
Seems neither wrong nor right, see comments.

solth:
> Edit: This seem to stem from the specific functionality attached to keys named as "Person", it does not happen if if define a "Person" key in the ruleset, even with different subkeys than the one coming from K10-plus. I have not tested it yet with other datasets yet where the ATS related NPE does not happen.

@BartChris thanks for this clarification. I think "Person" is a special case and separate from the issue I tried to fix with this pull request. 

Do you think the linked issue is sufficiently resolved with these proposed changes or are there other suggestions or change requests you would like to see? (e.g. as @matthias-ronge mentioned there seem to be other issues in the metadata handling which are not addressed by this pull request. It is merely ment to handle the exception that can occur when ruleset and xml transformation script do not match a little more gracefully. It does not fix the underlying problem.)

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessTextMetadata.java
solth: In this particular case I disagree. The additional `Objects.equals` is not required and unnecessarily complicates this check, since `InputType.INTEGER` is not null and therefor we do not need a null check here (calling the `equals` method on contants or enums when comparing with variables is an encouraged way for avoiding null pointer exceptions). So if you do not mind I would leave this part unchanged.

Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
solth: Yes, I think you are correct. This pull request is not meant to fix the underlying problem of the metadata handling but is just intended to better deal with the resulting null pointer exception.

markusweigelt (changes requested):
Does not work for me because I still get an NPE.

What have I done?
- Created a process with metadata group
- Check if the group is available in metadata editor
- Change the group name in `meta.xml` on filesystem
- Open the metadata editor again -> NPE

```
	at org.kitodo.production.forms.createprocess.ProcessSimpleMetadata.isRequired(ProcessSimpleMetadata.java:96)
	at org.kitodo.production.forms.createprocess.ProcessTextMetadata.isRequired(ProcessTextMetadata.java:30)
...	
```

markusweigelt:
Ok I have read over that. I will check this only for the import again.

markusweigelt:
Not existing top level fields working

`<kitodo:metadata name="notExistingField"><xsl:text>Lorem ipsum dolor sit amet</xsl:text></kitodo:metadata>`

but I am not able to import without NPE with non-existent top-level groups.

`
<kitodo:metadataGroup name="notExistingGroup">
     <kitodo:metadata name="notExistingField"><xsl:text>Lorem ipsum dolor sit amet</xsl:text></kitodo:metadata>
</kitodo:metadataGroup>
`
or

`<kitodo:metadataGroup name="notExistingGroup">
     <kitodo:metadata name="TitleDocMain"><xsl:text>Lorem ipsum dolor sit amet</xsl:text></kitodo:metadata>
</kitodo:metadataGroup>
`
producing following NPE

`
	at org.kitodo.production.forms.createprocess.ProcessTextMetadata.addLeadingZeros(ProcessTextMetadata.java:46)
	at org.kitodo.production.forms.createprocess.ProcessTextMetadata.<init>(ProcessTextMetadata.java:37)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createMetadataEntryEdit(ProcessFieldedMetadata.java:375)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createUndefinedMetadataTable(ProcessFieldedMetadata.java:258)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createMetadataEntryEdit(ProcessFieldedMetadata.java:385)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.createMetadataTable(ProcessFieldedMetadata.java:233)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.buildTreeNodeAndCreateMetadataTable(ProcessFieldedMetadata.java:156)
	at org.kitodo.production.forms.createprocess.ProcessFieldedMetadata.<init>(ProcessFieldedMetadata.java:134)
	at org.kitodo.production.helper.ProcessHelper.initializeProcessDetails(ProcessHelper.java:102)
	at org.kitodo.production.helper.ProcessHelper.transformToProcessDetails(ProcessHelper.java:74)
	at org.kitodo.production.helper.ProcessHelper.generateAtstslFields(ProcessHelper.java:161)
	at org.kitodo.production.services.data.ImportService.importProcessHierarchy(ImportService.java:528)
	at org.kitodo.production.forms.createprocess.CatalogImportDialog.getRecordHierarchy(CatalogImportDialog.java:181)
	at org.kitodo.production.forms.createprocess.CatalogImportDialog.getRecordById(CatalogImportDialog.java:247)
	at org.kitodo.production.forms.createprocess.CatalogImportDialog.search(CatalogImportDialog.java:108)
`

markusweigelt:
We could also provide a general error message here, indicating that most likely the Kitodo XML generated after the XSL transformation is faulty. The errors I have noticed include, for example, empty groups when fields imported from K10Plus are empty, and there is a condition for checking them in XSL but no condition for the group.

solth:
> We could also provide a general error message here, indicating that most likely the Kitodo XML generated after the XSL transformation is faulty. The errors I have noticed include, for example, empty groups when fields imported from K10Plus are empty, and there is a condition for checking them in XSL but no condition for the group.

That would be the optimal solution, in my opinion - show a warning dialog that something when wrong with the import, likeley related to the ruleset and the mapping file not being compatible, and that additional information can be found in the logs, where the exact error message (and perhaps stack trace) should be logged.

solth:
Closing this stale pull request. Will open an new one if updated solution is available.

PR #5609, opened 2023-03-28
Don't complain required fields from an altogether empty metadata group

### LINKED ISSUE #5297: Validation mis-enforces required fields in empty always-showing metadata group
**Describe the bug**
If you have an `alwaysShowing` nested metadata key, with sub-keys that are declared `minOccurs` > 0, in process creation, the validation complains about these fields to be empty and refuses creating the process.

**Expected behavior**
If the `alwaysShowing` metadata group is all empty, the validation should not complain about these fields to be filled. If there is any data in any sub-field of the metadata group, then the validation should require these fields to be filled.

matthias-ronge:
If for a metadata group, all children are empty, then the whole group is empty, and then, ignore 'minOccurs' from the child.

Fixes #5297

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessSimpleMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessSimpleMetadata.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.function.BiConsumer;
 
+import org.kitodo.api.dataeditor.rulesetmanagement.ComplexMetadataViewInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.MetadataViewInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.SimpleMetadataViewInterface;
 import org.kitodo.api.dataformat.Division;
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessSimpleMetadata.java
```diff
@@ -86,6 +87,12 @@ public boolean isUndefined() {
     }
 
     public boolean isRequired() {
+        if (container != null && container.getChildMetadata().isEmpty()) {
```
IkramMaalej, line 90: The required simple metadata of a 'docType' are also no longer required.
The simple metadata also has a container wich is the selected docType (volume...) , so this condition "Container != null && container.getChildMetadata().isEmpty()" is not precise enough to distinguish between simple metadata of  the selected docType or simple metadata of a metadata group.

IkramMaalej (changes requested):

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessSimpleMetadata.java
matthias-ronge: Since a docType is “simple”, it shouldn't have a container, or am I misunderstanding something?

Can you give me an example (a ruleset) where the problem occurs?

IkramMaalej: The doctype itself doesn't have a container, but the simple metadata like "HaupTitel" has a container, which is the doctype (e.g. volume)  and where all the metadata of the doctype are located.

IkramMaalej (approved):

PR #5613, opened 2023-03-30
Make behaviour of repeated import configurable

### LINKED ISSUE #5410: Make behaviour of repeated import configurable
Production makes it possible to carry out several imports “on top of each other”. If metadata is imported with different keys, these are aggregated in the display. However, if metadata with the same key is imported multiple times, the existing value for that key is replaced by the newly imported one.

In our use case, it is desirable that the additionally imported metadata is also displayed. The key should be repeated, the previously imported value should be kept as well.

As a result of the practice meeting, a solution was designed: The behavior of individual keys should be made controllable via the ruleset. In the `<editing>` section, it should be possible to specify for each key, whether it should be _replaced_ (default), _retained_ or _added_ when it is imported again.

matthias-ronge:
Adds a new control in the `<settings>` in the ruleset, which determines how metadata behaves in the "additional import" function. There are three settings, `reimport="replace"` will overwrite the existing metadata with the new ones (this is already the default case), `reimport="add"` will add the new metadata to the older ones, and `reimport="keep"` will keep the old existing metadata without replacing them. The renewed values are not imported at all.

Resolves #5410


Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
```diff
@@ -321,5 +326,50 @@ public boolean isAlwaysShowingForKey(String keyId) {
...
+                    currentMetadata.addAll(currentEntries);
+                    currentMetadata.addAll(updateEntries);
+                    break;
+                case KEEP:
+                    currentMetadata.addAll(currentEntries);
+                    break;
+                case REPLACE:
+                    currentMetadata.addAll(updateEntries.isEmpty() ? currentEntries : updateEntries);
+                    break;
+                default: throw new IllegalStateException("complete switch");
```
henning-gerhardt, line 370: Please use an other error message here. If someone get this message as he use a non valid entry it would be very difficult to find the right place and the causing issue. Maybe something like that `"Used not supported reimport case {}"` where`{}` is replaced by the wrong value?

henning-gerhardt (changes requested):
It is only a code review. I did not check any added nor changed functionality.

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
```diff
@@ -321,5 +326,50 @@ public boolean isAlwaysShowingForKey(String keyId) {
         return false;
     }
 
+    @Override
+    public int updateMetadata(Collection<Metadata> currentMetadata, String acquisitionStage,
+            Collection<Metadata> updateMetadata) {
+
+        final Function<String, MutableTriple<Collection<Metadata>, Reimport, Collection<Metadata>>> entryProducer = key -> {
+            var entry = new MutableTriple<Collection<Metadata>, Reimport, Collection<Metadata>>();
```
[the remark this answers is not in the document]
matthias-ronge: Here `var` abbreviates `MutableTriple<Collection<Metadata>, Reimport, Collection<Metadata>>`, therefore the use seems appropriate to me. @solth is this okay for you?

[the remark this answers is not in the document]
solth: I do think we should be careful and do not become "overzealous" when making use of the `var` type inference, but it would probably be acceptable in this particluar case because the type is right there on the right hand side of the variable assignement. 

On the other hand, this is perhaps a topic that should be discussed more thoroughly and the result of the discussion should be included in the coding guidelines. Some seem to consider it a "code smell" (https://medium.com/@spextreme/java-var-should-be-a-code-smell-17af0be79646) while others think the advantages outweigh the risks (https://www.beyondjava.net/should-you-embrace-var-in-java-10)

So perhaps the best solution would be to stick to explicit variable typing for now and only use `var` once we have come to a general decision concerning its usage.

markusweigelt (changes requested):

markusweigelt (changes requested):
Great work! Thx that you have implemented a reimport option that works even better and is simply more thought out.

Kitodo/rulesets/ruleset.xsd
```diff
@@ -347,6 +363,7 @@
         <xs:attribute type="xs:boolean" name="editable" default="true"/>
         <xs:attribute type="xs:boolean" name="excluded" default="false"/>
         <xs:attribute type="xs:boolean" name="multiline" default="false"/>
+        <xs:attribute type="ruleset:Reimport" name="reimport" default="replace"/>
```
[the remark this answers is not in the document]
matthias-ronge: So far it has been common that we always use the default which reflects the previous functionality. This is, so that after an update, the administrators have to make as few adjustments as possible to keep the system working as before. And the previous behavior was, that the metadata was replaced. And, I'd be `add` for the default behavior, by the way.

Kitodo-DataEditor/src/test/resources/testReimportOfMetadataModes.xml
```diff
@@ -0,0 +1,70 @@
...
+        <key id="metadataToKeepExceptInEditing">
+            <label>Metadata to keep, except in editing</label>
+        </key>
+    </declaration>
+
+    <editing>
+        <setting key="metadataToAdd" reimport="add" />
+        <setting key="metadataToReplaceExplicitly" reimport="replace" />
+        <setting key="metadataToKeep" reimport="keep" />
+
```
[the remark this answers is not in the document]
matthias-ronge: This is intentional. `<key id="metadataToReplaceByDefault">` has no `<setting>`, so the default is used.

IkramMaalej (changes requested):

IkramMaalej (commented):
When a metadata is configured as `reimport="add"` in the settings section, and this metadata is also configured with `maxOccurs="1"` for example in the restrictions section on a structural element, after an additional import, this metadata will be displayed twice and not once, which should be  the correct behavior.

Therefore,  the maximum number of occurrences is unchecked for the case of `reimport="add"`

Kitodo/rulesets/ruleset.xsd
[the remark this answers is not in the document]
IkramMaalej: Can not be the default value somehow configurable and not hard-coded?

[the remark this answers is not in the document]
BartChris: one question would be, wether this setting should be configurable
- ruleset-wide?
- project-wide?
- client-wide?
- system-wide?

I would probably also prefer to define it in the ruleset because it is a user decision and not an administartor decision in my opinion.

matthias-ronge:
I've modified the code to take this into account. It was a major change and I've now extracted the function into its own class. I've added cases to the test to test this new behaviour as well.

matthias-ronge:
@solth I tested locally, MetadataST finished without errors. I cannot say why it failed here.

solth:
> @solth I tested locally, MetadataST finished without errors. I cannot say why it failed here.

6th build was successful 😅

IkramMaalej (approved):

solth:
@BartChris do you have any further change requests or would you approve this pull request?

BartChris:
@solth I just had a note on the configurability of the reimport of metadata, but this is beyond the scope of this pull request. So i am fine with the pull request as it is right now.

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/ReimportMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Determines the result of re-importing metadata for a single metadata key.
+ */
+class ReimportMetadata implements Supplier<Collection<Metadata>> {
+
+    // Metadata key for which the result of the reimport is determined.
+    private final String key;
+
+    // existing metadata for this key before reimport
+    private Collection<Metadata> currentEntries;
```
solth, line 32:
```suggestion
    private final Collection<Metadata> currentEntries;
```

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/ReimportMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+
+    // configured reimport behavior
+    private Reimport reimport;
+
+    // metadata fetched on reimport
+    private Collection<Metadata> updateEntries;
```
solth, line 38:
```suggestion
    private final Collection<Metadata> updateEntries;
```

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
```diff
@@ -48,6 +51,11 @@ public class RulesetManagement implements RulesetManagementInterface {
      */
     private static final Logger logger = LogManager.getLogger(RulesetManagement.class);
 
+    /**
+     * Language used internally in places where language does not matter.
```
solth, line 55: Could you clarify this comment? "...in places where language does not matter" - so where _does_ it matter?

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/xml/Reimport.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.xml.bind.annotation.XmlEnum;
+import javax.xml.bind.annotation.XmlEnumValue;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+/**
+ * This class is a backing bean for the XML attribute reimport in the ruleset.
+ * With it, JAXB can map the attribute to an enum.
+ */
+@XmlEnum(String.class)
```
solth, line 23:
```suggestion
@XmlEnum()
```
Redundant default parameter assignenment

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
```diff
@@ -130,27 +141,13 @@ public ProcessFieldedMetadata(Division<?> structure, StructuralElementViewInterf
...
-            }
-        });
-        metadata.addAll(metadataToAdd);
-
-        TreeNode editedTreeNode = treeNode;
+    public int addMetadataIfNotExists(Collection<Metadata> potentialMetadataItems)
+            throws InvalidMetadataValueException, NoSuchMetadataFieldException {
 
+        preserve();
+        int count = rulesetService.updateMetadata(division.getType() ,metadata, "create", potentialMetadataItems);
```
solth, line 148:
```suggestion
        int count = rulesetService.updateMetadata(division.getType(), metadata, "create", potentialMetadataItems);
```

solth (commented):
@matthias-ronge thanks a lot for this valuable contribution. I have no objections, just some tiny suggestions to optimize code style in a few places.

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/RulesetManagement.java
matthias-ronge: comment extended

solth (approved):

andre-hohmann:
@matthias-ronge : Thanks a lot for the additional functionality. We talked about that in the Community Board and want to ask, if you could describe the settings `reimport="replace"`, `reimport="add"` and `reimport="keep"` in [Regelsatz 3.x](https://github.com/kitodo/kitodo-production/wiki/Regelsatz-3.x), as for example [Schlüssel mit besonderer Verwendung](https://github.com/kitodo/kitodo-production/wiki/Regelsatz-3.x#schl%C3%BCssel-mit-besonderer-verwendung). 

We asked ourselves, if a delete-function is realized by `reimport="replace"`: 
* If metadata in the source is deleted and an additional import executed, the metadata in Kitodo.Production is deleted (or replaced by "null").

Is this assumption correct?

PR #5614, opened 2023-03-31
Prevent wrong error message when deleting process

### LINKED ISSUE #5120: Unclear error message after deletion of a process
### Problem
After a process is deleted, an error message occurs: 
- `Fehler beim Hochladen von Home.`

It is not clear, why this message appears and what should be done, but it can be stated that: 
1. the process is **not available** anymore in the process list
2. the process is **not available** anymore in the index
3. ~~the folder with the meta.xml, ... is **still available** in the file system~~ 
Update 2022-05-03: This is due to the local configuration in the SLUB and must be regarded. 

### Solution 
When the error message is necessary, it should be clearly described, why it appears and what needs to be done by the user. 
When the error message is not necessary, it should not appear. 

### Example
FehlerLöschenVorgang

solth:
Fixes #5120 by checking whether a task is currently assigned to a user before passing the user object to the `uploadFromHome` method (which resulted in an NPE and produced the documented error message when no user was assigned)


matthias-ronge (approved):
Looks correct as per coding guidelines. Not tested

PR #5615, opened 2023-03-31
Display options in language switcher translated to each corresponding…

### LINKED ISSUE #5608: Language switcher not self-explanatory
The language switcher is displayed in the currently set language. To better depict the problem, lets assume we had more translations. Our language switcher would look like this:

ドイツ語 | 英語 | スペイン語 | **日本**

Correctly it should look like this:

Deutsch | english | Español | **日本**

This affects both the language switcher on the login page and the language switcher in the user menu (that is, it affects the properties files): The language names should be written in the corresponding language in all language files.

solth:
… language

Fixes #5608


matthias-ronge (approved):
Looks correct as per coding guidelines. Not tested

PR #5617, opened 2023-04-05
Fix bugs caused by file names with dots

### LINKED ISSUE #5278: File names with dots are not working
**Describe the bug**
If a file name contains a dot Kitodo cannot differentiate the names, does not display all the images in the metadata editor and throws an error on saving. 

"Ambiguous canonical file name part in the same physical division"

**To Reproduce**
1. Name your files test.a-1.tif and test.a-2.tif
2. Try to open a process with the images in the medatada editor.

**Expected behavior**
Kitodo should support dots in a file name or unallowed charcters should be documented as other characters have problems as well:
https://github.com/kitodo/kitodo-production/issues/5044 

**Release**
3.4.4 SNAPSHOT

IkramMaalej:
Fixes #5278

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -524,11 +524,7 @@ public class FileService {
      * @return the name of the file
      */
     public String getFileName(URI uri) {
-        String fileNameWithExtension = fileManagementModule.getFileNameWithExtension(uri);
-        if (fileNameWithExtension.contains(".")) {
-            return fileNameWithExtension.substring(0, fileNameWithExtension.lastIndexOf('.'));
-        }
-        return fileNameWithExtension;
+        return FilenameUtils.getBaseName(fileManagementModule.getFileNameWithExtension(uri));
     }
 
     /**
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -526,7 +526,7 @@ public class FileService {
     public String getFileName(URI uri) {
         String fileNameWithExtension = fileManagementModule.getFileNameWithExtension(uri);
         if (fileNameWithExtension.contains(".")) {
-            return fileNameWithExtension.substring(0, fileNameWithExtension.indexOf('.'));
+            return fileNameWithExtension.substring(0, fileNameWithExtension.lastIndexOf('.'));
         }
         return fileNameWithExtension;
     }
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/MetadataImageComparator.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/MetadataImageComparator.java
@@ -34,13 +34,6 @@ public class MetadataImageComparator implements Comparator<Object> {
         if (firstObject instanceof String && secondObject instanceof String) {
             firstName = (String) firstObject;
             secondName = (String) secondObject;
-
-            int firstNameLastPeriodIndex = firstName.lastIndexOf('.');
-            int secondNameLastPeriod = secondName.lastIndexOf('.');
-            if (firstNameLastPeriodIndex > -1 && secondNameLastPeriod > -1) {
-                firstName = firstName.substring(0, firstNameLastPeriodIndex);
-                secondName = secondName.substring(0, secondName.lastIndexOf('.'));
-            }
         } else if (firstObject instanceof URI && secondObject instanceof URI) {
             URI firstUri = (URI) firstObject;
             URI secondUri = (URI) secondObject;
--- a/Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/file/FileServiceIT.java
@@ -11,9 +11,11 @@
 
 package org.kitodo.production.services.file;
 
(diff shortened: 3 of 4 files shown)


Kitodo/src/main/java/org/kitodo/production/services/file/MetadataImageComparator.java
```diff
@@ -34,13 +34,6 @@ public int compare(Object firstObject, Object secondObject) {
         if (firstObject instanceof String && secondObject instanceof String) {
             firstName = (String) firstObject;
             secondName = (String) secondObject;
-
-            int firstNameLastPeriodIndex = firstName.lastIndexOf('.');
-            int secondNameLastPeriod = secondName.lastIndexOf('.');
-            if (firstNameLastPeriodIndex > -1 && secondNameLastPeriod > -1) {
-                firstName = firstName.substring(0, firstNameLastPeriodIndex);
-                secondName = secondName.substring(0, secondName.lastIndexOf('.'));
-            }
```
solth, line 43: I am not sure if this part should really be removed. Apparently file extensions are not to be considered when comparing images (maybe the `compare` method is meant to be invariant of invidual media variants/represenstations of one image/scan?)

@matthias-ronge even though this part of the code is quite old, do you remember why you stripped the file extension from the filenames before passing the parameters to the `compareImages` method?

matthias-ronge: This originates from 2.x, though [there](/kitodo/kitodo-production/blob/2.x/Goobi/src/de/sub/goobi/metadaten/MetadatenImagesHelper.java#L445) the code did read quite different. In case that the config parameter `image.sorting` is set to `number`, the images are compared by their integer value (`Integer.valueOf()`) of their file name, so the ending must be cut off to allow conversion to `int`.

henning-gerhardt (changes requested):
First: I can not reproduce the original error. In the issue was a possible solution described by Matthias but without an answer if this solution was helpful or not. Maybe BartChris should even test and review this changes?

Second: Adding intensive tests for the class MetadataImageComparator would be helpful to decide if the removed parts are really needed or not. The tests should include string and URI objects and the different kind of 'number' and 'alphanumeric' sorting like written down in the section of the `kitodo_config.properties` file.

BartChris:
I will have a look if i can still reproduce the error and wether this fixes it.

Kitodo/src/main/java/org/kitodo/production/services/file/MetadataImageComparator.java
BartChris: It seems to me that the stripping of the file extension is already happening earlier, when searchForMedia is called so that the removal of those lines might not have an effect in the code section discussed here.

https://github.com/kitodo/kitodo-production/blob/3ca7384e44f8681e9b49de91738ee05e2333584b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java#L1099-1104

BartChris: It seems to me that for the problem described in the linked issue the change in FileService:

```java
public String getFileName(URI uri) {
        String fileNameWithExtension = fileManagementModule.getFileNameWithExtension(uri);
        if (fileNameWithExtension.contains(".")) {
            return fileNameWithExtension.substring(0, fileNameWithExtension.lastIndexOf('.'));
        }
        return fileNameWithExtension;
    }
```
is not even relevant. The change which fixes the problem in the linked issue is the deletion of lines in the `MetadataImageComparator` because the file extension are already stripped of here as it seems to me while the files are collected.

https://github.com/kitodo/kitodo-production/blob/7e2caad5fdbf06929bf53418e86dd80b93c3ed35/Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java#L346-L355

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -526,7 +526,7 @@ public boolean canRead(URI uri) {
     public String getFileName(URI uri) {
         String fileNameWithExtension = fileManagementModule.getFileNameWithExtension(uri);
         if (fileNameWithExtension.contains(".")) {
-            return fileNameWithExtension.substring(0, fileNameWithExtension.indexOf('.'));
+            return fileNameWithExtension.substring(0, fileNameWithExtension.lastIndexOf('.'));
```
BartChris, line 529: It seems to me, that the intent of the getFileName method (before and after the change) is to get the name of the file without the extension. Could we not just rely on the `getBaseName`-method of the Commons IO-library here?

https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#getBaseName(java.lang.String)

Kitodo/src/main/java/org/kitodo/production/services/file/MetadataImageComparator.java
BartChris: In order to prevent the corruption of file name with dots, the convention should probably be the following: if a String is passed to the comperator, this String is always a filename without  an extension in order to not having to implement the removal of extensions in the Comparator which has no knowledge what kind of String is given to it.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
IkramMaalej: @BartChris : thanks for review!
Yes we can refactor the `getFileName` method and use `getBaseName` method of the Commons IO-library.

PR #5618, opened 2023-04-05
Fix number of scans not updating after uploading or deleting media in…

IkramMaalej:
… metadata editor


markusweigelt (approved):
Works for me. 

However, I think it's good to add a singular here. "1 Medien" does not sound good and shows that the software is not complete.

PR #5619, opened 2023-04-06
Remove unused javascript file

solth:
Remove unused javascript file `metadata.js`.

### DIFF
--- a/Kitodo/src/main/resources/messages/errors_de.properties
+++ b/Kitodo/src/main/resources/messages/errors_de.properties
@@ -12,7 +12,6 @@
 error=Fehler...
 
 # A
-anErrorOccurred=Ein Fehler ist aufgetreten\:
 errorAdding=Fehler beim Hinzuf\u00FCgen von ''{0}''
 authorityAssignedError=Berechtigung kann nicht gel\u00F6scht werden, da dieser noch Rollen zugewiesen sind.
 
--- a/Kitodo/src/main/resources/messages/errors_en.properties
+++ b/Kitodo/src/main/resources/messages/errors_en.properties
@@ -12,7 +12,6 @@
 error=Error...
 
 # A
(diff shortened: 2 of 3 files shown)


Kitodo/src/main/resources/messages/errors_de.properties
```diff
@@ -12,6 +12,7 @@
 error=Fehler...
 
 # A
+anErrorOccurred=Ein Fehler ist aufgetreten\:
```
henning-gerhardt, line 15: I did not found any usage of this key.

Kitodo/src/main/resources/messages/errors_en.properties
```diff
@@ -12,6 +12,7 @@
 error=Error...
 
 # A
+anErrorOccurred=An error occurred\:
```
henning-gerhardt, line 15: see above

Kitodo/src/main/resources/messages/errors_es.properties
```diff
@@ -12,6 +12,7 @@
 error=error...
 
 # A
+anErrorOccurred=Se ha producido un error\:
```
henning-gerhardt, line 15: see above

henning-gerhardt (changes requested):
Looks good but why a new error key was added which is not used is not clear for me.

solth:
@henning-gerhardt thanks for the review. You are right, this message key is not used currently. Actually, I did not _add_ it, though, but just renamed an existing message key to a more appropriate name (from `getNormDataRecordFailed` to `anErrorOccurred`). 

I did not remove it since I was going to use it in another PR, but I will add the key again (with the new name) in that new PR. For now, I have remove the message key as you suggested.

henning-gerhardt (approved):

PR #5620, opened 2023-04-06
Rename directories when changing process title

### LINKED ISSUE #5597: Aligning file directory names after changing a process name in the context of a projectconfiguration using the variable "processtitle" for subdirectories
**Describe the bug**
Changing a process name has no effect on the directory name for the images in a projectconfiguration with "processtitle" as part of the directory name - look at the screenshot of the configuration in Hamburg. 

**To Reproduce**
1. Changing a process name
2. Checking subdirectories in /images and stating unchanged names

**Expected behavior**
We miss the behaviour of K2 aligning the directory names according to the process name.

**Screenshots**
grafik

**Release**
Version 3.5

solth:
Fixes #5597

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
@@ -94,7 +94,7 @@ public interface FileManagementInterface {
      *            the URI to the resource to rename
      * @param newName
      *            the new name of the resource
-     * @return true, if successful, false otherwise
+     * @return URI of renamed resource
      */
     URI rename(URI uri, String newName) throws IOException;
 
--- a/Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
@@ -94,7 +94,7 @@ public interface FileManagementInterface {
      *            the URI to the resource to rename
      * @param newName
      *            the new name of the resource
-     * @return URI of renamed resource
+     * @return URI of renamed resource, or null if either 'uri' or 'newName' is null
      */
     URI rename(URI uri, String newName) throws IOException;
 
--- a/Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
@@ -94,7 +94,7 @@ public interface FileManagementInterface {
      *            the URI to the resource to rename
      * @param newName
      *            the new name of the resource
-     * @return URI of renamed resource, or null if either 'uri' or 'newName' is null
+     * @return URI of renamed resource
      */
     URI rename(URI uri, String newName) throws IOException;
 
--- a/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
+++ b/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
@@ -157,16 +157,11 @@ public class FileManagement implements FileManagementInterface {
             return null;
         }
 
-        String uriString = uri.toString();
         if (isDirectory(uri)) {
-            if (uriString.equals(newName + "/") || uriString.equals(newName)) {
-                return null;
-            } else {
-                return renameDirectory(uri, newName);
-            }
+            return renameDirectory(uri, newName);
         }
 
-        String substring = uriString.substring(0, uriString.lastIndexOf('/') + 1);
+        String substring = uri.toString().substring(0, uri.toString().lastIndexOf('/') + 1);
         if (newName.contains("/")) {
             newName = newName.substring(newName.lastIndexOf('/') + 1);
         }
@@ -189,6 +184,10 @@ public class FileManagement implements FileManagementInterface {
     }
 
     private URI renameDirectory(URI directoryPath, String newDirectoryName) throws IOException {
+        String uriString = directoryPath.toString();
+        if (uriString.equals(newDirectoryName + "/") || uriString.equals(newDirectoryName)) {
+            return null;
+        }
         File newDirectory = new File(KitodoConfig.getKitodoDataDirectory() + newDirectoryName);
         File oldDirectory = new File(KitodoConfig.getKitodoDataDirectory() + directoryPath);
         boolean success = oldDirectory.renameTo(newDirectory);
--- a/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
+++ b/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
@@ -157,6 +157,10 @@ public class FileManagement implements FileManagementInterface {
             return null;
         }
 
+        if (isDirectory(uri)) {
+            return renameDirectory(uri, newName);
+        }
+
         String substring = uri.toString().substring(0, uri.toString().lastIndexOf('/') + 1);
         if (newName.contains("/")) {
             newName = newName.substring(newName.lastIndexOf('/') + 1);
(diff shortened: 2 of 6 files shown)


henning-gerhardt (approved):
Changes look good and did work for entries with and without a `processtitle` inside. 

Maybe adding a check to compare old and new directory name and do only the renaming if they differ.

solth:
> Maybe adding a check to compare old and new directory name and do only the renaming if they differ.

Thanks for the review. There already is a similar check in the `ProcessForm` when saving a process that compares old and new process title and only triggers renaming actions if the titles differ: https://github.com/kitodo/kitodo-production/blob/766c5eb043fbe804f3bf2d6c1c7ee1869978c6f7/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java#L181

Ideally this would be checked inside the renaming function itself to cover all use cases but thanks to the check above the renaming function is already skipped when manually saving a single process without changing the title. Should I change this PR to adjust the code accordingly or leave that change to a separate PR?

henning-gerhardt:
As the renaming functionality of the FileManagement class from the Kitodo-Filemanagement module for files or directories could be called from everywhere I would add even there a check and doing the renaming of files or directories only if they differ. I think that this additional check should be done in the pull request as the comparing should be simple and easy. If this is more complicated than move it to a separate pull request.

solth:
@henning-gerhardt I added a check to skip renaming a directory if the new name equals the old name and a corresponding test. Since the behaviour of renaming _files_ is a little different and does not relate to the changes in this pull request I have not added a similar check, there.

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -157,6 +157,14 @@ public URI rename(URI uri, String newName) throws IOException {
             return null;
         }
 
+        if (isDirectory(uri)) {
+            if (uri.toString().equals(newName + "/")) {
```
henning-gerhardt, line 161: This did not work on my local system, as `newName` already ends with a `/` character and adding an additional `/` character let the comparing every time be false. Maybe reverting this last change is better as it could be difficult for now to find right solution.

solth: I simply extended this check to cover uri with and without trailing slash. Would that work for you?

henning-gerhardt: I will test this. Maybe moving the comparing of old and new directory inside the called method as there are the resulting paths created and must not recreated in on other way?

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -157,7 +157,16 @@ public URI rename(URI uri, String newName) throws IOException {
             return null;
         }
 
-        String substring = uri.toString().substring(0, uri.toString().lastIndexOf('/') + 1);
+        String uriString = uri.toString();
+        if (isDirectory(uri)) {
+            if (uriString.equals(newName + "/") || uriString.equals(newName)) {
+                return null;
```
matthias-ronge, line 163: Javadoc of `URI rename(URI uri, String newName)` in _org.kitodo.api.filemanagement.FileManagementInterface_ states:

    * @return true, if successful, false otherwise

Please update Javadoc in `FileManagementInterface`.

matthias-ronge (changes requested):

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
solth: Done

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
solth: Done

Kitodo-API/src/main/java/org/kitodo/api/filemanagement/FileManagementInterface.java
```diff
@@ -94,7 +94,7 @@ public interface FileManagementInterface {
      *            the URI to the resource to rename
      * @param newName
      *            the new name of the resource
-     * @return true, if successful, false otherwise
+     * @return URI of renamed resource
```
matthias-ronge, line 97: should state that the result may be `null`, and when

matthias-ronge (changes requested):

Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/FileManagement.java
```diff
@@ -179,6 +183,21 @@ public URI rename(URI uri, String newName) throws IOException {
...
+        }
+        File newDirectory = new File(KitodoConfig.getKitodoDataDirectory() + newDirectoryName);
+        File oldDirectory = new File(KitodoConfig.getKitodoDataDirectory() + directoryPath);
+        boolean success = oldDirectory.renameTo(newDirectory);
+        if (success) {
+            return newDirectory.toURI();
+        } else {
+            throw new IOException("Unable to rename directory '" + directoryPath + "'!");
+        }
+    }
```
henning-gerhardt, line 199: Maybe in this way:

```java
    private URI renameDirectory(URI directoryPath, String newDirectoryName) throws IOException {
        File newDirectory = new File(KitodoConfig.getKitodoDataDirectory() + newDirectoryName);
        File oldDirectory = new File(KitodoConfig.getKitodoDataDirectory() + directoryPath);

        if (! oldDirectory.equals(newDirectory)) {
            boolean success = oldDirectory.renameTo(newDirectory);
            if (success) {
                return newDirectory.toURI();
            } else {
                throw new IOException("Unable to rename directory '" + directoryPath + "'!");
            }
        } else {
            return newDirectory.toURI();
        }
    }
```

Returning `null` should be avoided and the changed JavaDoc comment could be changed to `URI of renamed resource`

henning-gerhardt, line 199: Maybe in this way:

```java
    private URI renameDirectory(URI directoryPath, String newDirectoryName) throws IOException {
        File newDirectory = new File(KitodoConfig.getKitodoDataDirectory() + newDirectoryName);
        File oldDirectory = new File(KitodoConfig.getKitodoDataDirectory() + directoryPath);

        if (! oldDirectory.equals(newDirectory)) {
            boolean success = oldDirectory.renameTo(newDirectory);
            if (success) {
                return newDirectory.toURI();
            } else {
                throw new IOException("Unable to rename directory '" + directoryPath + "'!");
            }
        } else {
            return newDirectory.toURI();
        }
    }
```

Returning `null` should be avoided and the changed JavaDoc comment could be changed to `URI of renamed resource`

solth: Changed

Kitodo-FileManagement/src/test/java/org/kitodo/filemanagement/FileManagementTest.java
```diff
@@ -150,6 +150,20 @@ public void shouldRenameFile() throws Exception {
...
+    @Test
+    public void shouldSkipRenamingDirectory() throws Exception {
+        String directoryName = "testDir";
+        URI resource = fileManagement.create(URI.create(""), directoryName, false);
+        assumeTrue(fileManagement.isDirectory(resource));
+        URI expectedUri = new File(KitodoConfig.getKitodoDataDirectory() + resource).toURI();
+        assertEquals("Renaming directory to the identical name should return null", expectedUri,
+                fileManagement.rename(resource, directoryName));
+        String directoryWithTrailingSlash = directoryName + "/";
+        assertEquals("Renaming directory to the identical name with trailing slash should return null", expectedUri,
```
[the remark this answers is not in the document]
solth: Done

Kitodo-FileManagement/src/test/java/org/kitodo/filemanagement/FileManagementTest.java
```diff
@@ -150,6 +150,20 @@ public void shouldRenameFile() throws Exception {
         assertTrue(fileManagement.fileExist(newUri));
     }
 
+    @Test
+    public void shouldSkipRenamingDirectory() throws Exception {
+        String directoryName = "testDir";
+        URI resource = fileManagement.create(URI.create(""), directoryName, false);
+        assumeTrue(fileManagement.isDirectory(resource));
+        URI expectedUri = new File(KitodoConfig.getKitodoDataDirectory() + resource).toURI();
+        assertEquals("Renaming directory to the identical name should return null", expectedUri,
```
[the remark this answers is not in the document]
solth: Done

henning-gerhardt (approved):

PR #5623, opened 2023-04-13
Implement <key use="processTitle">

### LINKED ISSUE #4459: <key use="processTitle"> not implemented
A `<key>` declared as `use="processTitle"` in the ruleset, should be set to the process title on process creation. However, this seems not to be implemented, except for newspaper processes. (No reference to `RulesetManagementInterface.getFunctionalKeys(FunctionalMetadata.PROCESS_TITLE)` found in code.)

**Goal:** Implement writing of process title to declared keys.

matthias-ronge:
Resolves #4459


IkramMaalej (commented):
Thanks for this PR!
Here are some comments:

1. I don't think it's a good idea that the values of the metadata configured with `<key use="processTitle"/>`  are adjusted in the backend without informing the user => Better way: the fields of these metadata are disabled and synchronised  with the process title field (every time , if the process title is edited these fields will also be updated).
2. If these fields are already filled in (with imported data or manually) this function will not be applied, then this behavior should also be documented.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -632,13 +635,41 @@ private void saveProcessHierarchyMetadata() {
...
     }
 
+    private void setProcessTitleMetadata(Workpiece workpiece) {
+        Collection<String> keysForProcessTitle = rulesetManagement.getFunctionalKeys(FunctionalMetadata.PROCESS_TITLE);
+        if (!keysForProcessTitle.isEmpty()) {
+            String processTitle = currentProcess.getProcess().getTitle();
+            addAddableMetadataRecursive(workpiece.getLogicalStructure(), keysForProcessTitle, processTitle);
+            addAddableMetadataRecursive(workpiece.getPhysicalStructure(), keysForProcessTitle, processTitle);
+        }
+    }
```
IkramMaalej, line 656: The `addAddableMetadataRecursive` Method is not the appropriate  method to update the configured Metadata with `use="processTitle"` because these metadata can also be required metadata wich are also  configured with `maxOccurs="1"` and `minOccurs="1"` and have to occur only one time. Therefore,  they are  not inclued in the addable metadata list.

matthias-ronge: Good catch! I changed that.

matthias-ronge:
Your remarks are all valuable.

Issue #4459 is _only about_ making the behaviour that already works for Newspaper processes available for other processes as well. It may be necessary to _somehow_ get the process title into the meta.xml, so that it can be used there by scripts in the workflow and by the XSLT when exporting, for example to determine file system paths that contain the process title, for other processes as well, if the existing Newspaper workflow should be reused for other non-newspaper media.

The other functions are all worth considering, but then when they are needed, they should be requested as an issue, and there must be funding for their development. All of the other functional considerations would also affect Newspaper processes, where these functions have not yet existed either.

1. That doesn’t yet happen in newspaper processes either.
2. If the field is already filled, the function is not required. (I consider such a case to be a non-matching ruleset—vs—import mapping configuration.)
3. That's true, but that hasn't happened in the newspaper processes either, and apparently nobody has been bothered by it yet.
4. You can specify that in the ruleset if you want. (`<setting editable="false"/>` or `<setting excluded="true"/>`)

IkramMaalej (approved):

PR #5628, opened 2023-04-17
Inactive templates

### LINKED ISSUE #3311: Remove switch to create unreachable production templates

solth:
Fixes #3311


BartChris:
Activating and deactivating the templates works good! But after this change it is possible to not only show a deactivated template, but also to create processes from it. Should that be possible?

solth:
> Activating and deactivating the display of inactive templates works good! But after this change it is possible to not only show a deactivated template, but also to create processes from it. Should that be possible?
> An indication in the UI, if a template is active or inactive (we have something like that for projects) would also be good.

Thanks alot for the review. You are absolutely correct that deactivated templates should not be available offered for process creation (they are already skipped when creating a process for a specific project via the process list). I will make a corresponding change and then re-request a review from you.

PR #5631, opened 2023-04-19
Remove unneccessary null-checks

matthias-ronge:
It seems that 'settings' can never be null, as other parts of the code rely on it not being null, without checking for it. The commit doesn’t touch `public boolean isRequired()` as it is already handled in PR #5609.


IkramMaalej:
I think that this check is still needed because a NPE still occurs when trying to import some data using a mapping file that creates metadata groups and metadata entries within the group that are both unknown to the ruleset . See #5603 https://github.com/kitodo/kitodo-production/issues/5217#issuecomment-1174818329

IkramMaalej (commented):
I think that this check is still needed because a NPE still occurs when trying to import some data using a mapping file that creates metadata groups and metadata entries within the group that are both unknown to the ruleset . See #5603 https://github.com/kitodo/kitodo-production/issues/5217#issuecomment-1174818329

PR #5637, opened 2023-04-26
Display progress when creating newspaper processes

### LINKED ISSUE #3676: No progress in creating newspaper processes
If newspaper processes are created, no progress is made in the Task Manager. It is continuously at 0%. However, the process runs in the background. Updating the page (<kbd>F5</kbd>) doesn't make anything better.

**Goal:** The progress (percentage) increases.

matthias-ronge:
Fixes #3676


solth (approved):
Works like a charm!

Issue #3722, opened 2020-06-04
Calendar cannot be saved


matthias-ronge:
In version 2, the calendar can be downloaded and uploaded again in another session. You don't have to do all the editing at once. In addition, the (scripted) creation of the XML file offers further options for importing calendar data. The function no longer exists.

![Screenshot from version 2](https://user-images.githubusercontent.com/3040657/83736726-56157f80-a652-11ea-8059-2aa5392270a6.jpg)

**Goal:** It is cumbersome to download the file locally and upload it again. But saving should be possible.

henning-gerhardt:
Please do not store this calendar data on server this. This would need a lot more logic behind (serving data, update data, remove data, user access, permissions to this data, permissions for users to interact with this data, ...)

matthias-ronge:
My idea is to save a `publication-history.xml` in the newspaper's process directory. The calendar editor would then have the same buttons as the metadata editor: cancel, save, save and close. There is exactly one publication history per newspaper. Perhaps you can then later use the `publication-history.xml` to insert editions into the newspaper afterwards, which is still an unimplemented feature, but a desiration.

henning-gerhardt:
Please keep in your mind that 
- not every "newspaper" is a newspaper
- not every newspaper has a regular publication course
- not every newspaper starts and ends in a year, like season publications from August to July next year and then from September to July the next year after (08/1900 - 07/1901 and 09/1901 - 07/1902)
- and maybe many more things which I forget now

matthias-ronge:
That is all correct, but has nothing to do with the problem here. The issue is that you cannot interrupt the editing of the appearance history and continue later.

matthias-ronge:
Completed.

PR #5639, opened 2023-04-26
Bring back calendar download as XML

matthias-ronge:
One part of #3722

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Course.java
+++ b/Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Course.java
@@ -34,6 +34,7 @@ import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.logging.log4j.util.Strings;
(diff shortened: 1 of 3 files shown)


oliver-stoehr (approved):

Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Course.java
```diff
@@ -34,6 +34,7 @@
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.logging.log4j.util.Strings;
```
henning-gerhardt, line 37: Why are you using the `isBlank` method from a Logging framework instead of the Apache Commons Language framework like in all the other places?

Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/CourseToGerman.java
```diff
@@ -20,6 +20,7 @@
 import java.util.Set;
 import java.util.TreeSet;
 
+import org.apache.logging.log4j.util.Strings;
```
henning-gerhardt, line 23: See above

PR #5640, opened 2023-04-26
Bring back calendar import as XML

matthias-ronge:
The other part of #3722


IkramMaalej:
Can you please upload an example  xml-file so I can use it to review this PR.

oliver-stoehr (approved):
Works fine with files generated by download button in #5639

PR #5644, opened 2023-04-28
Return import configurations sorted by title

solth:
Currently, the entries in the pulldown menu for default import configurations in the project edit page are unsorted:

Bildschirmfoto 2023-05-02 um 10 34 49

This PR adds a small change to return import configurations sorted alphabetically from service class in order to make user experience less confusing.


henning-gerhardt (approved):
Changes are okay and include a test. Works for me on correct page.

PR #5651, opened 2023-05-04
Update ElasticSearch client to fix vulnerability

solth:
Update ElasticSearch client library to prevent potential arbitrary code execution exploit.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
         <saxon.version>9.9.1-8</saxon.version>
         <log4j.version>2.19.0</log4j.version>
         <junit.version>5.9.2</junit.version>
-        <elasticsearch.version>7.17.7</elasticsearch.version>
+        <elasticsearch.version>7.17.8</elasticsearch.version>
         <maven-failsafe-plugin.version>2.22.2</maven-failsafe-plugin.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -65,7 +65,7 @@
         <saxon.version>9.9.1-8</saxon.version>
         <log4j.version>2.19.0</log4j.version>
         <junit.version>5.9.2</junit.version>
-        <elasticsearch.version>7.17.7</elasticsearch.version>
+        <elasticsearch.version>7.17.8</elasticsearch.version>
```
henning-gerhardt, line 68: Why we are not switching to latest release aka 7.17.10 (at 05/05/2023) of elastic search?

solth: I just used and tested the first newer version that was marked as fixing the mentioned vulnerability to minimize the potential for incompatibilities, but I guess if ElasticSearch adheres to the semantic versioning guidelines, 7.17.10 should work, too.

solth: Just tested it and version 7.17.10 does indeed work as well without problems, so I changed the version in the pull request as you suggested.

henning-gerhardt (approved):
I tested the new client version with an old (7.17.7) and new (7.17.10) ES server and I got no issues: searching, updating, indexing. But I tested this only with a few data and not with our complete data set.

PR #5657, opened 2023-05-08
Task action processor

markusweigelt:
This pr add the new TaskActionProcessor and following TaskActions to process.

- `COMMENT` action adds a comment to the task
- `ERROR_OPEN` action lock the task and add an error comment when task status is OPEN or INWORK
- `ERROR_CLOSE` action set the task status of logged task to OPEN
- `PROCESS` action set task status of open task to INWORK
- `CLOSE` action close a task

Please consider the updated documentation https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ#processors and for simplified use of ActiveMQ testing, the Docker image and the Java client https://github.com/slub/kitodo-production-activemq.


Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+ * This is a web service interface to modify task status.
+ */
+public class TaskActionProcessor extends ActiveMQProcessor {
+
+    public static final String KEY_CORRECTION_TASK_ID = "correctionTaskId";
+    public static final String KEY_MESSAGE = "message";
+    public static final String KEY_TASK_ACTION = "action";
+    public static final String KEY_TASK_ID = "id";
+    private final TaskService taskService = ServiceManager.getTaskService();
+    private WorkflowControllerService workflowControllerService;
```
solth, line 48:
```suggestion
    private final WorkflowControllerService workflowControllerService;
```

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+        try {
+            taskAction = TaskAction.valueOf(mapMessageObjectReader.getMandatoryString(KEY_TASK_ACTION));
+        } catch (IllegalArgumentException e) {
+            throw new ProcessorException("Unknown task action " + taskActionField);
+        }
+
+        try {
+            Task currentTask = taskService.getById(taskId);
+            if (Objects.isNull(currentTask)) {
+                throw new ProcessorException("Task with id " + taskId + "not found.");
```
solth, line 81:
```suggestion
                throw new ProcessorException("Task with id " + taskId + " not found.");
```

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+     * this service.
+     */
+    public TaskActionProcessor() {
+        super(ConfigCore.getOptionalString(ParameterCore.ACTIVE_MQ_TASK_ACTION_QUEUE).orElse(null));
+        workflowControllerService = new WorkflowControllerService();
+    }
+
+    /**
+     * This is the main routine processing incoming messages. It gets the task id and the task action for processing.
+     * Every action has its own behavior so please read the comment on the action for more information.
```
solth, line 62: I am not sure if I missed something but where are these comments explaining each action? Do yo mean the exception messages?

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    private void markErrorCommentAsCorrected(Task currentTask, Integer correctionTaskId)
+            throws DAOException, DataException, IOException {
+        List<Comment> comments = ServiceManager.getCommentService().getAllCommentsByTask(currentTask);
+        Optional<Comment> optionalComment;
+        optionalComment = comments.stream().filter(currentTaskComment -> CommentType.ERROR.equals(
+                currentTaskComment.getType()) && !currentTaskComment.isCorrected() && correctionTaskId.equals(
+                currentTaskComment.getCorrectionTask().getId())).findFirst();
+        if (!optionalComment.isEmpty()) {
```
solth, line 207:
```suggestion
        if (optionalComment.isPresent()) {
```

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Comment.java
```diff
@@ -234,12 +235,22 @@ public Task getCorrectionTask() {
...
     public void setCorrectionTask(Task correctionTask) {
         this.correctionTask = correctionTask;
     }
 
+    /**
+     * Check if the comment has a correction task which differs from the current task.
+     *
+     * @return true if the comment has a separate correction task
+     */
+    public boolean hasSeparateCorrectionTask() {
```
solth, line 250: I do not understand the requirement for this method. I thought the `correctionTask` of a comment is _always_ different from it's `currentTask`, and in addition to that _always_ has a lower `ordering` than the `currentTask`. 

@IkramMaalej is that correct?

That's how adding a correction comment via the metadata editor or task details page works at the moment, anyway: only workflow tasks **prior** to the current task are offered as candiates for the "correction task" of a comment. AFAIK that's the whole reasoning behind correction comments and workflows and I wouldn't want to change this rule and logic just to enable updating process states via ActiveMQ. Instead, a `ProcessorException` should be thrown when trying to create a correction comment with a correction task that equals the current task.

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
```diff
@@ -349,39 +349,51 @@ public void unassignTaskFromUser(Task task) throws DataException {
...
         taskService.save(currentTask);
 
-        Task correctionTask = comment.getCorrectionTask();
-        correctionTask.setProcessingStatus(TaskStatus.OPEN);
-        correctionTask.setProcessingEnd(null);
-        correctionTask.setCorrection(true);
-        taskService.save(correctionTask);
-
-        lockTasksBetweenCurrentAndCorrectionTask(currentTask, correctionTask);
+        if (comment.hasSeparateCorrectionTask()) {
```
solth, line 365: Unsure about this part, see comment above at `Comment.java`.

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
```diff
@@ -349,39 +349,51 @@ public void unassignTaskFromUser(Task task) throws DataException {
...
      *
      * @param comment
-     *              as Comment object
+     *         as Comment object
      */
-    public void solveProblem(Comment comment) throws DataException, DAOException, IOException {
-        closeTaskByUser(comment.getCorrectionTask());
+    public void solveProblem(Comment comment, TaskEditType taskEditType)
+            throws DataException, DAOException, IOException {
+        if (comment.hasSeparateCorrectionTask()) {
```
solth, line 384: See above. Not sure if "correction comments" make sense if their `currentTask` and `correctionTask` are the same.

solth (commented):
Thanks for this pull request to provide this valuable feature. Thumbs up for the multiple tests, too.

I have not tested the functionality yet but have just checked the code so far. Looks good, but I am unsure about the `Comment.hasSeparateCorrectionTask` method (see comment below). Perhaps you could provide some further explanation for why this is required.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
markusweigelt: No i mean the JavaDoc of the the task action enum.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Comment.java
markusweigelt: I added a new behavior to handle errors without specific correction task.

The behavior with the selection of the correction task in creating comment form remains. For external systems that report an error via the new queue, however, it is up to them whether they specify a correction task identifier or not. If none is set then the current task is used as correction task.

fyi @IkramMaalej

IkramMaalej: The concept behind a  correction comment that an error schould be found in a specific workflow step (current task) in a previous workflow step (correction task) and a correction workflow will be started.

I would suggest to implement the need to report an error from external systems  otherwise. 
When `correctionTaskId` is set, then it is simple. A correction comment will be created and a correction workflow will be started. 
If no `correctionTaskId` is set, a simple comment will be created with an additional label (for example `Error_OPEN_EXTERNAL_SYSTEM` or `Error_CLOSE_EXTERNAL_SYSTEM`...) to the comment's message and the status of the current task will be changed (for example to `locked` in this case :  https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ#error_open) . So, we can keep the same behaviour of the correction workflows also for external systems.

markusweigelt: > If no correctionTaskId is set, a simple comment will be created with an additional label (for example Error_OPEN_EXTERNAL_SYSTEM or Error_CLOSE_EXTERNAL_SYSTEM...) to the comment's message and the status of the current task will be changed (for example to locked in this case : https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ#error_open) . So, we can keep the same behaviour of the correction workflows also for external systems.

For that i added a new task edit type QUEUE so that you can see that the change of task take place from the queue. In addition, the specification of a message is reporting for the reporting of an error over the new processor. So i think these labels are not transparent or user-friendly. 

To prevent npe and fill the empty field in database i added the new behaviour with `correctionTaskId` same like `currentTaskId`. I can change the behaviour to the result that correction task id is not set. If i get an approve for that on this PR i will change that. I don't think the implementation with the labels makes sense so i so I will not adjust this for now.

Correct me if I have misunderstood something :)

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/CommentDAO.java
```diff
@@ -54,6 +55,18 @@ public List<Comment> getAllByProcess(Process process) {
...
 
+    /**
+     * Get all comments by task ordered by id ascending.
+     *
+     * @param task
+     *         The current task to get the comments for
+     * @return List of comments
+     */
+    public List<Comment> getAllByTask(Task task) {
+        return getByQuery("FROM Comment WHERE currentTask_id = :taskId ORDER BY id ASC",
```
matthias-ronge, line 66: On Linux, table names are case sensitive and we decided that they should be all lowercase. So I assume, this must be `FROM comment` instead.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+
+        if (Objects.nonNull(comment)) {
+            ServiceManager.getCommentService().saveToDatabase(comment);
+        }
+    }
+
+    private static Comment buildComment(Task currentTask, String message) {
+        Comment comment = new Comment();
+        comment.setProcess(currentTask.getProcess());
+        comment.setAuthor(ServiceManager.getUserService().getCurrentUser());
```
matthias-ronge, line 144: Does this work? Does Active MQ have a “current user”? (Just asking. Not necessary to change if it is okay.)

Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -599,8 +598,6 @@ taskProcessPropertyColumns=
 # Active MQ services need an existing user as whom to act in the system
 activeMQ.user=testAdmin
 
-# You can provide a topic that Kitodo reports results and status messages to
-#activeMQ.results.topic=KitodoProduction.ResultMessages.Topic
```
matthias-ronge, line 603: Please undo remove this

matthias-ronge (changes requested):

Kitodo/src/main/resources/kitodo_config.properties
markusweigelt: Sorry accidentally deleted.

matthias-ronge (approved):

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/CommentDAO.java
markusweigelt: I changed this but we should change also the other tables names, cause table name of every comment query starts with the upper case.

matthias-ronge: Very surprising. Either there is a graceful handling in Hibernate, or nobody ever noticed it doesn’t work?

matthias-ronge: There are even totally wrong table names here. It’s a chance that this is dead code. I create a separate pull request for that.

henning-gerhardt: This query is a HQL (Hibernate Query Language) and not a SQL query and an HQL is not case-sensitive or get formatted into the right and correct database query language. But if you create a raw query (not here but on other places) or writing a flyway migration you are writing an SQL query which is case-sensitive.

henning-gerhardt: > There are even totally wrong table names here. It’s a chance that this is dead code. I create a separate pull request for that.

Do not mix HQL with SQL queries! You can and maybe even must do things in HQL in an other way then in SQL! You can of course write everything in SQL but then you lost the independency from the underlaying database.

matthias-ronge: https://github.com/kitodo/kitodo-production/pull/5692

markusweigelt: So i think it depends on your character configuration of database. If it ends with `_ci` so it will be case insensive. The [installation documentation](https://github.com/kitodo/kitodo-production/wiki/Installationsanleitung-f%C3%BCr-Kitodo.Production-3.4) has not such a specification. IMO the default is mysql specific and default is with `_ci` https://dev.mysql.com/doc/refman/8.0/en/charset-server.html.

henning-gerhardt: @matthias-ronge and @markusweigelt : I can only repeat: this query strings are written in HQL and not SQL, and so the queries itself are not relaying on the real used database and how the database is configured.

matthias-ronge: Then my suggestion for change was wrong. I didn’t knew about HQL at all.

solth:
@henning-gerhardt would you mind reviewing this pull request? As I am not very familiar with ActiveMQ, I am not very well suited for this review, I think.

henning-gerhardt:
@solth , @markusweigelt : I can review this pull but reviewing it will take some time.

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
```diff
@@ -350,39 +350,51 @@ public void unassignTaskFromUser(Task task) throws DataException {
      *
      * @param comment as Comment object
      */
-    public void reportProblem(Comment comment) throws DataException {
+    public void reportProblem(Comment comment, TaskEditType taskEditType) throws DataException {
```
henning-gerhardt, line 353: As you change the method signature by adding a new parameter please adjust the Javadoc.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+            if (Objects.isNull(currentTask)) {
+                throw new ProcessorException("Task with id " + taskId + " not found.");
+            }
+
+            processAction(mapMessageObjectReader, taskAction, currentTask);
+            ServiceManager.getProcessService().saveToIndex(currentTask.getProcess(), true);
+
+            for (Task task : currentTask.getProcess().getTasks()) {
+                // update tasks in elastic search index, which includes correction comment status
+                taskService.saveToIndex(task, true);
```
henning-gerhardt, line 89: Why are all tasks of a process are saved to index? Why not only the current task? If a process as 12 tasks then 11 tasks without any change are saved to index. This is maybe on a single process not a big deal but by processing hundreds or thousand processes per day this is a big influence.

henning-gerhardt (changes requested):
This is the first round of reviewing this pull request and did only contain a read code review. Testing and getting experience with this pull request is next.

But I'm unsure about the setting task state to 'lock' as this is dangerous and was a sign of a big error in the workflow in the past. I will talk to you in next days.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
markusweigelt: Thx. I think I copied that and didn't think about it enough and adjust that.

markusweigelt: I have changed that. Now only if `correction id` is defined all tasks are updated in the index. If no `correction id` is set only the current task updates.

markusweigelt:
@henning-gerhardt I have added your comments from the review and hints from our conversation.

Task Action `ERROR_OPEN` 
- now only works in task processing status `INWORK`
- If no correction identifier is set, the task remains in processing state `INWORK`

Task Action `ERROR_CLOSE` 
- If no correction identifier is set, the task processing state should be `INWORK`
- If correction identifier is set, the task processing state should be `LOCKED`

The tests and the documentation linked above is adapted too.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,219 @@
...
+            Task currentTask = taskService.getById(taskId);
+            if (Objects.isNull(currentTask)) {
+                throw new ProcessorException("Task with id " + taskId + " not found.");
+            }
+
+            processAction(mapMessageObjectReader, taskAction, currentTask);
+
+            if (mapMessageObjectReader.hasField(KEY_CORRECTION_TASK_ID)) {
+                for (Task task : currentTask.getProcess().getTasks()) {
+                    taskService.saveToIndex(task, true);
```
henning-gerhardt, line 88: Saving all task to index of a process looks for me still like solving a hidden issue, even this is only called when a correction id is submitted.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/TaskActionProcessor.java
```diff
@@ -0,0 +1,219 @@
...
+
+    private void actionProcess(Task currentTask, User currentUser) throws DataException {
+        currentTask.setProcessingStatus(TaskStatus.INWORK);
+        currentTask.setEditType(TaskEditType.QUEUE);
+        currentTask.setProcessingTime(new Date());
+        taskService.replaceProcessingUser(currentTask, currentUser);
+        if (Objects.isNull(currentTask.getProcessingBegin())) {
+            currentTask.setProcessingBegin(new Date());
+            taskService.save(currentTask);
+        }
```
henning-gerhardt, line 178: Task get only saved on the first time ever. If this task is executed a second time, the change is only submitted to the index and not to the database which is resulting in different task state between database and index. I would recommend to set the processing begin every time the task is set to "in process" and save the task even every time.

henning-gerhardt (changes requested):
This is now a functional review. I tested a few use scenarios but I did not test the usage of the correction id! Without a correction id the new functionality is working even I'm not convinced by setting a task to "LOCKED" (red) as this is in my current experience world wrong: after a "DONE" (green) task only a "OPEN" (yellow) or "INPROCESS" (orange) task state an follow and never a "LOCKED" (red) state.

One bigger issue is that if a task is set to "INPROCESS" a second time the state is not saved to the database and the new processing time is even not updated.

I did even read your documentation in the wiki (thumbs up for this) but is was for me at least at two places confusing or not understandable to read:

ERROR_OPEN (documentation at https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ#error_open)

Some kind of confusing to read "This action set the processing status of a task to LOCKED when correctionTaskId is set and creates an error comment." and later "When correctionTaskId is set the correction task is set to OPEN."

ERROR_CLOSE (documentation at https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ#error_close)

It is difficult to understand or confusion to read (at least for me): Name of the action ingest that the task got closed. But first sentence says "OPEN". Next sentence says "LOCKED" (kind of close), except under some conditions is set or stay "INWORK" and last sentence says "DONE" (kind of close). 

Maybe you can explain this two actions in a better or other way. Maybe even renaming the action names to more meaning and understandable names.

markusweigelt:
@henning-gerhardt have incorporated your comments and hints. If you are satisfied with it, then please approve.

henning-gerhardt (approved):
Works for me in my usage scenarios without using a correction id.

solth:
@markusweigelt please rebase against current master.

markusweigelt:
Merged against the current master. The current error in the CI is unrelated to this PR.

solth:
@markusweigelt I am not sure but I think the following line can potentially throw NullPointerExceptions now that `Comment.correctionComment` can be null, couldn't it?

```

markusweigelt:
@solth I resolved a potential null pointer exception and checked if there were more occurrences with 'getCorrectionTask,' but I didn't find any.

PR #5659, opened 2023-05-09
Cleanup of message properties

solth:
- translate remaining German message keys
- remove some unused message keys
- delete key duplicates from Spanish translations


markusweigelt (approved):
Thx! Works for me.

Issue #3392, opened 2020-04-03
Metadata editor is small after many mistakes


matthias-ronge:
After many errors have occurred …

![Screenshot: many error messages showing](https://user-images.githubusercontent.com/3040657/78376130-2accd080-75ce-11ea-9085-3833c1865151.png)

… the metadata editor is small and lots of white space remains down. It never changes after that:

![Screenshot: a small metadata editor and a lot of white space](https://user-images.githubusercontent.com/3040657/78376268-5d76c900-75ce-11ea-9db4-a6ab2bf86915.png)

**Goal:** The metadata editor should use the available screen space again after error messages have gone.

Kathrin-Huber:
The + icon for adding Metadata dissapears

matthias-ronge:
Yes, I noticed that, too.

oliver-stoehr:
A possible solution might be:
Add a click listener to `body` or something similar and check for every event if the real target is the close button of the messages. (It is not possible to add a listener directly to the close button without changing the primefaces component.)
If the close button was clicked, execute the JS `resize` method.

andre-hohmann:
@matthias-ronge : Does this behaviour still occur?

PR #5660, opened 2023-05-10
Show pagination format error

### LINKED ISSUE #5658: Stacktrace is displayed after pagination format error.
**Describe the bug**
If the input in pagination start value is incorrectly formed according to the selected pagination mode a stacktrace is displayed. Occurs when the input string is unsuccessfully converted to int.

**To Reproduce**
1. Go to 'Metadaeditor - Pagination'
2. Set pagination to 'Arabic'
3. Select one or more pages for pagination.
4. Enter a String with non-numerical characters, e.g. "test" in 'Pagination start value'.
5. Stacktrace is displayed.

**Expected behavior**
A error message should be displayed, instead of the stacktrace.

**Screenshots**

**Release**
3.5.0, (current master branch not tested yet).

 - Windows 11
 - Firefox

pontus-osterdahl:
This pull requests catches the number format exception when malformed pagination is attempted and displays a error message instead of the stacktrace.

* Catch NumberFormatException and print error message
* Add paginationFormatError message

Fixes #5658


solth (changes requested):
Thanks for the pull request. It does indeed fix the linked issue (there is just one small error in the German translation). 

This solution does emphasize the importance of fixing #3392, however, because the "Apply" button in the pagination panel is unreachable after closing the pagination start value error message. I guess the optimal solution would be to display any errors (pagination, validation or other errors) in a popup dialog instead, but that should probably be done when dealing with #3392.

solth (approved):

Issue #5626, opened 2023-04-14
Subordinate processes cannot be linked to superordinate process in another project


andre-hohmann:
**Describe the bug**
If a subordinate process (for example: volume) is created and the superordinate process exists, but in another project, the two processes are not linked during the import. 
In spite, a second superordinate is created, it the import depths is `2`. 
If the process is linked manually, the following message occurs: 
- `The search completed, but nothing was found. You can only find processes that have the same project and rule set.`

**To Reproduce**
1. Search an existing process of a volume and copy the Identifier - for example the PPN
2. Import the metadata of the process in a different project then the one of the superordinate process - import depths is `2`
3. See the message with the number of imported processes - it will show: `2 processes retrieved from K10Plus-SLUB-PICA`
4. Try to link the subordinate process with the existing superordinate process - the message `The search completed, but nothing was found. You can only find processes that have the same project and rule set.` is shown

**Expected behavior**
It should be possible to link subordinate processes of different projects with the same superordinate process, regardless of its project affiliation.
This has worked in earlier versions. 

**Screenshots**
![FehlerUeberordnung01](https://user-images.githubusercontent.com/14835770/232036119-70d2c13e-97ef-4f4e-adcc-fcd6ef255120.jpg)

**Release**
KITODO.PRODUCTION Version 3.6.0-SNAPSHOT

**Desktop (please complete the following information):**
 - OS: Windows
 - Chrome

BartChris:
I am wondering wether the two projects must have the same ruleset in order for that to work. @matthias-ronge made a comment on that here: https://github.com/kitodo/kitodo-production/issues/4941#issuecomment-1025616071.

andre-hohmann:
Yes, the ruleset must be the same, but relations between processes of different projects with the same ruleset should be possible. 

In my examples, the projects have even the same process template "Standard". 
It seems to me that in https://github.com/kitodo/kitodo-production/issues/4941#issuecomment-1025616071 the creation of a process via the parent process (periodical, newspaper, ...) is described. I want to create processes for volumes of multivolume works, which can be imported from K10plus. But in the end, the rules should be the same. 

I can find migrated processes of volumes, which belong to different projects (with the same ruleset), but are linked to the same superordinate process. 
I am also quite sure, that this was possible in earlier versions. We are working with the current master and maybe there is a new mistake.

BartChris:
This would effect two constellations, as far as i can see. 
[...]

PR #5663, opened 2023-05-10
Allow linking of processes to parents in different project

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/5626


andre-hohmann:
@BartChris : Thanks a lot for the draft! 
We applied it and the first tests were successful - from my point of view. However, i can only assess it only from the users' perspective regarding the functionality and I hope that i have taken all aspects into account. 

Here the result of my first observations. 
The children processes are now always linked to the existing parent process, regardless 
- of the project affiliation of the existing and new processes 
- of the project authorizations of the users

Kitodo/src/test/resources/metadata/9/meta.xml
henning-gerhardt, line 1: Why get this file be deleted?

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -964,6 +965,18 @@ private Process loadParentProcess(Ruleset ruleset, int projectId, String parentI
         return parentProcess;
     }
 
+    private List<ProcessDTO> sortProcessesByProjectID(List<ProcessDTO> processDTOs, int projectId) {
```
henning-gerhardt, line 968: You are adding a sorting here but without to show the project information in the list of possible parent processes - so I don't see any benefit of sorting here.

BartChris: You are right that it does not make sense in case of manual process linking. The origin of that is the idea, that when doing automatic linking during import we have to decide to which parent a child gets linked (if there are multiple). I had the idea of having the parents in the same project to be listed first (and therefor be linked) to not just have a random selection.

henning-gerhardt: Do you mean with automatic linking the mass import feature? If yes, as we did not use this feature, I'm not aware of possible issues there. But then your sorting logic is useful and I pray that even then the right parent process is sorted first.

As we do only manual linking with one parent process over all projects we are not effected by this additional sorting and not showing project information. This could / should be added in a different pull request if this information need to be displayed.

BartChris: Yes, this is mostly for mass import. I added an unit test to ensure that sorting works correctly.

BartChris:
As said, i would propose the sorting logic to automatically link to the parent in the same project. We cannot eliminate all potential problems here, because there might even be multiple possible parent processes in the same project. But this is the risk you have to take when you rely on mass import.

However, in order to cover the case of having multiple possible parent processes for manual process creation, i would like to adapt the logic to return all found processes from the search so that the user can actually select which specific project he or she wants to link to. Only in case of mass import the system has to select one process. This would have to be addressed in a different pull request. (which should also introduce a way to display the project of the process a user is going to link to).

henning-gerhardt:
@BartChris : did you work on this pull request? We are using this pull request already in our productive system as without this changes we could not have go productive.

henning-gerhardt:
@BartChris : Sorry for pinging you. Any news regarding this pull request?

BartChris:
@henning-gerhardt sorry for letting you wait. I will look into this but i fear we also have to take into accout the whole discussion here which has brought up some conflicting ideas: 
https://github.com/kitodo/kitodo-production/discussions/5635

henning-gerhardt:
I don't think that all points of the discussion need be implemented in first approach. As in the discussion is written that an iterative process with this pull request as first iteration and the other points in other later pull requests. It this is correct @andre-hohmann  and @solth ?
We at SLUB need this changes for 3.6.0 release and later releases. If this pull request is not going into the master branch I will fork this changes into the SLUB fork and must and will maintain this changes for SLUB only.

BartChris:
I looked into the discussion again and what i get from it is, that in the SLUB you would allow linking also to inactive projects. (To quote Andre: " Aus meiner Sicht ist es nicht sinnvoll, ein deaktiviertes Projekt erneut zu aktivieren, weil ein Vorgang zur Verknüpfung benötigt wird." https://github.com/kitodo/kitodo-production/discussions/5635#discussioncomment-6107541) so the current behaviour of the pull request would be correct in that regard. 

The main problem i see right now is wether we should allow linking to processes in projects the user does not have access to. Would it be in line with the "iterative" approach to allow the linking for now by bringing the changes of this PR into master and then work on another pull request which introduces the new authority to do that? Or should we wait for a pull request which introduces the new authority for adding links to parent processes?
The implementation of the new permission requires elaborating the specfic behaviour. Should a user without this permission not see the linkable processes at all? Or should he or she be informed, that there might be parent processes but that they cannot be linked? How and when does this happen? What happens during mass import?

henning-gerhardt:
I read this sentence from Andre too but at SLUB we did not close any project as we have corrections in any project at any time. But as there are more concerns which may must be discussed, so this must be wait until the discussion is more clear. I'm hoping that the discussion goes on.

andre-hohmann:
Thanks, @henning-gerhardt and @BartChris for the summary. 

I am not sure, who decides, if the iterative approach can be applied. I think this must be decided by conses (Release Management, participants of this discussion, Community Board).
I just know, that we can work with this current draft, as we do not apply the mass import. In my opinion, it would be preferable to apply the PR in the master. 

Probably, it could be officially asked, if the current PR can be merged to the master. If this is not possible, we would have precise requirements from the release management and we will apply this changes into the SLUB-fork. 
We cannot work with Kitodo.Production 3 without this functionality. 

We could set this discussion to the list of the Community Board [Themensammlung Community Board](https://github.com/kitodo/kitodo-production/wiki/Themensammlung-Community-Board) and discuss it there. 
Regardless of whether the pull request is accepted or not, a consistent solution to this problem should be found. 

The challenge is to find a funder, who will finance the implementation of the requirements of all stakeholders. Especially when authorizations, mass import, ... are taken into account.

solth:
In my opinion even an iterative approach should not undermine the current permission system. For me, to make this pull request viable, the following adjustments would have to be made:
1. add the special permission for linking with otherwise unavailable parent processes
2. show otherwise unavailable process candidates in the list of parent candidates in "Titelsatzverknüpfung", but display an "!" icon next to them to distinguish them from other regular processes and to show that these processes are not accessible to the user via the process or task lists
   - if the user _does_ have the special permission from 1.: activate selection of these processes in question in the tab "Titelsatzverknüpfung"
   - if the user does _not_ have the special permission from 1.: deactivate the processes in question (show them as unselectable/grey) in "Titelsatzverknüpfung" and expand the tooltip to inform the user that he lacks the special permission required to link processes to other processes to which he does not have regular access
5. do _not_ consider otherwise unavailable processes as potential parents during mass import, since this would mean additional work and adjustments and the SLUB does not require this part of the implementation; this is something that could be added in a _later_ iteration

In any case, we have to keep in mind that if a user makes a mistake and links to the _wrong_ parent process from another project (to which he normally does not have access) using this new special permission, he could not remove the faulty link/fix the error by himself, because process links can currently only be removed from within the metadata editor and only to subordinate/child processes, not to superordinate/parent processes. So in this case he would have to inform some other user with appropriate project access to remove the wrong process link for him. Although I think that this would be a bad user experience and desgin, it would be acceptable to merge, since it would still fulfill the basic requirement of the SLUB while _not_ undermining the permission system. Improving upon this shortcoming could be done in a subsequent iteration, too.

andre-hohmann:
Thank you very much for the extensive requirement description, @solth. 
We will examine the requirements in the SLUB Dresden and will decide about the next steps. 

@BartChris : Thanks a lot for the draft! As far as you are not interested yourself in merging the PR, you do not need  to invest more time in it. 
We can assess in the Community Board, if other institutions are interested in a solution, or if this is only relevant for SLUB Dresden.

solth:
Dear @andre-hohmann I hope it is obvious that only my description is extensive, but the actual change requests themselves are rather small. I just wanted to prevent any misunderstandings so I explained my ideas very "verbosely". 

As far as I am concerned I think we would merely need to add an SQL script for the new permission and then add one more function to the "Titlesatzverknüpfung" that - upon selecting a candidate as parent process - checks whether
- the user has _regular_ access to the process 👉 display process normally 
- the user does _not_ have regular access, but does have the new _permission_ to link to other projects processes 👉 display the process and activate selection, but add the "!" icon with tooltip like "This process belongs to another project which is not assigned to you. You can link these processes but won't be able to remove the link yourself in case of an erroneous process association".
- the user has _neither_ regular access _nor_ the new permission 👉 display the process, but deactivate it for selection (grey) and add a tooltip to the "!" icon like "This process belongs to another project which is not assigned to you. To be able link these processes anyway, please contact the system administrator to receive the appropriate permission."

One thing I forgot above is that even if the user _does_ have the new permission, he should never see any processes of projects that belong to different _clients_, but if I understand the current changes by @BartChris correctly this restriction has already been incorporated.

solth:
@andre-hohmann I have forked the branch of @BartChris and added the mentioned authority to link to parent processes of unassigned projects. The branch can be found here: https://github.com/effective-webwork/kitodo-production/tree/process_parent_link 

It is also rebased against the current master and adjusts the "TitleLinkVerknüpfung" tab as described above:
Bildschirmfoto 2023-11-02 um 09 42 42
(as you can see I haven't been able to add an icon with tooltip to the menu option, so I instead added a hint in brackets)

@andre-hohmann  and @henning-gerhardt could you check if this meets your requirements?

andre-hohmann:
Dear @solth , thanks a lot the improvement! 

Unfortunately, i cannot check it yet, because we do not apply the current version of Kitodo.Production. We will update our experimental system as soon as possible, but i cannot estimate the duration. When the current version is available, i will check the improvement from the users' perspective. 

Nevertheless, i have the following questions: 
1. Is the field "`TitleLinkVerknüpfung`" (select process) above the one shown in "Titelsatzverknüpfung" - see: [Create process - volume](https://github.com/kitodo/kitodo-production/wiki/Mehrb%C3%A4ndige-Werke#band)?
2. Does the user has to select the process every time, or only, when there are several ones? In my opinion, the superordinate process should be immediately selected, if there is only one. Otherwise, i suppose that the users will be annoyed.

solth:
>     1. Is the field "`TitleLinkVerknüpfung`" (select process) above the one shown in "Titelsatzverknüpfung" - see: [Create process - volume](https://github.com/kitodo/kitodo-production/wiki/Mehrb%C3%A4ndige-Werke#band)?
> 

I meant "Titlesatzverknüpfung", not "TitleLinkVerknüpfung", sorry for the confusion. 
Perhaps I cropped the screenshot too much, so here is a larger version of it and yes, it shows the same part you mentioned above:

Bildschirmfoto 2023-11-03 um 12 22 18

---

>     2. Does the user has to select the process every time, or only, when there are several ones? In my opinion, the superordinate process should be immediately selected, if there is only one. Otherwise, i suppose that the users will be annoyed.

As far as I can tell the process is not selected automatically, even if the search returned just a single process. But in my opinion that is another topic or suggestions for improvement and does not have to do with linking processes across multiple projects.

---

Anyway, I reviewed this pull request once more and partly have to adjust my opinion of it, I think. I still maintain that we should add the new permission/authority to control whether someone is allowed to link processes to parent processes of unassigned projects eventually, as described and implemented above. 

This would however mean a change to the database, which would mean this fix could not be part of planned bugfix release 3.6.1 (see https://github.com/kitodo/kitodo-production/milestone/21) - because bugfix releases cannot contain any changes to the database for the sake of compatibility - but would only just be included in the next major release 3.7.0. Since 3.7.0 will not be released before Q4 2023/Q1 2024 and I assume that would be too late for you to be of any use, I would instead merge the original pull request by @BartChris (once it is rebased against the current master) into the 3.6.x branch for 3.6.1, which hopefully can be released very soon, and include the new permission in the release 3.7.0. 

@andre-hohmann Would that solve the issue for you or are you not going to use any official release anyway, but instead are planing to continue building your own release of selected pull requests, in which case you would merge the branch this pull request is based on into your own custom version anyway and including it in any official releases wouldn't be of such a high priority for you?

solth:
After discussing this with @henning-gerhardt I have decided to drop this issue from the 3.6.1 release because the SLUB is probably not going to use this bug fix release and will wait for 3.7.0.

andre-hohmann:
@solth : As i wrote above, we will think about the next steps. 

It is not of such high priority for us, because we can apply the draft of @BartChris. 

Of course, we want to use an official release as soon as possible, but we want to test it carefully. As we are now getting used to Kitodo Production 3 in general and the colleagues have many questions, there is not enough time for testing of new versions.

BartChris:
@solth I tested the changes on your branch and they work good for me. The message that the user does not have access to a specific project is useful, but i think we should definitely show the name of the project as well.

Another thing i noticed is that my current implementation shows a difference in behavior depending on the trigger for the parent search. 

When a process is imported using "import depth" 2 or above, the search for the parent is triggered automatically. and then only one process is selected via the dropdown as defined here:

https://github.com/kitodo/kitodo-production/blob/9221a28a33d54c039041c83efe4303239c0edd0d/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L967

When the search is triggered manually all found processes are returned to the dropdown. The behaviour should probably be more consistent.

andre-hohmann:
@solth : Thanks a lot for your investigations and improvements!

@henning-gerhardt has updated our experimental system to 3.6 and now i could apply/test the changes. 
This was necessary, because i did not comprehend the changes completely. 
Now, i understand the changes better and in general it works, but i have some questions. 

To avoid further misunderstandings, it might be better have a look on it during an online meeting. I am just not sure yet, who should participate besides us two. I can imagine, that @BartChris and @henning-gerhardt could be interested in it. 
The results could be published here as another comment.

andre-hohmann:
Thanks to all for your support and sorry for the inconveniences. 

After a short e-mail conversation, this PR can be merged from my point of view. The main goal has been reached and the discussions on this PR and the issue are very long and confusing. 

Any further improvements should be described and/or discussed separately.

solth:
@BartChris I updated my branch https://github.com/effective-webwork/kitodo-production/tree/process_parent_link. Could you test whether this now sufficiently addresses the issues you pointed out (project title, preselect allowed process)?

BartChris:
@solth Your changes work good for me. However i encountered another problem. The authority check only takes place when you manually search for the parent process. It does however not happen if the automatic search for the parent is triggered using `loadParentProcess`. This method is also accessed during mass import:

https://github.com/effective-webwork/kitodo-production/blob/a2d577933fc878ea198ef9328d591f881662e77e/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L959-970

We should probably also add the checks you added in the `TitleRecordLinkTab` to the `ImportService` class so that a user cannot link to a process which belongs to a project he or she does not have access to.

BartChris:
superseded by https://github.com/kitodo/kitodo-production/pull/5869

Issue #5661, opened 2023-05-10
Different process states for done processes


henning-gerhardt:
With development of Kitodo.Production 3.x the state value of processes was changed so that done processes (all tasks in the process workflow are done) got the new value `100000000000` instead of `100000000` from 2.x. But this change was forgotten while development the migration of newspaper like processes or serial publications. If a new hierarchy process are created while this migration this process got the value `100000000` when all child processes are done.

Instead of dealing with two different values the new value should only be used and if possible all old values should be changed to the new value.

PR #5664, opened 2023-05-11
Replace old with new value for sortHelperStatus

henning-gerhardt:
Fixes: #5661 

- replacing old value `100000000` with new value `100000000000` on all needed places
- replacing the magic numbers with enumeration entries
- changing the values in the database

It is not necessary to reindex all processes after this change is live but I would recommend it. 

Searches in database and elastic search are still using both values to determinate completed processes. This can maybe changed in later versions.

### DIFF
--- a/Kitodo-DataManagement/src/main/resources/db/migration/V2_123__Replace_old_sorthelperstatus_value.sql
+++ b/Kitodo-DataManagement/src/main/resources/db/migration/V2_123__Replace_old_sorthelperstatus_value.sql
@@ -1,0 +1,21 @@
+--
+-- (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+--
+-- This file is part of the Kitodo project.
+--
+-- It is licensed under GNU General Public License version 3 or later.
+--
+-- For the full copyright and license information, please read the
+-- GPL3-License.txt file that was distributed with this source code.
+--
+
+-- 1. Disable safe updates
+--
+SET SQL_SAFE_UPDATES = 0;
+
+-- 2. Replace old sortHelperStatus from "100000000" to "100000000000"
+UPDATE process SET sortHelperStatus = "100000000000" WHERE sortHelperStatus = "100000000";
+
+-- 3. Enable safe updates
+--
+SET SQL_SAFE_UPDATES = 1;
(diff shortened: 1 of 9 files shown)


thomaslow:
Hi Henning, this is just a suggestion, but you could also use the recursive SQL query to determine whether a parent process has no unfinished tasks in any child processes, see:

https://github.com/kitodo/kitodo-production/blob/5425b6f2e330d58df0b4ce1059549f264127fae0/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/TaskDAO.java#L152-L195

At the time of implementing it, I noticed the "sortHelperStatus" field in the database, but didn't really understand how and when it was used. Since my goal was only to allow sorting of the process list, I didn't touch the "sortHelperStatus"-part of the code.

Anyway, the recursive SQL query should give you the same information as the "sortHelperStatus" without the need to keep that information up-to-date in the database.

henning-gerhardt:
Hi @thomaslow : thank you for your suggestion but this out of scope of this pull request as I only want to fix the issue and not change the calculation and behaviour of this field.

The `sortHelperStatus` field was introduced in 1.x to speed up the displaying and using of the process status. Before this field was introduced the calculation was done on demand and was very very slow. So displaying 100 rows of processes takes one or two minutes and f.e. creating a list of 10.000 processes as an excel sheet takes 15 and more minutes. The pre-calculation was significant faster with the negative side of more complexity. But I think this compromise is still good. 

Your suggestion would revert this change back to an on demand calculation. With the help of recursive query (which may not work on other database management systems) it may even so fast like the pre-calculation without the additional complexity but how looks this on the database side if there are a lot of this queries are run at the same time as many users (more then f.e. 20 users) are doing things in the software which are using this recursive query?

If such a change should happen than the query should work on so many database management systems as possible and the database systems even should handle this kind of usage. Until now there are no real hardware / virtual ware requirements for running Kitodo.Production in different scales.

thomaslow:
@henning-gerhardt You are correct: A fully on-demand calculation would be problematic for filtering or sorting. 

I just wanted to let you know, that there is a new method that can calculate this information very quickly via the recursive query, and that the same information is also available in the ElasticSearch index as the field "progressCombined" (because it was necessary for sorting processes). 

Of course, I understand that this is out of the scope of this pull request. It is just an idea for future improvements.

solth (approved):

PR #5665, opened 2023-05-16
Refactor SecurityConfig class

henning-gerhardt:
- splitting content of big method `configure` into smaller methods without loosing the readable code.
- reducing the allowed checkstyle violations by one

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
+++ b/Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
@@ -99,29 +99,137 @@ public class SecurityConfig extends WebSecurityConfigurerAdapter {
         http.sessionManagement().maximumSessions(1).sessionRegistry(getSessionRegistry())
                 .expiredUrl(LOGIN_PAGE);
 
-        // viewAll... Authority to view list of entities
-        // view...... Authority to view entity at edit page
-        // edit...... Authority to change and save entities at edit page
+        // site specific rules
+        authorizeSpecificPages(http);
+
+        // more general rules should be at end
+        authorizeGeneralPages(http);
+
+        http.addFilterAfter(new SecurityObjectAccessFilter(), FilterSecurityInterceptor.class);
+
+        handleFormLogin(http);
+    }
+
+    private void authorizeSpecificPages(HttpSecurity http) throws Exception {
+        authorizePageClientEdit(http);
+
+        authorizePageIndexing(http);
+
+        authorizePageProcesses(http);
+
+        authorizePageProcessEdit(http);
+
+        authorizePageProjects(http);
+
+        authorizePageProjectEdit(http);
+
+        authorizePageTemplateEdit(http);
+
+        authorizePageDocketEdit(http);
+
+        authorizePageRulesetEdit(http);
+
+        authorizePageWorkflowEdit(http);
+
+        authorizePageTasks(http);
+
+        authorizePageUsers(http);
+
+        authorizePageRoleEdit(http);
+
+        authorizePageLdapGroupEdit(http);
+    }
+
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
```diff
@@ -99,99 +99,182 @@ protected void configure(HttpSecurity http) throws Exception {
...
-                EDIT_CLIENT + CLIENT_ANY,
-                VIEW_CLIENT + GLOBAL,
-                VIEW_CLIENT + CLIENT_ANY)
+        // site specific rules
+        authorizeSpecificPages(http);
 
-            .antMatchers("/pages/indexingPage.jsf").hasAnyAuthority(
-                "editIndex_" + GLOBAL,
-                "viewIndex_" + GLOBAL)
+        // more general rules should be at end
```
solth, line 105: Just out of curiosity: why should general rules be at the end?

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
```diff
@@ -99,99 +99,182 @@ protected void configure(HttpSecurity http) throws Exception {
...
-                VIEW_ALL_TASKS + CLIENT_ANY)
+        authorizePageProjects(http);
+
+        authorizePageProjectEdit(http);
+
+        authorizePageTemplateEdit(http);
+
+        authorizePageDocketEdit(http);
+
+        authoruzePageRulesetEdit(http);
```
solth, line 130:
```suggestion
        authorizePageRulesetEdit(http);
```

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
```diff
@@ -99,99 +99,182 @@ protected void configure(HttpSecurity http) throws Exception {
...
+                VIEW_WORKFLOW + GLOBAL,
+                VIEW_WORKFLOW + CLIENT_ANY);
+    }
 
-            .antMatchers("/pages/images/**").permitAll()
-            .antMatchers("/javax.faces.resource/**", "**/resources/**").permitAll()
-            .antMatchers("/js/modeler.js").permitAll()
-            .antMatchers("/js/toggle.js").permitAll()
-            .anyRequest().authenticated();
+    private void authoruzePageRulesetEdit(HttpSecurity http) throws Exception {
```
solth, line 195:
```suggestion
    private void authorizePageRulesetEdit(HttpSecurity http) throws Exception {
```

solth (changes requested):
Looks good, just one small typo.

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
henning-gerhardt: See the examples in https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/config/annotation/web/builders/HttpSecurity.html#authorizeHttpRequests(org.springframework.security.config.Customizer) and there  `Note that the matchers are considered in order. Therefore, the following is invalid because the first matcher matches every request and will never get to the second mapping`

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
henning-gerhardt: Typo is fixed

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
henning-gerhardt: Typo is fixed

solth (approved):

PR #5668, opened 2023-05-22
Do not reset form data on changing ldap group on user edit page

### LINKED ISSUE #4933: Choosing an LDAP group for a new user resets input
When creating a new user, and setting an LDAP group for him or her, the values already entered are lost.

Screen cast

**Goal:** Do not loose values when changing the LDAP group

solth:
Fixes #4933


matthias-ronge (approved):
Formally correct. I didn’t test it.

PR #5672, opened 2023-05-30
Fix thumbnail scrolling

### LINKED ISSUE #5670: Scroll-panel for images in metadata editor does not work correctly
**Describe the bug**
In the "detail view" of the metadata editor, the scroll-panel on the rights side does not work correctly. 
~~1. If an image is selected in the scroll-panel, no image is shown in the "Detail view", but the image in the structure tree is selected. Therefore, it is possible/necessary to click additionally the image in the structure tree to show the image file in the "detail view".~~ 
Solved by: 
- #5655

2. If the image is selected in the structure tree, the correct image is shown in the detail view, but the scroll-panel shifts to the last images. 
3. If an image is selected in the scrollbar, the selected images in the scroll-panel stays not in focus, but the first image is shown again. 
Therefore it is necessary to scroll back to the selected image in the scroll-panel. 

See GIF below. 

**To Reproduce**
1. Open the metadata editor 
3. Go to "detail view"
5. Scroll down to an image in the panel and select it

**Expected behavior**
If an image in the sin the panel is selected, 
1. ~~the selected image should be shown directly in the detail view.~~
Solved by: 
- #5655
2. the sin the panel  should always be in focus of the selected image and should not shift to the beginning or the end of the scrollbar. 

**Screenshots**
ScrollbarMetadateneritor

**Release**
KITODO.PRODUCTION Version 3.6.0-SNAPSHOT

 - Windows
 - Chrome

markusweigelt:
- Change refreshing element
- Adapt the scroll function to the improved structure

Fixes #5670


solth (approved):
Thanks for this fix! Looks correct and seems to work. :+1:

andre-hohmann (changes requested):
@markusweigelt : 

- It works perfectly, if you are navigating in the "scroll-bar". 
- If you select an object in the structure tree, it works well until object number 100 (roughly). Then, the selected object is out of focus. It seems, as if with every object, the focused object moves down. You should test this with a process of at least 200 objects. Could you check this again?

Kitodo/src/main/webapp/WEB-INF/resources/js/scroll.js
```diff
@@ -157,7 +157,7 @@ function initializeStructureTreeScrolling() {
 }
 
 function scrollToPreviewThumbnail(thumbnail, scrollable) {
-    let thumbnailHeight = thumbnail.parent().parent().height();
+    let thumbnailHeight = thumbnail.closest(".thumbnail-parent").height();
```
oliver-stoehr, line 160:
```suggestion
    let thumbnailHeight = thumbnail.closest(".thumbnail-parent").outerHeight(true);
```
You could use here `outerHeight` to consider the elements margin and change the following rule in kitodo.css:3167.
```
#imagePreviewForm\:thumbnailWrapper .thumbnail-parent {
    display: block;
    margin-bottom: var(--group-margin);
    margin-left: auto;
    margin-right: auto;
}
```
The `scrolltoPreviewThumbnail` method did not take the spacing between the elements into account when computing the scrollTop value. These changes should do the trick.

oliver-stoehr (changes requested):

markusweigelt:
Added @oliver-stoehr suggestions, tested the change with more than 200 images and it worked for me.

andre-hohmann (approved):
@markusweigelt : Thanks a lot! 
From the users' perspective, it works fine now.

oliver-stoehr (approved):

Issue #5676, opened 2023-06-01
Marking a correction comment as "resolved" does not update the index properly


oliver-stoehr:
**Describe the bug**
When a correction comment is marked as "resolved", the task and/or process index is not updated with the new correction comment status. The index still contains the status "2" (process has at least one open correction comment).

**To Reproduce**
1. Create a new correction comment
2. The task list and process list show the red "C"/"K".
3. Mark the correction comment as resolved.
4. The task list and process list still show the red "C"/"K" and not the expected grey "C"/"K".
The process widget does however show the correct grey "C"/"K". (The info is here correctly fetched from database and not from index, like on the task list and process list.

**Release**
3.6.0-SNAPSHOT

**Desktop (please complete the following information):**
 - OS: macOS 13.3
 - Browser: Safari 16.4

PR #5679, opened 2023-06-01
Improve correction icon

oliver-stoehr:
Replace the correction "C"/"K" with a more universal icon independent of the user language. The new icon matches the style of the other icons used in Kitodo and has the same size as the other icons, thus not breaking the list layout.
The character is replaced by an exclamation icon in a circle:
Bildschirmfoto 2023-06-01 um 17 06 52
The color scheme (red/gray) for active/resolved correction comments is kept.

### DIFF
--- a/Kitodo/src/main/resources/messages/messages_de.properties
+++ b/Kitodo/src/main/resources/messages/messages_de.properties
@@ -227,7 +227,6 @@ copyRolesToClient=Rollen zum Mandanten hinzuf\u00FCgen
 copyTemplates=Projekt inkl. Templates kopieren
 correction=Korrektur
 correctionFor=Korrektur f\u00FCr Schritt
-correctionK=K
 correctionMessage=Korrekturmeldung
 correctionMessageSend=Korrekturmeldung senden
 correctionMessageSendAllTasks=Korrekturmeldung f\u00FCr alle Schritte senden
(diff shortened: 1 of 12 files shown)


markusweigelt (changes requested):
Thx for implementation.

In my case it does not work in list views. In the comment column only the speech bubble is visible even if a correction task is active. In task list i do not have such a column to show comment or error icon. How did you set this up these columns?

image

oliver-stoehr:
@markusweigelt There is no comment column in the tasks widget, but only in the processes widget, the designated processes page and the designated tasks page.

I recently discovered that the comments might not be updated in index properly (see https://github.com/kitodo/kitodo-production/issues/5676). Is the displayed icon also wrong when you rebuild your index?

markusweigelt:
Unfortunately, the main icon within the process list widget does not change. When hovering over the comment icon, the list of recent comments opens, and that's where you can finally see the correction message.

image

I couldn't find any additional locations other than the dashboard (as seen in the screenshot). Could you please provide me with more information about where to find the corresponding tasks and process page? We can also clarify this briefly via webcall.

oliver-stoehr:
@markusweigelt I took another look at the processes widget and fixed the icon in the processes widget.
You can enable the comments column on the processes page and the tasks pages by enabling it via the "configure columns" menu next to the "extended search" button.

markusweigelt:
@oliver-stoehr Thank you, now it's working for me. The columns in the task and process overviews also display the icon.

markusweigelt (approved):

Issue #5580, opened 2023-03-06
Remove dead code of interfaces (Kitodo-API)


matthias-ronge:
Production consists of a program core and modules. These are addressed via `interface`s. These interfaces were designed from scratch in a very, very early development stage of Production v. 3. Over time, the functionality of the individual modules has become clearer. It showed that part of the interfaces are dead code, which should be removed.

 - [ ] **Remove ID of the _command interface_.** The command interface is used to run scripts. An ID and a command line are passed. The ID is generated, passed, and returned in the program core, but it is not for anything further.
 - [ ] **Remove the _data editor interface_.** The data editor interface was designed to call the metadata editor. It has two methods, the first for read-only viewing of an XML file in 2.x format. The second method is used to edit an XML file in the metadata editor and returns whether the editing was successful. The interface has two sub-interfaces: structure data editor and metadata editor. The whole data editor interface is dead code.
   * The read-only viewing of a 2.x XML file has never been implemented and will become less necessary the further away from previous software versions.
   * The function to start the metadata editor, which returns the result of the editing, is not compatible with the reality of a JSF web application. When a user clicks a button to launch the metadata editor, the action must finish, after which JSF redirects the browser to the metadata editor. The function cannot wait for a result at this point, otherwise the redirect cannot take place. The feature was never implemented.
   * The two sub-interfaces do not contain functions.
 - [ ] **Remove unused parts of the _image management interface_.** The image management interface offers four methods: `changeDpi`, `createDerviative`, `getScaledWebImage`, `getSizedWebImage`.
   * It should be clarified whether the `changeDpi` does. It is not documented exactly how it is intended: When changing a 300 DPI image to 150 DPI, does it become half as big (half as many pixels)? Or does it double in size when trying to print it, keeping its number of pixels? In the first case, the function would be identical to `createDerivative`, in the second case the question arises, whether that is needed by anybody.
   * The `createDerivative` and `getScaledWebImage` functions are functionally equivalent. Therefore, the `getScaledWebImage` function is no longer offered in the front-end. However, it exists in the interface and is represented in the database. This is dead code and can be removed.
 - [ ] **Remove the _OCR interface_.** The OCR interface was never implemented, and can be removed.

PR #5681, opened 2023-06-02
Remove ID of the command interface

matthias-ronge:
Part of #5580


Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/CommandService.java
```diff
@@ -22,8 +22,6 @@
 
 class CommandService {
 
-    private Random random = new Random(1000000);
```
henning-gerhardt, line 25: Class import of java.util.Random is not necessary anymore and can be removed.

henning-gerhardt (changes requested):
Changes look good and did work. Remove the now unneeded class import statement from the CommandService class of the Kitodo-FileManagement module and the pull request is fine.

PR #5682, opened 2023-06-02
Remove dead code of data editor interface

matthias-ronge:
Part of #5580


matthias-ronge:
Unfortunately the branch has strayed too far from the trunk. It is no longer possible to understand how the desired change can be achieved under the given circumstances. The pull request is closed. It needs to be redeveloped.

solth:
Parts of the code that were to be removed via this pull request were still used in by the Swiss Federal Archives. That is one reason it hasn't been merged, sorry for the late reply.

PR #5684, opened 2023-06-02
Remove the OCR interface

matthias-ronge:
Part of #5580


markusweigelt (approved):
Should work! I haven't tested anything because if code builds and selenium test runs error free it shouldn't be a problem.

Maybe link the issue #5580 for automatic close after merge.

PR #5692, opened 2023-06-07
Use correct spellings of database table names

matthias-ronge:
Database table names are case-sensitive on Linux. So, I wonder why this ever worked …


henning-gerhardt:
This is working and will ever working as this queries are written in HQL (Hibernate Query Language) and not in SQL! HQL queries will be translated to the right language of the used database system so you do not need to care about f.e. case sensitive things.

henning-gerhardt:
Even the "wrong" table names are correct as they are not table names from the database they are the used bean names.

matthias-ronge:
Thank you for the clarification. I didn’t knew about HQL.

PR #5697, opened 2023-06-13
OCR-D Workflow integration

markusweigelt:
- Create OCR Profile entity with title and selected file of directory configured with property `directory.ocr.profiles`
- Add new authorities to manage OCR Profile
- Select OCR Profile in process template
- Select individual OCR Profile in process -> default is the process template OCR Profile
- Add OCR Profile as shell script variable via `{ocrprofilefile}` placeholder
- Add project DMS export path as shell script variable via `{projectdmsexportpath}` placeholder
- Added documentation among others for the placeholder https://github.com/kitodo/kitodo-production/wiki/Workflow-Editor#platzhalter


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/OCRWorkflow.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public Boolean isActive() {
+        if (Objects.isNull(this.active)) {
+            this.active = true;
+        }
+        return this.active;
+    }
+
+    /**
+     * Set ruleset as active.
```
matthias-ronge, line 69: `Sets whether OCR workflow is active.`

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/OCRWorkflow.java
```diff
@@ -0,0 +1,115 @@
...
+public class OCRWorkflow extends BaseBean {
+
+    @Column(name = "title")
+    private String title;
+
+    @Column(name = "file")
+    private String file;
+
+    @Column(name = "active")
+    private Boolean active = true;
```
matthias-ronge, line 34: What if `active` is `null`?
→ `private boolean active = true;`

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/OCRWorkflow.java
```diff
@@ -0,0 +1,115 @@
...
+        this.file = file;
+    }
+
+    /**
+     * Check if ocr workflow is active.
+     *
+     * @return true or false
+     */
+    public Boolean isActive() {
+        if (Objects.isNull(this.active)) {
```
matthias-ronge, line 62: No need for this if you make `active` a `boolean` primitive.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
```diff
@@ -62,6 +62,10 @@ public class Process extends BaseTemplateBean {
     @JoinColumn(name = "docket_id", foreignKey = @ForeignKey(name = "FK_process_docket_id"))
     private Docket docket;
 
+    @ManyToOne
```
matthias-ronge, line 65: Please add corresponding `@OneToMany` relationship in `OCRWorkflow`.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/OCRWorkflow.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.ForeignKey;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "ocrworkflow")
+public class OCRWorkflow extends BaseBean {
```
matthias-ronge, line 25: The term _workflow_ is already in use in Production with a different meaning. Maybe we can find a not-yet-used name for this class, like _OCR configuration_ (`OcrConfig`).

Kitodo/src/main/java/org/kitodo/production/dto/OCRWorkflowDTO.java
```diff
@@ -0,0 +1,99 @@
...
+package org.kitodo.production.dto;
+
+/**
+ * OCR workflow DTO object.
+ */
+public class OCRWorkflowDTO extends BaseDTO {
+
+    private String file;
+    private String title;
+    private Boolean active = true;
```
matthias-ronge, line 21: as above, should be `boolean` primitive

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/OCRWorkflow.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+
+        if (object instanceof OCRWorkflow) {
+            OCRWorkflow ruleset = (OCRWorkflow) object;
```
henning-gerhardt, line 104: `ruleset` as variable name? Maybe use `ocrworkflow`?

Kitodo-DataManagement/src/main/resources/db/migration/V2_124__Add_authorities_to_manage_ocr_workflow.sql
```diff
@@ -0,0 +1,41 @@
...
+INSERT IGNORE INTO authority (title) VALUES ('editOCRWorkflow_clientAssignable');
+INSERT IGNORE INTO authority (title) VALUES ('deleteOCRWorkflow_clientAssignable');
+INSERT IGNORE INTO authority (title) VALUES ('viewOCRWorkflow_clientAssignable');
+
+-- Add table "ocrworkflow"
+CREATE TABLE IF NOT EXISTS ocrworkflow (
+    id INT(10) NOT NULL AUTO_INCREMENT,
+    title varchar(255) NOT NULL,
+    file varchar(255) NOT NULL,
+    active tinyint(1) DEFAULT NULL,
```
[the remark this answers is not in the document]
henning-gerhardt: If `one` means true why should a new entry per default is active?

Kitodo/src/main/java/org/kitodo/production/dto/OCRWorkflowDTO.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param title
+     *            as String
+     */
+    public void setTitle(String title) {
+        this.title = title;
+    }
+
+    /**
+     * Check if ruleset is active.
```
henning-gerhardt, line 63: `ruleset` ?

Kitodo/src/main/java/org/kitodo/production/dto/OCRWorkflowDTO.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @return whether ruleset is active or not
```
henning-gerhardt, line 65: `ruleset`?

Kitodo/src/main/java/org/kitodo/production/dto/OCRWorkflowDTO.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public Boolean isActive() {
+        return this.active;
+    }
+
+    /**
+     * Set if ruleset is active.
```
henning-gerhardt, line 72: `ruleset`?

Kitodo/src/main/java/org/kitodo/production/dto/OCRWorkflowDTO.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param active
+     *            whether ruleset is active or not
```
henning-gerhardt, line 75: `ruleset`?

Kitodo/src/main/java/org/kitodo/production/forms/OCRWorkflowEditView.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Get list of ocr workflow filenames.
+     *
+     * @return list of ocr workflow filenames
+     */
+    public List<Path> getOCRWorkflowFilenames() {
+        try (Stream<Path> ocrWorkflowPaths = Files.walk(Paths.get(ConfigCore.getParameter(ParameterCore.DIR_OCR_WORKFLOWS)))) {
+            return ocrWorkflowPaths.filter(f -> f.toString().endsWith(".sh"))
```
henning-gerhardt, line 91: Regarding to `.sh` file extension: what about non Unix systems which did not use `.sh` as file extension like on Windows with `.bat` or similar?

Kitodo/src/main/java/org/kitodo/production/forms/OCRWorkflowEditView.java
```diff
@@ -0,0 +1,119 @@
...
+                    .map(Path::getFileName).sorted().collect(Collectors.toList());
+        } catch (IOException e) {
+            Helper.setErrorMessage(ERROR_LOADING_MANY, new Object[] {ObjectType.OCR_WORKFLOW.getTranslationPlural() },
+                    logger, e);
+            return new ArrayList<>();
+        }
+    }
+
+    /**
+     * Get importConfiguration.
```
henning-gerhardt, line 101: `importConfiguration`?

Kitodo/src/main/java/org/kitodo/production/forms/OCRWorkflowEditView.java
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @return value of importConfiguration
```
henning-gerhardt, line 103: `importConfiguration`?

Kitodo/src/main/java/org/kitodo/production/forms/OCRWorkflowEditView.java
```diff
@@ -0,0 +1,119 @@
...
+     */
+    public OCRWorkflow getOcrWorkflow() {
+        return ocrWorkflow;
+    }
+
+    /**
+     * Set ocrWorkflow.
+     *
+     * @param ocrWorkflow as org.kitodo.data.database.beans.OCRWorkflow
```
henning-gerhardt, line 112: Is the full package name needed?

Kitodo/src/main/java/org/kitodo/production/forms/OCRWorkflowListView.java
```diff
@@ -0,0 +1,78 @@
...
+@Named("OCRWorkflowListView")
+@SessionScoped
+public class OCRWorkflowListView extends BaseForm {
+    private OCRWorkflow ocrWorkflow;
+    private static final Logger logger = LogManager.getLogger(OCRWorkflowListView.class);
+    private final String ocrWorkflowEditPath = MessageFormat.format(REDIRECT_PATH, "ocrWorkflowEdit");
+
+
+    /**
+     * Get import configurations.
```
henning-gerhardt, line 46: `import configuration`?

Kitodo/src/main/java/org/kitodo/production/forms/OCRWorkflowListView.java
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @return import configurations
```
henning-gerhardt, line 48: `import configurations`?

Kitodo/src/main/java/org/kitodo/production/forms/OCRWorkflowListView.java
```diff
@@ -0,0 +1,78 @@
...
+            Helper.setErrorMessage(ERROR_LOADING_MANY,
+                    new Object[] { ObjectType.OCR_WORKFLOW.getTranslationPlural() }, logger, e);
+            return new ArrayList<>();
+        }
+    }
+
+
+    public String newOCRWorkflow() {
+        return ocrWorkflowEditPath;
+    }
```
henning-gerhardt, line 63: Is this correct?  Method `newOCRWorkflow` is returning content of `ocrWorkflowEditPath`? Maybe only the variable naming must be adjusted.

Kitodo/src/main/java/org/kitodo/production/services/ocr/OCRWorkflowService.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Constructor necessary to use searcher in child classes.
+     */
+    private OCRWorkflowService() {
+        super(new OCRWorkflowDAO());
+    }
+
+    /**
+     * Return singleton variable of type OCRDWorkflowService.
```
henning-gerhardt, line 38: `OCRDWorkflowService`? Did you mean `OCRWorkflowService`?

Kitodo/src/main/java/org/kitodo/production/services/ocr/OCRWorkflowService.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @return unique instance of OCRDWorkflowService
```
henning-gerhardt, line 40: `OCRDWorkflowService`?

henning-gerhardt (changes requested):
This is an only read code review. I did not test it nor I checked logic in deep. 

Beside the remarks as comments there is one other: some method use `OCR` in big letters and sometimes in camel case `Ocr` and - if I'm not wrong - sometimes in small letters `ocr`. Maybe this can be unified to one style?

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
markusweigelt: We don't have such a corresponding relationship for similar examples, for example Ruleset or Docket and we don't need this relationship atm. If the relationship is needed then we should add it later and provide an appropriate migration script. I want to customize the data structure only for the cases which are really needed.

@solth What do you mean?

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/OCRWorkflow.java
bertsky: That ambiguity is [well known](https://ocr-d.de/en/spec/glossary#ocr-d-workflow) and oft discussed. So far, I have never seen a good counter proposal. _OCR configuration_ gets closest indeed – I could live with that.

Within [ocrd_kitodo](https://github.com/slub/ocrd_kitodo/) documentation, though, we decided to use _workflow_ and always disambiguate between (organisational) Kitodo Workflow and (technical) OCR Workflow.

henning-gerhardt:
> > Rename ocr to ocrd
> 
> I cannot find a review/discussion which motivated that change. Since in ocrd_kitodo we decided to not make Kitodo depend on OCR-D whatsoever, but stay generic, I recommend against it. So far, technically, AFAICS there is nothing OCR-D-specific here. (Any OCR backend could use OCR workflow file selection.) 

In my opinion this pull request is a OCR-D specific integration and not a generic one as in my opinion every integration has its own needs: some are more complex some are more simple. OCR-D is needed code changes other did not require that.

For example  the Zeutschel ABBYY is a lot simpler to integrate and did not need any code change of Kitodo.Production and worked like that: calling a shell script which is starting a simple java application on the Kitodo.Production host is starting the OCR. The OCR by ABBYY is controlled by the Zeutschel Manager software and closing the executing task through the ActiveMQ message queue from Kitodo.Production.

bertsky:
> In my opinion this pull request is a OCR-D specific integration and not a generic one as in my opinion every integration has its own needs: some are more complex some are more simple. OCR-D is needed code changes other did not require that.

Just because this PR is the first to integrate OCR workflow files does not mean it is not generic. Again, there is nothing specific to OCR-D here. The contents of the files is arbitrary as far as Kitodo is concerned, and the OCR is still integrated merely as a ScriptTask.

> For example the Zeutschel ABBYY is a lot simpler to integrate and did not need any code change of Kitodo.Production and worked like that: calling a shell script which is starting a simple java application on the Kitodo.Production host is starting the OCR. The OCR by ABBYY is controlled by the Zeutschel Manager software and closing the executing task through the ActiveMQ message queue from Kitodo.Production.

Zeutschel/ABBYY decided they do not want to concern the users with details like selecting OCR workflows. That's understandable for a commercial service which is technically backed by a black box anyway. But they _could_ have offered a workflow selector, too. Any OCR backend could.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/OCRWorkflow.java
bertsky: Agreed, _configuration_ may not be so good after all. Yes, _OCR profile_ would also be acceptable, if properly explained/documented, as indeed would have to be done in any case.

Erikmitk:
@markusweigelt Maybe you can explain where the OCR-D specificity comes from? I don't see it either. What would prevent someone from using whatever OCR they use now?

markusweigelt:
I assumed that the workflows were OCR-D specific and that's why I had refactored it as part of the adjustements from the review. I think we all can agree on `OCR Profile`. I will then adjust that accordingly.

matthias-ronge (commented):
**TL;DR Technically, the result of the reviews is OK. But I wonder whether we need the functionality in this form.**

What basically strikes me about this pull request is that a lot of effort is put into it for little additional functionality. Moreover, this can only be used with a special external application. At least if I understood it correctly. For me, it looks like this:

In my opinion, the `ocrProfile` is technical metadata. The goal is, that this metadata can be defined in the production template. I would choose the following solution, without touching the source code at all: The metadata is created with the desired value in the ruleset with the `domain="technical"`. (An additional wrapping ruleset can be created for exactly this setting using includes, if different settings for the same ruleset are to be stored in the system.) The ruleset with the metadata is used in the production template. So, when a process is created, the `ocrProfile` is taken over into the process as the corresponding metadata. Here it can be used.
I see the directory for OCR profiles, that can be configured here, completely outside of KitodoProduction.


I think of a completely different (better) implementation: If you could edit a page of metadata in the production templates. So that you can start the metadata editor in the production template and set any metadata (allowed in the ruleset) there. This is then passed into all created processes.

That's my opinion on this. Technically, I haven't tried the pull request. The code looks OK when I looked at it.

BartChris:
Hi all,

as i already tested out the OCR-D suite (https://github.com/slub/ocrd_manager, https://github.com/slub/ocrd_controller) in our Kitodo test system and it works good an i am therefor trying to deploy that in production.  I would appreciate if we could move forward in enabling a better integration of Kitodo and OCR-D. and want to make some comments here based on practical experience.

I got the OCR-D integration running without using the changes of this pull request and i want to point out that we mostly have to think about the indented users of the proposed implementation.

Right now i mostly think about two scenarios when doing an OCR in the context of out mass digitization
1) Printed Documented with a simple layout (maybe from 1800 until today). For that maybe applying the default workflow provided in the OCR-D manger repository is already enough

https://github.com/slub/ocrd_manager/blob/main/workflows/ocr-workflow-default.sh

2) More complex layouts which would require preprocessing and more steps from the OCR-D workflow:
 
https://ocr-d.de/en/workflows
I would probably apply the more complex workflows to older materials with more complex workflows.

Apart from that i could derive most of the information from the metadata we encode as part of the description in Kitodo, mainly the language (English, German..) and the Script type (Antiqua, Fraktur). Iam already injecting those parameters on the fly in my predefined workflows.

The question for me then is: What is the additional value of tying a specific workflow to a production template? A production template could be associated with a project which holds many different type of materials which require different types of OCR. In general, a production template is determined by different considerations than the selection of a specific OCR.

This pull request make the "OCR profile" also selectable by process. This is more useful, but do the content editors really have the necessary knowledge to make an informed decision here? I do not see myself or another person with more knowledge configuring OCR profiles for single processes.

The only possible scenario i can think of is that a user can - on the basis of the project - define, that this material - although it is quiet young requires a more sophisticated workflow. This would mean that in the end i have two profiles: "Simple Layout" and "Complex layout". And this could probably be recorded, as Matthias suggests, in the technical metadata.

Having that information stored, it is possible to attach to those "Profile" to a specific file with workflow instructions. But it is probably not ultimately necessary since my script which is  called from Kitodo already contains a lot of Business logic and could get the workflow specification from a lot of different places.

Having said that, i would prefer not spending a lot of discussion whether the approach here is to OCR-D specific or generic enough as a lot of issues (how does reporting back to Kitodo work etc. and we already have some good proposals here: https://github.com/kitodo/kitodo-production/pull/5657 ) are much more important than the selection of a workflow file.

markusweigelt:
> I got the OCR-D integration running without using the changes of this pull request and i want to point out that we also have to think about the indented users of the proposed implementation.
> 
> Right now i mostly think about two scenarios when doing an OCR in the context of out mass digitization
> 
> 1. Printed Documents with a simple layout (maybe from 1800 until today). For that maybe applying the default workflow provided in the OCR-D manger repository is already enough
> 
> https://github.com/slub/ocrd_manager/blob/main/workflows/ocr-workflow-default.sh
> 
> 2. More complex layouts which would require preprocessing and more steps from the OCR-D workflow:
> 
> https://ocr-d.de/en/workflows I would probably apply the more complex workflows to older materials with more complex layouts.

> Apart from that i could derive most of the information from the metadata we encode as part of the description in Kitodo, mainly the language (English, German..) and the Script type (Antiqua, Fraktur). I am already injecting those parameters on the fly in my predefined workflows.
> 
> The question for me then is: What is the additional value of tying a specific OCR-workflow ("OCR profile") to a production template? 



>A production template could be associated with a project which holds many different type of materials which require different types of OCR. In general, a production template is determined by different considerations than the selection of a specific OCR.
> 




> This pull request make the "OCR profile" also selectable by process. This is more useful, but do the content editors really have the necessary knowledge to make an informed decision here? I do not see myself or another person with more knowledge configuring OCR profiles for single processes.
> 

> 
> Having that information stored, it is possible to attach to those "Profile" specific file with workflow instructions and specify those files in Kitodo. But it is probably not ultimately necessary since my script which is called from Kitodo already contains a lot of Business logic and could get the workflow specification (encoded in a file) from a lot of different places.
> 
> I would probably agree that the approach here is generic enough since i could also use the information in the "OCR Profile" file to send specific instruction to _any_ OCR server or process. I am not sure if it is really necessary, but nothing prevents the encoding of OCR server specific information (e.g. an Abby Server) in the file. Or one can choose not to select a profile at all because it is not necessary in the Institution's OCR setup. Kitodo stays agnostic in that regard. Although i do not see that many use cases for the given implementation, i just see it as support for an additional scenario of interacting with an OCR system. (by passing it workflows which are stored in Kitodo), which still allows for other, different ways of using an external OCR service.
> 

> PS: This is probably a limited view based on my practical experiences. I am sure there are a lot of scenarios where having a lot of profiles to select from is useful. But i am wondering if a sophisticated workflow specification will ever happen inside of Kitodo and not in another (e.g. OCR-D-based) software.

Erikmitk:
Let me preface this by saying that everything depends on how you use Kitodo.Production and how (digitization) workflows are organized in each individual institution.
 
> PS: This is probably a limited view based on my practical experiences. I am sure there are a lot of scenarios where having a lot of profiles to select from is useful. But i am wondering if a sophisticated workflow specification will ever happen inside of Kitodo and not in another (e.g. OCR-D-based) software.

**If you know what kind of material you are digitizing than you could make an informed decision about which specific OCR-workflow you want to use as early as creating the project.** So you can run every process through the tailored OCR workflow and it will work out just fine. Outliers can be dealt with individually. If you use whatever default workflow is available from the start, you then have to individually re-OCR each process with a different OCR-workflow if you'll have figured out later on that there's a better workflow available. This is completely independent from OCR-D, because you could use different engines or OCR models or whatever even with any kind of script task or processing as you choose.

**But how do you make an informed decision about which OCR-Workflow to choose in the first place?!** 

> This would mean that in the end i have two profiles: "Simple Layout" and "Complex layout". And this could probably be recorded, as Matthias suggests, in the technical metadata.

This view is too limited. **You can have complex layouts, simple layouts, workflows for specific material from specific time frames and even handwriting.** The main idea would be to share OCR profiles and give better default workflows for a wide variety of use cases. The choice we have (OCR-D specific) is very limited and will be expanded. Additionally, the idea would be to share workflows and models to gather experience in the community. Not every complex layout is the same and it can be useful to have a more specifically trained workflow at hand. **Complex layouts in what kind of complexity? Mix of text and images? Very narrow tables? Overlapping Multi-Columns?** These are not the same. If someone has found a useful workflow it could be shared publicly with examples etc. Not every institution has to experiment to find the perfect workflow for common material types or eras themselves. **This can be a shared effort.** So one **could** end up with a wide variety of possible workflows, parameters or models. That's what this OCR-profile selector is for. The choice to put this in the ruleset therefore does not seem viable, because that is to heavy on manual labor in maintaining these choices and keeping the actual workflows and options in sync. Just drop a new workflow-profile in the directory and it is there to be used.


In the context of OCR-D the main vision is to get rid of workflow choice completely. The project aims for auto-correecting and -optimizing workflows so that your only choice is to use it or not. But as we're not there yet, we need to have the option to chose a OCR-profile that can be used to select a corresponding OCR-workflow. And again, this could be OCR-D or any other script call that can be tuned by specifying a OCR profile that can be read and used accordingly.

Hope this makes sense!

BartChris:
@Erikmitk I completely agree with what you are saying. I think as well that the exchange of standards and Best practices how to approach different types of materials would be one of the most important outcomes of such an undertaking. 
And i think you are also right in saying that there could be more profiles than just "easy" and "complex". 

I would say that the most important thing in general is moving forward in integrating OCR-D or any other technology in Kitodo so that we have a basis for knowledge sharing. With regard to that PR:

bertsky:
@BartChris thanks for your feedback! 

I still see one point which @Erikmitk has not addressed:
> Apart from that i could derive most of the information from the metadata we encode as part of the description in Kitodo, mainly the language (English, German..) and the Script type (Antiqua, Fraktur). I am already injecting those parameters on the fly in my predefined workflows.

Note: we are currently [preparing](https://github.com/slub/ocrd_manager/issues/67) to change our default OCR workflow(s) to include that metadata information (eg. Tesseract model switch block). 

You can pass that info to the Manager from the Kitodo Script via parameters:

    /usr/local/kitodo/scripts/script_ocr_process_dir.sh ... --workflow {ocrprofilefile} --lang $(meta.topstruct.DocLanguage) --script $(meta.topstruct.slub_script)

See [database entries for default Kitodo workflow in the demo](https://github.com/slub/ocrd_kitodo/blob/ad712ac6d0ad896532c4dbd482ef8456fb79c3ac/_resources/kitodo-sample/sql/kitodo_post_init.sql#L46).

In a custom/external Kitodo instance, you need to [configure your Kitodo workflows](https://github.com/kitodo/kitodo-production/wiki/Workflow-Editor) to include an OCR-D step. That wiki link also points to [documentation on all the available placeholders](https://github.com/kitodo/kitodo-production/wiki/Platzhalter).

markusweigelt:
Closed cause new implementation without the use of an explicit entity with PR #5809

PR #5701, opened 2023-06-26
Rename media files corresponding to their order value

### LINKED ISSUE #5601: Rename jpegs after changing sequence in Kitodo 3
In Kitodo 2 was a subpoint of pagination to regenerate filenames after changing the sequence of the scans. 
It was useful after adding or deleting pictures. Especially when URN granular is generated afterwards based on the filenames. 
As far as I see there was no implementation of this feature in Kitodo 3. 
There is no possibility to delete or add pictures afterwards either. 

Is there by any chance a possibilty that I just overlook the features or is there no implementation yet?

solth:
This PR adds a button to the metadata editor allowing users to rename all media files of the process according to the corresponding physical divisions `ORDER` attribute. 

Bildschirmfoto 2023-06-26 um 09 54 05

The new filename will be padded with `0` on the left to reach a filename length that can be configured for each project in the project settings. 

Bildschirmfoto 2023-06-26 um 09 55 38

The button is only displayed when the current user has the (new) permission to rename media files, which is not assigned to any role by default.

Bildschirmfoto 2023-06-26 um 09 55 02

After renaming the media files has finished, a result dialog will be displayed, showing the number of files renamed in all configured process folders.

Bildschirmfoto 2023-06-26 um 09 57 00

Filename changes will be reverted if the user does not explicitely save changes in the metadata editor. This approach is consistent with other functionalities in the editlor like uploading or removing images, which will also be reverted when leaving the metadata editor without clicking the "Save" button.

Fixes #5601


markusweigelt (approved):
Code looks good and it works for me!

Suggestions:
- Handling of error if number of filename length in project is too low for all images
- A test for resetting the media names when user not save changes after media renaming.

How do you do this?
After media renaming i restarted the application. All images where found but the meta.xml doesn't changed. How do you save the current state?

markusweigelt:
I approved the PR last week, but I am still struggling with the scenario where the media is renamed and not saved, or the `preDestroy` method is not executed due to errors from the server and/or client side.

However, I was unable to produce such an problem during my tests by stopping the server immediately after the replacement. Therefore, I will maintain my approval from my side.

PR #5741, opened 2023-08-02
Change version 'org.apache.xmlgraphics' to '2.8' and add removed depe…

IkramMaalej:
…ndencies


solth:
This is goind to fix https://cwe.mitre.org/data/definitions/20.html and https://cwe.mitre.org/data/definitions/918.html, right @IkramMaalej ?

solth:
@henning-gerhardt would you mind providing a review for this pull request?

PR #5745, opened 2023-08-09
Sort template projects by title

### LINKED ISSUE #5645: Sorted project list on process template create and view

On opening the details of a process template the assigned projects are unsorted. 

Screenshot_2023-04-28_15-32-18

On creating a new process template the projects are unsorted too.

Screenshot_2023-04-28_15-38-01

List of projects on create and viewing process templates should be sorted by project title.

This issue could be a bug to but I'm unsure if this sorting by design or not.

### LINKED ISSUE #5728: Sort projects in process template view by project title
Currently the list of projects in the process template view are sorted by the project ids or unsorted at all. See

Screenshot_2023-07-21_09-50-29

This did not only effect on creating new process templates even the display on already created process templates.
Screenshot_2023-10-02_16-06-17

The list of projects should be sorted by project title.

solth:
Fixes #5728 
Fixes #5645

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/SelectProjectDialogView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/SelectProjectDialogView.java
@@ -14,7 +14,9 @@ package org.kitodo.production.forms.createprocess;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.faces.context.FacesContext;
 import javax.faces.view.ViewScoped;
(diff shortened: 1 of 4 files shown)


henning-gerhardt:
For creating / editing new process templates the sorting of projects is now correct. But this change did not fixes the issue #5728 as the sorting in the overview is still incorrect. In the past I missed to add a screenshot (which I did now) and even here:
Screenshot_2023-10-02_16-06-17

Sorry for the late reply and the missing screenshot :-(

solth:
> But this change did not fixes the issue #5728 as the sorting in the overview is still incorrect.

You added this aspect to the existing issue 5728, but isn't that already described in the separate different issue #5645 ? So this pull request does indeed resolve issue 5728 in it's original form and description, I think.

henning-gerhardt:
Issue #5728 is a full duplicate of #5645 which was even opened by me. 

How should we go?
- Changing both to only one aspect (#5728 for editing and #5645 for the overview), so this pull request is then valid for #5728?
- Closing #5728 as duplicate of #5645?

solth:
@henning-gerhardt in my opinion #5728 and #5645 originally described two different aspects and were not actually duplicates: 5728 is about the order of assigned projects in the "template edit" page, while 5645 described the order of projects in the row expansion of the template list. 

Since both pages load the data from different sources (template edit page loads the whole template bean from the database while template list loads data from the index), they require distict solutions. Therefore I think we should treat them as similar, but separate issues that can and have to be resolved individually. Then, this PR would still fix 5728 while the other issue would remain open.

solth:
@henning-gerhardt no need to change the issue descriptions. I have added a commit so that this pull request should cover both lists and issues.

henning-gerhardt (approved):
Changes look good in code and functionality as now the project lists are sorted.

Issue #5647, opened 2023-05-02
NPE while creating Import Configuration


henning-gerhardt:
**Describe the bug**
Creating a new import configuration can cause a null pointer exception like

```
java.lang.NullPointerException
	at java.base/java.lang.String.contains(String.java:2036)
	at org.kitodo.production.helper.Helper.appendUnusedInsertions(Helper.java:493)
	at org.kitodo.production.helper.Helper.getTranslation(Helper.java:479)
	at org.kitodo.production.forms.validators.ImportConfigurationMappingValidator.validate(ImportConfigurationMappingValidator.java:60)
	at org.kitodo.production.forms.validators.ImportConfigurationMappingValidator.validate(ImportConfigurationMappingValidator.java:30)
	at javax.faces.component._ComponentUtils.callValidators(_ComponentUtils.java:291)
	at javax.faces.component.UIInput.validateValue(UIInput.java:489)
	at org.primefaces.component.picklist.PickList.validateValue(PickList.java:116)
	at javax.faces.component.UIInput.validate(UIInput.java:746)
	at org.primefaces.component.picklist.PickList.validate(PickList.java:182)
	at javax.faces.component.UIInput.processValidators(UIInput.java:293)
	at javax.faces.component.UIComponentBase.processValidators(UIComponentBase.java:1458)
	at javax.faces.component.UIComponentBase.processValidators(UIComponentBase.java:1458)
	at javax.faces.component.UIComponentBase.processValidators(UIComponentBase.java:1458)
	at org.primefaces.util.ComponentUtils.processValidatorsOfFacetsAndChilds(ComponentUtils.java:247)
	at org.primefaces.component.api.UITabPanel.processValidators(UITabPanel.java:1107)
	at javax.faces.component.UIForm.processValidators(UIForm.java:210)
```

**To Reproduce**
1. Create new Import configuration
2. Insert somthing for title and description field
3. Select Fileupload as Configuration type
4. Click on Save Button
5. Get the null pointer exception

**Expected behavior**
In case of a configuration error an error message should appear and not a null pointer exception

**Release**
3.6.0-SNAPSHOT (master branch from March 1th 2023 commit 5b60965afb8a9512b0021b416ad98291971e66a4)

solth:
[...]

PR #5746, opened 2023-08-09
Create title for new processes from uploaded files without parent pro…

solth:
…cess automatically

Titles for processes created from uploaded files are currently only generated automatically when the uploaded XML file also contains the parent record of the process. For processes created from files _without_ parent record the title generation has to be triggered manually by clicking gear symbol next to the title input field.

The change in this pull request ensures that process title generating rules from the ruleset are also applied automatically when the uploaded XML contains only one record without its parent.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/MetadataImportDialog.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/MetadataImportDialog.java
@@ -80,9 +80,14 @@ public abstract class MetadataImportDialog {
             }
         }
 
-        if (StringUtils.isBlank(tempProcess.getAtstsl()) && Objects.nonNull(parentTempProcess)) {
-            ProcessHelper.generateAtstslFields(tempProcess, Collections.singletonList(parentTempProcess),
-                    ImportService.ACQUISITION_STAGE_CREATE, true);
+        if (StringUtils.isBlank(tempProcess.getAtstsl())) {
+            if (Objects.nonNull(parentTempProcess)) {
+                ProcessHelper.generateAtstslFields(tempProcess, Collections.singletonList(parentTempProcess),
+                        ImportService.ACQUISITION_STAGE_CREATE, true);
+            } else {
+                ProcessHelper.generateAtstslFields(tempProcess, null,
+                        ImportService.ACQUISITION_STAGE_CREATE, true);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


solth:
@BartChris would you mind doing a review for this pull request?

BartChris:
i will look into it :+1:

BartChris:
While looking into it the file upload also stumbled over the Nullpointer exception described here and pointed out  the roor cause:
https://github.com/kitodo/kitodo-production/issues/5647

BartChris:
I have a problem understanding the constellation here. It appears to me as if the constellation is the following:
If a file is uploaded which does not contain a parent record the generation of the title (and therefor this pull request) seem to work. But if i upload a file which contains the child record and a higherlevel-Identifier (but not the full parent record), the import only works if the parent's record ID is already in the Kitodo Database. 
If the parent record does not exist, the Import dialog shows two data entries (for child and parent) but both contain the same data (for the child) and same Identifier.
image

So for me it does not work to to upload a metadata entry which contains a parent identifier (but not the parent record). It seems like Kitodo is trying to import from the uploaded XML file when it accounteres an higherlevelIdentifier. But since there is no actual data in the xml for the parent that does not work.

If the current Pull request is only about fixing the generation of the title generation for xml files without any higherlevelIdentifier in them, we could adress those other issues seperately.

BartChris (approved):

solth:
@BartChris thank you very much for the review. Yes, for retrieving a separate parent record from the same, uploaded XML file an additional mapping file has to be configured in the corresponding "Import configuration":

Bildschirmfoto 2023-08-18 um 16 54 20

In principle the system should not even _try_ to extract a parent record from an uploaded file when this parameter is not set in the import configuration, so that seems to be a bug. The question is whether I created that bug with this pull request. Otherwise it should be possible to reproduce this behaviour with the same file and configuration in the current master branch. Could you confirm if the problem exists there as well?

(for the record: yes, this pull request only aims to correctly create the process _title_ of a single record in an uploaded XML file and not to fix the problem you discovered and described above)

BartChris:
Ok, i made the mistake, that i indeed set a mapping file here. 

image

If no mapping file is selected then Kitodo is not trying to import the parent record. And this works in master and your pull request branch. So this is then a configuration error on my side. So Kitodo works correct in this regard and this is probably a question of documentation again.

solth:
@BartChris thanks alot for the additional test. I think we could easily improve this - in a separate PR - by adding another check to the "import configuration form" during saving that verifies the selected "parent mapping file" is different from the regular mapping files of that configuration - what do you think?

BartChris:
Yes, that would be a good idea.

PR #5747, opened 2023-08-10
Print comments onto docket

oliver-stoehr:
Print comments onto the docket instead of the old, unused 'wikifield'.

To test this, make sure you have the updated docket.xsl in your local directory. Then use the "print docket" button (printer icon) in the process list on any process with at least one comment.

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/docket/DocketData.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/docket/DocketData.java
@@ -39,8 +39,8 @@ public class DocketData {
     /** The creation Date of the process. */
     private String creationDate;
 
-    /** A comment. */
-    private String comment;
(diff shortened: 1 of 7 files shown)


solth (approved):
Code looks good and comments are now output to the docket correctly again! 👍

solth:
From my perspective this looks good, but I am not sure if the old `wikiField` property of `Process.java` still needs to be taken into account as well. There is not functionality in Kitodo 3 to write any comments to this field anymore, but maybe this is still required for old processes migrated from Kitodo 2.x?

solth:
@kitodo/kitodo-community-board do you think we still need to consider the old "wikifield" from processes migrated from Kitodo 2.x? Or is it unlikely to still print dockets with metadata for those old processes?

PR #5748, opened 2023-08-10
Improve status color

oliver-stoehr:
Improve color for task status "open". To make the colors easier to distinguish, the statuses have now the colors:
- Locked: Orange
- **Open: Light Blue**
- In Processing: Light Green
- Done: Green
Bildschirmfoto 2023-08-10 um 16 12 30

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/processEdit/taskList.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/processEdit/taskList.xhtml
@@ -38,7 +38,7 @@
                               style="background-color:
                                 #{item.processingStatus.title == 'statusDone' ? 'var(--green)' : ''}
                                 #{item.processingStatus.title == 'statusInProcessing' ? 'var(--light-green)' : ''}
-                                #{item.processingStatus.title == 'statusOpen' ? 'var(--light-orange)' : ''}
+                                #{item.processingStatus.title == 'statusOpen' ? 'var(--blue)' : ''}
                                 #{item.processingStatus.title == 'statusLocked' ? 'var(--orange)' : ''};"
                 />
                 <p:panel rendered="#{SecurityAccessController.hasAuthorityToEditTask()}">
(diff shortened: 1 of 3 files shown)


solth:
@kitodo/kitodo-community-board the status color for open tasks has been critisized for being too similar to the color of locked tasks for a long time. IMO changing it to light blue does make those status sufficiently distinguishable, but I think your opinion in this matter is more relevant. Could you weigh in on whether you think this is an appropriate improvement in this regard?

matthiaskissler:
I am one of those people affected by color vision impairment (Rot-Grün Sehschwäche). The improvement as shown by oliver-stoehr would be more than welcome.

oliver-stoehr:
I also tested the darker blue, because I realized the light blue is too similar to the "closed" green – even without any vision impairment.  
Bildschirmfoto 2023-08-17 um 10 31 20
@matthiaskissler is this combination also easy to distinguish for you?

matthiaskissler:
The darker blue would be even better because the contrast is higher than with the light blue

oliver-stoehr:
I changed the color to the darker blue.

andre-hohmann:
Although i am (not yet) affected, i can understand the requirement and support the search for a solution. If blue is chosen, i would prefer the light blue. 

additionally to the question dark blue/light blue, we should also consider the following aspects (thanks to @henning-gerhardt for the hints): 

1. _Documentation_: This affects maybe several documentation. The official one seems not to be affected only in a minor way, but there could be some mayor effects on internal documentations. 
2.  _Wording_: The colors have influence on how people describe the status of processes, as for example "If the process is green", if "Scanning" is red". In the beginning, this may irritate users. 
3. _Habits_: Some colleagues may be irritated by the change itself. In some cases, this has implications for acceptance. 

The aspects can be adapted, respectively new habits will be developed with time. Nevertheless i cannot assess the impact on all institutions, which apply Kitodo.Production. The state "open" is an important state, as it is the basis of accepting a task and thus affects many users. 

Maybe the colors of the states "open" and "in process" can be exchanged. 
Aspect 1 is still valid and aspects 2 and 3 are just not so "harsh". It would have at least more similarity to Kitodo.Production 2. 

I agree with @solth that this should be discussed in the community board and maybe even feedback of the community is necessary.

subhhwendt:
The community board confirm the suggested colour scheme by default and welcomes your offer to create an additional option in kitodo_config.properties to store an individual mix for a (backward) change in the entire installation.

solth:
I will close this pull request until the option to configure the color scheme is included, as @subhhwendt suggested.

PR #5756, opened 2023-08-16
Refactor guard clauses into their own class

matthias-ronge:
I was just reading this [article (in German) about Guard Clauses](https://www.heise.de/hintergrund/Fehlersuche-Guard-Clauses-den-Code-auf-Fehler-vorbereiten-9241426.html?seite=all) and I think it makes sense. It is extremely rare that such checks are already in our code. I moved the first tests into a separate class and refactored them from the code. My suggestion is to use them more regularly.


matthias-ronge:
- moved class to `utils` package
- tests added
- yes, I am asking for opinions here @solth @BartChris @markusweigelt @oliver-stoehr 
- and yes, if we decide to use this, we should use it in way more places. This PR is just a first throw.

matthias-ronge:
Even if the request for a generous implementation of Guards did not make it into the development fund, would you merge this anyway? Then, we could at least _start_ using them.

Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
```diff
@@ -0,0 +1,184 @@
+package org.kitodo.utils;
```
henning-gerhardt, line 1: Missing license header.

Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
```diff
@@ -0,0 +1,184 @@
...
+import static org.junit.Assert.fail;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+
+import org.junit.Test;
+
+public class GuardTest {
```
henning-gerhardt, line 14: Some notes:
- write a single test method for a single test scenario and not all scenarios in one test method
- exception handling can be done in a lot more readable way. See https://www.baeldung.com/junit-assert-exception for Junit 5 and Junit 4 usage - where I prefer Junit 5

henning-gerhardt (changes requested):

matthias-ronge:
No idea why the build fails here. It works fine for me, both in Eclipse and in the shell:

    [INFO] Running org.kitodo.utils.GuardTest
    [INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.224 s -- in org.kitodo.utils.GuardTest

Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
matthias-ronge: License header inserted.

Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
matthias-ronge: - Tests divided.
- I used Junit 5. Yes, I recently discovered that there is this new variant. Much nicer!

Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+class GuardTest {
```
henning-gerhardt, line 26: Maybe change the visibility of the class and the methods to "public" - I got som warnings from CodeQL as the new test class is not in use.

henning-gerhardt (approved):
Only a small note - can may be ignored if CodeQL is not complaining.

Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
```diff
@@ -0,0 +1,184 @@
...
+    void isNotNullShouldFailForNullObjekt() {
+        IllegalArgumentException e = assertThrows(IllegalArgumentException.class, () -> {
+            Object object = null;
+            Guard.isNotNull("object", object);
+        }, "should throw IllegalArgumentException");
+        assertEquals("'object' must not be null", e.getMessage(), "should provide an understandable description");
+    }
+
+    @Test
+    @Tag("isPositive(double)")
```
stweil, line 115: This  and seven more similar lines cause warnings:
```
Jul 05, 2024 6:24:52 PM org.junit.jupiter.engine.descriptor.JupiterTestDescriptor lambda$getTags$1
WARNING: Configuration error: invalid tag syntax in @Tag("isPositive(double)") declaration on [void org.kitodo.utils.GuardTest.isPositiveDoubleShouldFailForZero()]. Tag will be ignored.
```

Should these lines be removed?

PR #5757, opened 2023-08-17
Fix incorrect sorting by process title in task list

### LINKED ISSUE #5751: Incorrect sorting by column "process" in the list of tasks
**Describe the bug**
When a list of tasks is sorted by column "Process", not all tasks are displayed in the correct order. This occurs more often in extensive list of tasks (for example 10.000), but also in shorter lists. 

The processes in the process list are sorted correctly. 

**To Reproduce**
1. Show task list
2. Click on column "Process"
3. Click on column "Process" (to check sorting a-z and z-a)
4. Check the order of the processes by process title
5. See wrong sorting of the processes

**Expected behavior**
The column "process" in the task list should be ordered correctly by characters and numbers of the process titles - according to the sorting of the process in the process list.

**Screenshots**
Example with ~12.000 hits
exampleSortingTasklist01

Example with ~90 hits
exampleSortingTasklist02

**Release**

 - Windows
 - Chrome

thomaslow:
Closes #5751

Sorting based on an ElasticSearch "text" field means that the list is sorted based on the analyzed string (after stop-word-removal and stemming). This will cause an incorrect sort order. This pull request fixes the incorrect sorting for the task list for both columns "process title" and "last editing user" by using the corresponding "keyword" fields instead.


solth:
@thomaslow thanks a lot for the fix. Could you please rebase your branch against the current master where the retrieval of the Chrome driver for Selenium tests has been fixed?

solth (approved):

PR #5769, opened 2023-09-05
Filter client, role and project information in user list

solth:
Most pages and lists in Kitodo.Production are currently filtered by the current client of the the current user. The only exception to this are the client, role and project columns in the user list, which currently show _all_ assigned clients, roles and projects, no matter which information should actually be visible to the user according to his own clients and permissions. This undermines the client system and potentially even poses a security issue.

This pull request filters the columns accordingly so that only those information are displayed in the columns mentioned above to which the user has access via the clients and roles assigned to him.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/RoleService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/RoleService.java
@@ -11,6 +11,8 @@
 
 package org.kitodo.production.services.data;
 
+import static org.kitodo.constants.StringConstants.COMMA_DELIMITER;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/RoleService.java
```diff
@@ -160,6 +161,12 @@ public List<Role> getAllRolesByClientId(int clientId) {
      * @return String containing role titles
      */
     public static String getRoleTitles(List<Role> roles) {
-        return roles.stream().map(Role::getTitle).collect(Collectors.joining(", "));
+        if (ServiceManager.getSecurityAccessService().hasAuthorityGlobalToViewRoleList()) {
+            return roles.stream().map(Role::getTitle).collect(Collectors.joining(", "));
+        } else {
+            Client currentClient = ServiceManager.getUserService().getSessionClientOfAuthenticatedUser();
+            return roles.stream().filter(role -> role.getClient().equals(currentClient)).map(Role::getTitle)
+                    .collect(Collectors.joining(", "));
```
markusweigelt, line 169: Use constant here.

markusweigelt (changes requested):
Missing joining parameter in `ProjectService.java` line 356

markusweigelt (approved):

PR #5772, opened 2023-09-07
Fix Task Manager crashing

### LINKED ISSUE #5771: Task Manager fails with ArrayIndexOutOfBoundsException
**Describe the bug**
In a productive environment with many scripts, sooner or later, Task Manager cannot be opened anymore, showing an error like this:

image_2023_09_06T12_52_33_324Z

Tasks do still run, but cannot be observed anymore. Reloading the page or logging of and logging back on doesn’t fix it, the error is visible to all users. Needs Tomcat restart.

matthias-ronge:
Fixes #5771

Adds logging of exceptions being caught.


matthias-ronge:
This pull request is a draft, until success of patch is confirmed

PR #5776, opened 2023-09-08
Synchronize accesses to 'taskList'

### LINKED ISSUE #5771: Task Manager fails with ArrayIndexOutOfBoundsException

matthias-ronge:
Fixes #5771

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/tasks/TaskManager.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/tasks/TaskManager.java
@@ -76,7 +76,7 @@ public class TaskManager {
      */
     public static void addTask(EmptyTask task) {
         LinkedList<EmptyTask> tasks = singleton().taskList;
-        synchronized(tasks) {
+        synchronized (tasks) {
             tasks.addLast(task);
         }
     }
@@ -97,7 +97,7 @@ public class TaskManager {
      */
     static void addTaskIfMissing(EmptyTask task) {
         LinkedList<EmptyTask> tasks = singleton().taskList;
-        synchronized(tasks) {
+        synchronized (tasks) {
             if (!tasks.contains(task)) {
                 int pos = lastIndexOf(TaskState.WORKING) + 1;
                 try {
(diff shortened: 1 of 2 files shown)


henning-gerhardt:
I can review this pull request after beginning of October 2023.

solth:
@matthias-ronge please rebase against current master.

henning-gerhardt (changes requested):
I'm unsure over your changes and maybe I need more knowledge about thread safe programming to interpret your changes better.

Did I read your changes correct: your changes block parallel access of the thread unsafe list inside the methods after every call get a reference to the list. One access call is manipulating the list and after this the next one (undefined which one) is acting on this list.

My approach was to block the call of the methods itself, manipulate the list and then let the next caller access the changed list independing what the method is doing.

Maybe it is very helpful for understanding this issue is to add tests which shows without your changes that the list can get be broken and your changes fixes this and the list is correct in any case.

matthias-ronge:
I see the main problem as a race condition **between** the `TaskSitter` class **and** the `TaskManager` class. The key point of this pull request is, that all the source code, which accesses the non-thread-safe list, is synchronized on **one** object. Intuitively, I use the task list object itself to synchronize on.

Your approach of simply declaring the methods as `synchronized` would synchronize the methods on TaskSitter on the TaskSitter object (`TaskSitter.this`), **but** the methods of TaskManager on the TaskManager object (`TaskManager.this`). This means that you would have had **two** semaphores, and this can still lead to a race condition.

I can report that this patch is in use as on a customer server, and the problems have not occurred since then.

henning-gerhardt (commented):
I can not judge if this changes are correct even with the explanations from @matthias-ronge as my knowledge of parallel programming is to low.

sVnsation:
Changes look good to me. Everytime the `taskList` is accessed by `TaskSitter` or `TaskManager` it's locked to this single thread only with the synchronized keyword. The next thread can only access the `taskList` when the previous thread is finished.

matthias-ronge:
Thanks @sVnsation for the cross-check. @solth, would it be okay to apply the change?

PR #5779, opened 2023-09-11
Fixes #5778

### LINKED ISSUE #5778: TaskScripThread missing translation, and finished at 0%
**Describe the bug**
The `TaskScriptThread` executes automatic workflow steps. It is displayed in the task manager. Here, the verbalization is missing ("Automatic task:"), and the finished task has 0% progress. These are small things, but it's not nice, and 0% can also be irritating, as if the task didn't work correctly (but it's just showing wrong).

**To Reproduce**
1. Use a workflow with an automatic task
2. Let a process with that workflow perform an automatic task.

**Expected behavior**
If task has completed, should show 100% progress. Add translation to messages file: “Automatic task”

**Screenshots**
taskscriptthread_missing_translation and finished at 0%

matthias-ronge:
Fixes #5778

### DIFF
--- a/Kitodo/src/main/resources/messages/messages_de.properties
+++ b/Kitodo/src/main/resources/messages/messages_de.properties
@@ -1084,6 +1084,7 @@ taskManagerIsRunning=Aufgaben nicht automatisch starten
 taskManagerIsStopped=Aufgaben automatisch starten
 taskManager=Taskmanager
 taskSaving=Aufgabe wird gespeichert...
+taskScriptThread=Skript-Aufgabe
 taskStatus=Aufgabenstatus
 tasks=Aufgaben
 technicalData=Technische Daten
--- a/Kitodo/src/main/resources/messages/messages_en.properties
+++ b/Kitodo/src/main/resources/messages/messages_en.properties
@@ -1085,6 +1085,7 @@ taskManagerIsRunning=do not start tasks automatically
 taskManagerIsStopped=start tasks automatically
 taskManager=Task manager
 taskSaving=Saving task...
+taskScriptThread=Script task
 taskStatus=task status
 tasks=Tasks
 technicalData=Technical data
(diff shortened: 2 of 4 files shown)


Kitodo/src/main/resources/messages/messages_es.properties
```diff
@@ -1083,6 +1083,7 @@ taskManagerIsRunning=No iniciar las tareas automáticamente
 taskManagerIsStopped=Inicie las tareas automáticamente
 taskManager=Gestor de tareas
 taskSaving=La tarea está guardada...
+taskScriptThread=Aŭtomata tasko
```
solth, line 1086: Taking a look at the Spanish words for `task` in the previous and next lines of this file gives the impression that "tasko" may not be the best translation here. Since @danilopenagos kindly offered to contribute the Spanish translation for new messages, maybe you could just leave this line out?

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -1085,6 +1085,7 @@ taskManagerIsRunning=do not start tasks automatically
 taskManagerIsStopped=start tasks automatically
 taskManager=Task manager
 taskSaving=Saving task...
+taskScriptThread=Automatic task
```
solth, line 1088: See above.

Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -1085,6 +1085,7 @@ taskManagerIsStopped=Aufgaben automatisch starten
 taskManager=Taskmanager
 taskSaving=Aufgabe wird gespeichert...
 taskStatus=Aufgabenstatus
+taskScriptThread=Automatische Aufgabe
```
solth, line 1088: The translation of this key should either incorporate the information about the scirpt - so for example "Automatische Skriptaufgabe" - or the message key should be changed to better reflect the message itself (e.g. "automaticTask" or something similar). Either way we should not continue the questionable tradition of Kitodo to add messages where key and value or are so different.

solth (changes requested):

Kitodo/src/main/resources/messages/messages_es.properties
matthias-ronge: I clicked next to it and got Esperanto instead of Español. I think, should be: Tarea automática

Kitodo/src/main/resources/messages/messages_de.properties
matthias-ronge: Renamed `TaskScriptThread` to `AutomaticTasksThread`.

solth (approved):

PR #5783, opened 2023-09-13
Remove test for indexing

matthias-ronge:


matthias-ronge:
@solth apparently, you have to activate the branch `hibernate-search` at Codacy

solth:
> @solth apparently, you have to activate the branch `hibernate-search` at Codacy

done

matthias-ronge:
@solth The _CI / build_ has not been executed though. (I did hope, it would then.) Do you have to enable it as well, or do I need to recreate the pull requests now? Do you know or can you check?

henning-gerhardt:
CI will only be trigged in master branch as limited in https://github.com/kitodo/kitodo-production/blob/master/.github/workflows/main.yml#L9-L10 ... maybe removing this limit should fix it

solth:
@matthias-ronge I think @henning-gerhardt is right. Perhaps you can activate CI for your branch yourself by adding `hibernate-search` to that line in the .yml file? If not, I would suggest to open a separate PR to change this line in the main.yml file.

solth (commented):
Adding the branch name to the configuration seems to have had the intended effect, the tests were performaned for this pull request 👍 

I would suggest not to _remove_ the whole test, though, but instead just `@ignore` it until hibernate search is implemented and the mass indexer can be used to manually re-index the system.

PR #5794, opened 2023-09-29
Abort password change when old password is not correct

### LINKED ISSUE #5787: Misleading message while password change
**Describe the bug**
If a normal user (not an administrative user) is changing his own password two messages appear:
- one is saying that the old and the new password did not match
- second is saying that the password was successful changed

Using the old password afterwards is not possible any more and only the new password is accepted.

I don't know if this is happening only if the user is connected to LDAP service or even on database only password usage.

**To Reproduce**
1. Login as a regular user
2. Switch to the user properties
3. Change the password (entering old and twice new password) 
4. See the misleading error message

**Expected behavior**
Comparing that the old used password is correct and differ from the new password is right but the error message should not appear in this case. If the old password is incorrect than this message should appear and the password should not be changed.

**Screenshots**
Screenshot_2023-09-26_10-04-46

**Release**

solth:
Fixes #5787


solth:
Closed as updated PR exists in #5807

PR #5798, opened 2023-10-05
Remove unused statistic button

### LINKED ISSUE #5721: Statistic button in project list without functionality
**Describe the bug**
The statistic button 

missing_action

in the project list did not have any mouse hover description nor a functionality behind (see https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/webapp/WEB-INF/templates/includes/projects/projectList.xhtml#L147-L150)

**Expected behavior**
Add the missing functionality or remove the button 

**Release**

henning-gerhardt:
fixes #5721 

Decided to remove the button as needed functionality is not known.


solth:
Sorry for not merging this PR, yet. I am just hesitant because I think we should probably be able to add some useful functionality to this button with the existing functions (like automatically filtering the process list by project and then opening one of the already available statistic popups) which would perhaps be more useful than removing it altogether.

henning-gerhardt:
As I don't know who and when this missing functionality is added removing this not functional button is better then have this button and confused users. When this functionality is added back, adding back the button should be a easy task.

solth:
You are probably right.

PR #5800, opened 2023-10-05
Adjust metadata replacement

### LINKED ISSUE #5799: Meta data of newspaper issues can not be accessed on variable replacement
**Describe the bug**
We a calling a lot of scripts with variables which get there data from the meta data of a process. This did work so well for monographs and volumes of multi volumes or periodicals. But this did not work for newspaper like issues or other more deep structured media types.

Our script call is f.e.

 `/some/path/to/script -language ${meta.DocLanguage} -type ${meta.slub_script}` 

and this get replaced to f.e. 

`/some/path/to/script -language ger -type Antiqua`

but on  newspaper like issues it is only called 

`/some/path/to/script -language  -type` 

and the called script is failing because of the missing parameters.

**To Reproduce**
1. Have a script task which is using variable replacement from meta data
2. Execute the task on a newspaper issue and see how the script is called

**Expected behavior**
Even on newspaper like issues or similar deeper structured data the replacement of variables with data from meta data file should work.

**Release**

henning-gerhardt:
fixes #5799

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
@@ -397,6 +397,9 @@ public class VariableReplacer {
                 if (!allChildren.isEmpty()) {
                     allFirstchildValue = MetadataEditor.getMetadataValue(allChildren.get(0), variableFinder.group(5));
                     if (Objects.isNull(allFirstchildValue)) {
+                        allFirstchildValue = determineReplacementForTopstruct(variableFinder, dollarSignIfToKeep);
+                    }
+                    if (Objects.isNull(allFirstchildValue) || StringUtils.isEmpty(allFirstchildValue)) {
                         List<LogicalDivision> firstChildChildren = allChildren.get(0).getChildren();
                         if (!firstChildChildren.isEmpty()) {
                             allFirstchildValue = MetadataEditor.getMetadataValue(firstChildChildren.get(0), variableFinder.group(5));
--- a/Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
@@ -393,8 +393,17 @@ public class VariableReplacer {
         switch (metadataLevel) {
             case ALL:
                 List<LogicalDivision> allChildren = workpiece.getLogicalStructure().getChildren();
-                String allFirstchildValue = allChildren.isEmpty() ? null
-                        : MetadataEditor.getMetadataValue(allChildren.get(0), variableFinder.group(5));
+                String allFirstchildValue = null;
+                if (!allChildren.isEmpty()) {
+                    allFirstchildValue = MetadataEditor.getMetadataValue(allChildren.get(0), variableFinder.group(5));
+                    if (Objects.isNull(allFirstchildValue)) {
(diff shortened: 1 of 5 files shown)


matthias-ronge:
This fix is a functional change. Previously, with $meta.x variables only the first child was checked, and if this is empty, the variable is resolved on the top structure.

With this change, the first grandchild is now also taken into account, which would result in a different output value in other constellations:

```
📄 topstruct
  ├─ 📝 md: Foo
  └─ 📄 firstchild
      └─ 📄 firstgrandchild
          └─ 📝 md: Bar
```
Before: $(meta.md) → “Foo” (firstchild has no ‘md’, so resolve from topstruct)  
After: $(meta.md) → “Bar” (firstchild has no ‘md’, found on firstgrandchild, value from topstruct ignored)

Maybe fix like this:
```java
       if (!allChildren.isEmpty()) {
           allFirstchildValue = MetadataEditor.getMetadataValue(allChildren.get(0), variableFinder.group(5));
/*»*/      if (Objects.isNull(allFirstchildValue)) {                                                     
/*»*/          allFirstchildValue = determineReplacementForTopstruct(variableFinder, dollarSignIfToKeep);
/*»*/      }
           if (Objects.isNull(allFirstchildValue)) {
               List<LogicalDivision> firstChildChildren = allChildren.get(0).getChildren();
               if (!firstChildChildren.isEmpty()) {
                   allFirstchildValue = MetadataEditor.getMetadataValue(firstChildChildren.get(0), variableFinder.group(5));
               }
           }
       }
```

This should maintain the original behaviour, and fix your problem as well.

matthias-ronge (changes requested):
See comment above

henning-gerhardt:
My intention was to look first into the first child and then in the first grandchild and if both locations are not returning an result look into the top element. 

But your code change suggestion is not working without an other needed change: `determineReplacementForTopstruct` method is returning an empty string if in the top element not result is found (hidden by the `dollarSignIfToKeep` variable which is used as return value in not found case and the `dollarSignIfToKeep` variable is an empty string in many cases) so the following check condition must be extended to check for an empty string to get the right meta data.

Kitodo/src/main/java/org/kitodo/production/helper/VariableReplacer.java
```diff
@@ -387,6 +387,24 @@ private String determineReplacementForMetadata(Matcher variableFinder) {
...
 
+    private String determinateReplacementForAll(Matcher variableFinder, String dollarSignIfToKeep) {
+        List<LogicalDivision> allChildren = workpiece.getLogicalStructure().getChildren();
+        String allFirstchildValue = null;
+        if (!allChildren.isEmpty()) {
+            allFirstchildValue = MetadataEditor.getMetadataValue(allChildren.get(0), variableFinder.group(5));
+            if (Objects.isNull(allFirstchildValue)) {
+                allFirstchildValue = determineReplacementForTopstruct(variableFinder, dollarSignIfToKeep);
+            }
+            if (Objects.isNull(allFirstchildValue) || StringUtils.isEmpty(allFirstchildValue)) {
```
matthias-ronge, line 398: [`StringUtils.isEmpty()`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty-java.lang.CharSequence-) implies a `null` check, so you can remove the `Objects.isNull()` check

matthias-ronge (changes requested):
Just one more little thing.

matthias-ronge (approved):

PR #5803, opened 2023-10-05
Visualisation of audio as a waveform

markusweigelt:
This PR adds a waveform to an audio element in the media view and includes tools for navigating within the waveform e.g. zoom, checkbox for centered cursor or forward and backward jumping.

The waveform in its default state without zoom.
image

This is what it looks like when something has been zoomed into the waveform.
image

In the project settings, the waveform must be enabled for this. This setting is only displayed if an audio folder has also been configured. By default, the waveform is disabled.
image

The [Wavesurfer](https://wavesurfer.xyz) library is used for drawing the waveform. The library is well-supported and has been around for a few years. Additionally, the developer is very responsive to issues or inquiries.

During the development, I encountered the following problem, which has now also been included in the [Wavesurfer FAQ](https://github.com/katspaugh/wavesurfer.js#faq).

> **There is a mismatch between my audio and the waveform. How do I fix it?**
> If you're using a VBR (variable bit rate) audio file, there might be a mismatch between the audio and the waveform. This can be fixed by converting your file to CBR (constant bit rate). See https://github.com/katspaugh/wavesurfer.js/issues/2890#issuecomment-1601067822 for details.

Another potential issue in the future could be that the files become too large. However, there is an option to pre-generate the peaks of the waveform. Evaluating these pre-generated files would need to be implemented as needed and is not part of this PR.

> **Does wavesurfer support large files?**
> Since wavesurfer decodes audio entirely in the browser using Web Audio, large clips may fail to decode due to memory constraints. We recommend using pre-decoded peaks for large files (see [this example](https://wavesurfer.xyz/examples/?predecoded.js)). You can use a tool like [bbc/audiowaveform](https://github.com/bbc/audiowaveform) to generate peaks.

During development, I worked with an audio file that was approximately 30 minutes long and 30 MB in size. There were no issues regarding the length of the file. Additionally, in the current implementation, the peaks are cached on the client-side for use in the metadata editor.


henning-gerhardt:
Good addition to the project. 

Some remarks:
- Why is a configuration flag needed in the project settings? Should displaying the waveform not enough by the used mime type of the file?
- Waveviewer/Wavesurfer is licensed under condition as BSD-3-Clause license. This license should be added to the repo or? Maybe even add a new file which contains which file from "outside" is licensed under which condition and where to found is (project home page or something similar).

markusweigelt:
> Good addition to the project.
> 
> Some remarks:
> 
> * Why is a configuration flag needed in the project settings? Displaying this or not could not be determinated by the mime type of the file?
> * Waveviewer/Wavesurfer is licensed under condition as BSD-3-Clause license. This license should be added to the repo or? Maybe even add a new file which contains which file from "outside" is licensed under which condition and where to found is (project home page or something similar).

markusweigelt:
> * Why is a configuration flag needed in the project settings? Displaying this or not could not be determinated by the mime type of the file?

The configuration makes sense if only the audio files should be captured but no waveform is required, e.g.  chapters or sequences should not be captured.

> * Waveviewer/Wavesurfer is licensed under condition as BSD-3-Clause license. This license should be added to the repo or? Maybe even add a new file which contains which file from "outside" is licensed under which condition and where to found is (project home page or something similar).

Yes, that definitely needs to be done. Currently, there are no direct templates available. I will take a look at how other repositories handle this, or do you have a template that we can use as a reference?

markusweigelt:
> > * Waveviewer/Wavesurfer is licensed under condition as BSD-3-Clause license. This license should be added to the repo or? Maybe even add a new file which contains which file from "outside" is licensed under which condition and where to found is (project home page or something similar).
> 
> Yes, that definitely needs to be done. Currently, there are no direct templates available. I will take a look at how other repositories handle this, or do you have a template that we can use as a reference?

I have added the LICENSE file directly to the "lib/wavesurfer" folder. According to my understanding of this license, that should be sufficient.

In the future, you could potentially consider using a JavaScript package manager for these libraries. This would require adding an additional build step for downloading the libraries. With this approach, you could easily manage the libraries and their versions.

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/partials/media-detail.xhtml
```diff
@@ -16,14 +16,45 @@
...
+            <p:outputPanel rendered="#{fn:startsWith(selectedGalleryMediaContent.mediaViewMimeType, 'audio') and DataEditorForm.galleryPanel.isAudioMediaViewWaveform()}">
+                <div id="audioWaveformTools" >
+                    <div>
+                        <label><h:outputText value="#{msgs.audioWaveformToolsZoom}"/>: <input type="range" min="0" max="250" value="0"/></label>
+                    </div>
+                    <div>
+                        <label><h:outputText value="#{msgs.audioWaveformToolsCenteredCursor}"/>: <input type="checkbox" checked="" value="autoCenter"/></label>
+                    </div>
+                    <div>
+                        <p:commandButton title="#{msgs.audioWaveformToolsJumpBackFiveSeconds}"
```
oliver-stoehr, line 38: These buttons should have the attribute `type="button"`, when they only trigger client side actions. Otherwise they will submit their form and cause an unnecessary update.

Kitodo-DataManagement/src/main/resources/db/migration/V2_125__Add_audio_media_view_waveform_to_project.sql
```diff
@@ -0,0 +1,14 @@
...
+--
+-- It is licensed under GNU General Public License version 3 or later.
+--
+-- For the full copyright and license information, please read the
+-- GPL3-License.txt file that was distributed with this source code.
+--
+
+--
+-- Migration: Add column for state of audio waveform in media view to project table.
+ALTER TABLE project ADD mediaView_audio_waveform TINYINT(1) NOT NULL DEFAULT 0
```
oliver-stoehr, line 14: Missing semicolon at the end of line

Kitodo/src/main/webapp/WEB-INF/resources/js/media_detail_audio_waveform.js
```diff
@@ -0,0 +1,119 @@
...
+               self.#buildTimeout = setTimeout(function() {
+                    self.#build();
+               }, 500);
+            }, {once: true});
+
+        }
+    }
+
+    #build() {
+        let self = this
```
oliver-stoehr, line 50: Missing semicolon

Kitodo/src/main/webapp/WEB-INF/resources/js/media_detail_audio_waveform.js
```diff
@@ -0,0 +1,119 @@
...
+            waveToolsSlider.addEventListener('input', (e) => {
+                const minPxPerSec = e.target.valueAsNumber;
+                self.#wavesurfer.zoom(minPxPerSec);
+            });
+
+            waveToolsContainer.querySelectorAll('input[type="checkbox"]').forEach((input) => {
+                input.onchange = (e) => {
+                    self.#wavesurfer.setOptions({
+                        [input.value]: e.target.checked,
+                    })
```
oliver-stoehr, line 99: Missing semicolon

Kitodo/src/main/webapp/WEB-INF/resources/js/media_detail_audio_waveform.js
```diff
@@ -0,0 +1,119 @@
...
+                }
+            })
```
oliver-stoehr, line 101: Missing semicolon

oliver-stoehr (changes requested):
I found only a few minor things mainly regarding code style. Functionality is fine!

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/partials/media-detail.xhtml
markusweigelt: Nice! It had worked for me as well without type by manually stopping the propagation. Now I've added the type and removed the stopping. Thanks for the suggestion.

oliver-stoehr (approved):

PR #5804, opened 2023-10-06
Add security policy

solth:
Add security policy with information about supported versions of Kitodo.Production. 

As discussed with @kitodo/kitodo-board only the version which is currently in development - 3.7 right now - will receive active development while the last release (3.6 at the moment of this writing) will get security fixes.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+The following versions of Kitodo.Production are currently being supported with ongoing feature development and/or security updates.
+
+| Version | Active Development       |      Security Fixes      |
+|-----| :----------------------: |:------------------------:|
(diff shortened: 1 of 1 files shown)


solth:
@stefanCCS thanks for the review. The e-mail address (and the whole template) is more or less the same as in Kitodo.Presentation (just a little simplified because of one less column), see https://github.com/kitodo/kitodo-presentation/blob/master/SECURITY.md 

I neither know whether the e-mail address actually exists nor who reads it but I assumed it would be the appropriate address since it's already in use for presentation. Perhaps @sebastian-meyer can answer this question?

The check marks (and "multiplication x") are not displayed locally because they are Github emojis that only work on this platform (see this cheatsheet for details: https://github.com/ikatyang/emoji-cheat-sheet/blob/master/README.md)

stefanCCS (approved):

PR #5809, opened 2023-10-11
OCR-D Workflow integration (without entity implementation)

markusweigelt:
This pull request re-implements the PR #5697 without the use of an explicit OCR-D workflow entity. 

The workflows are currently selected from a directory that can be configured via the `kitodo_config.properties`. In the future, it should be possible to load OCR-D workflows through an OCR-D manager endpoint. This PR provides the preparation for this by using `Pair` objects to combine ID and name instead of using a simple string list.

- Add new authority to assign OCR-D Workflow 
- OCR-D workflow directory can be configured using the `ocrd.workflows.directory` parameter
- Choose the OCR-D workflow from the workflows directory in the process template, or replace it within the process details.
- Incorporate the OCR-D workflow identifier as a shell script variable using the `(ocrdworkflowid)` placeholder
- Added documentation among others for the placeholder https://github.com/kitodo/kitodo-production/wiki/Workflow-Editor#platzhalter


Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -888,6 +890,7 @@ position=Position
 ppn=PPN
 priority=Priorit\u00E4t
 process=Vorgang
+processConfig.useProcessTemplateOcrdWorkflow=aus Produktionsvorlage
```
solth, line 893: I am not sure this is a good message key: the value does not contain any connection to OCR-D workflows, so the message key should probably reflect this? Perhaps something like `processConfig.fromTemplate` would be more appropriate? (we already have too many message keys were keys and values do not really match and I think when introducing new message keys we should try to do better)

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -889,6 +891,7 @@ position=Position
 ppn=PPN
 priority=Priority
 process=Process
+processConfig.useProcessTemplateOcrdWorkflow=From process template
```
solth, line 894: see above

Kitodo/src/main/webapp/WEB-INF/templates/includes/processEdit/details.xhtml
```diff
@@ -46,6 +47,12 @@
                     <p:ajax event="change" oncomplete="toggleSave()"/>
                 </p:selectOneMenu>
             </div>
+
+            <div jsf:rendered="#{SecurityAccessController.hasAuthorityToAssignOcrdWorkflow()}">
```
solth, line 51: I didn't know JSF attributes can be used in standard html tags. Very interesting! 👍

solth (changes requested):
Code looks good, just some very minor change suggestions. (Please rebase your branch against the current master.)

This is just a code review. Functional review will follow.

Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -71,6 +71,9 @@ directory.modules=/usr/local/kitodo/modules/
 # Falls Dateien zum Debuggen / Tracen geschrieben werden sollen, hier ein Verzeichnis angeben
 directory.debug=/usr/local/kitodo/debug/
 
+# Absolute path to the directory that the OCR-D workflow files will be
+# read from. It must be terminated by a directory separator ("/").
+ocrd.workflows.directory=/usr/local/kitodo/ocrd/workflows/
```
solth, line 76: Are there going to be other OCRD files apart from workflows in the future? Otherwise I am a little confused about this nested folder structure - why not use just `ocrd_worfklows` or something like that?

Kitodo/src/main/java/org/kitodo/production/services/ocr/OcrdWorkflowService.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Get list of OCR-D workflows from directory.
+     * <p>
+     * The files are relative paths to the OCR-D workflow directory.
+     * </p>
+     *
+     * @return list of OCR-D workflows
+     */
+    public List<Pair> getOcrdWorkflows() {
```
solth, line 64: My IDE tells me raw use of parameterized class `Pair` is discouraged. Perhaps you could replace this and similar other parts with `List<Pair<?, ?>`?

Kitodo/src/main/java/org/kitodo/production/services/ocr/OcrdWorkflowService.java
```diff
@@ -0,0 +1,102 @@
...
+                            .collect(Collectors.toList());
+                } catch (IOException e) {
+                    logger.error(e.getMessage(), e);
+                }
+            }
+        }
+        return new ArrayList<>();
+    }
+
+    private static Pair getImmutablePairFromPath(Path filePath, Path ocrdWorkflowsDirectory) {
```
solth, line 81: See above.

Kitodo/src/main/java/org/kitodo/production/services/ocr/OcrdWorkflowService.java
```diff
@@ -0,0 +1,102 @@
...
+        return ImmutablePair.of(path, path);
+    }
+
+    /**
+     * Get an OCR-D workflow by identifier.
+     *
+     * @param ocrdWorkflowId The OCR-D workflow identifier
+     * @return The OCR-D workflow
+     */
+    public Pair getOcrdWorkflow(String ocrdWorkflowId) {
```
solth, line 94: See above.

Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -874,6 +876,38 @@ public List<Project> getProjects() {
         return ServiceManager.getProjectService().getAllForSelectedClient();
     }
 
+    /**
+     * Get list of OCR-D workflows for select list.
+     *
+     * @return list of OCR-D workflows
+     */
+    public List<Pair> getOcrdWorkflows() {
```
solth, line 884: See below.

Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -874,6 +876,38 @@ public List<Project> getProjects() {
...
+        return ServiceManager.getOcrdWorkflowService().getOcrdWorkflows();
+    }
+
+    /**
+     * Get the OCR-D workflow.
+     *
+     * @return Immutable key value pair
+     */
+    public Pair getOcrdWorkflow() {
```
solth, line 893: See below.

Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -874,6 +876,38 @@ public List<Project> getProjects() {
...
+        return ServiceManager.getOcrdWorkflowService().getOcrdWorkflow(process.getOcrdWorkflowId());
+    }
+
+    /**
+     * Set the OCR-D workflow.
+     *
+     * @param ocrdWorkflow
+     *         The immutable key value pair
+     */
+    public void setOcrdWorkflow(Pair ocrdWorkflow) {
```
solth, line 903: See below.

Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
```diff
@@ -274,6 +275,38 @@ public List<Workflow> getWorkflows() {
         return ServiceManager.getWorkflowService().getAvailableWorkflows();
     }
 
+    /**
+     * Get list of OCR-D workflows for select list.
+     *
+     * @return list of OCR-D workflows
+     */
+    public List<Pair> getOcrdWorkflows() {
```
solth, line 283: See below.

Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
```diff
@@ -274,6 +275,38 @@ public List<Workflow> getWorkflows() {
...
+        return ServiceManager.getOcrdWorkflowService().getOcrdWorkflows();
+    }
+
+    /**
+     * Get the OCR-D workflow.
+     *
+     * @return Immutable key value pair
+     */
+    public Pair getOcrdWorkflow() {
```
solth, line 292: See below.

Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
```diff
@@ -274,6 +275,38 @@ public List<Workflow> getWorkflows() {
...
+        return ServiceManager.getOcrdWorkflowService().getOcrdWorkflow(template.getOcrdWorkflowId());
+    }
+
+    /**
+     * Set the OCR-D workflow.
+     *
+     * @param ocrdWorkflow
+     *         The immutable key value pair
+     */
+    public void setOcrdWorkflow(Pair ocrdWorkflow) {
```
solth, line 302: See below.

solth (commented):
Code seems to work as intended. (Just some small comments - see below). I was able to configure a OCRD workflow for a template and later change it for individual processes. 

Just one small remark I think could be improved upon: 
Bildschirmfoto 2023-11-27 um 11 19 17
I think the `from Template` option should contain information about _which_ OCRD workflow it is that will be derived from the template, otherwise this might not be very informative (a user would first have to check the template to see if it defines a workflow appropriate for the current process and whether he has to change something in the process or not)

Also, there is a conflict in `ProcessForm.java` that needs to be resolved.

Kitodo/src/main/resources/kitodo_config.properties
markusweigelt: Currently, no additional files are planned. The path is adjustable, but I have now changed the default to your suggestion.

markusweigelt:
> I think the from Template option should contain information about which OCRD workflow it is that will be derived from the template, otherwise this might not be very informative (a user would first have to check the template to see if it defines a workflow appropriate for the current process and whether he has to change something in the process or not)

Good point! I have now added the selected file from the template. If there is no selection in the project template, the message of no selection will be shown.

solth (approved):

PR #5810, opened 2023-10-11
Filtering of single and multiple select fields

markusweigelt:
In the last conversation about a project, it was noted that the more options assigned to a key, the more cumbersome it becomes to find the right option.

PrimeFaces 8 already provides an implementation for making the select field searchable/filterable.
https://primefaces.github.io/primefaces/8_0/#/components/selectonemenu?id=filtering
https://primefaces.github.io/primefaces/8_0/#/components/selectmanymenu?id=filtering

I have made this configurable for select fields through the `filterable` attribute in the rule set. In the settings of a key, you can now set the attribute to `filterable="true"`. This will display the following filter:

filter-select

By default, the filter is not displayed and must be explicitly activated. After activation, the filter is displayed, and it searches for all entries that begin with the input.

There are also other Primeface filter modes that allow the entered search term to match not only at the beginning but also at the end or anywhere within the option being searched for. Additionally, you can set a custom filter mode.

If needed, these filter modes can also be made configurable, or the default filter mode of "only at the beginning" can be changed to "anywhere." It depends on which option is more prevalent.

### DIFF
--- a/Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/KeyView.java
+++ b/Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/KeyView.java
@@ -154,6 +154,15 @@ class KeyView extends AbstractKeyView<KeyDeclaration> implements DatesSimpleMeta
         return settings.isEditable(declaration.getId());
     }
 
+    @Override
+    public boolean isFilterable() {
+        InputType inputType = getInputType();
+        if (InputType.MULTIPLE_SELECTION.equals(inputType) || InputType.ONE_LINE_SINGLE_SELECTION.equals(inputType)) {
+            return settings.isFilterable(declaration.getId());
+        }
+        return false;
+    }
+
     /**
      * Checks if a URI is in the configured namespace, if one has been
      * specified. Typically, a namespace is used as a URL prefix. For namespaces
--- a/Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/dataeditor/rulesetmanagement/SimpleMetadataViewInterface.java
@@ -118,6 +118,13 @@ public interface SimpleMetadataViewInterface extends MetadataViewInterface {
     boolean isEditable();
 
     /**
+     * Returns whether values under this key can be filtered in this view.
+     *
+     * @return whether values can be filtered
+     */
+    boolean isFilterable();
+
+    /**
      * Returns whether the value corresponds to the value range. The value range
      * can be determined in various ways. Integers or dates must parse, it may
      * be that the value must be in a list or is checked against a regular
--- a/Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/Settings.java
+++ b/Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/Settings.java
@@ -107,6 +107,22 @@ public class Settings {
     }
 
     /**
+     * Whether the key is filterable.
+     *
+     * @param keyId
+     *            key for which the query is
+     * @return whether the key is filterable
+     */
+    boolean isFilterable(String keyId) {
+        if (currentSettings.containsKey(keyId)) {
+            return currentSettings.get(keyId).isFilterable();
+        } else {
+            return true;
+        }
+    }
+
+
+    /**
      * Whether the key is excluded.
      *
      * @param keyId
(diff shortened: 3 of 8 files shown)


andre-hohmann (changes requested):
@markusweigelt : Thanks for the PR! That will help a lot to handle the missing project-specific values and roles of persons and corporations. 

From the users' perspective, it works. I would prefer to apply the filter mode `contains` instead of `startsWith`. 

For example: If the collection "Handschriften" is assigned to an object, another collection like "Französische Handschriften", "Deutsche mittelalterliche Handschriften", ... is assigned, too. It would be convenient, if all collections, which contain "Handschriften" are displayed. 

Additionally, we use prefixes like "Projekt: ..." or "LDP: ..." to distinct collections. These prefixes must be known when appliying `startsWith`, which becomes difficult the more collections are configured. 

Would it be possible to make it configurable by applying the attributes `filterable="contains"`, `filterable="contains"`, ... or must it be configured in the code/backend?

markusweigelt:
@andre-hohmann I changed the default filter mode to 'contains.' I believe it makes the most sense this way. If there is a need, the modes could be made configurable. Or do you think the need already exists here?

matthias-ronge:
> Are there currently fields where it makes sense to check for `startsWith` or `endsWith`?

Ruleset already supports `<pattern>` matching

markusweigelt:
> > Are there currently fields where it makes sense to check for `startsWith` or `endsWith`?
> 
> Ruleset already supports `<pattern>` matching

As I understand the pattern, it is only applicable for validation, but not for filtering on the client side. Correct me if I'm wrong.

matthias-ronge:
The check happens inside Tomcat, it’s not Javascript. When I wrote it, it should be that the metadata editor doesn’t even allow the value, but I think it was changed so these checks only apply during validation. I don’t know why it was changed

markusweigelt:
However, this pull request only concerns the client-side and the filtering of the results displayed by the application. I can't imagine right now how the pattern before validation would help. Or, in principle, does it essentially do what the current implementation does?

IkramMaalej (changes requested):

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/KeyView.java
```diff
@@ -157,7 +157,8 @@ public boolean isEditable() {
     @Override
     public boolean isFilterable() {
         InputType inputType = getInputType();
-        if (InputType.MULTIPLE_SELECTION.equals(inputType) || InputType.ONE_LINE_SINGLE_SELECTION.equals(inputType)) {
+        if (InputType.MULTIPLE_SELECTION.equals(inputType) || InputType.MULTI_LINE_SINGLE_SELECTION.equals(
```
IkramMaalej, line 160: Sorry, I think this edit doesn't make sense because `MULTI_LINE_SINGLE_SELECTION` maps to an input type `selectOneRadio` which cannot be filtered (`selectOneRadio` does not support the "filter" and "filterMatchMode" attributes).

```
case MULTI_LINE_SINGLE_SELECTION:
                return "oneRadio";
```

matthias-ronge (approved):

BartChris:
Also not directly related to the pull request but since this is about usabilty. It is great that this PR allows us to filter a long list, but unfortunately the editor does not immediatly see which values are already selected if the selected values are at the end of the list.  Has it been considered to use https://www.primefaces.org/showcase-v8/ui/input/checkboxMenu.xhtml which seem to be a drop-in-replacement if i am not mistaken.

markusweigelt:
@BartChris The English translation of the German text is: "The design decisions that have been made and why these components were chosen are beyond my knowledge. It is true that it's about user-friendliness, but only for a small aspect of filtering existing lists. The mentioned component does not have such filtering, so I believe the context is not applicable here and should be discussed or submitted elsewhere, for example, in the community board.

IkramMaalej (approved):

Kitodo-DataEditor/src/main/java/org/kitodo/dataeditor/ruleset/KeyView.java
markusweigelt: Yes, there was a reason. I vaguely remember checking on that. During the review, I didn't find it right away, and I assumed I had simply forgotten it.

stweil:
@markusweigelt, @solth, there is a bug report on the community mailing list:
```
wir haben die neue Version Kitodo Production 3.7.0 bei uns installiert. Nun erscheint bei Aufruf des Metadateneditors sowie beim Anlegen von Vorgängen folgende Fehlermeldung:

Exception message: Receiver class org.kitodo.dataeditor.ruleset.KeyView does not define or inherit an implementation of the resolved method 'abstract boolean isFilterable()' of interface org.kitodo.api.dataeditor.rulesetmanagement.SimpleMetadataViewInterface.
```

Is this related to the PR?

markusweigelt:
Currently, I am no longer involved in the project, so I can only make a remote diagnosis. 

It sounds like that. We use a cherry-pick approach for these changes in our SLUB because, for certain reasons, we do not use the current version but still need this feature. However, it was tested and developed against the state at that time. It is not out of the question that there is a bug here. Were the modules deleted or updated beforehand to ensure that the new JARs are being loaded?

solth:
> Were the modules deleted or updated beforehand to ensure that the new module JAR's are being loaded?

That was my first thought as well after reading the mail on the community list, since old/outdated modules have often been the cause of errors after the application has been updated in the past. I will write an corresponding answer.

solth:
Ok, @matthias-ronge beat me to it!

PR #5811, opened 2023-10-13
Always reset password input fields when opening 'change password' dialog

### LINKED ISSUE #5808: Password change dialog keeps data
**Describe the bug**
If the password change dialog is called a second time the old entered values are still available. This is happening in the administrative and in the user call of this dialog.

Administrative call:
Screenshot_2023-10-10_13-55-55

User call:

Screenshot_2023-10-10_13-56-41

**To Reproduce**
1. Go to the password dialog
2. Enter the password successful
3. Re-open the password dialog

**Expected behavior**
Opening the password change dialog should start in any case with no values entered.

**Release**

solth:
Fixes #5808


henning-gerhardt (approved):
With this changes the old entered passwords are not visible / accessible any more.

Issue #3762, opened 2020-06-10
Metadata editor crashes with metadata group not defined in the ruleset


matthias-ronge:
**Edit:** _General explaination:_ If the ruleset module gets a key ID that isn’t defined in the ruleset, it considers it as a text metadata. This goes back to days where the reading was that the modules should be independent of each other, and thus the ruleset module only gets the ID of a key, but didn’t have its class on its classpath, and so couldn’t check if it is complex. So, it returns a text box to be rendered, and the JSP fails to put a metadata group into the text string of the text box.

The assumption behind this behaviour was that, if ever there is an entry in the metadata which is missing in the ruleset, this is likely to be a value; because metadata groups are unlikely to appear “by accident”, but maybe an entry does, if it was removed from the ruleset in belief that it had never been used, but it was used in one single process that the editor of the ruleset didn’t know about. Unlike 2.x, this shouldn’t crash the metadata editor anymore, which it did in thie past.

----

Trying to start the metadata editor with a migrated process. Thumbnail images have been generated. Structure tee and metadata panel are already showing, gallery not yet. Then the page is reset and gives this error:

Date/time: 2020-06-10 14:43:53  
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:77.0) Gecko/20100101 Firefox/77.0  
User IP: 192.168.91.1  
Request URI: /kitodo/pages/metadataEditor.jsf  
Ajax request: Yes  
Status code: 500  
Exception type: class javax.faces.application.ViewExpiredException  
Exception message: View "/pages/metadataEditor.jsf" could not be restored.  
Exception UUID:  
Stack trace:

```
    javax.faces.application.ViewExpiredException: View "/pages/metadataEditor.jsf" could not be restored.
    	at org.apache.myfaces.lifecycle.RestoreViewExecutor.execute(RestoreViewExecutor.java:181)
    	at org.apache.myfaces.lifecycle.LifecycleImpl.executePhase(LifecycleImpl.java:195)
    	at org.apache.myfaces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:142)
    	at javax.faces.webapp.FacesServlet.service(FacesServlet.java:204)
    	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:292)
    	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:207)
[...]

PR #5831, opened 2023-11-03
Prevent NPE

matthias-ronge:
Do not crash the building of error messages it an argument is `null`.

The message building checks the result string, if all `insertion`s have been inserted in the string. If not, they are appended. Example:
“Metadata entry {0} not allowed for division {1}” ["TitleDocMain", "ephemera", "ruleset_default", "IOException: Cannot read ruleset_default.xml"] → “Metadata TitleDocMain not allowed for division ephemera: ruleset_default: IOException: Cannot read ruleset_default.xml”.

If an `insertion` is `null`, the test `if (!messageBuilder.toString().contains(insertion))` (next line) will throw a `NullPointerException`. Error message building shouldn’t throw an exception by itself.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/Helper.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/Helper.java
@@ -490,6 +490,7 @@ public class Helper {
         StringBuilder messageBuilder = new StringBuilder(message);
         for (String insertion : insertions) {
             String separator = ": ";
+            insertion = Objects.toString(insertion);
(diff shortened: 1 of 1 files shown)


solth:
Does this fix or is related to #3762 or #5217 ? Could you provide some description?

matthias-ronge:
> Does this fix or is related to #3762 or #5217 ?
No.

> Could you provide some description?
See above

solth (approved):

PR #5834, opened 2023-11-09
Use type track for audio and video MIME types instead of type page for physical division type

markusweigelt:
- Use track type for audio and video MIME types
- Add MediaUtils class to handle media specific functionalities
- Tests for these functionalities
- Add supported video and audio formats to `kitodo_fileFormats.xml`


solth (commented):
Code looks good, just one very minor point about import order in one file, if I am nitpicking.

PR #5842, opened 2023-11-27
Complete DFG-Viewer ruleset's XSLT file

matthias-ronge:
So far, the repository only contain a placeholder XSLT file for the DFG-Viewer ruleset. This pull request replaces that with the full XSLT file.

_I don't remember the origin of this file, I picked this up and cleaned it up a bit. Glory to whom the glory belongs._

### DIFF
--- a/Kitodo/src/main/resources/xslt/dfg-viewer.xsl
+++ b/Kitodo/src/main/resources/xslt/dfg-viewer.xsl
@@ -11,13 +11,136 @@
  * GPL3-License.txt file that was distributed with this source code.
  *
 -->
-<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+<xsl:stylesheet version="2.0"
+                xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/resources/xslt/dfg-viewer.xsl
```diff
@@ -11,13 +11,136 @@
  * GPL3-License.txt file that was distributed with this source code.
  *
 -->
-<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+<xsl:stylesheet version="2.0"
+                xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+                xmlns:mets="http://www.loc.gov/METS/"
+                xmlns:xlink="http://www.w3.org/1999/xlink"
```
solth, line 17:
```suggestion
```
XLink namespace is not used in this xslt file and can be removed from the declaration.

solth (changes requested):
Thanks for the pull request. I think it's a very good idea to have instantly useful rulesets and mapping files in the repository! Just one small change request, otherwise this looks very good.

solth (approved):

PR #5845, opened 2023-11-30
Display tooltip instead of the overlay in preview 

markusweigelt:
- The tooltip can be enabled through the project settings. By default, the overlay will still be displayed.

image

- The tooltip is only displayed when an image type has been selected for the preview view.
- It´s configurable that the image of media view is rendered in the tooltip (Hence, a preloading mechanism has been implemented to load the media view image only when the tooltip is about to be displayed) 
- The size of the tooltip depends on its content. Images are restricted to a maximum size of 40% of the viewport height.
-  Improve userinterface of folder use to add config for tooltip in project settings.

**Without tooltip**

image

image

**With tooltip**

image

image


solth (commented):
Looks good and works as advertised! 

I do have one remark/suggestion about the configuration, though. If I am not mistaken you end up with 3 possible options that can be configured - "overlay", "overlay + preview tooltip", "overlay + media view tooltip".

You use _two_ switches to control these three options of the thumbnail behaviour on mouse hover, though:

**1. Just overlay (old behaviour):**
Bildschirmfoto 2024-01-08 um 12 46 00

**2. Overlay + Preview Tooltip:**
Bildschirmfoto 2024-01-08 um 12 46 05

**3. Overlay + Media View Tooltip:**
Bildschirmfoto 2024-01-08 um 12 45 54

**4. ???:**
Bildschirmfoto 2024-01-08 um 12 03 31

So to me it seems this fourth switch setting either represents an invalid configuration or one that is already covered by a different state of the switches (probably the same as option 3, "Overlay + Media View Tooltip").

Therefore I think it would be better to have just _one_ configuration element - a pulldown menu with three options or a set of radio buttons, for example - instead of the _two_ separate switches to avoid confusion. 

Alternatively, you could keep the switches but deactivate the second switch until the first is set to "true". This seems to be more of a hassle, though, since you would also have to potentially update the value of the second switch (to "false") again when the first is set to "false" again.

markusweigelt:
@solth Thank you for the feedback. As suggested, the hover behavior in the preview will now be configured through a selection field.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/enums/PreviewHoverMode.java
```diff
@@ -0,0 +1,21 @@
...
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.data.database.enums;
+
+public enum PreviewHoverMode {
+
```
solth, line 15: Could you add some short description comment to the individual options in this enum?

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/partials/media-list-overlay.xhtml
```diff
@@ -13,19 +13,67 @@
...
-        <h:outputText rendered="#{media.type eq 'AUDIO'}">
-            #{msgs.audio} #{media.shortId}
-        </h:outputText>
-        <h:outputText rendered="#{media.type eq 'IMAGE'}">
-            #{msgs.image} #{media.shortId}, #{msgs.page} #{media.orderlabel}
-        </h:outputText>
-    </p:outputPanel>
+        xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
+        xmlns:a="http://xmlns.jcp.org/jsf/passthrough">
+
```
solth, line 21:
```suggestion
    <!--@elvariable id="media" type="org.kitodo.production.forms.dataeditor.GalleryMediaContent"-->
```
Adding this comment helps the IDE evaluating the properties of this EL variable.

solth (commented):
Looks very good now! Just two more very tiny suggestions, if you don't mind!

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/enums/PreviewHoverMode.java
markusweigelt: Now documented.

solth (approved):

PR #5848, opened 2023-12-05
Change version to 3.6.1

solth:


henning-gerhardt (approved):
Changes looks good. Module and war file build as release 3.6.1. Did not check the application itself as we behind the current 3.6.0 version.

PR #5863, opened 2023-12-22
Handle potential ConfigException during catalogue configuration import

### LINKED ISSUE #5862: Unhandled `ConfigException` when trying to convert missing `kitodo_opac.xml` to new `ImportConfigurations`
**Describe the bug**
When trying to convert the entries of the XML configuration file `kitodo_opac.xml` to new instances of type `Import Configuration`, an unhandled `ConfigException` is thrown. 

**To Reproduce**
1. Make sure no `kitodo_opac.xml` file exists
2. Navigate to `Projects` page and there to the `Import configurations` tab
3. Click the `Import catalogue configurations` button

The reason for this seems to be that the corresponding method (`startCatalogConfigurationImport`) only catches one type of possible exception (`org.apache.commons.configuration.ConfigurationException`) but not the other (`org.kitodo.exceptions.ConfigException`)

**Expected behavior**
If no file `kitodo_opac.xml` is found when clicking the button mentioned above, the resulting exception should be handled appropriately instead of showing its stack trace.

**Screenshots**
Bildschirmfoto 2023-12-22 um 15 06 40

**Release**
3.6.1

solth:
Fixes #5862 by handling `ConfigException` in addition to `ConfigurationException`.

The reason for both exception being handled in separate `catch` blocks is that the `ConfigurationException` is thrown when syntactical errors in the XML file are encountered and this "cause" is then also displayed (in addition to the exception message itself): 
Bildschirmfoto 2023-12-22 um 16 04 56

The `ConfigException` on the other hand is thrown when the whole XML file is missing. In this case, the "cause" of the exception is identical to the exception message itself, so there is no need to show both of them:
Bildschirmfoto 2023-12-22 um 16 03 53

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ImportConfigurationListView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ImportConfigurationListView.java
@@ -24,6 +24,7 @@ import org.apache.logging.log4j.Logger;
 import org.kitodo.config.OPACConfig;
 import org.kitodo.data.database.beans.ImportConfiguration;
 import org.kitodo.data.database.exceptions.DAOException;
+import org.kitodo.exceptions.ConfigException;
 import org.kitodo.exceptions.ImportConfigurationInUseException;
 import org.kitodo.production.enums.ObjectType;
 import org.kitodo.production.helper.Helper;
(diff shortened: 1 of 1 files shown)


matthias-ronge:
It is one of the oddities of Java, that if an exception doesn’t have a _cause_, the cause is set to `this`, not `null`. This is why you experience it that way.

matthias-ronge (approved):

PR #5895, opened 2024-01-25
[3.6.x] Catch Exceptions in indexing processes, log warning and continue

matthias-ronge:
If the indexing has a problem reading a METS file, you will get a condition like depicted below, and there is nothing you can do. There is also no additional information displayed as to what is causing this:

Screenshot

Cases in which the index becomes corrupt in a productive environment have become rare, fortunately! But having to rebuild the index, for whatever reason, is associated with downtime in the production environment during which colleagues cannot work. Therefore, I have followed the procedure here when the METS file is missing: An error is now logged and indexing continues for this process without METS data. The error is logged at `WARN` level, and you can take care of this process manually later.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -71,6 +71,7 @@ import javax.xml.xpath.XPathFactory;
 
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.io.IOUtils;
+import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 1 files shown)


solth (approved):
It works and the code looks good. 

We do loose the information in the frontend that there is a problem with one (or more) of the processes during indexing, though, but I think that is an acceptable tradeoff for now. 

I wonder if in a follow-up improvement we could somehow _keep_ the red x or some other indicator that informs the user/admin to check the log for details. (I would also like to see the detailed exception message in the log; with the change in this pull request it is masked by a generic `Cannot read metadata file for indexing` message that makes it a little more difficult to resolve potential issues)

But as I said that can be done in a separate pull request and I think the error handling here is a great improvement. Thanks @matthias-ronge

PR #5897, opened 2024-01-25
[3.6] password change fix

solth:
See #5807


henning-gerhardt (approved):
Changes looks correct ported.

PR #5899, opened 2024-01-25
[3.6] Relocate mysql dependency

solth:
See #5849


markusweigelt:
I can only assess it to a limited extent but i think if the test passes, there should be no problem

markusweigelt (approved):

PR #5900, opened 2024-01-25
[3.6.] Updating Bouncy Castle dependency

solth:
See #5867


markusweigelt (approved):
I can only assess it to a limited extent but i think if the test passes, there should be no problem

PR #5908, opened 2024-02-01
Enable debug logging from KitodoRestClient

matthias-ronge:
Adds the possibility to to enable logging from KitodoRestClient. This helps to debug in case of connection problems. Two log levels are supported. `DEBUG` will log the HTML response, including the query URL, the status code and the status text. `TRACE` will also log the result body.


henning-gerhardt (approved):
Changes did work and changes are according to the coding guidelines. 

Logging of used URIs is helpful but I did not see any future value of the trace log level - maybe this is only helpful if you have really connection issues.

Issue #5906, opened 2024-01-30
Long numbers are not written into the Elasticsearch Index


BartChris:
**Describe the bug**
If you have a long number, e.g. a catalog identifier like `991022551489706476` those number does not get indexed into the Elasticsearch index. I encountered that problem, when i tried to import a child process and Kitodo did not find a parent process although the parent process exists. The problem is that for those long  numbers no value gets indexed.The metadata section of the indexed document looks like this. (The relevant field here is CatalogIDDigital)

```json 
                    {
                        "name": "Author",
                        "content": "Martin Chemnitz"
                      },
                      {
                        "name": "CatalogIDDigital"
                      },
                      {
                        "name": "DocLanguage",
                        "content": "lat"
                      }

```

The problem is that those numbers have the Java type `Long` and when constructing the Index we currently only look for strings and integers:

https://github.com/kitodo/kitodo-production/blob/39a881e8e0eb9381d8c2fbd52ecfcabfb26bdc72/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java#L2121-L2123

We probably should also keep Longs here by explicitely allowing that. 

```java
 Object value = xmlJSONObject.get(key);
            if (value instanceof String || value instanceof Integer || value instanceof Long) {
                json.put(prepareKey(key), value);
            }
```
When applying this change the CatalogIDDigital has a value in Elasticsearch.
```json
  {
                        "name": "CatalogIDDigital",
                        "content": 9910211111112122
                      },
                      {
                        "name": "TitleDocMain",
                        "content": "... Pars Examinis decretorum Concilii Tridentini"
                      },

```
This fix enables me to find my parent process again, but i am not sure wether we better convert the value to a string or how to differentiate this from other cases where we need a Long. (we probably do not need Longs for identifiers, since we are not doing any calculations on them).

**To Reproduce**
1. Put a long number into a metadatafield
2. Save the record
3. Check the Elasticsearch Index. The long number is missing from the metadata

**Expected behavior**
Long numbers shoutd not be lost when indexing the document.

**Release**
3.6.1, Master

@solth I put this into the 3.6.2 milestone because i consider it as a quite serious bug. Feel free to take it out, if you are not planning to include this in the next bugfix release.

solth:
Fixed by #5936

PR #5911, opened 2024-02-05
[3.6] Allow datatype Long when indexing

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/5906

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -2105,7 +2105,7 @@ public class ProcessService extends ProjectSearchService<Process, ProcessDTO, Pr
         while (keys.hasNext()) {
             String key = keys.next();
             Object value = xmlJSONObject.get(key);
-            if (value instanceof String || value instanceof Integer) {
+            if (value instanceof String || value instanceof Integer || value instanceof Long) {
                 json.put(prepareKey(key), value);
             } else if (value instanceof JSONObject) {
                 JSONObject jsonObject = (JSONObject) value;


BartChris:
@solth Strange, i will check. Would it be an option to cast all Longs to a String before indexing? I am wondering if there is a scenario where we actually need the number value in the index.

BartChris:
@solth I implemented exactly that. While doing that i also encountered that Kitodo breaks when searching for very long numbers using the search field:

grafik

grafik
I 
because Elasticsearch tries to cast the search value for searching in the ID field.

solth:
@BartChris thanks a lot for looking into this and for the fix. I agree with you that casting numerical metadata values to String is a good solution, since no numerical operations are done on the values in the index anyway (e.g. queries that retrieve documents from ElasticSearch with numerical metadata values in a given range, for example). 

Would you mind opening a corresponding pull request with the same fix against the master branch?

BartChris:
@solth I targeted master as well and changed the test to adapt to the changes in master: https://github.com/kitodo/kitodo-production/pull/5936

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -647,7 +648,7 @@ public List<ProcessDTO> findByAnything(String searchQuery) throws DataException
                 ProcessTypeField.PROJECT_TITLE.getKey(),
                 ProcessTypeField.COMMENTS.getKey(),
                 ProcessTypeField.WIKI_FIELD.getKey(),
-                ProcessTypeField.TEMPLATE_TITLE.getKey()).operator(Operator.AND);
+                ProcessTypeField.TEMPLATE_TITLE.getKey()).operator(Operator.AND).lenient(true);
```
solth, line 651: Just out of curiosity: why is this necessary?

BartChris: The id field of the process is stored as type "long" in the elasticsearch index "kitodo_process". When you query for a very long number in the ID field (which happens automatically when you search for all fields in the top left or search for parent processes), Elasticsearch tries to cast this number to a Long which does not work. Lenient true ignores exceptions regarding type casting.

solth (approved):

PR #5913, opened 2024-02-08
Add a script to empty the database

matthias-ronge:
Adds a developer tool. It is not a functional change to the program. The attached script is made available for administration. It is not called by the program.


solth:
If it's not called or needed by the application itself and just a tool for developers, I think this script should be placed in the https://github.com/kitodo/kitodo-contrib repository instead. Please open a pull request over there (you could add it to the "db" folder, for example)

PR #5914, opened 2024-02-08
Delete legacy SQL files

matthias-ronge:
**Delete 'default_new.sql'**
The script contains a database version that is reached after applying a few database migration steps. It is not clear which database version is contained. The file was never updated as further migration steps were added. It is unusable for creating a database that matches the application version.

**Delete 1.11.x CE migration files**
These files were used in 2016 to migrate the 1.11x CE version. They no longer have anything to do with the current development of 3.x. They are still available in the 2.x branch. Here, they are deleted because they served their purpose in the past.


henning-gerhardt (approved):
Changes are okay as this files still exists in 2.x branch and they are only needed in 2.x.

PR #5917, opened 2024-02-13
[3.6] Update spring security

solth:
Bump Spring security version to fix security issue. See #5912 for details.


markusweigelt (approved):
It works for me when all tests have run successfully.

PR #5923, opened 2024-02-19
Update location of chrome driver for testing

solth:


henning-gerhardt (approved):
Looks good and the selenium tests are working again. Maybe using one of the json end points from https://github.com/GoogleChromeLabs/chrome-for-testing?tab=readme-ov-file#json-api-endpoints (f.e. https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions-with-downloads.json) for catching the uri's instead of hard coding them? This must not be done yet but maybe in future?

solth:
Good idea! As you can see in the code (https://github.com/kitodo/kitodo-production/blob/c723b1c5e6327f4546ac6479c53928cdcc7c578b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java#L42) we already use one of those JSON endpoints (https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions.json) to determine the latest _version_ of Chrome, but not to retrieve the complete URL. I will merge this PR so upcoming builds don't fail but will then try if I can implement your suggestion afterwards, if it's feasible.

PR #5504, opened 2023-01-12
Add media partials for audio and video

markusweigelt:
**What is a 'mediaPartial'?**

- Media partial consist of logicalDivision with a label and a physical division with link to media file, start time and duration

```
<mets:structMap TYPE="PHYSICAL">
    <mets:div ID="PHYS_0000" TYPE="physSequence">
        ...
        <mets:div ID="PHYS_XXXX" ORDER="X" TYPE="track">
            <mets:fptr>
                <mets:area FILEID="FILE_XXXX_DEFAULT" BEGIN="00:00:00" BETYPE="TIME" EXTENT="00:02:30" EXTTYPE="TIME"></mets:area>
            </mets:fptr>
            <mets:fptr FILEID="FILE_XXXX_THUMBS"/>
        </mets:div>
        ...    
    </mets:div>
</mets:structMap>
```

**What is implemented in the PR?**
- Add the `use` with name `mediaPartial` -> divisions marked with this `use` are utilized as media partials
- Divisions with at least one child division with the  `use` `mediaPartial` are active to create media partial for assigned audio and video - only with this condition a button is displayed in media detail (between the media and thumbnail stripe) to create a media partial for audio and video
- When media partials for audio or video are present, the media partials panel is rendered between the media and thumbnail stripe
- With each addition, deletion, or editing of a media partial, they are sorted based on their start time, and the duration until the next media partial or to the duration of media is calculated

**How to?**
- Extend the AV Documentation with documentation for Media Partials
https://github.com/kitodo/kitodo-production/wiki/Verwendung-von-Audio-oder-Video#nutzung-der-medienteile-media-partials

PR #5942, opened 2024-02-22
Improvements media partials

markusweigelt:
- Support for various input formats for start time, which will be converted to the fixed time format `HH:mm:ss.SSS`. For example, `00:00:30` will be converted to `00:00:30.000`, and `1:1:1.1` will be converted to `01:01:01.100` (https://github.com/kitodo/kitodo-production/pull/5504#issuecomment-1908175639)
- Enhance the detail of media partials:
  - Introduce a media partial icon in the heading, providing information on start time and duration
  - Restrict media playback to the specified time range

image

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryMediaContent.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryMediaContent.java
@@ -218,7 +218,9 @@ public class GalleryMediaContent {
     }
 
     /**
-     * @return
+     * Get the media partial of physical division.
+     *
+     * @return The media partial or null
      */
     public MediaPartial getMediaPartial() {
         if (isMediaPartial()) {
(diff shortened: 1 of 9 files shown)


solth:
Please rebase against current master to use up-to-date Chrome-testing-url (see #5923)

markusweigelt:
@solth I rebased the branch against the master. The error which came up over the CI it is not related to the current changes afaik.

markusweigelt:
It was indeed a problem that I caused, but it has been fixed with the recent changes.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryMediaContent.java
```diff
@@ -24,8 +24,10 @@
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.kitodo.api.dataformat.MediaPartial;
 import org.kitodo.api.dataformat.View;
 import org.kitodo.production.enums.MediaContentType;
+import org.kitodo.production.helper.metadata.MediaPartialHelper;
```
solth, line 30:
```suggestion
```
Import seems not to be used

solth (commented):
Looks good and works as advertised. Just found one small unused import that should be removed.

solth (approved):

PR #5944, opened 2024-02-23
Use efficient isEmpty call over inefficient empty string comparison

henning-gerhardt:
`isEmpty` call is more efficient than a complex `equals` call on strings if only comparison is on empty or not empty string.

Discovered by CodeQL scan.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -1183,14 +1183,14 @@ public class FileService {
                     physicalDivision.getMediaFiles().put(entry.getKey(), mediaFile);
                 }
                 String fileCanonical = subfolder.getCanonical(mediaFile);
-                if ("".equals(unitCanonical)) {
+                if (unitCanonical.isEmpty()) {
                     unitCanonical = fileCanonical;
                 } else if (!unitCanonical.equals(fileCanonical)) {
                     throw new InvalidImagesException("Ambiguous canonical file name part in the same physical division: \""
                             + unitCanonical + "\" and \"" + fileCanonical + "\"!");
                 }
             }
-            if (physicalDivision.getMediaFiles().size() > 0 && "".equals(unitCanonical)) {
+            if (!physicalDivision.getMediaFiles().isEmpty() && unitCanonical.isEmpty()) {
                 throw new InvalidImagesException("Missing canonical file name part in physical division " + physicalDivision);
             }
             canonicals.add(unitCanonical);
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -1183,14 +1183,14 @@ public class FileService {
                     physicalDivision.getMediaFiles().put(entry.getKey(), mediaFile);
                 }
                 String fileCanonical = subfolder.getCanonical(mediaFile);
-                if (unitCanonical.isEmpty()) {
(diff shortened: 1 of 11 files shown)


Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1183,14 +1183,14 @@ private List<String> getCanonicalFileNamePartsAndSanitizeAbsoluteURIs(Workpiece
                     physicalDivision.getMediaFiles().put(entry.getKey(), mediaFile);
                 }
                 String fileCanonical = subfolder.getCanonical(mediaFile);
-                if ("".equals(unitCanonical)) {
+                if (unitCanonical.isEmpty()) {
```
solth, line 1186: Calling `equals` on a non-empty String literal instead of String variable (that could be `null`) implicitly prevents `NullPointerExceptions`. 

Since this null-save comparison was made here and because I cannot say whether it was done to explicitly deal with null values, I would prefer to guarantee the same behaviour and either add an explicit null-check
`Objects.isNull(unitCanonical) || unitCanonical.isEmpty()`
or use the null-save 
`StringUtils.isBlank(unitCanonical)`
here.

(Same goes for the next to changes, see below)

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -1183,14 +1183,14 @@ private List<String> getCanonicalFileNamePartsAndSanitizeAbsoluteURIs(Workpiece
...
                     unitCanonical = fileCanonical;
                 } else if (!unitCanonical.equals(fileCanonical)) {
                     throw new InvalidImagesException("Ambiguous canonical file name part in the same physical division: \""
                             + unitCanonical + "\" and \"" + fileCanonical + "\"!");
                 }
             }
-            if (physicalDivision.getMediaFiles().size() > 0 && "".equals(unitCanonical)) {
+            if (!physicalDivision.getMediaFiles().isEmpty() && unitCanonical.isEmpty()) {
```
solth, line 1193: Same here, see above.

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
```diff
@@ -426,7 +426,7 @@ public int getProgress(ObjectType currentType, PushContext pollingChannel) throw
     public String createMapping() throws IOException, CustomResponseException {
         for (String mappingType : KitodoRestClient.MAPPING_TYPES) {
             String mapping = readMapping(mappingType);
-            if ("".equals(mapping)) {
+            if (mapping.isEmpty()) {
```
solth, line 429: Same here, see above.

solth (changes requested):

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
henning-gerhardt: Changed.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
henning-gerhardt: Changed.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
henning-gerhardt: Good catch. I will change it to `StringUtils.isBlank()` call on all 3 mentioned places.

solth (approved):

PR #5946, opened 2024-02-24
Add missing hashCode methods

henning-gerhardt:
In this classes are `equals()` methods implemented without necessary  ` hashCode()` method. `hashCode()` methods are needed by HashMap or similar classes on usage. `Objects.hashCode` is available since Java 6 and can create unique results instead of self-implementing.

Discovered by CodeQL.


solth (approved):

solth:
@henning-gerhardt would you mind rebasing your branch one more time?

PR #5949, opened 2024-02-26
Improve calendar tests

oliver-stoehr:
Improve stability and speed of selenium test for calendar and check if number of issues is correct.

Part of https://github.com/orgs/kitodo/projects/8


henning-gerhardt (approved):
Changes looks good. Tests are running, so merging should be fine.

Kitodo/src/test/java/org/kitodo/selenium/CalendarST.java
```diff
@@ -75,6 +75,7 @@ public void createProcessFromCalendar() throws Exception {
         calendarPage.addBlock();
         calendarPage.addIssue("Morning issue");
         calendarPage.addIssue("Evening issue");
+        assertEquals("Number of issues in the calendar does not match", 4, calendarPage.countIssues());
```
stweil, line 78: This assertion seems to fail randomly:
```
Error:    CalendarST.createProcessFromCalendar:78 Number of issues in the calendar does not match expected:<4> but was:<3>
```

Here are two different results for the same code: [fail](https://github.com/kitodo/kitodo-production/actions/runs/10144546187/job/28048260829#step:14:43940), [pass](https://github.com/stweil/kitodo-production/actions/runs/10144545791/job/28048260031#step:14:43946)

PR #5954, opened 2024-02-26
Fix potential input resource leak

matthias-ronge:
See https://github.com/kitodo/kitodo-production/pull/5951#review-thread-or-comment-id-959855209


henning-gerhardt (approved):
Good and more readable replacement.

solth (changes requested):
Please remove imports `java.io.BufferedReader`, `java.util.List` and `java.util.stream.Collectors` from this class, which are not required anymore after this change.

solth (approved):

PR #5956, opened 2024-02-26
Fix dereferenced variable may be null

matthias-ronge:
See https://github.com/kitodo/kitodo-production/pull/5951#review-thread-or-comment-id-959855212


henning-gerhardt (approved):
Changes looks good from reading.

PR #5965, opened 2024-02-29
Save SM links in order

### LINKED ISSUE #5964: SM links are not saved in order
**Describe the bug**
As per [METS-Anwendungsprofil, section 2.3.2.1](https://dfg-viewer.de/fileadmin/groups/dfgviewer/METS-Anwendungsprofil_2.3.1.pdf#page=16): “The order of the `mets:smLink` for each logical structure must correspond to the correct arrangement of the physical structure elements within that logical structure.” Presentation relies on this fact to pick the first image as the preview thumbnail.

This requirement is not fulfilled with Production. During save, the references for the `smLink` elements are not sorted by the `ORDER` of the linked physical divisions, and the `SmLink` objects are generated in a parallelizing stream, hence the output order is non-deterministic.

**To Reproduce**
1. Save METS file.
2. Send to Presentation.
3. Chances are (but not necessarily) you see a random preview image in the images list.

**Expected behavior**
METS file shall be compliant to “METS-Anwendungsprofil” specification.

matthias-ronge:
Fixes #5964

### DIFF
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/DivXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/DivXmlElementAccess.java
@@ -15,6 +15,7 @@ import java.math.BigInteger;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


matthias-ronge:
I think the fact that the Selenium test fails here has nothing to do with my change.

solth (approved):

PR #5974, opened 2024-03-04
Remove unused method 'parseDate'

solth:
Removes unused method `CalendarForm.parseDate` and thereby implicitely resolves security weaknesses [CWE-190](https://cwe.mitre.org/data/definitions/190.html), [CWE-192](https://cwe.mitre.org/data/definitions/192.html), [CWE-197](https://cwe.mitre.org/data/definitions/197.html) and [CWE-681](https://cwe.mitre.org/data/definitions/681.html)


matthias-ronge (approved):
I completely agree with removing dead code. It looks like this really is a relic. The message keys are no longer in the message files.

PR #5978, opened 2024-03-04
Remove execute right from shell files

henning-gerhardt:
After running the unit tests (f.e. mvn clean verify) the file `Kitodo/src/test/resources/scripts/script_createDirMeta.sh` was marked as modified in GIT repository. The only unit test which is using this file is the FileServiceTest class but there are right adding execute right before the test and removed after the test (but before the asserts are made). The issue did not come from this test class the issue come from the point that the files are in GIT repository checked in with execute rights. So removing the execute rights from this files unit test execution let the GIT repository clean / unmodified.

Running the all-tests profile (f.e. mvn clean verify -Pall-tests) shows two more issues with created files and directories but this is an other story and pull request.


stweil:
I don't understand why the execute right should be removed. If the unit test removes the execute right, I'd consider that as a bug of the unit test.

henning-gerhardt:
It looks like that in previous commits the execute right change was unintended added to the repository. The test itself is right and tests / test suites should set and remove the execution rights from this files (like other tests too). Having execution rights on files inside a repository are - in my opinion - a bad way and a security issue.

stweil:
> Having execution rights on files inside a repository are - in my opinion - a bad way and a security issue.

I could simplify the test scripts a lot by adding execute permission to all shell scripts in pull request #6111.

Could you please explain why you think that this would be "a bad way and a security issue"? My change allows removing much code (including conditional code), so I think it's a good thing.

Erikmitk:
> Could you please explain why you think that this would be "a bad way and a security issue"?

I think setting execute permission (`chmod +x`) needs to be done *only* in the environment the script is actually run and not in the repo. It is a security issue because if you clone the repository and there's a malicious or unintended executable script in it then you have a potential security problem. **That can be mitigated by a sensible default i.e. non-executable scripts.**

Therefore setting the necessary permission in the test code is reasonable and prevents unforeseen consequences.

stweil:
I don't understand this argument and never saw such recommendations anywhere else. Any script can be executed, no matter whethere it has execute permission or not. If it does not have execute permission, you just have to run it with `bash SCRIPT`.

stweil:
There is indeed a security issue with the current code which sets the file permissions during the test. The code uses this pattern for a number of scripts when the test user runs tests:

1. Add execute permission to SCRIPT.
2. Run SCRIPT.
3. Remove execute permission from SCRIPT.

Step 1 changed the permission to 777 (read/write/execute for any user). So a malicious user on the build system could write a program which watches SCRIPT until step 1 was done, then modify SCRIPT to run malicious code before step 2 was run. Then this malicious code is executed in the context of the test user.

Erikmitk:
> Any script can be executed, no matter whether it has execute permission or not. If it does not have execute permission, you just have to run it with bash SCRIPT.

Yes, that's an intentional command that is executed. The security issue is not that a script *can* be executed. The issue is that it is executable *by default*. That's main concern for this issue. And of course it's a tradeoff between security and convenience. From a security standpoint nothing should be executable by default unless it's an intentional decision or the source is verifiable by an integrity check. It is more convenient and easier to ignore that though. That's a decision that can be made. I'd argue in favor of being conservatively in favor of security.

> So a malicious user on the build system…

You can stop the scenario right there because that's the root problem in this case. You can construct any possible scenario from this so it is not meaningful in this context. :)

stweil:
I am sorry, but I still don't understand your argument. In my example above I have explained a scenario on a build host with more than a single user where the current code is a real risk if there are malicious users. My pull request #6111 fixes this.

How could a script with execute permissions be misused? Why do other widely used repositories have executable scripts without fearing security issues?

Erikmitk:
> In my example above I have explained a scenario on a build host with more than a single user where the current code is a real risk if there are malicious users. My pull request https://github.com/kitodo/kitodo-production/pull/6111 fixes this.

The build host is a sandboxed and highly secured environment. That's effort that is not useful. How is this creating any value?

> How could a script with execute permissions be misused?

Like I said, setting **reasonable defaults** is an important part of security mitigation. Making files executable by default is not a sensible setting, in my opinion. In our case, scripts that modify the environment for the application should be used and modified with **intention** i.e. **setting permission when and where they are needed**. We obviously disagree and can argue ourselves in circles. :)

> Why do other widely used repositories have executable scripts without fearing security issues?

I don't know and we have no control over them. Widely used repositories have **plenty of security issues all the time** and they should step up their security game if you ask me. Not sure what we gain by doing what other's are doing when there's a case to be made for more sensible and secure developmental practices.

Erikmitk:
@solth needs to make a decision here. :)

solth:
I am not an security expert but I can see valid points made on both sides. That said, I tend to agree with @Erikmitk that having reasonable defaults for execution rights in place makes sense.

Removing unnecessary code is always an improvement, but since the code in question does indeed serve a purpose this is a different story, even if that purpose is limited. An argument could be made that implementing this feature to add and remove execution rights via Java code may not be worth the effort to begin with, but since we already _have_ the system in place, I do not see the advantage of discarding it. Therefor I opt for this solution over #6111

solth (approved):

PR #5993, opened 2024-03-06
Fix IllegalStateException: No logical node selected!

matthias-ronge:
Exception occurs when opening Metadata Editor, because calling method expects `null` in no-data case, but receives IllegalStateException:

Stacktrace

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
@@ -255,7 +255,7 @@ public class DataEditorService {
         TreeNode selectedLogicalNode = dataEditor.getStructurePanel().getSelectedLogicalNode();
 
         if (Objects.isNull(selectedLogicalNode)) {
-            throw new IllegalStateException("No logical node selected!");
+            return null;
         }
 
         if (!(selectedLogicalNode.getData() instanceof StructureTreeNode)) {


Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
solth, line 1: I think this can potentially create a `NullPointerException` because the result of `DataEditorService.getStructuralElementView` is also passed as last parameter to `ProcessService.findLinkableChildProcesses` where `.contains` is called on it without a null-check:
https://github.com/kitodo/kitodo-production/blob/3d4fd478e9d1a17a1472544d21e0ffc50d0386d3/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java#L785

matthias-ronge: No, because `allowedSubstructuralElements` is the result of `Map.keySet()` which cannot be `null`.

BTW, this is not the problem here. The problem was that the calling method _expects_ `null` in case there is nothing (and there are severeal more correct `return null;` statements below in `getStructuralElementView()`), but in this one place the function throws an `IllegalStateException` _instead of_ returning `null`.

solth: Maybe, but `AddDocStrucTypeDialog.search` _definitely_ calls a method on the result of `DataEditorService.getStructuralElementView` (`DataEditorService.getStructuralElementView(this.dataEditor).getAllowedSubstructuralElements()`, to be specific), that will definitely result in a `NullPointerException` when `getStructuralElementView` is changed to return `null`:

https://github.com/kitodo/kitodo-production/blob/0e6e0ee5dfe7e88ebb07889bb052b32e5536928d/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java#L634

It looks to me that the other cases where `null` is returned are problematic as well, because the resulting `NullPointerException` is not handled anywhere.

matthias-ronge:
I also can't rule out that this will cause an error somewhere else, but for me it solved the specific stack trace problem (and maybe that's already helpful, for example to be able to see an error message). There may be more problems, but merging this would still be an improvement on before, not a deterioration overall, I think.

matthias-ronge:
It looks like the code that caused the bug is gone. I'm closing the pull request as no longer relevant.

Issue #5850, opened 2023-12-06
Maven reports issues with used plugins


henning-gerhardt:
Running Maven build reports the warning

`Plugin validation issues were detected in 6 plugin(s)`

If you run the Maven build with the property `-Dmaven.plugin.validation=verbose` you get the following output

```
Plugin validation issues were detected in 6 plugin(s)

 * org.apache.maven.plugins:maven-resources-plugin:3.1.0
  Declared at location(s):
   * org.kitodo:kitodo-production:3.7.0-SNAPSHOT (pom.xml) @ line 941
   * org.kitodo:kitodo-data-format:3.7.0-SNAPSHOT (Kitodo-DataFormat/pom.xml) @ line 73
  Used in module(s):
   * org.kitodo:kitodo-api:3.7.0-SNAPSHOT (Kitodo-API/pom.xml)
   * org.kitodo:kitodo-data-management:3.7.0-SNAPSHOT (Kitodo-DataManagement/pom.xml)
   * org.kitodo:kitodo-query-url-import:3.7.0-SNAPSHOT (Kitodo-Query-URL-Import/pom.xml)
   * org.kitodo:kitodo-command:3.7.0-SNAPSHOT (Kitodo-Command/pom.xml)
   * org.kitodo:kitodo-file-management:3.7.0-SNAPSHOT (Kitodo-FileManagement/pom.xml)
   * org.kitodo:kitodo-data-format:3.7.0-SNAPSHOT (Kitodo-DataFormat/pom.xml)
   * org.kitodo:kitodo-data-editor:3.7.0-SNAPSHOT (Kitodo-DataEditor/pom.xml)
   * org.kitodo:kitodo-image-management:3.7.0-SNAPSHOT (Kitodo-ImageManagement/pom.xml)
   * org.kitodo:kitodo-longtermpreservation-validation:3.7.0-SNAPSHOT (Kitodo-LongTermPreservationValidation/pom.xml)
   * org.kitodo:kitodo-persistentidentifier:3.7.0-SNAPSHOT (Kitodo-PersistentIdentifier/pom.xml)
   * org.kitodo:kitodo-docket:3.7.0-SNAPSHOT (Kitodo-Docket/pom.xml)
   * org.kitodo:kitodo-xml-schemaconverter:3.7.0-SNAPSHOT (Kitodo-XML-SchemaConverter/pom.xml)
[...]

PR #6000, opened 2024-03-08
Fix Tomcat start with Eclipse broken, local-config ignored

### LINKED ISSUE #5999: Tomcat start with Eclipse on Windows broken, local-config ignored
**Describe the bug**
Tomcat no longer starts from Eclipse under Windows. The reason turns out to be that the `kitodo_config.properties` from the `config-local` directory is no longer deployed, but rather the one from `Kitodo/src/main/resources`. Without the local path information, Tomcat will not start, not least because it can no longer find the modules directory.

**To Reproduce**
1. Launch Tomcat [from Eclipse](/kitodo/kitodo-production/wiki/Developer_3.x-Getting-started-Eclipse-Windows)
2. See error

**Expected behavior**
Build must use `kitodo_config.properties` from the `config-local`.

**Release**
Current master (3.7.0-SNAPSHOT)

 - OS: Windows 10
 - Browser: Firefox 123.0.1

**Additional context**
When Tomcat starts and logging was set to `debug` level before, you will find a log entry like:

    Config - Using configuration from file:/<workspace-path>/.metadata/.plugins/org.eclipse.wst.server.core/tmp0/wtpwebapps/kitodo/WEB-INF/classes/kitodo_config.properties

Previously, this file was a copy of `config-local/kitodo_config.properties`, and now it is a copy of `Kitodo/src/main/resources/kitodo_config.properties`.

matthias-ronge:
Fixes #5999. ~Does undo commit 8062b8caa66b08aee2d428b03e1f77235a005fe4 which was merged in PR #5851 and was intended as a fix for #5850. As a consequence, issue #5850 must be reopened.~

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
         <elasticsearch.version>7.17.10</elasticsearch.version>
         <maven-failsafe-plugin.version>2.22.2</maven-failsafe-plugin.version>
 
-        <!-- maven-resources-plugin version > 3.2.0 introduces changes that
+        <!-- maven-resources-plugin versions greater 3.2.0 introduce changes that
              stop the build in Eclipse on Windows from using "config-local" -->
         <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
 
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,11 @@
         <junit.version>5.9.2</junit.version>
         <elasticsearch.version>7.17.10</elasticsearch.version>
         <maven-failsafe-plugin.version>2.22.2</maven-failsafe-plugin.version>
-        <maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
+
+        <!-- maven-resources-plugin version > 3.2.0 introduces changes that
+             stop the build in Eclipse on Windows from using "config-local" -->
+        <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
+
         <maven-jar-plugin.version>3.3.0</maven-jar-plugin.version>
         <maven-war-plugin.version>3.4.0</maven-war-plugin.version>
         <mockito.version>5.0.0</mockito.version>


henning-gerhardt:
Can you try to elaborate which of the changes is responsible for your issue? I think that is related to the updated maven plugins but as I have no access to a Windows based development system I can't test or elaborate which one is responsible.

matthias-ronge:
> Can you try to elaborate which of the changes is responsible for your issue?

Basically yes, but if those updates are not critical to security, I'd like to do that later. I currently have a whole series of projects in the line.

henning-gerhardt (approved):
Even with old version 3.2.0 the configuration files are copied or not copied depending on used development profile or not and if config-local directory is existing or not the right configuration files was inside the created war file. But this should be confirmed on an other Unix-based system (MacOS/iOS and or an other Linux system maybe?).

pom.xml
```diff
@@ -68,7 +68,11 @@
         <junit.version>5.9.2</junit.version>
         <elasticsearch.version>7.17.10</elasticsearch.version>
         <maven-failsafe-plugin.version>2.22.2</maven-failsafe-plugin.version>
-        <maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
+
+        <!-- maven-resources-plugin version > 3.2.0 introduces changes that
```
solth, line 72: @matthias-ronge If 3.2.0 introduces those breaking changes, why are you changing to exactly that version here? I suspect either the comment or the version itself in line 74 is wrong. Perhaps you meant version 3.*3*.0 that breaks the functionality?

solth:
> But this should be confirmed on an other Unix-based system (MacOS/iOS and or an other Linux system maybe?).

I can confirm that Kitodo compiles and runs on MacOS with the the changes in this pull request and the correct local configuration is used. Therefor I am fine with merging this PR when the question with the apparent typo in the comment (see above) has been resolved.

pom.xml
matthias-ronge: versions **greater** 3.2.0—I will replace the `>` symbol.

PR #6006, opened 2024-03-20
Add missing filterable attribute in ruleset.xsd

### LINKED ISSUE #6005: Ruleset.xsd did not allow filterable attribute
**Describe the bug**
With PR #5810 it is now possible to add an automatic filtering for single and multiple select fields. This is done by adding a `filterable` attribute with value `true` in the used ruleset. But if you validate the ruleset with the ruleset.xsd schema file, validation is failing because the `filterable` attribute is not allowed.

**To Reproduce**
1. add `filterable` attribute to at least one single or multiple select field / meta data in the ruleset file.
2. Look in the application that this is working
3. Validate the ruleset file against the schema file ruleset.xsd (f.e. `xmllint --schema ruleset.xsd slub.xml`) and see the reported errors.

**Expected behavior**
Ruleset.xsd file must know about the now allowed `filterable` attribute`.

**Additional context**
It should be discussed how new features of the ruleset are implemented and verified. Maybe the needed classes for accessing the ruleset should work like accessing METS/MODS data where the responding classes generated from the corresponding xsd files. It could be possible that there are more mismatches between the rules in the ruleset.xsd file and in the accessing implementation.

henning-gerhardt:
Fixes #6005

### DIFF
--- a/Kitodo-DataEditor/src/test/resources/ruleset.xsd
+++ b/Kitodo-DataEditor/src/test/resources/ruleset.xsd
@@ -49,7 +49,7 @@
                 URI; if the type is specified, this applies. If a namespace has been specified, Production looks for an
                 XML file in the same directory whose file name is the same as the last segment of the namespace URI
                 and, if it finds it, makes the namespace elements available as a selection list.
-                
+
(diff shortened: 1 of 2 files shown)


matthias-ronge (approved):
Thanks for the completion.

PR #6007, opened 2024-03-20
Add url encoding for SearchInterfaceTypes other than SRU

### LINKED ISSUE #5945: No URL-encoding of queries over OAI and CUSTOM interfaces
**Describe the bug**
Queries that are send through CUSTOM and OAI interfaces are not URL-encoded. This results in a `java.lang.IllegalArgumentException: Ilegal cahracter` error for queries containing special characters like for example spaces. The user is presented with a stack trace that includes the key used to access the catalog.

**To Reproduce**
1. Create an CUSTOM or OAI interface, have a working import configuration etc. and a nice project.
2. Go to 'Create new process', choose your working import configuration
2. Enter an ID that contains a special character (like a blank space), click on 'Search'
4. See error, ask everyone watching your screen to forget the credentials displayed in the error message

**Expected behavior**
Special characters should be URL-encoded and not result in an error.

**Screenshots**
image
image

**Release**
3.6.0

 - OS: any
 - Browser: any
 - Version: any recent

**Additional context**
This is a problem we are facing with signatures that contain a space character in one part of its name. But generally every signature in our system consists of multiple parts, separated by spaces. So we already had to work around this by introducing a separator character in our database to make any queries possible at all.

gerlingsm:
Add url encoding for idPrameter, prefix and identifier for SearchInterfaceTypes other than SRU

 fix #5945


Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
```diff
@@ -269,11 +269,10 @@ private DataRecord performQueryToRecord(DataImport dataImport, String queryURL,
         String prefix = Objects.nonNull(idPrefix) && !identifier.startsWith(idPrefix) ? idPrefix : "";
         String idParameter = SearchInterfaceType.OAI.equals(dataImport.getSearchInterfaceType()) ? OAI_IDENTIFIER
                 : dataImport.getIdParameter();
-        String queryParameter = idParameter + EQUALS_OPERAND + prefix + identifier;
         if (SearchInterfaceType.SRU.equals(interfaceType)) {
-            fullUrl += URLEncoder.encode(queryParameter, encoding);
+            fullUrl += URLEncoder.encode(idParameter + EQUALS_OPERAND + prefix + identifier, encoding);
         } else {
-            fullUrl += queryParameter;
+            fullUrl += URLEncoder.encode(idParameter, encoding) + EQUALS_OPERAND + URLEncoder.encode(prefix + identifier, encoding);
```
BartChris, line 275: It is probably not necessary to encode the parameter name as well. I am not sure but i could maybe also lead to problems server side? But i am not sure about that and have not tested it.

gerlingsm: I would agree that in most cases (including the issue in question) it would be sufficient to just encode the value.
As far as I know, however, it is technically possible or permissible to use these characters in the parameter name as well. Therefore, my suggestion was to also encode this as a precaution.

BartChris: I suppose it is highly unlikely that people will use parameter names which contain special charcters and standard names like "id" or "title" will not be changed on encoding. So normal search backend will not be affected by your change i suppose. People who are using parameter names with special characters can probably also handle them on the Backend. Therefor i agree with you.

gerlingsm:
> Looks good. I am wondering whether we also need to encode the query parameter values here:
> 
> https://github.com/kitodo/kitodo-production/blob/83b1b9fcbb586add543a55c8d7844d0d82077e36/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java#L417
> 
> I suppose this is right now only used in SRU context, where we can search for different values than the ID. But it is probably a good idea to implment encoding here as well for future usages.

I have updated the fix/url-encoding branch to also fix this for other searchinterfaces.
@BartChris do you think, this is sufficient?

BartChris (approved):

BartChris:
Changes look good and work. Regarding the encoding of the search string

https://github.com/kitodo/kitodo-production/blob/25cd0888a4543ad58adb559b40cdb43f2408002f/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java#L421-L427

i noticed that

1. there is never a search conducted with using another interface than SRU (apart from the search by ID of course)
2. we are joining the search String by "AND" although it is impossible to have more one query keyword
3. Other interface are probably not joining their queries by "AND"

But i would suggest keeping it the way it is done in the PR to have it documented in the code that encoding is necessary for all interfaces.

solth:
@gerlingsm would you mind opening a corresponding pull request with the same fix against the 3.6.x branch?

gerlingsm:
> @gerlingsm would you mind opening a corresponding pull request with the same fix against the 3.6.x branch?

This is not a problem, I have prepared a corresponding pull request #6060

mnscholz:
Unfortunately, this change breaks our current Kitodo config and we are no longer able to import catalog data.

We use the CUSTOM import method to get data from our catalog (Sisis). We need to set a search query parameter with a `=` in order to search the catalog ID only in a certain index field:
grafik

I think the change is overshooting: Query parameter names (left hand side of the `=`) are fixed strings that are manually configured and thus can be escaped by the user if necessary (maybe a hint would be sufficient). At least the CUSTOM method should make as little assumptions as necessary and be as flexible as possible.

BartChris:
> I suppose it is highly unlikely that people will use parameter names which contain special charcters and standard names like "id" or "title" will not be changed on encoding. So normal search backend will not be affected by your change i suppose. People who are using parameter names with special characters can probably also handle them in encoded form on the Backend. Therefor i agree with you.

So I was probably wrong here...

One option would be to define an intermediary endpoint that removes the encoding again and have Kitodo communicate only with that endpoint. The other option would be to drop the encoding of the parameter name altogether. I think we can probably change this, as it was only introduced as a precaution. What do you think @gerlingsm @solth

solth:
> What do you think @gerlingsm @solth

Yes, I think you are right. The previous assumption stands in direct contrast to the issue @mnscholz has reported. Restricting the URL enconding to the parameter value should resolve both issues, #5945 for @gerlingsm and #6474 for @mnscholz, and could also be back-ported to Kitodo.Production 3.8 because no database migration is necessary.

BartChris:
I committed a Pull Request which tries to adress that: https://github.com/kitodo/kitodo-production/pull/6477

Issue #6009, opened 2024-03-21
Bug when switching the separator for Mass Import


BartChris:
**Describe the bug**
There is a feature of switching the CSV seperator in the mass import page to match the column seperator of the imported csv file.
![image](https://github.com/kitodo/kitodo-production/assets/7516499/f786eb48-f30e-454c-a6f2-d06d738a10a5)
When using this feature the list with the metadata keys (representing the current column headings) is recreated using List.of

https://github.com/kitodo/kitodo-production/blob/83b1b9fcbb586add543a55c8d7844d0d82077e36/Kitodo/src/main/java/org/kitodo/production/forms/massimport/MassImportForm.java#L129

Using List.of here creates a problem because the new list is immutable so no new items can be added to the list. When using the "Add Metadata" feature
![image](https://github.com/kitodo/kitodo-production/assets/7516499/3eb04b73-3a24-42bb-a9e4-077eaf4f1a96)

afterwards an exception is thrown because Kitodo tries to add items to an now immutable list.

https://github.com/kitodo/kitodo-production/blob/83b1b9fcbb586add543a55c8d7844d0d82077e36/Kitodo/src/main/java/org/kitodo/production/forms/massimport/AddMetadataDialog.java#L116-117

![image](https://github.com/kitodo/kitodo-production/assets/7516499/f69fee12-5327-4a8e-b65c-76ab9bb19681)

**To Reproduce**
1. Upload a csv file seperated by comma (",")
2. Switch the seperator to from semicolon to comma (";")
3. Try to add metadata

**Expected behavior**
The list with the metadata keys probably has to stay mutable after changing the csv separator.

**Release**
Kitodo Production 3.6.2, Master

PR #5961, opened 2024-02-27
Add tests for mass import

solth:
This pull request adds unit, integration and Selenium tests for mass import.

Part of https://github.com/orgs/kitodo/projects/8

PR #6012, opened 2024-03-22
Allow quotes in imported csv files and make mass import more resilient

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/6009

This PR supersedes the closed Pull Requests https://github.com/kitodo/kitodo-production/pull/6011 and https://github.com/kitodo/kitodo-production/pull/6010 and does three things:
- it makes the mass import more resilient by filtering out empty lines in the imported csv file which lead to an exception right now. (see https://github.com/kitodo/kitodo-production/pull/6010)
- it fixes the bug described in https://github.com/kitodo/kitodo-production/issues/6009 by changing the method how the seperator for the import is switched
- it allows for quoted values in the CSV file, which can be used when the value contains potential seperators ("," or ";"). Values which contain a seperator break the import in current master. This PR fixes that. 

Marking as draft until tests can be added after  https://github.com/kitodo/kitodo-production/pull/5961 is merged.


Kitodo/src/main/java/org/kitodo/production/services/data/MassImportService.java
```diff
@@ -100,10 +85,24 @@ public List<CsvRecord> parseLines(List<String> lines, String separator) {
...
+                        inQuotes = false;
+                    } else if (String.valueOf(c).equals(separator) && !inQuotes) {
+                        cells.add(new CsvCell(currentCell.toString()));
+                        currentCell.setLength(0); // Reset currentCell
+                    } else {
+                        currentCell.append(c);
+                    }
+                }
+                cells.add(new CsvCell(currentCell.toString()));
+                records.add(new CsvRecord(cells));
```
BartChris, line 104: we could make that more resilient by not adding a record where a quoted value has not closing quotation. 

```
1233,TEST,Name
1234,'TEST,Name
```
```java
 if (!inQuotes){
                    cells.add(new CsvCell(currentCell.toString()));
                    records.add(new CsvRecord(cells));
                }
```

henning-gerhardt:
Is [OpenCSV](https://opencsv.sourceforge.net/) project / library not an option to be used instead of coding this again?

BartChris:
Yes, it definitely is. While fixing the above bugs i was already wondering wether we will end up fixing a lot of bugs manually, which might already be taken care of by a library. I was just not sure, when the point is reached where we should bring in a new dependency. And if this might bring in new problems. I will look into it.

henning-gerhardt:
I prefer using libraries instead of writing the same solution for my issue again and if I encounter issues I would solve them in the library or in the source data. But in the end it is a decision by the developer.

henning-gerhardt:
Just a note: OpenCSV was not in use (and the dependency could be removed) until this pull request and the dependency is quite old: version 3.10 from July 2017 - maybe updating the version to a more recent version like 5.9 from November 2023?

henning-gerhardt:
A small suggestion how the method `parseLines` in the `MassImportService` class can may look:

```java
    public List<CsvRecord> parseLines(List<String> lines, String separator) throws IOException, CsvException {
        List<CsvRecord> records = new LinkedList<>();
        CSVParser parser = new CSVParserBuilder()
                .withSeparator(separator.charAt(0))
                .withQuoteChar('\"')
                .build();
        try (BufferedReader reader = new BufferedReader(new StringReader(String.join("\n", lines)));
             CSVReader csvReader = new CSVReaderBuilder(reader)
                .withSkipLines(0)
                .withCSVParser(parser)
                .build()) {
            for (String[] entries : csvReader.readAll()) {
                List<CsvCell> cells = new LinkedList<>();
                for (String value : entries) {
                    cells.add(new CsvCell(value));
                }
                records.add(new CsvRecord(cells));
            }
        }
        return records;
    }
```

henning-gerhardt:
And a more compact variant:

```java
    public List<CsvRecord> parseLines(List<String> lines, String separator) throws IOException, CsvException {
        List<CsvRecord> records;
        CSVParser parser = new CSVParserBuilder()
                .withSeparator(separator.charAt(0))
                .withQuoteChar('\"')
                .build();
        try (StringReader reader = new StringReader(String.join("\n", lines));
             CSVReader csvReader = new CSVReaderBuilder(reader)
                .withSkipLines(0)
                .withCSVParser(parser)
                .build()) {
            records = csvReader.readAll().stream().map(csvLines -> Arrays.stream(csvLines)
                    .map(CsvCell::new)
                    .collect(Collectors.toCollection(LinkedList::new)))
                    .map(CsvRecord::new).collect(Collectors.toCollection(LinkedList::new));
        }
        return records;
    }
```

BartChris:
Thanks a lot for the support, this looks way better. Your second suggestion probably needs some finetuning to work, so i used the more classical approach for now. There are still some problems when i have a very large csv (2000 and more lines) and switch the separator multiple times. Somehow the table data gets lost. I will look into that next week.

henning-gerhardt:
You are welcome. I looked only the reading part and I think there is room left for more improvements:
- maybe the `StreamReader` must be wrapped in a `BufferedReader` for more amount of data
- maybe converting the `UploadedFile` from PrimeFaces to a list of strings back to a `Reader` object as `CSVReader` can directly use a `InputBufferStream`: `CSVReaderBuilder`(`UploadedFile`.`getInputStream()`).build(). Extracting the header line should even be possible with the OpenCSV library with the same delimiters and quote qualifiers.

BartChris:
> Thanks a lot for the support, this looks way better. Your second suggestion probably needs some finetuning to work, so i used the more classical approach for now. There are still some problems when i have a very large csv (2000 and more lines) and switch the separator multiple times. Somehow the table data gets lost. I will look into that next week.

This happens also in current master. When having 2000 entries in my CSV file i can only switch the separator once. After doing an additional switch the table loses its data.

solth:
> This happens also in current master. When having 2000 (500 are working) entries in my CSV file i can only switch the separator once. After doing an additional switch the table loses its data.

@BartChris If this happens in the master as well, the problem is not introduced by the changes in this pull request, right? Is there another reason why it is still a draft?

BartChris:
@solth Sorry, i did not get to that again. I will add tests next week and then this should be ready for review.

Kitodo/src/main/java/org/kitodo/production/forms/massimport/MassImportForm.java
```diff
@@ -54,9 +56,10 @@ public class MassImportForm extends BaseForm {
     private ImportConfiguration importConfiguration;
     private UploadedFile file;
     private String csvSeparator = ";";
-    private String previousCsvSeparator = null;
     private List<String> metadataKeys = new LinkedList<>(Collections.singletonList("ID"));
     private List<CsvRecord> records = new LinkedList<>();
+    private String importedCsvHeaderLine = "";
+    private List<String> importedCsvLines = new LinkedList<>();
```
solth, line 62: Should `importedCsvHeaderLine` and `importedCsvLines` also be reset in `resetValues`, perhaps? Otherwise, these values could remain after consecutive file uploads of empy CSV files, couldn't they? Or is that perhaps even intended?

Kitodo/src/main/java/org/kitodo/production/services/data/MassImportService.java
solth: Did you abandon this idea?

Kitodo/src/test/resources/scripts/script_createDirMeta.sh
solth, line 1: I think this change was probably commited accidentally. There is an ongoing discussion about this topic and multiple pull requests are offering distinct ways to deal with it. See #5978 and #6111 for details. So I wouldn't change the execution rights of this file in this pull request, because its not related to mass import directly.

solth (commented):
Looks good and seems to work. Just a few little questions and remarks.

Kitodo/src/test/resources/scripts/script_createDirMeta.sh
BartChris: Yes, this was committed accidentally. When running tests locally, this file often gets changed and is then included in the commit if one is not careful. I would be happy if we could find a way to resolve this annoyance (in whatever way).

Kitodo/src/main/java/org/kitodo/production/forms/massimport/MassImportForm.java
BartChris: I agree. It is more consistent when to reset those values as well.

solth (approved):

PR #6013, opened 2024-03-25
Run Kitodo Script commands via Active MQ

### LINKED ISSUE #5980: API to trigger actions, e.g., DMS Export
## Description

An interface (API) that allows (external systems) to authenticate and trigger certain actions, e.g., a DMS Export for a certain process in Kitodo.

Currently we publish a finished process to our presentation platform, need the mets file for that and hence have a script that imitates a user, signing in and click DMS-Export. This is cumbersome and the script needs to be adapted if changes in the GIU happens after an update. An interface would be more appropriated here.

For this specific task, the exp. costs are probably low.

matthias-ronge:
This development allows to run Kitodo Script commands using the Active MQ interface.

For example, to export processes, you send a Map Message with the content:

    id        := "ID string used with Active MQ logging. Must be present, but is meaningless"
    script    := "action:export exportImages:true"
    processes := 4

`processes` can be a single integer, or string representing an integer, or a list of either, or string with several integers separated by non-digit character(s).

Resolves #5980

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
+++ b/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
@@ -1,0 +1,62 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.production.interfaces.activemq;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import javax.jms.JMSException;
+
+import org.kitodo.config.ConfigCore;
+import org.kitodo.config.enums.ParameterCore;
+import org.kitodo.data.database.beans.Process;
+import org.kitodo.data.database.exceptions.DAOException;
+import org.kitodo.data.exceptions.DataException;
+import org.kitodo.exceptions.InvalidImagesException;
+import org.kitodo.exceptions.MediaNotFoundException;
+import org.kitodo.exceptions.ProcessorException;
+import org.kitodo.production.services.ServiceManager;
+import org.kitodo.production.services.command.KitodoScriptService;
+import org.kitodo.production.services.data.ProcessService;
+
+/**
+ * Executes instructions to start a Kitodo Script command from the Active MQ
+ * interface. The MapMessage must contain the command statement in the
+ * {@code script} argument. You pass a list of the process IDs as
+ * {@code processes}.
+ */
+public class KitodoScriptProcessor extends ActiveMQProcessor {
+
+    private final KitodoScriptService kitodoScriptService = ServiceManager.getKitodoScriptService();
+    private final ProcessService processService = ServiceManager.getProcessService();
+
+    public KitodoScriptProcessor() {
+        super(ConfigCore.getOptionalString(ParameterCore.ACTIVE_MQ_KITODO_SCRIPT_QUEUE).orElse(null));
+    }
+
+    @Override
+    protected void process(MapMessageObjectReader ticket) throws ProcessorException, JMSException {
+        try {
+            String script = ticket.getMandatoryString("script");
+            Collection<Integer> processIds = ticket.getCollectionOfInteger("processes");
+            List<Process> processes = new ArrayList<>(processIds.size());
+            for (Integer id : processIds) {
+                processes.add(processService.getById(id));
+            }
+            kitodoScriptService.execute(processes, script);
+        } catch (DAOException | DataException | IOException | InvalidImagesException | MediaNotFoundException e) {
+            throw new ProcessorException(e);
+        }
+    }
+}
--- a/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
+++ b/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
@@ -19,6 +19,7 @@ import java.util.List;
 
 import javax.jms.JMSException;
 
+import org.apache.commons.lang3.ArrayUtils;
 import org.kitodo.config.ConfigCore;
 import org.kitodo.config.enums.ParameterCore;
 import org.kitodo.data.database.beans.Process;
@@ -48,12 +49,12 @@ public class KitodoScriptProcessor extends ActiveMQProcessor {
 
     @Override
     protected void process(MapMessageObjectReader ticket) throws ProcessorException, JMSException {
-        final List<String> allowedCommands = Arrays.asList(ConfigCore.getParameter(
-            ParameterCore.ACTIVE_MQ_KITODO_SCRIPT_ALLOW).split(","));
+        final String[] allowedCommands = ConfigCore.getStringArrayParameter(
+            ParameterCore.ACTIVE_MQ_KITODO_SCRIPT_ALLOW);
         try {
             String script = ticket.getMandatoryString("script");
             int space = script.indexOf(' ');
-            if (!allowedCommands.contains(script.substring(7, space >= 0 ? space : script.length()))) {
+            if (!ArrayUtils.contains(allowedCommands, script.substring(7, space >= 0 ? space : script.length()))) {
                 throw new IllegalArgumentException((space >= 0 ? script.substring(0, space) : script)
                         + " is not allowed");
             }
--- a/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
+++ b/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
@@ -13,6 +13,7 @@ package org.kitodo.production.interfaces.activemq;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 
@@ -47,8 +48,15 @@ public class KitodoScriptProcessor extends ActiveMQProcessor {
 
     @Override
     protected void process(MapMessageObjectReader ticket) throws ProcessorException, JMSException {
+        final List<String> allowedCommands = Arrays.asList(ConfigCore.getParameter(
+            ParameterCore.ACTIVE_MQ_KITODO_SCRIPT_ALLOW).split(","));
         try {
             String script = ticket.getMandatoryString("script");
+            int space = script.indexOf(' ');
+            if (!allowedCommands.contains(script.substring(7, space >= 0 ? space : script.length()))) {
+                throw new IllegalArgumentException((space >= 0 ? script.substring(0, space) : script)
+                        + " is not allowed");
+            }
             Collection<Integer> processIds = ticket.getCollectionOfInteger("processes");
             List<Process> processes = new ArrayList<>(processIds.size());
             for (Integer id : processIds) {
--- a/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
+++ b/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
@@ -16,7 +16,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
--- a/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
+++ b/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
@@ -16,6 +16,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -24,9 +25,9 @@ import java.util.stream.Collectors;
 import javax.jms.JMSException;
 import javax.jms.MapMessage;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.util.Strings;
 import org.kitodo.utils.Guard;
 
 public class MapMessageObjectReader {
@@ -128,7 +129,7 @@ public class MapMessageObjectReader {
     public Collection<Integer> getCollectionOfInteger(String key) throws JMSException {
         Collection<String> collectionOfString = getCollectionOfString(key);
         List<Integer> collectionOfInteger = collectionOfString.stream()
-                .flatMap(string -> Arrays.stream(string.split("\\D+"))).filter(Strings::isNotBlank)
+                .flatMap(string -> Arrays.stream(string.split("\\D+"))).filter(StringUtils::isNumeric)
                 .map(Integer::valueOf).collect(Collectors.toList());
         return collectionOfInteger;
     }
--- a/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
+++ b/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
@@ -11,13 +11,16 @@
 
 package org.kitodo.production.interfaces.activemq;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.jms.JMSException;
 import javax.jms.MapMessage;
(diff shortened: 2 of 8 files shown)


Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
```diff
@@ -110,6 +113,53 @@
...
+     * @param key
+     *            the name of the set to return
+     * @return the set requested
+     * @throws JMSException
+     *             can be thrown by MapMessage.getObject(String)
+     */
+    public Collection<Integer> getCollectionOfInteger(String key) throws JMSException {
+        Collection<String> collectionOfString = getCollectionOfString(key);
+        List<Integer> collectionOfInteger = collectionOfString.stream()
+                .flatMap(string -> Arrays.stream(string.split("\\D+"))).map(Integer::valueOf)
```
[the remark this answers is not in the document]
matthias-ronge: Fixed below

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
```diff
@@ -110,6 +114,53 @@
...
+     *            the name of the set to return
+     * @return the set requested
+     * @throws JMSException
+     *             can be thrown by MapMessage.getObject(String)
+     */
+    public Collection<Integer> getCollectionOfInteger(String key) throws JMSException {
+        Collection<String> collectionOfString = getCollectionOfString(key);
+        List<Integer> collectionOfInteger = collectionOfString.stream()
+                .flatMap(string -> Arrays.stream(string.split("\\D+"))).filter(Strings::isNotBlank)
+                .map(Integer::valueOf).collect(Collectors.toList());
```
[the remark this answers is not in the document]
matthias-ronge: False positive. Cannot throw NumberFormatException, because `string.split("\\D+")` will only return sequences of digits or empty strings, in case of non-digit characters at the beginning or end of the string. The latter are filtered out by `.filter(Strings::isNotBlank)`, thus leaving no chance for NumberFormatException to arise.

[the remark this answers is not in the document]
solth: Why not replace `.filter(Strings::isNotBlank)` with `.filter(StringUtils::isNumeric)`? That would take care of empty Strings and `values` as well (exactly what `isNotBlank` does), would perform an explicit number check and would be better readable and understandable, IMHO. 

It also would not require any new dependencies as we already use the Apache Commons library and the `StringUtils` class in multiple other places in the code.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
```diff
@@ -110,6 +114,53 @@ public String getMandatoryString(String key) throws JMSException {
...
+        return collectionOfInteger;
```
solth, line 135:
```suggestion
        return collectionOfString.stream()
                .flatMap(string -> Arrays.stream(string.split("\\D+"))).filter(StringUtils::isNumeric)
                .map(Integer::valueOf).collect(Collectors.toList());
```
No need to assigne the result of the stream operation to a variable and then return the variable in the next line - just inline it, e.g. directly return the result.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
```diff
@@ -11,19 +11,23 @@
...
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.jms.JMSException;
 import javax.jms.MapMessage;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.util.Strings;
```
solth, line 30:
```suggestion
import org.apache.commons.lang3.StringUtils;
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;
```
Import `StringUtils` instead of `Strings` since it provides an explicit numeric check. (see below)

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
```diff
@@ -11,19 +11,23 @@
 
 package org.kitodo.production.interfaces.activemq;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
```
solth, line 19:
```suggestion
```
`List` import unused when directly returning result of stream operation (see below)

solth (changes requested):
Looks good (apart from some small suggestions), but lacks a test. Please provide a test for this new functionality and update the documentation in the wiki.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
```diff
@@ -110,6 +113,53 @@ public String getMandatoryString(String key) throws JMSException {
...
+     * @return the set requested
+     * @throws JMSException
+     *             can be thrown by MapMessage.getObject(String)
+     */
+    public Collection<Integer> getCollectionOfInteger(String key) throws JMSException {
+        Collection<String> collectionOfString = getCollectionOfString(key);
+        List<Integer> collectionOfInteger = collectionOfString.stream()
+                .flatMap(string -> Arrays.stream(string.split("\\D+"))).filter(Strings::isNotBlank)
+                .map(Integer::valueOf).collect(Collectors.toList());
+        return collectionOfInteger;
```
solth, line 134:
```suggestion
        return collectionOfString.stream().flatMap(string -> Arrays.stream(string.split("\\D+")))
                .filter(StringUtils::isNumeric).map(Integer::valueOf).collect(Collectors.toList());
```

solth (changes requested):

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/MapMessageObjectReader.java
matthias-ronge: I used to do it this way, but then I should do it the other way around. Now you want me to do it the other way around again. No problem, I just need to know.

solth (approved):
Thank you for incorporating the change requests.

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
```diff
@@ -0,0 +1,440 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.kitodo.MockDatabase;
+import org.kitodo.SecurityTestUtils;
+import org.kitodo.data.database.beans.Process;
+import org.kitodo.production.services.ServiceManager;
+import org.kitodo.production.services.command.KitodoScriptService;
+
+public class KitodoScriptProcessorIT {
```
henning-gerhardt, line 35: Can this new test class please written with Junit5 instead of Junit4? Thank you.

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
```diff
@@ -0,0 +1,440 @@
...
+            }
+        });
+
+        // carry out test
+        underTest.process(testData);
+
+        // check results
+        assertEquals("should have passed the script to be executed", scriptResult.get(0), "action:test");
+        assertEquals("should have passed one process", processesResult.size(), 1);
+        assertEquals("should have passed process 1", processesResult.get(0).getId(), (Integer) 1);
```
henning-gerhardt, line 438: Unneeded cast `(Integer)  1` .

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -85,7 +85,7 @@ public static KitodoScriptService getInstance() {
     /**
      * Private constructor. Use {@link #getInstance()} to get the instance.
      */
-    private KitodoScriptService() {
+    protected KitodoScriptService() {
```
henning-gerhardt, line 88: Why are you changing the constructor from `private` to `protected`?

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
henning-gerhardt: The test class must be adjusted as PR #6093 is merged and this test class will not work with Junit5.

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
```diff
@@ -0,0 +1,440 @@
...
+            }
+        };
+
+        // the object to be tested
+        KitodoScriptProcessor underTest = new KitodoScriptProcessor();
+
+        // organize return of results
+        List<String> scriptResult = new ArrayList<>();
+        List<Process> processesResult = new ArrayList<>();
+        Field serviceField = KitodoScriptProcessor.class.getDeclaredField("kitodoScriptService");
```
henning-gerhardt, line 422: Is using the Mockito framework may a better / cleaner solution than using field injection here?

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
matthias-ronge: This was done so that the `KitodoScriptProcessorIT` can create a derived class from this one, that does not perform any functionality, but changes a state in the test class, so that you can check that the code to be tested is working correctly.

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
matthias-ronge: I don't know exactly what that means. I've never used it before. A quick look, it might be helpful, maybe. Unfortunately, I don't have the time to learn this right now, so maybe at some point, I can do a change pull request again. Or, can you suggest what the code should look like?

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
matthias-ronge: The cast is needed, otherwise I get the error: _The method assertEquals(String, Object, Object) is ambiguous for the type KitodoScriptProcessorIT._

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
henning-gerhardt: I will look into this and if I get no right solution in the next time your way here is okay. It can be changed in the future too.

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
```diff
@@ -0,0 +1,440 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.kitodo.MockDatabase;
+import org.kitodo.SecurityTestUtils;
+import org.kitodo.data.database.beans.Process;
+import org.kitodo.production.services.ServiceManager;
+import org.kitodo.production.services.command.KitodoScriptService;
+
+public class KitodoScriptProcessorIT {
+    @BeforeAll
```
henning-gerhardt, line 36: This should be executed before any test is executed or?
```suggestion
    @BeforeEach
```

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
```diff
@@ -0,0 +1,440 @@
...
+    public void prepare() throws Exception {
+        MockDatabase.startNode();
+        MockDatabase.insertProcessesForWorkflowFull();
+        SecurityTestUtils.addUserDataToSecurityContext(ServiceManager.getUserService().getById(1), 1);
+    }
+
+    @AfterAll
```
henning-gerhardt, line 43: This should be executed after each test or?
```suggestion
    @AfterEach
```

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
matthias-ronge: It doesn't matter here, as there is only one test included. As far as I can see, the database is only created once for all other test classes, not for each test. That's why I would do the same here.

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
matthias-ronge: It doesn't matter here, as there is only one test included. As far as I can see, the database is only removed once for all other test classes, not for each test. That's why I would do the same here.

Kitodo/src/test/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessorIT.java
henning-gerhardt: With `BeforeEach` and `AfterEach` would ensure, that every test is executed in the same conditions. Okay, there is currently only one test in this test class but it must / should be changed on adding an other test where I think this would got forgotten.

henning-gerhardt (approved):
Changes looks good but I can not test the changes by real samples but the tests should cover the introduced new processor. After merging the documentation https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ should / must be updated for the new ActiveMQ processor so anyone can create the correct message queue messages.

henning-gerhardt:
A hopefully important information and maybe discussion decision:

It could be a dangerous to enable feature through the configuration option as any KitodoScript can be executed for an unlimited list of process ids. You only needed access to the used ActiveMQ server and send to this queue a message. There is no restriction like in the UI where you need at least a login and proper rights to execute KitodoScripts for amount of or all projects and their processes. 

As everyone how can send ActiveMQ messages to that queue can execute the "hidden" KitodoScript action `deleteProcess` with a unlimited number of process ids and with this an anonymous person can delete all the data. Until now it was not necessary to introduce a security concept on external accessible APIs like in the UI  but with introducing this possibility it is needed. With the current not really existing security concept you can only enable this feature including dangerous KitodoScripts or you can not use this feature at all. Deleting this KitodoScript `deleteProcess` is a bad option as is could be really needed if f.e. 300 - 400 wrong created newspaper issues must be deleted. It is the only way to do this without deleting process by process in the UI which take a very long time (even deleting a lot of processes through the KitodoScript needs a lot of time but deleting one by one takes "years" of manual deleting).

At least a note be applied to the part inside `kitodo_config.properties` which hints to the dangerous part of this feature. A security concept for using the application from outside - until now only for ActiveMQ but needed for future developments like HTTP based REST APIs - is needed and even the development of this security access system (may re-use existing concepts and implementations instead of re-do all this again with all possible bugs and security holes).

matthias-ronge:
I can't really imagine that it happens _by accident_, that a Kitodo script is parameterized like that. Yes, and maybe it's even intended to delete processes, who knows. On our systems, only defined ports are accessible, so no one can activate anything other than `localhost`, not anything else connect to Active MQ. Of course, we can add an additional explanation to the `kitodo_config.properties` file, what do you think:

```txt
# You can provide a queue from which messages are read to run a Kitodo Script
# •-------------------------------------------------------------------------•
# | CAUTION: This can be used to use an unknown Kitodo script with the name |
# | "action:deleteProcesses", and this could result in data destruction!    |
# •-------------------------------------------------------------------------•
#activeMQ.kitodoScript.queue=KitodoProduction.KitodoScript.Queue
```

henning-gerhardt:
The text looks good but I don't know if we really want to hint to the dangerous script call.

If you run everything on one host / virtual machine / real system and access everything only by localhost the risk is lower to to them who run this parts on different hosts  / virtual machines / real systems as they need this splitting.

matthias-ronge:
I take your concerns seriously and added a whitelist to the configuration so that only explicitly permitted commands can be started remotely. I think this should relieve all worries.

Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -639,6 +639,13 @@ activeMQ.user=testAdmin
 # You can provide a queue from which messages are read to process task actions
 #activeMQ.taskAction.queue=KitodoProduction.TaskAction.Queue
 
+# You can provide a queue from which messages are read to run a Kitodo Script
+#activeMQ.kitodoScript.queue=KitodoProduction.KitodoScript.Queue
+
+# The Kitodo Script commands authorized to be executed must be named here:
+activeMQ.kitodoScript.allow=createFolders,export,searchForMedia
```
henning-gerhardt, line 646: As other configuration options this should be disabled too.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+    private final KitodoScriptService kitodoScriptService = ServiceManager.getKitodoScriptService();
+    private final ProcessService processService = ServiceManager.getProcessService();
+
+    public KitodoScriptProcessor() {
+        super(ConfigCore.getOptionalString(ParameterCore.ACTIVE_MQ_KITODO_SCRIPT_QUEUE).orElse(null));
+    }
+
+    @Override
+    protected void process(MapMessageObjectReader ticket) throws ProcessorException, JMSException {
+        final List<String> allowedCommands = Arrays.asList(ConfigCore.getParameter(
```
henning-gerhardt, line 51: Are cases considered when the configuration option is not set (`#activeMQ.kitodoScript.allow=...`) or not set to a value (`activeMQ.kitodoScript.allow=`)?

henning-gerhardt:
> I take your concerns seriously and added a whitelist to the configuration so that only explicitly permitted commands can be started remotely. I think this should relieve all worries.

Thanks. Such an allow list is a simple way to archive this - better then nothing at all.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
henning-gerhardt: Maybe you should use the `getStringArrayParameter()` method to retrieve the list as of strings. Then you did not even need to call the `split(",")` method as this is already done. I don't know how the handling of not set or empty set configuration value is.

matthias-ronge: ConfigCore.getParameter() should NoSuchElementException. But if it would return `null`, String.split() would throw a NullPointerException, and if it would return empty string, any command would be forbidden. So it is safe.

matthias-ronge: I don't like `getStringArrayParameter()` because it returns an array, not a list object, and—unintuitively—uses the ampersand as separator character.  I also couldn't figure out how it behaves with empty parameters. Do you really want that?

henning-gerhardt: I'm not a friend of reimplementing code again. Maybe move this code into a method of the `ConfigCore` class so the implementation can be reused on other places instead of hiding it in this place - maybe already the code is created on other places.

henning-gerhardt (approved):

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/KitodoScriptProcessor.java
matthias-ronge: I used the old code. We can then change this globally later, but the change would introduce differences in config, and I thought there was a reason not to use the separator comma for something.

henning-gerhardt:
@matthias-ronge @solth The [ActiveMQ developer documenation](https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ) is not yet updated for this new feature.

Issue #5969, opened 2024-03-01
Updating H2 database


henning-gerhardt:
## Description

The current version of the H2 database being utilized in integration and Selenium-based tests has several critical security issues. Upgrading to the latest version of the H2 database poses challenges beyond merely updating the version number in the pom.xml file, as significant changes have been implemented, not all of which may directly impact our H2 usage.

## Expected Benefits of this Development

The H2 database serves as the backbone for crucial integration and Selenium tests. However, with an outdated version, there's no guarantee that bugs, including security vulnerabilities, will be addressed. Given that software undergoes continuous development with new features and bug fixes, it's imperative to ensure we stay updated to mitigate risks effectively.

## Estimated Costs and Complexity

The estimated costs and complexity fall within the range of low to medium.

PR #6015, opened 2024-03-27
H2 upgrade

henning-gerhardt:
Fixes: #5969 

Main work was made by @solth. I'm just "fixed" the selenium tests by applying the other changes to the used selenium hibernate configuration.

### DIFF
--- a/Kitodo/src/test/resources/selenium/resources/hibernate.cfg.xml
+++ b/Kitodo/src/test/resources/selenium/resources/hibernate.cfg.xml
@@ -22,11 +22,14 @@
         <!-- SQL - Settings -->
         <property name="dialect">org.hibernate.dialect.H2Dialect</property>
         <property name="hibernate.connection.driver_class">org.h2.Driver</property>
+
         <property name="hibernate.connection.username">sa</property>
         <property name="hibernate.connection.password"/>
-        <property name="hibernate.connection.url">
-            jdbc:h2:tcp://localhost/mem:test;MODE=MariaDB;DATABASE_TO_LOWER=TRUE;DB_CLOSE_DELAY=-1
-        </property>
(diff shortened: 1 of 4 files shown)


Kitodo/src/test/resources/selenium/resources/hibernate.cfg.xml
```diff
@@ -22,11 +22,14 @@
         <!-- SQL - Settings -->
         <property name="dialect">org.hibernate.dialect.H2Dialect</property>
         <property name="hibernate.connection.driver_class">org.h2.Driver</property>
+
         <property name="hibernate.connection.username">sa</property>
         <property name="hibernate.connection.password"/>
-        <property name="hibernate.connection.url">
-            jdbc:h2:tcp://localhost/mem:test;DB_CLOSE_DELAY=-1
-        </property>
+        <property name="hibernate.connection.url">jdbc:h2:tcp://localhost/mem:test;DB_CLOSE_DELAY=-1;MODE=MariaDB;DATABASE_TO_LOWER=TRUE;CASE_INSENSITIVE_IDENTIFIERS=TRUE;NON_KEYWORDS=USER</property>
```
solth, line 28: Just out of curiosity: is this additional `tcp` in the connection url required because Selenium accesses the database via the browser e.g. http, thus tcp?

solth (approved):

Kitodo/src/test/resources/selenium/resources/hibernate.cfg.xml
henning-gerhardt: I did not see an additional connection uri. It is the same as before only with the needed additional parameters like on the other places. Why only a tcp based connection is working on the selenium configuration is out of my knowledge but was before this changes already in use. Maybe the tcp connection is used here because of the used Tomcat container where no other connections are possible.
Or did I read your comment wrong?

solth: Ah, yes, you are right, the `tcp` was already there, I missed that.

PR #6018, opened 2024-04-02
Import service Selenium test

solth:
This PR adds a Selenium test for the import of hierarchical processes and fixes two other existing Selenium tests.

Part of https://github.com/orgs/kitodo/projects/8

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/test/utils/TestConstants.java
+++ b/Kitodo/src/test/java/org/kitodo/test/utils/TestConstants.java
@@ -1,0 +1,34 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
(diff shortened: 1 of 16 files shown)


Kitodo/src/test/java/org/kitodo/test/utils/TestConstants.java
```diff
@@ -0,0 +1,34 @@
...
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.test.utils;
+
+/**
+ * This class is used to group constants used in various tests in one place from which they can be retrieved.
+ */
+public class TestConstants {
```
henning-gerhardt, line 17: Maybe define this class as `public final class TestConstants`  like in https://www.baeldung.com/java-constants-good-practices described? There are some more notices to constant classes.

henning-gerhardt (approved):
Changes looks good. Only a small remark but not so important to request a change.

solth:
> Changes looks good. Only a small remark but not so important to request a change.

Thanks for the quick review. I will include your suggested change into this pull request since you already provided the solution and there is no extensive refactoring necessary. Additional optimizations for constant usage can be added in subsequent iterations and pull requests.

PR #6019, opened 2024-04-02
Show growl message when layout settings were loaded

solth:
This pull request adds a popup to the UI when opening a process in the metadata editor, informing the user when previously saved metadata layout settings for the current task were loaded and applied to make it more intelligible why the editor might look different in certain situations. 

Bildschirmfoto 2024-04-02 um 15 56 01


BartChris:
@solth I am not sure what is going on here, but the message is always in German for me, although all the other UI elements and notifications are in English. 

I debugged the code and when the message gets loaded 
https://github.com/solth/kitodo-production/blob/1584885f7cb74e79948aa6048f70ddf0f4f916eb/Kitodo/src/main/java/org/kitodo/production/helper/Helper.java#L508 `desiredLanguage` is always `de`. Switching the language multiple times does not change that. Can you reproduce that? Does it have something to do with the time, when this is loaded (during the view action)?

solth:
@BartChris thanks for pointing that out. I was able to reproduce the error - but the other way around! For me, the message stays in English, even if I switch to German. I think this might for some reason take the users browser or OS settings (which might be set to English on my current system) instead of the Kitodo settings. I will change this PR to draft until I can find an fix the problem.

solth:
@BartChris it seems the method `Helper.desiredLanguage()` - which is used by `getTranslation()` - can return the wrong local in some instances. I fixed the issue in this pull request by directly using the `LocaleHelper.currentLocale` to determine the correct current language and it seems to work. Do you agree with this solution?

BartChris (approved):
Works for me now! I am wondering wether we should make the message disappear after some time? It might get annoying if you always have to close it by hand. But maybe this is a tradeoff between information for non experienced users and power users.

PR #6020, opened 2024-04-02
Relocate dependencies

henning-gerhardt:
Relocate some XML based dependencies as needed for switch to Java 17 and Jakarta support without changing existing code.

### DIFF
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -77,7 +77,7 @@
         </dependency>
         <dependency>
             <groupId>org.glassfish</groupId>
-            <artifactId>javax.json</artifactId>
+            <artifactId>jakarta.json</artifactId>
         </dependency>
         <dependency>
             <groupId>org.hibernate</groupId>
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -81,6 +81,10 @@
         </dependency>
         <dependency>
             <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
             <artifactId>hibernate-c3p0</artifactId>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -501,6 +501,18 @@ from system library in Java 11+ -->
             </dependency>
             <dependency>
                 <groupId>org.hibernate</groupId>
+                <artifactId>hibernate-core</artifactId>
+                <version>${hibernate.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <!-- can be removed with Hibernate 6.x -->
+                        <groupId>javax.xml.bind</groupId>
+                        <artifactId>jaxb-api</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>org.hibernate</groupId>
                 <artifactId>hibernate-c3p0</artifactId>
                 <version>${hibernate.version}</version>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -54,7 +54,7 @@
         <hamcrest.version>2.2-rc1</hamcrest.version>
         <hibernate.version>5.6.10.Final</hibernate.version>
         <jaxb.api.version>2.3.3</jaxb.api.version>
-        <jaxb.glassfish-runtime.version>2.3.6</jaxb.glassfish-runtime.version>
+        <jaxb.glassfish-runtime.version>2.3.9</jaxb.glassfish-runtime.version>
         <jaxb2-basics-runtime.version>1.11.1</jaxb2-basics-runtime.version>
         <jaxen.version>1.2.0</jaxen.version>
         <jhove.version>1.20.1</jhove.version>
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
         <org.apache.avalon.framework.version>4.3.1</org.apache.avalon.framework.version>
         <hamcrest.version>2.2-rc1</hamcrest.version>
         <hibernate.version>5.6.10.Final</hibernate.version>
-        <jaxb.api.version>2.3.1</jaxb.api.version>
+        <jaxb.api.version>2.3.3</jaxb.api.version>
         <jaxb.glassfish-runtime.version>2.3.6</jaxb.glassfish-runtime.version>
         <jaxb2-basics-runtime.version>1.11.1</jaxb2-basics-runtime.version>
         <jaxen.version>1.2.0</jaxen.version>
@@ -226,8 +226,8 @@
             <!-- Include jaxb runtime dependencies which have been removed
 from system library in Java 11+ -->
             <dependency>
-                <groupId>javax.xml.bind</groupId>
-                <artifactId>jaxb-api</artifactId>
+                <groupId>jakarta.xml.bind</groupId>
+                <artifactId>jakarta.xml.bind-api</artifactId>
                 <version>${jaxb.api.version}</version>
             </dependency>
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -485,8 +485,8 @@ from system library in Java 11+ -->
             </dependency>
             <dependency>
                 <groupId>org.glassfish</groupId>
-                <artifactId>javax.json</artifactId>
-                <version>1.1.4</version>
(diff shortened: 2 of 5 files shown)


solth:
@thomaslow would you mind taking a look at this? I know you said you would not have time to do reviews for Kitodo at the moment we last talked, but since this is a smaller PR concerning a topic you have already invested some time investigating I think you'd be the best suited for this and perhaps might do an exception?

thomaslow:
Hi Henning, Arved asked me to review your pull request, because I previously experimented with upgrading Kitodo-Production to Java 17 and Tomcat 10. I do have a preliminary branch that allows to start Kitodo-Production with Java 17 and Tomcat 10 (untested unit and integration tests). However, I mixed changes related to Java 17 and Tomcat 10. I don't know which of those changes can be applied safely for Java 17 while keeping backwards compatibility to Java 11.

Anyway, I found this [Eclipse reference](https://wiki.eclipse.org/Jakarta_EE_Maven_Coordinates) that lists all Maven Package changes to the Jakarta-Namespace. Maybe you find that helpful.

henning-gerhardt:
@thomaslow : thanks for your input and your link which should be helpful. This pull request is independent of the Java 11-> 17 and Tomcat 9 -> 10 issue but is part of it. 

My approach was the following:

step one: update / relocate all the dependencies which can be  moved to the new groupId and / or arctifactId where the development of the dependencies is going on which must be done time by time.
step two: do the javax -> jakarta namespace renaming which is possible with Java 11. I don't know if this is possible with all the used dependencies or not
step three: do the Java 11 -> 17 needed update including the needed dependency updates
step four: do the Tomcat 9 -> 10 update

Step 3 and 4 can maybe switched but it is even possible that the javax to jakarta migration need even the Java 11 to 17 update if one or more dependencies made trouble.

I did myself an approach for the javax to jakarta migration (see https://github.com/kitodo/kitodo-production/compare/master...slub:kitodo-production:javax_to_jakarta) but I'm got some issues with Hibernate and I stopped the migration at this point as Matthias is doing the Hibernate Search integration right now and I don't want to made him more work to integrate my javax -> jakarta changes into his work.

thomaslow:
Hi Henning, are you still working on these changes or do you force-push your commits to be up-to-date with master?

henning-gerhardt:
If I'm working on this this pull request would not visible or on draft mode. I'm only refreshing all my pull request after changes on the main / master branch to keep them up to date.

thomaslow (approved):
I tested your changes on a local deployment of kitodo-production and could not observe any problems.

A minor note is that the change from `javax.xml.bind:jaxb-api:2.3.1` to `jakarta.xml.bind-api:2.3.3` is actually not making any difference in the final war file. The resolved dependencies are the same. The old `javax.xml.bind:jaxb-api:2.3.1` is still included indirectly via `hibernate-core:5.6.10.Final`. The newer dependency `jakarta.xml.bind-api:2.3.3` was already included before via `org.glassfish.jaxb:jaxb-runtime:2.3.6`.

Technically, having two jar files in the classpath that provide the same files for the same namespace (`javax.xml.bind`) is not a good idea. 

You could try to fix that by specifically excluding the old dependency `javax.xml.bind:jaxb-api:2.3.1` from the hibernate pom import. Of course, that would assume hibernate is compatible with the newer version. 

Personally, I would leave it as is. It seems to have worked fine for a long time already.

henning-gerhardt:
Thanks for hinting this out @thomaslow !

I will try to adjust this PR als dublicated / similar jar files are bad (jaxb-api) and if a dependency is already provided by an other this is not needed (even Maven is not using it).

henning-gerhardt:
@thomaslow : I updated my PR according to your suggestions. If one or both additional commits are not wanted I would remove them.

henning-gerhardt:
Updating `org.jvnet.jaxb` to version `2.0.9` including the replacement to use the commons-lang for generating the `equals` and ` hash` methods will reduce the error counting in CodeQL for this generated classes as this new generation way will generate Javadoc for this methods. I discovered this a few days ago. So this pull request did not only adjust the used dependencies it will reduce the CodeQL discovered issues.

Kitodo-DataManagement/pom.xml
```diff
@@ -77,7 +77,11 @@
         </dependency>
         <dependency>
             <groupId>org.glassfish</groupId>
```
stweil, line 79: This still needs an update.

pom.xml
```diff
@@ -479,8 +473,8 @@ from system library in Java 11+ -->
             </dependency>
             <dependency>
                 <groupId>org.glassfish</groupId>
-                <artifactId>javax.json</artifactId>
-                <version>1.1.4</version>
+                <artifactId>jakarta.json</artifactId>
+                <version>1.1.6</version>
             </dependency>
```
stweil, line 478: I use this new code:
```
            <dependency>
                <groupId>jakarta.json</groupId>
                <artifactId>jakarta.json-api</artifactId>
                <version>2.1.3</version>
            </dependency>
            <dependency>
                <groupId>org.eclipse.parsson</groupId>
                <artifactId>jakarta.json</artifactId>
                <version>1.1.6</version>
            </dependency>
```

stweil (changes requested):

pom.xml
henning-gerhardt: Sorry, you missed the point: this is not a pull request to push the dependencies to latest version. I just updated the artifact and group id to the new used names without breaking any compatibility with the current code and provide a way to use renamed JavaEE namespace in future.  This is only one step in a list of many steps and not the end.

stweil (approved):
I think this PR can be merged.

pom.xml
```diff
@@ -220,11 +219,6 @@
...
             <!-- Include jaxb runtime dependencies which have been removed
```
solth, line 220: Do we still need this comment when the corresponding dependency is removed?

solth (commented):
Looks good, just one small remark, see below.

solth:
> Your exclusion rule for `javax.xml.bind:jaxb-api:2.3.1` works for me. Now, the final war file only contains the newer jakarta jar.
> 
> I'm not sure whether the explicit pom dependency declaration for `jakarta.xml.bind-api:2.3.3` should be removed or not. What is Kitodo's general strategy here? Removing it means that there is one less dependency declaration, meaning pom files would be more compact and there would be one less specific artifact version, which has to be maintained. On the other hand, the source code explicitly imports classes from this package. Because of that, I think it would generally be a good idea to keep this dependency in pom files, even though it would not be strictly necessary due to the transitive dependency via glassfish or hibernate.
> 
> The recommended strategy of the official [`maven-dependency-plugin`](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-mojo.html) is that all artifacts whose classes are explicitly imported in source code should also be listed as dependencies in the pom file, even if they are included transitively via other artifacts. The corresponding maven command `mvn dependency:analyze` reports theses missing dependencies declarations as warnings, now also including `jakarta.xml.bind-api:2.3.3`, amongst many other similar warnings.
> 
> @solth What do you think?

Thanks a lot for the explanation @thomaslow I understand that both approaches have their advantages (streamlining pom files vs. explicit dependencies for directly used packages and classes) and would agree with you that we should stick with keeping those dependencies in the pom file, especially when this is an officially recommended strategy.

pom.xml
henning-gerhardt: I don't know if this comment is only for the direct following mentioned dependency or more. I can remove it if you want.

henning-gerhardt:
> Thanks a lot for the explanation @thomaslow I understand that both approaches have their advantages (streamlining pom files vs. explicit dependencies for directly used packages and classes) and would agree with you that we should stick with keeping those dependencies in the pom file, especially when this is an officially recommended strategy.

@solth : Did I understand you correct, that I should revert my commit https://github.com/kitodo/kitodo-production/pull/6020/commits/f46777d80d9ce4eadc919d1ef338f726832a44cb in total? If so then your commenting is obsolete or?

solth:
> @solth : Did I understand you correct, that I should revert my commit [f46777d](https://github.com/kitodo/kitodo-production/commit/f46777d80d9ce4eadc919d1ef338f726832a44cb) in total? If so then your commenting is obsolete or?

@henning-gerhardt yes, please revert that commit.

henning-gerhardt:
> > @solth : Did I understand you correct, that I should revert my commit [f46777d](https://github.com/kitodo/kitodo-production/commit/f46777d80d9ce4eadc919d1ef338f726832a44cb) in total? If so then your commenting is obsolete or?
> 
> @henning-gerhardt yes, please revert that commit.

Done. I must rebase the branch in any way.

PR #5998, opened 2024-03-07
Allow multiple values for same key in mass import csv

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/5996

PR #6031, opened 2024-04-11
[3.6] Backport of - Allow multiple values for same key in mass import csv 

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/5998


BartChris:
I have no idea why the selenium tests for the Calendar are failing in the CI, locally they seem to work. I have not been able to fix them.

```
Error:    CalendarST.createProcessFromCalendar Multiple Failures (2 failures)
	org.awaitility.core.ConditionTimeoutException: Condition with alias 'Wait for openCalendarLink to be clickable' didn't complete within 5 seconds because condition with lambda expression in org.kitodo.selenium.testframework.pages.ProcessesPage that uses org.openqa.selenium.WebElement was not fulfilled.
	org.openqa.selenium.NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":"#editForm\:cancel"}
  (Session info: chrome=123.0.6312.122)
```

oliver-stoehr:
I tried to reproduce the error on my local machine, but was not able to trigger the exception reliably.
Since this error occurs in `NewspaperProcessesGeneratorIT` it should be independent of the calendar selenium tests.
However could not find the exact source of this error. It might be the result of another test not cleaning up properly/modifying the file `…/10/meta.xml` used in this test.

BartChris:
So the problem came from the new test in the Import Service integration tests. The tests changed the meta/10.xml so i reset this file after all tests of this class are finished. The other changes to the Calendar tests are not strictly necessary for this PR but since the Selenium tests failed two times i think it is maybe a good idea to bring them in as well.

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/CalendarPage.java
```diff
@@ -257,51 +190,69 @@ public List<String> getMetadata(String issueName) {
      * Click cancel button and leave calendar.
      */
     public void closePage() {
-        buttonCancel.click();
+        getById(BUTTON_CANCEL).click();
+    }
+
+    public int countIssues() {
```
solth, line 196: I think this method should be removed as it seems never to be used.

solth (approved):

PR #5784, opened 2023-09-13
[hibernate search] Remove Index Must Be Updated warning and blank indexing page

matthias-ronge:
Issue #5760 1b)

Issue #5760, opened 2023-08-23
Replace native ElasticSearch integration with Hibernate Search


matthias-ronge:
This development aims to replace the native ElasticSearch integration with the Hibernate Search framework, which uses ElasticSearch. Little shall change for those starting the application out.

The exchange should take place in two steps:

**1. A version is created in which the existing ElasticSearch integration has been removed. This requires the following steps:**

- [ ] 1a) remove test for indexing
- [ ] 1b) remove indexing page
- [ ] 1c) remove Hibernate usage (remove DTO objects, unify saveToDatabase and saveToIndex in save function)
- [ ] 1d) remove database column `indexAction` from all tables and objects

**Goal:** The application runs alone with the database. Only the filters and advanced search may not work.

**2. A version is created in which Hibernate Search has been implemented. This requires the following steps:**

- [ ] 2a) Insert Hibernate Search framework. Annotate table columns.
- [ ] 2b) Implement bulk indexing (reimplement indexing page)
- [ ] 2c) Implement / customize search. The filters are working again.
- [ ] 2d) The advanced search works again.
- [ ] 2e) Search in non-database data (METS files) in a meaningful way
- [ ] 2f) optimization of the search

**Goal is to fix a lot of errors and problems with indexing, see #5546.**

_This development, funded by the Kitodo e.V. Development Fund_

stweil:
> This development aims to replace the native ElasticSearch integration with the Hibernate Search framework, which uses ElasticSearch. Little shall change for those starting the application out.

Isn't there consense that Elasticsearch shall be replaced by OpenSearch? Then the text should be "[...] which uses OpenSearch".

henning-gerhardt:
> > This development aims to replace the native ElasticSearch integration with the Hibernate Search framework, which uses ElasticSearch. Little shall change for those starting the application out.
> 
> Isn't there consense that Elasticsearch shall be replaced by OpenSearch? Then the text should be "[...] which uses OpenSearch".

There is only the consense to use ElasticSearch or OpenSearch with HibernateSearch.

stweil:
Then the text should be "[...] which works with either Elasticsearch or OpenSearch".

@matthias-ronge, what is the status of this pull request and the whole work on support for HibernateSearch?

solth:
@stweil We are in the middle of the review. If you want to support us with the review, you are very welcome to do so.

BartChris:
[...]

PR #6032, opened 2024-04-12
[hibernate search] Introduce a common interface for DTO objects and data beans

matthias-ronge:
Issue #5760 1c) -- part 1

Follow-up pull request to #5784 ([immediate diff](/matthias-ronge/kitodo-production/compare/5760_1b...5760_1c_1))

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -139,13 +139,13 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
     private List<Map<String, Object>> metadata;
 
     @Transient
-    private Integer numberOfMetadata;
+    private int numberOfMetadata;
 
     @Transient
-    private Integer numberOfImages;
+    private int numberOfImages;
 
     @Transient
-    private Integer numberOfStructures;
+    private int numberOfStructures;
 
     @Transient
     private String baseType;
@@ -622,6 +622,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
      * Get OCR-D workflow identifier.
      *
      * @return The OCR-D workflow identifier
+     * @see <a href="https://ocr-d.de/en/workflows">OCR-D Workflow Guide</a>
      */
     public String getOcrdWorkflowId() {
         return ocrdWorkflowId;
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -430,7 +430,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
     }
 
     /*
-     * If list is empty just set, if not first clear and next set.
+     * Set batches, if list is empty just set, if not first clear and next set.
      */
     @Override
     @SuppressWarnings("unchecked")
@@ -459,8 +459,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
     /**
      * Set comments.
      *
-     * @param comments
-     *            as List of Comment objects
+     * @param comments as List of Comment objects
      */
     public void setComments(List<Comment> comments) {
         this.comments = comments;
@@ -493,8 +492,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
     /**
      * Set exported.
      *
-     * @param exported
-     *            as boolean
+     * @param exported as boolean
      */
     public void setExported(boolean exported) {
         this.exported = exported;
@@ -512,8 +510,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
     /**
      * Set metadata.
      *
-     * @param metadata
-     *            as Map
+     * @param metadata as Map
      */
     public void setMetadata(List<Map<String, Object>> metadata) {
         this.metadata = metadata;
@@ -551,8 +548,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
     /**
      * Set inChoiceListShown.
      *
-     * @param inChoiceListShown
-     *            as java.lang.Boolean
+     * @param inChoiceListShown as java.lang.Boolean
      */
     public void setInChoiceListShown(Boolean inChoiceListShown) {
         this.inChoiceListShown = inChoiceListShown;
@@ -560,9 +556,8 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
 
     /**
      * Determines whether or not two processes are equal. Two instances of
-     * {@code Process} are equal if the values of their {@code Id},
-     * {@code Title}, {@code OutputName} and {@code CreationDate} member fields
-     * are the same.
+     * {@code Process} are equal if the values of their {@code Id}, {@code Title},
+     * {@code OutputName} and {@code CreationDate} member fields are the same.
      *
      * @param object
      *            An object to be compared with this {@code Process}.
@@ -622,7 +617,6 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
      * Get OCR-D workflow identifier.
      *
      * @return The OCR-D workflow identifier
-     * @see <a href="https://ocr-d.de/en/workflows">OCR-D Workflow Guide</a>
      */
     public String getOcrdWorkflowId() {
         return ocrdWorkflowId;
@@ -632,7 +626,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
      * Set the OCR-D workflow identifier.
      *
      * @param ocrdWorkflowId
-     *            The identifier of the OCR-D workflow
+     *         The identifier of the OCR-D workflow
      */
     public void setOcrdWorkflowId(String ocrdWorkflowId) {
         this.ocrdWorkflowId = ocrdWorkflowId;
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -126,13 +126,13 @@ public class Process extends BaseTemplateBean {
     private List<Map<String, Object>> metadata;
 
     @Transient
-    private int numberOfMetadata;
+    private Integer numberOfMetadata;
 
     @Transient
-    private int numberOfImages;
+    private Integer numberOfImages;
 
     @Transient
-    private int numberOfStructures;
+    private Integer numberOfStructures;
 
     @Transient
     private String baseType;
@@ -606,7 +606,7 @@ public class Process extends BaseTemplateBean {
      *
      * @return Amount of structure elements
      */
-    public int getNumberOfStructures() {
+    public Integer getNumberOfStructures() {
         return numberOfStructures;
     }
 
@@ -615,7 +615,7 @@ public class Process extends BaseTemplateBean {
      *
      * @return Amount of meta data elements
      */
-    public int getNumberOfMetadata() {
+    public Integer getNumberOfMetadata() {
         return numberOfMetadata;
     }
 
@@ -624,7 +624,7 @@ public class Process extends BaseTemplateBean {
      *
      * @param numberOfMetadata Integer value of amount of meta data elements
      */
-    public void setNumberOfMetadata(int numberOfMetadata) {
+    public void setNumberOfMetadata(Integer numberOfMetadata) {
         this.numberOfMetadata = numberOfMetadata;
     }
 
@@ -633,7 +633,7 @@ public class Process extends BaseTemplateBean {
      *
      * @return Integer value of amount of images
      */
-    public int getNumberOfImages() {
+    public Integer getNumberOfImages() {
         return numberOfImages;
     }
 
@@ -642,7 +642,7 @@ public class Process extends BaseTemplateBean {
      *
      * @param numberOfImages Integer value of amount of images
      */
-    public void setNumberOfImages(int numberOfImages) {
+    public void setNumberOfImages(Integer numberOfImages) {
         this.numberOfImages = numberOfImages;
     }
 
@@ -651,7 +651,7 @@ public class Process extends BaseTemplateBean {
      *
      * @param numberOfStructures Integer value of amount of structure elements
      */
-    public void setNumberOfStructures(int numberOfStructures) {
+    public void setNumberOfStructures(Integer numberOfStructures) {
         this.numberOfStructures = numberOfStructures;
     }
 
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
+import javax.el.PropertyNotWritableException;
 import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
@@ -38,6 +39,7 @@ import org.kitodo.data.database.persistence.HibernateUtil;
 import org.kitodo.data.database.persistence.ProcessDAO;
 import org.kitodo.data.elasticsearch.index.converter.ProcessConverter;
 import org.kitodo.data.interfaces.BatchInterface;
+import org.kitodo.data.interfaces.ClientInterface;
 import org.kitodo.data.interfaces.DocketInterface;
 import org.kitodo.data.interfaces.ProcessInterface;
 import org.kitodo.data.interfaces.ProjectInterface;
@@ -277,6 +279,23 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
         this.project = (Project) project;
     }
 
+    /**
+     * Specifies the project to which the process belongs.
+     * 
+     * <p>
+     * <b>API Note:</b><br>
+     * This function exists because Faces does not recognize the more generic
+     * function {@link #setProject(ProjectInterface)} as a setter for the
+     * property {@code project} and otherwise throws a
+     * {@link PropertyNotWritableException}.
+     *
+     * @param project
+     *            project to which the process should belong
+     */
+    public void setProject(Project project) {
+        this.project = project;
+    }
+
     @Override
     public Ruleset getRuleset() {
         return this.ruleset;
@@ -287,6 +306,23 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
         this.ruleset = (Ruleset) ruleset;
     }
 
+    /**
+     * Sets the business domain specification this process is using.
+     * 
+     * <p>
+     * <b>API Note:</b><br>
+     * This function exists because Faces does not recognize the more generic
+     * function {@link #setRuleset(RulesetInterface)} as a setter for the
+     * property {@code ruleset} and otherwise throws a
+     * {@link PropertyNotWritableException}.
+     *
+     * @param ruleset
+     *            the business domain specification
+     */
+    public void setRuleset(Ruleset ruleset) {
+        this.ruleset = ruleset;
+    }
+
     @Override
     public Docket getDocket() {
         return docket;
@@ -295,6 +331,24 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
     @Override
     public void setDocket(DocketInterface docket) {
         this.docket = (Docket) docket;
+    }
+
+    /**
+     * Sets the docket generation statement to use when creating a docket for
+     * this process.
+     * 
+     * <p>
+     * <b>API Note:</b><br>
+     * This function exists because Faces does not recognize the more generic
+     * function {@link #setDocket(DocketInterface)} as a setter for the property
+     * {@code docket} and otherwise throws a
+     * {@link PropertyNotWritableException}.
+     *
+     * @param docket
+     *            the docket generation statement
+     */
+    public void setDocket(Docket docket) {
+        this.docket = docket;
     }
 
     /**
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -19,7 +19,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
-import javax.el.PropertyNotWritableException;
 import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
@@ -287,7 +286,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
      * This function exists because Faces does not recognize the more generic
      * function {@link #setProject(ProjectInterface)} as a setter for the
      * property {@code project} and otherwise throws a
-     * {@link PropertyNotWritableException}.
+     * {@code PropertyNotWritableException}.
      *
      * @param project
      *            project to which the process should belong
@@ -314,7 +313,7 @@ public class Process extends BaseTemplateBean implements ProcessInterface {
      * This function exists because Faces does not recognize the more generic
      * function {@link #setRuleset(RulesetInterface)} as a setter for the
      * property {@code ruleset} and otherwise throws a
-     * {@link PropertyNotWritableException}.
+     * {@code PropertyNotWritableException}.
      *
      * @param ruleset
      *            the business domain specification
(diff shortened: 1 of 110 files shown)


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
```diff
@@ -17,7 +17,9 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
+import javax.el.PropertyNotWritableException;
```
henning-gerhardt, line 22: Backport comment from #6085 : Instead of using this exception from the javax namespace an exception class from the Kitodo namespace should be used. Maybe this new exception class is extending this class and if so the needed javax -> jakarta change for this exception class in the near future would only happen at one file instead of many places / files.

I'm writing this only on this place but should be changed on all places.

Kitodo-DataManagement/src/main/java/org/kitodo/data/interfaces/WorkflowInterface.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.data.interfaces;
+
+import java.util.Objects;
+
+import org.apache.logging.log4j.util.Strings;
```
henning-gerhardt, line 16: Backport of comment from #6085 : 

A string comparing method from the log4j.util.Strings package is used here instead of using a similar method from the commons.lang3.StringUtils which is more common in use.

I'm writing this only on this place but should be changed on all places.

Kitodo-DataManagement/src/main/java/org/kitodo/data/interfaces/WorkflowInterface.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param status
+     *            as String
+     * @throws IllegalArgumentException
+     *             if {@link WorkflowStatus} has no constant with the specified
+     *             name
+     * @deprecated Use {@link #setStatus(WorkflowStatus)}.
+     */
+    @Deprecated
+    default void setWorkflowStatus(String status) {
+        setStatus(Strings.isNotEmpty(status) ? WorkflowStatus.valueOf(status) : null);
```
henning-gerhardt, line 85: Backport of comment from #6085:

Maybe replace `String.isNotEmpty` from log4j package with a `! StringUtils.isBlank` from commons.lang3.StringUtils package?

I'm writing this only on this place but should be changed on all places.

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -332,35 +312,68 @@ public TaskDTO convertJSONObjectToDTO(Map<String, Object> jsonObject, boolean re
...
      * @param jsonObject the jsonObject retrieved from the ElasticSearch index for a task
-     * @param taskDTO the taskDTO
+     * @param taskInterface the taskInterface
      */
-    private void convertTaskProjectFromJsonObjectToDTO(Map<String, Object> jsonObject, TaskDTO taskDTO) throws DataException {
-        ProjectDTO projectDTO = new ProjectDTO();
-        projectDTO.setId(TaskTypeField.PROJECT_ID.getIntValue(jsonObject));
-        projectDTO.setTitle(TaskTypeField.PROJECT_TITLE.getStringValue(jsonObject));
-        taskDTO.setProject(projectDTO);
+    private void convertTaskProjectFromJsonObjectToInterface(Map<String, Object> jsonObject,
```
henning-gerhardt, line 370: You are naming the variables by the class name. This could be done but it is not so pretty. Instead if naming the variable `taskInterface` a `task` variable from the  `TaskInterface` class should be more readable.

Maybe even do not mention the 'interface' using in method names. Is there no better naming / wording for methods and variables possible where the interface classes are used? 

I'm writing this only on this place but should be changed on all places.

Kitodo/src/test/java/org/kitodo/selenium/LoginST.java
```diff
@@ -1,44 +0,0 @@
-/*
```
henning-gerhardt, line 1: Is removing this selenium test intended? The login itself should be tested or not?

henning-gerhardt (changes requested):
This is a code reading review only. I did not the the functionality itself nor I'm discussing the system architecture changes.

Kitodo-DataManagement/src/main/java/org/kitodo/data/interfaces/WorkflowInterface.java
matthias-ronge: This is actually always a mistake. Maybe isomeone one day feels compelled to create a checkstyle rule against this?

Kitodo-DataManagement/src/main/java/org/kitodo/data/interfaces/WorkflowInterface.java
matthias-ronge: As far as I can see, this is the more correct choice in all places.

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
matthias-ronge: These names came about through a RegEx replace from DTO to Interface. I completely agree that variables shouldn't have …interface in their names. I have completely changed this.

Kitodo/src/test/java/org/kitodo/selenium/LoginST.java
matthias-ronge: Yes, this test abolition is intentional.

The test checked whether the login was aborted with a warning, if the index was incomplete. This was a bad situation because the application display was based on the index and so incorrect or incomplete data was displayed. This situation will no longer exist in the future, as indexing will only be effective for searching metadata (and possibly speeding up the database search for special columns or queries), but in future, the main application will work without an index.

Whether a login is generally possible is tested in _every_ Selenium test, as the user must log in first.

Kitodo-DataManagement/src/main/java/org/kitodo/data/interfaces/WorkflowInterface.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param status
+     *            as String
+     * @throws IllegalArgumentException
+     *             if {@link WorkflowStatus} has no constant with the specified
+     *             name
+     * @deprecated Use {@link #setStatus(WorkflowStatus)}.
+     */
+    @Deprecated
+    default void setWorkflowStatus(String status) {
+        setStatus(StringUtils.isNotBlank(status) ? WorkflowStatus.valueOf(status) : null);
```
henning-gerhardt, line 85: `StringUtils.isBlank` and `StringUtils.isNotBlank` are checking for white spaces. Is this white space check needed? If not `StringUtils.isEmpty` and ` StringUtils.isNotEmpty` is much faster.

The maybe wrong usage is in more places added but I'm writing this only once.

Maybe through my previous comment for the log4j String usage and a wrong example the usage is here used. Sorry, my fault :-(

Kitodo/src/main/java/org/kitodo/production/converter/IllustratedSelectItemConverter.java
```diff
@@ -30,7 +30,7 @@ public class IllustratedSelectItemConverter implements Converter, Serializable {
 
     @Override
     public Object getAsObject(FacesContext context, UIComponent component, String value) {
-        if (StringUtils.isNotEmpty(value)) {
+        if (StringUtils.isNotBlank(value)) {
```
henning-gerhardt, line 33: Is this and even on other places the replacement from `isNotEmpty` to `isNotBlank` indented? `isNotEmpty` is much faster and did not allow white spaces,  `isNotBlank` is even checking for white space usage and should only used if white spaces should be allowed.

There are more places where this replacement is happening but I'm writing this only once.

Kitodo/src/test/java/org/kitodo/production/services/data/TemplateServiceIT.java
```diff
@@ -92,12 +92,12 @@ public void shouldCheckForUnreachableTasksWithException() throws Exception {
 
     @Test
     public void shouldHasCompleteTasks() throws Exception {
-        TemplateDTO templateDTO = templateService.findById(1);
-        boolean condition = templateService.hasCompleteTasks(templateDTO.getTasks());
-        assertTrue("Process DTO doesn't have complete tasks!", condition);
+        TemplateInterface template = templateService.findById(1);
+        boolean condition = templateService.hasCompleteTasks(template.getTasks());
+        assertTrue("Process Interface doesn't have complete tasks!", condition);
```
henning-gerhardt, line 97: Even you only changed from `DTO` to `Interface` inside the message maybe remove even the `Interface` so the error message is more readable like "Process doesn't have complete tasks!" as the used object type is not needed.

Kitodo/src/test/java/org/kitodo/production/services/data/TemplateServiceIT.java
```diff
@@ -92,12 +92,12 @@ public void shouldCheckForUnreachableTasksWithException() throws Exception {
...
 
-        templateDTO = templateService.findById(3);
-        condition = templateService.hasCompleteTasks(templateDTO.getTasks());
-        assertFalse("Process DTO has complete tasks!", condition);
+        template = templateService.findById(3);
+        condition = templateService.hasCompleteTasks(template.getTasks());
+        assertFalse("Process Interface has complete tasks!", condition);
```
henning-gerhardt, line 101: See my comment above but with a different error message: "Process has complete tasks!"

Kitodo/src/main/java/org/kitodo/production/converter/IllustratedSelectItemConverter.java
matthias-ronge: In all places where a string is checked for emptiness, something is done with it afterwards if it is not empty. For example, an `enum` constant is looked up from it. I have not found a case where an empty string should be processed differently than a string that only contains non-printing characters. Therefore, I think `isNotBlank()` is the right function here. Correct me if I am wrong.

Regarding the runtime: If the string is `null` or empty, it is identical, if it consists of non-printing characters, it is theoretically O(𝑛), but if the first printing character is found, the test stops anyway, so I assume a runtime of O(1). In any case, this is nothing that should noticeably slow down the application!

Kitodo/src/test/java/org/kitodo/production/services/data/TemplateServiceIT.java
matthias-ronge: You're right, that's ugly. I changed it.

Kitodo/src/main/java/org/kitodo/production/converter/IllustratedSelectItemConverter.java
henning-gerhardt: In case that the value is  `null` or an empty string the behaviour is identical. If the first character is a non white space there is one iteration in the `isBlank` method which can be ignored. The case that in the beginning of a string a lot of white spaces occur is not so reasonable/realistic in most usage scenarios.

So ignore my comment regarding the `isBlank` / `isEmpty` methods. I must read the wrong method in the past how the `isBlank` method is working.

oliver-stoehr (commented):
The code style looks good, the amount of comments is very nice.

Can you explain the purpose of this common interface? I've looked into HibernateSearch, but not in depth, yet. My understanding was, that one of the main advantages is, that HibernateSearch can work directly with the beans and does not need any extra layer. So my hope was, that we could remove the DTO objects without a replacement. 
Adding a new interface seems to me like just "shifting" the amount and complexity from our existing DTO objects to the new interface. But since I did not fully understand how HibernateSerdach works, I might be missing something essential!

matthias-ronge:
Yes, the DTOs shall be deleted, and many more files will be deleted later, but to reduce the pull requests’ size, I didn’t delete them yet. There will be deleted in a separate pull request.
Yes, it is possible to delete the interface completely then, which would mean to move the Javadoc comments to the beans and remove the interface. This is a decision of our own taste as developers. First, introducing the interface was helpful for me to see and get the beans and DTOs really work equally (they did for 96%, but there were inconsitencies). So if there will be a majority for removing the interface later, I can do that. But it should happen after the other removals have happened.

henning-gerhardt:
I don't think that is a good idea to access the Hibernate data directly in the UI - from a security view. There was many exploits in the used UI frameworks, bad written UI code and other mistakes which made it possible to access the database through the used ORM like Hibernate and UI to access (read / write / delete) the database data. From this point I would not recommend the way to expose the Hibernate objects to the UI and use a wrapping mechanism (like used POJO's / DTO's right now) to made the database data available in the UI. So removing them is a bad idea from my view of security to web applications.

matthias-ronge:
You were right about what you say, but unluckily, this happens anyway in Kitodo.Production, and the DTO objects here have a different meaning. They were the database objects, when loaded from index, for filling the tables. Removing the index objects when deleting ist straight foreward.

Kitodo-DataManagement/src/main/java/org/kitodo/data/interfaces/BatchInterface.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
```
stweil, line 9: Es wäre schön, wenn wir uns von diesem Text, der leider momentan durchgängig verwendet wird, lösen und statt dessen den empfohlenen und üblichen Text für die GPL verwenden könnten.

Rechtlich macht das keinen Unterschied, aber perspektivisch könnte man die GPL3-License.txt loswerden. Und prinzipiell halte ich es für wünschenswert, sich an vertraute allgemeine Standards zu halten statt eigene zu erfinden.

stweil: Siehe auch PR #6110.

stweil (changes requested):
Ich schreibe deutsch, weil vermutlich an der Diskussion keine englischsprachigen TN beteiligt sind und weil ich glaube, dass es so für alle einfacher ist.

Mit 22 Commits ist dieser PR ziemlich umfangreich und nicht so einfach zu prüfen. Mein erster Eindruck ist, dass etliche Commits zusammengefasst werden sollten.

Eine Grundanforderung ist, dass jeder einzelne Commit funktionierenden Code als Ergebnis haben muss. Wird dies nicht eingehalten, so macht das beispielsweise ein `git bisect` unnötig schwer. Ein Commit mit Änderungen und ein weiterer Commit, der das nachträglich korrigiert, sollten also zusammengefasst werden.

Die gewünschte Funktionalität – Umstellung auf Hibernatesearch – ist auf mehrere Pull Requests verteilt. Das ist in Ordnung. Aber ich würde gerne sehen, was letztlich das Endergebnis sein wird. Dafür benötige ich einen Branch oder einen Draft Pull Request, der zeigt, wie dieses Endergebnis aussieht und mit dem ich lokal die komplette neue Funktionalität ausprobieren kann. Ist das bereits möglich, und ich habe nur übersehen, wie es geht?

Kitodo-DataManagement/src/main/java/org/kitodo/data/interfaces/BatchInterface.java
matthias-ronge: Das hat mit diesem Pull Request aber nichts zutun. Sollten wir an anderer Stelle besprechen

matthias-ronge:
@stweil Generell stimme ich deiner Meinung zu. Allerdings ist diese Entwicklung zu groß, um Commits nur in dem Moment durchzuführen, in dem der Code funktioniert. Zum Review schlage ich vor, dass du die Commits ignorierst und nur den resultierenden Code gegenliest. Vielleicht ist jeder der Pull Requests für die Hibernate-Search-Entwicklung, so wie du es denkst, „ein Commit“.

Es sollte möglich sein, den Branch auszuchecken und auszuführen. Jeder Pull Request löst eine bestimmte Aufgabe, dieser hier führt „nur“ die Schnittstellenschicht ein. Es sollten also beim Ausführen des Codes keine funktionalen Unterschiede zu erkennen sein.

solth:
I understand @henning-gerhardt's concern but agree with @matthias-ronge that the DTOs in Kitodo.Production never really offerend any additional security. If I understand the concept of DTOs correctly they are actually not really intended as a security factor anyway, but rather to keep a low data profile when transfering objects between a client and a server in a network is expensive. That is how they are described on [Wikipedia](https://en.wikipedia.org/wiki/Data_transfer_object) and by their original designer Martin Fowler [here](https://martinfowler.com/bliki/LocalDTO.html), at least. 

Wikipedia, for example, states that "This pattern is often incorrectly used outside of remote interfaces." Martin Fowler writes "DTOs are called Data Transfer Objects because their whole purpose is to shift data in expensive remote calls." and further "Not just do you not need them in a local context, they are actually harmful both because a coarse-grained API is more difficult to use and because you have to do all the work moving data from your domain or data source layer into the DTOs." (the article linked above is old, but a good read and contains more interesting remarks about the concept of data transfer objects!) I therefor think that DTOs were not correctly used in Kitodo.Production to begin with, so getting rid of them is a good first step to reduce unnecessary complexity, IMHO.

I also agree with @oliver-stoehr that replacing each DTO with a corresponding new interface should be avoided and just shifts the code overhead to a differently named class. Instead, we should take this opportunity to create a cleaner implementation and take advantage of HibernateSearch's ability to directly manage index and database objects at the same time. Since @matthias-ronge already mentionend that removing the interfaces is an option for a follow-up pull request I can accept introducing them temporarily, even though I am not convinced it is necessary or the best approach. I think it might complicate the implementation of the next pull requests in the HibernateSearch project, but in the end that is up to Matthias.

One little remark concerning these individual pull requests against the `hibernate-search` branch: for these the reviews should focus on the larger concept of the development and not on code style details like method length, missing java docs etc. These more formal criteria should be checked when the final pull request from the `hibernate-search` branch back against the `master` branch is opened.

solth (commented):
I have a hard time seeing the point of these new data interfaces. After inspecting the files I am not convinced they provide any new value, but just unnecessarily increase the complexity. Until the DTO classes are removed, the data interfaces might be implemented by two classes, each, but in my opinion the introduction of the interfaces is redundant and not a good idea. 

Interfaces in Java should be used to give developers a choice which implementation of the interface they choose - for example `java.util.ArrayList` or `java.util.LinkedList` as specific implmentations of the interface `java.util.LinkedList` - when passing parameters to a method that specifies an interface as a parameter. 

In this case (Kitodo) however, once the DTOs are removed, a developer will always want to instantiate for example just a `Process`, not a `ProcessInterface`. The individual entities in Kitido.Production, like Projects, Templates, Processes, Tasks etc. are clearly defined. There won't be alternative implementations for these concepts, so in my opinion it doesn't make sense to introduce these more general interfaces here which would suggest alternative implementations are desirable.

Since these preliminary reviews are not supposed to check the code style or implementation details, but rather to determine the legitimacy of fundamental concept, I think it has to be a valid outcome of the review to reject the proposed solution, if the reviewer believes it goes in the wrong direction.

I would be willing to approve these changes under the clear condition that they are just a temporary solution and will be removed together with the DTO classes in a follow-up pull request.

I would be very interested to read what the other developers think about this approach. @henning-gerhardt , @thomaslow , @Erikmitk , @oliver-stoehr , @BartChris , what is your opinion on this?

henning-gerhardt:
I agree with the  @solth opinion regarding to introducing of the new interfaces. I see even no benefits or advantages to introduce them at this point as I did not see any different implementations of this classes in the future as they only used internal / inside the core application itself. If I missed the point than I need at least an explanation why they are really needed.

thomaslow:
I quickly skipped over the proposed changes. I agree that introducing interface classes that have a 1:1 relationship to their only subclasses is not really necessary. Usually, I like interfaces. In this case, however, the contract between the business logic and the database of Kitodo.Production is already achieved through the ORM model with hibernate beans. In my opinion, there are no additional interface classes necessary.

In the case of DTO classes, I partially agree with @henning-gerhardt. There definitely should be DTO classes. However, the existing DTO classes are not relevant anymore, since they were used between ElasticSearch and the business logic. I believe, this part (choosing the fields that are being indexed and converted to JSON) can now be controlled via Hibernate-Search. Based on the [@Indexed](https://docs.jboss.org/hibernate/stable/search/api/org/hibernate/search/mapper/pojo/mapping/definition/annotation/Indexed.html) annotation, Hibernate-Search will most certainly generate a similar JSON document as before, which provides the same security as before, meaning sensitive information (like passwords) are not leaked to the search engine. Therefore, the existing DTO classes may be removed.

Instead, someday, _in a galaxy far far away..._, there should be DTO classes between the business logic and the user interface, such that it would not be possible anymore to trigger numerous ORM-based SQL queries from the user interface and potentially build a REST API that can be used to move Kitodo towards the current generation of web technologies. However, this has nothing to do with the migration to Hibernate-Search and should not be part of this pull request.

Erikmitk:
I don't have anything to add. I agree with the comments made by @henning-gerhardt, @thomaslow and @solth that an additional interfaces makes no sense in this context. :)

matthias-ronge:
There are two groups of classes, database objects and index objects. (Here, I have the impression that we think the same thing, they are _called_ DTOs but the name is not _the concept behind DTO_, so here it is just called that.)

The introduction of the interface is aimed at standardizing the "almost"—but not everywhere exactly—equal classes of objects from the database and objects from the index so that they become interchangeable, and is documentation of what the methods must return. This was also necessary in order to be able to achieve the alignment.

It is a basic tool of refactoring to first encapsulate the section to be newly implemented in an interface, then you can use it based on the interface and reimplement it. This is what happened here.

After the work has been done successfully, there is the option of keeping the interface or removing it. This is undecided here, I can be happy with both.

solth:
@matthias-ronge I understand what the intention of these new data interfaces is, but they are unecessary in the long run, as discussed above. When the DTO classes are removed, the interfaces are obsolete and should be removed. Since all reviewers seem to agree on this, I would say this is very much decided. 

This means they can only be temporary addition. Therefore I think first opening a large pull request introducing a temporary construct and later opening more pull requests that remove this construct again is not very helpful and just multiplies the workload for you to maintain the individual branches and PRs and for the reviewers to perform uncessary reviews.

We did agree on splitting all changes of the hibernate search implementation into multiple pull requests to reduce the size of each individual PR, but the individual pull requests should all just introduce parts that are required for the final implementation and not contain large refactorings of the architecture that are later reverted.

Additionally, I don't think it makes sense to concentrate on fixing all tests for intermediate development states. Builds for pull requests against the `hibernate-search` branch do not have to pass all checks, in my opinon. Those checks are only necessary to pass once the final pull request against the master is openend.

matthias-ronge:
In the early days of 3.x development, we placed a lot of emphasis on mapping all functionality using interfaces. When we were removing the legacy UGH library, I was even explicitly asked to represent the entire interface of the library as a graphic, which took me eight hours.

I observe that the strategy for dealing with interfaces is changing. I know that there are reasons for and against them, and it doesn't bother me if we discard them in the end. However, at this point in development, they were necessary because: The beans and the DTO objects were largely congruent, but they differed in some details.
* the DTO objects returned strings or integers in some places, or wanted them as setters, where the beans return database objects
* the DTO objects returned true or false (present or null/empty) in some places, where the beans contain database objects
* the DTO objects returned information that was stored in the database in a different way.

With the introduction of the interfaces, I smoothed out these points and made the objects 100 percent interchangeable, which is necessary for the existing functionality to continue to work. This involved a lot of research to map the functionality properly. Because of the extensive changes, it was only possible for me to manage the brain load in this way; it's not just technical.

> I don't think it makes sense to concentrate on fixing all tests for intermediate development states.

I took a pragmatic approach here and analyzed tests that failed. In some cases, these led me to points where the functionality behaved differently than expected or described. So they did what they were supposed to and I was able to correct errors. However, there are also tests that are no longer useful in the change if they tested functionality that was implemented but never used outside of the tests. I deleted these tests. (Less here, but also in the following pull requests.) Thirdly, I temporarily `@Ignore`d a number of tests because they cannot work without an available index because they require index fields that are not currently available. However, these should work again at the end of development.

In principle, I don't think it's wrong if the tests that _should_ work also _do_ pass. If we should deviate from this, I would need to know because I assume that it is a condition for the review. However, I think that this could lead to a pile of unseen errors that then have to be resolved later. I think that this will give us a trustworthy result in the end, which is probably what everyone wants.

matthias-ronge:
By recent agreement with Release Management, the provision of the code for the entire first part of the development has now been made into one joint pull request #6209. The individual parts pull requests will be closed.

PR #6034, opened 2024-04-12
Remove unused class `FacesUtils`

solth:
This pull request removes the unused class `FacesUtils`.


matthias-ronge (approved):
Yes, this is dead code. It was once used to download the Calendar, which is now implemented using Faces’ `DefaultStreamedContent.builder()`

PR #6050, opened 2024-04-23
[hibernate search] Introduce interfaces for the database- and search-related functions of the services

matthias-ronge:
Issue #5760 1c) -- part 2

1c) part 2 introduces interfaces for the database- and search-related functionality of the data services, that by now still use the search index. The data services implement _a lot_ of functionality, some of which is thoroughly covered by tests, but otherwise is not used in the code.

The aim of this step was to clarify and document which of these functions are actually used in the source code—and you can see that there isn't that much—and how they work, in particular to clarify the implicit functionality: Does a function search on _all_ data? Or does it access the user session in the background, and look _only for the data for the client_, for which the logged-in user is currently working? _Or are there other criteria_, if so, which ones, and where are they passed?

You can see that this is not implemented uniformly everywhere, and not every function name describes exactly, what the function actually does. Clarifying this was a prerequisite for a re-implementation with identical functionality, which will follow in the next step.

For better reviewability, this part is provided as a separate pull request.

Follow-up pull request to #6032 ([immediate diff](/matthias-ronge/kitodo-production/compare/5760_1c_1...5760_1c_2))


matthias-ronge:
This pull request accidently was proposed to go into master. Closed.

PR #6053, opened 2024-04-25
Improve main

matthias-ronge:
The main class was well hidden as `KitodoVersionListener`. I renamed it to `KitodoProduction` and moved it. (Git doesn't get this on top of each other and show them as new, but it is from the previous class `KitodoVersionListener`. If you look at the individual commits, it becomes more understandable.) The instance exposes its essential properties, and the main class now handles essential functions of starting and stopping Active MQ and stopping the task manager, that was previously scattered throughout the application. The version is now no longer returned statically, but rather out of the main class.

I find it tidier this way.


henning-gerhardt (approved):
It is a code reading review only. I can not test the changes it self. But from reading everything looks good.

solth:
@matthias-ronge please rebase against current master, since this pull request is the next merge candidate.

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
```diff
@@ -0,0 +1,157 @@
...
+ * Kitodo.Production is the workflow management module of the Kitodo suite. It
+ * supports the digitization process of various types of materials.
+ */
+@WebListener
+public class KitodoProduction implements ServletContextListener, HttpSessionListener {
+    private static final Logger logger = LogManager.getLogger(KitodoProduction.class);
+    private static CompletableFuture<KitodoProduction> instance = new CompletableFuture<>();
+    private ServletContext context;
+    private Optional<Manifest> manifest;
+    private KitodoVersion version = new KitodoVersion();
```
solth, line 50:
```suggestion
    private final KitodoVersion version = new KitodoVersion();
```

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
```diff
@@ -0,0 +1,157 @@
...
+            throw new UndeclaredThrowableException(e);
+        }
+    }
+
+    /**
+     * Returns the servlet context.
+     * 
+     * @return the servlet context
+     */
+    public ServletContext getServletContext() {
```
solth, line 111: Do we need this method? My IDE shows it is unused, and the two interfaces that this method implements do not seem to require the method either.

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
```diff
@@ -0,0 +1,157 @@
...
+        return context;
+    }
+
+    /**
+     * Returns the manifest. If there is one. Otherwise, it can also be empty.
+     * 
+     * @return the manifest
+     */
+    public Optional<Manifest> getManifest() {
```
solth, line 120: See above.

solth (commented):
Just some small remarks. @matthias-ronge I know you just moved functionality to the new `KitodoProduction.java` file and didn't author it, but I think this is a good occasion to check the code for potential improvements, if they are low cost.

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
```diff
@@ -0,0 +1,157 @@
...
+    @Override
+    public void contextInitialized(ServletContextEvent event) {
+        context = event.getServletContext();
+        manifest = retrieveManifestFileAsStream(context);
+        manifest.ifPresent(version::setupFromManifest);
+        instance.complete(this);
+        startActiveMQ();
+    }
+
+    private static final Optional<Manifest> retrieveManifestFileAsStream(ServletContext context) {
```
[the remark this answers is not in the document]
matthias-ronge: I once did read that one should make all methods used in the constructor final, because if they are overloaded in a subclass, it could have confusing effects. I always do that; if the methods are under the constructor, you can see nicely in the “outline” what belongs to the constructors and where the actual class begins. For me, that is an improvement in the clarity of the code. In this case, it is particularly exciting because the “constructor” is not a constructor in the Java sense, but a method that is called by the Servlet Container to construct the class.

outline_

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
matthias-ronge: It's just a getter. You might want to get the servlet context to get the application's working directory, for example, or the server version. It's here for convenience. At the moment, it's not used yet.

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
matthias-ronge: It is my idea that the main class of the application also provides essential characteristics of the application via getters. The manifest is such.

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
solth: Please remove the method if it's unused and not required at the moment. Doesn't matter if it's just a getter. It should be added if required in the future.

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
solth: Please remove unused getters if they are not technically required, see above.

Kitodo/src/main/java/org/kitodo/production/KitodoProduction.java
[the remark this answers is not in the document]
solth: I am not convinced this having a structure outline in your IDE is a good enough reason to overrule this formal recommendation, but I can live with it if it really helps you with.

solth (changes requested):
Please remove unused getters.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/ActiveMQDirector.java
```diff
@@ -65,21 +64,6 @@ public class ActiveMQDirector implements Runnable, ServletContextListener {
     private static Session session = null;
     private static MessageProducer resultsTopic;
 
-    /**
-     * The method is called by the web container on startup
-     * and is used to start up the active MQ connection. All processors from
-     * {@link #services} are registered.
-     */
-    @Override
-    public void contextInitialized(ServletContextEvent initialisation) {
```
solth, line 74: Not sure how Checkstyle did let this pass, but removing this code renders the corresponding import statements for `ServletContextEvent`, `ServletContextListener` and `WebListener` obsolete. Therefore, these now-unused import statements should be removed.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/ActiveMQDirector.java
```diff
@@ -49,8 +49,7 @@
  * The class ActiveMQDirector also provides a basic ExceptionListener
  * implementation as required for the connection.
  */
-@WebListener
-public class ActiveMQDirector implements Runnable, ServletContextListener {
```
solth, line 53: Please adjust JavaDoc as this class does not implement `ServletContextListener` anymore.

solth (changes requested):
Some small adjustments required.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/ActiveMQDirector.java
matthias-ronge: Yes, that's right.

Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/ActiveMQDirector.java
```diff
@@ -23,10 +23,6 @@
 import javax.jms.MessageConsumer;
 import javax.jms.MessageProducer;
 import javax.jms.Session;
-import javax.servlet.ServletContextEvent;
-import javax.servlet.ServletContextListener;
-import javax.servlet.annotation.WebListener;
-
```
solth, line 29: This empty line should probably be re-instated to fulfill Checkstyle requirements.

solth (approved):

PR #6058, opened 2024-05-02
Inactive projects

### LINKED ISSUE #5686: It should not be possible to create processes in an inactive project
**Describe the bug**
It is possible to create processes for an inactive project, using normal process creation and mass import. After process creation in those projects the processes are not visible in the process list.
Apart from that, inactive projects should not be displayed by default to have a coherent behaviour between projects and process templates. 

grafik

**To Reproduce**
1. Mark an project as inactive.
2. Create a process for that project either using manual process creation or mass import.

**Expected behavior**
It should be impossible to create processes for an inactive project. The display behavior should be the same as it is implemented for process templates.

**Release**
Master

solth:
Fixes #5686 

**Note:** the changes in this pull request do _not_ hide deactivated projects, but only disable the buttons for creating processes in the "Actions" column. IIRC it was recently suggested to add the "Filter" field of the process, task and user lists to the project, template and workflow lists as well. Hiding deactivated projects by default - as @BartChris suggested in #5686 - and adding a switch to optionally display them should be added in a separate pull request that implements said feature, instead of introducing the deprecated "Filter" pulldown menu currently still available in the template lists footer in the project list.


BartChris (approved):
Works for me!

PR #6083, opened 2024-05-16
Handle exception during Kitodo Script execution

### LINKED ISSUE #5976: Unhandled `NullPointerException` when using KitodoScript `addDataToIssues`
**Describe the bug**
A `NullPointerException` is thrown when applying the KitodoScript `addDataToIssues` to selected processes that do not contain structure elements with the name supplied using the parameter `type`.

**To Reproduce**
1. Go to 'Processes' page
2. Select some processes via checkbox
3. Click on 'Actions' menu in the footer
4. Select 'Execute KitodoScript' from list of options
5. Select action 'addDataToIssues'
6. Replace 'PeriodicalIssue' as value of 'type' parameter with some string that does not exist as structure element in selected processes
7. Click button 'Execute Kitodo script'
8. See error

**Expected behavior**
The `NullPointerException` should be prevented or handled. Instead of seeing a stack trace, the user should be informed about the nature of the problem in a popup dialog or error message.

**Screenshots**
Bildschirmfoto 2024-03-04 um 11 58 30

**Release**
3.7.0-SNAPSHOT

solth:
Fixes #5976 

Also makes the "error-messages" list scrollable since it might contain many errors for many processes selected in the process list:

Bildschirmfoto 2024-05-16 um 15 01 38

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/exceptions/KitodoScriptExecutionException.java
+++ b/Kitodo/src/main/java/org/kitodo/exceptions/KitodoScriptExecutionException.java
@@ -1,0 +1,23 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.exceptions;
+
+public class KitodoScriptExecutionException extends Exception {
+
+    /**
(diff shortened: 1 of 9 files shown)


BartChris (approved):
The intended functionality is achieved: Adding data to a non-existing structural elements does no longer throw a Nullpointer exception. If multiple messages are displayed, the message is scrollable. 
Code looks good. One might discuss wether it is necessary to catch the `KitodoScriptExecutionException` in case of calling `deleteData`, because as far as i can see, the exception will never be thrown in that case since no `type` is specified there. But maybe it is good to have the exception in place if this is changed in the future. (e.g. throwing an exception if a non existing metadata should be deleted)

PR #6084, opened 2024-05-16
[3.6] Handle exception during Kitodo Script execution

solth:
Backport of #6083


BartChris (approved):
Changes work here as well.

PR #6085, opened 2024-05-17
[hibernate-search] Base search services on database

matthias-ronge:
Issue #5760 1c) -- part 3

In this part, the search functions in the services that have previously accessed the index, are implemented on the database.

Follow-up pull request to #6051 ([immediate diff](/matthias-ronge/kitodo-production/compare/5760_1c_2...5760_1c_3))


Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
```diff
@@ -17,7 +17,9 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
+import javax.el.PropertyNotWritableException;
```
henning-gerhardt, line 22: Exception class is only used for / in documentation. Should we introduce really this dependency? Maybe we should use our own exception class.

This is used in many places but I will comment it only here. The change should be made on all places.

Kitodo/src/main/java/org/kitodo/production/dto/ProjectDTO.java
```diff
@@ -11,13 +11,23 @@
...
 package org.kitodo.production.dto;
 
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
+import java.util.Objects;
+
+import org.apache.logging.log4j.util.Strings;
```
henning-gerhardt, line 21: Why using a string check method from log4j instead of more common lang3.StringUtils package?

This is change should be applied on all places but I'm pointing on this only here.

Kitodo/src/main/java/org/kitodo/production/dto/ProjectDTO.java
```diff
@@ -277,4 +287,33 @@ public boolean hasProcesses() {
     public void setHasProcesses(boolean hasProcesses) {
         this.hasProcesses = hasProcesses;
     }
+
+    @Override
+    public Date getStartDate() {
+        try {
+            return Strings.isNotEmpty(this.startDate) ? new SimpleDateFormat(DATE_FORMAT).parse(this.startDate) : null;
```
henning-gerhardt, line 294: `Strings.isNotEmpty` from log4j package can maybe replaced by a method from commons.lang3.StringUtils (isBlank maybe)?

This is change should be applied on all places but I'm pointing on this only here.

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
```diff
@@ -60,38 +56,34 @@ public HSSFWorkbook getResult() {
...
      */
-    public BoolQueryBuilder getQueryForFilter(ObjectType objectType) {
-        BoolQueryBuilder query = new BoolQueryBuilder();
-
-        try {
-            query = ServiceManager.getFilterService().queryBuilder(this.filter, objectType, false, false);
-        } catch (DataException e) {
-            logger.error(e.getMessage(), e);
+    public String getQueryForFilter(Class<? extends BaseBean> objectType) {
+        String query = "FROM Process AS process";
```
henning-gerhardt, line 74: I'm hoping that the created query is a Hibernate query in HQL and not a raw SQL query as raw SQL queries are not portable.

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
```diff
@@ -60,38 +56,34 @@ public HSSFWorkbook getResult() {
...
         }
-
         if (!this.showClosedProcesses) {
-            query.mustNot(ServiceManager.getProcessService().getQueryForClosedProcesses());
+            query += operator + " process.sortHelperStatus != '100000000000'";
+            operator = " AND";
         }
         if (!this.showInactiveProjects) {
-            query.mustNot(ServiceManager.getProcessService().getQueryProjectActive(false));
+            query += operator + " process.project.active = 0";
```
henning-gerhardt, line 85: So far I know is "active" a boolean type and comparison should be done by `true` or `false` instead of `1` or `0` values.

Type checking will be more important on more recent  versions of Hibernate and / or H2 database.

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
```diff
@@ -60,38 +56,34 @@ public HSSFWorkbook getResult() {
...
+        String query = "FROM Process AS process";
+        String operator = " WHERE";
+        if (!StringUtils.isBlank(filter)) {
+            query += operator + " process.title LIKE '%" + filter + "%'";
+            operator = " AND";
         }
-
         if (!this.showClosedProcesses) {
-            query.mustNot(ServiceManager.getProcessService().getQueryForClosedProcesses());
+            query += operator + " process.sortHelperStatus != '100000000000'";
```
henning-gerhardt, line 81: For `100000000000` value should  the constant value from the class be used.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -581,454 +489,162 @@ public void refresh(Process process) {
...
-     *            the id of the allowed project
-     * @param rulesetId
-     *            the id of the allowed ruleset
-     * @return found processes
-     * @throws DataException
-     *             if the search engine fails
-     */
-    public List<ProcessDTO> findLinkableParentProcesses(String searchInput, int projectId, int rulesetId)
+    @Override
+    public List<Process> findLinkableParentProcesses(String searchInput, int projectId, int rulesetId)
```
BartChris, line 537: This behaviour will be changed in release 3.7.0: https://github.com/kitodo/kitodo-production/pull/5869 
If the logic gets changed or re-implemented here, the concept of the linked PR should better guide the implementation.

matthias-ronge: Question. Do I understand correctly that this is just a hint, it is not a request to change anything in this pull request?

BartChris: When @solth merges the changes of https://github.com/kitodo/kitodo-production/pull/5869 into master, which is happening soon, it would probably be the best to rebase your changes based on the changes there. (After merging of #5869 it will be possible to link to processes in another project so the search for parents should include processes in other projects). If you prefer continue working on the concept in current master (search happens only in the same project) the merge with the new changes has to happen later. I guess it depends on what you prefer.

matthias-ronge: As long as the changes have not been merged into the branch I developed them for, I cannot merge master into the current branch. I have to wait and see, I think. The merge has to happen at some point, that's clear. So, thanks for the explanation.

Kitodo/src/main/java/org/kitodo/production/helper/SearchResultGeneration.java
matthias-ronge: Yes, it are all HQL queries

PR #6102, opened 2024-06-07
Add tests für CalendarService

oliver-stoehr:
Add unit and integration tests for CalendarService.


henning-gerhardt (commented):
Is it possible to convert this new written tests to Junit5?

henning-gerhardt (approved):

PR #6106, opened 2024-06-18
Extracts the renaming of processes into the process service

matthias-ronge:
I came across a functionality implemented in a form class that it shouldn't be, so I extracted it. It should be functionally equivalent.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
@@ -87,6 +87,7 @@ public class ProcessForm extends TemplateBaseForm {
     private Property property;
     private final FilterMenu filterMenu = new FilterMenu(this);
     private final transient FileService fileService = ServiceManager.getFileService();
+    private final transient ProcessService processService = ServiceManager.getProcessService();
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -270,73 +271,18 @@ private boolean renameAfterProcessTitleChanged() {
...
-    }
-
-    private void renameDefinedDirectories() {
-        String[] processDirs = ConfigCore.getStringArrayParameter(ParameterCore.PROCESS_DIRS);
-        for (String processDir : processDirs) {
-            // TODO: check it out
-            URI processDirAbsolute = ServiceManager.getProcessService().getProcessDataDirectory(process)
-                    .resolve(processDir.replace("(processtitle)", process.getTitle()));
-
-            File dir = new File(processDirAbsolute);
```
solth, line 329: Please remove now unused `File` import

solth (changes requested):
Looks good, apart from one small unused import statement that should be removed from `ProcessForm` (see below). Thanks for moving some business logic from the backing bean to a service class!

solth (approved):

PR #6110, opened 2024-06-29
Remove duplicate license file

stweil:
Update also the GPL 3 to the latest license text.


stweil:
I'm afraid this pull request only addresses a small part of the current license issues.

- There is still the problem that it is no clear statement which license applies to Kitodo as a whole (see issue #1301 which was closed to early in my opinion). 233 files in the repository have no explicit license statement. I don't think that it would be good to add such a statement to each of these files. A better solution would be to explain the license(s) in the README.
- GitHub is confused if there is more than a single license file in the repository root unless there is a file named LICENSE, LICENSE.txt, LICENSE.md or LICENSE.rst. My pull request restores this confusion which is bad.

We could create a `license` directory again, move all license files to this directory and link `license/GPL3-License.txt` to `LICENSE.txt`. But what was the reason for the currently used license headers? 1518 files refer to GPL3-License.txt and don't use the recommended header for GPL-3 files.

My favourite solution:
1. Use the commonly used header text which is recommended for GPL-3 files.
2. Add `LICENSE.md` for GPL-3.
3. Clarify all relevant licenses in `README.md`.

Would this be okay for everybody? If yes, I could start with 1 and 2. As far as I know 3 still needs a statement from the Kitodo Vorstand.

PR #6111, opened 2024-07-01
 Add execute permission to shell scripts and simplify test code

stweil:


stweil:
This pull request is an alternative for #5978.

stweil:
With this PR further simplifications are possible because several scripts exist more than once in the repository, some like for example `script_deleteSymLink.sh` even have three incarnations. All those duplicates could be removed.

solth:
Closing this pull request because the alternative approach #5978 was selected and merged.

PR #6114, opened 2024-07-05
pom.xml: Replace maven.compiler.{source,target} by maven.compiler.release

stweil:
This fixes a lot of build warnings with Java compilers newer than version 11.

As the change is supported since Java 9, it also works with Java 11.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,8 +36,7 @@
         <phase.prop>none</phase.prop>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.release>11</maven.compiler.release>
         <maven.compiler.debug>true</maven.compiler.debug>
         <maven.compiler.debuglevel>lines,vars,source</maven.compiler.debuglevel>
         <maxAllowedViolations>0</maxAllowedViolations>


stweil:
The commit fixes these build warnings:
```
WARNING system modules path not set in conjunction with -source 11
```

Please merge.

henning-gerhardt:
Compiling output did change from

`[INFO] Compiling 384 source files with javac [debug:lines,vars,source target 11] to target/classes`

to 

`[INFO] Compiling 384 source files with javac [debug:lines,vars,source release 11] to target/classes`

It may have the same meaning and same binary result but it could even mean something different.

stweil:
I have Apache Maven 3.8.7 (Debian bookworm).

stweil:
I think it is a warning from the Java compiler, so it depends on the Java version (openjdk 17.0.11 2024-04-16
 in my case).

henning-gerhardt:
> I think it is a warning from the Java compiler, so it depends on the Java version (openjdk 17.0.11 2024-04-16 in my case).

So far as I know there is currently no real support for Java versions greater then 11 in usage and even for compiling / building the application. Your fix should be applied at least on switching to Java 17 (or newer) as I don't know what are the consequences on using this right now with Java 11

stweil:
Related documentation: https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html.

stweil:
> It may have the same meaning and same binary result but it could even mean something different.

It means the same , but "unlike the [-source and -target options](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html), the compiler will detect and generate an error when using APIs that don't exist in previous releases of Java SE". This is a desired additional meaning.

stweil:
> Your fix should be applied at least on switching to Java 17

No, it works since Java 9. I don't think that anybody would use Java 8 or older with Kitodo.

stweil:
@solth, please merge.

stweil:
`mvn clean dependency:analyze` und andere Aufrufe mit OpenJDK 22 Compiler erzeugt diese Warnungen:

```
[WARNING] Systemmodulpfad ist nicht zusammen mit -source 11 festgelegt
  Wenn Sie den Speicherort der Systemmodule nicht festlegen, kann dies zu Klassendateien führen, die auf JDK 11 nicht ausgeführt werden können
    --release 11 wird anstelle von -source 11 -target 11 empfohlen, weil dadurch der Speicherort der Systemmodule automatisch festgelegt wird
```

Die hier verwendete Konfiguration entspricht dem Hinweis in den Warnungen und wird seit JDK 9 unterstützt.

JDK 9 und JDK 11 erzeugen offenbar keine Warnungen, aber JDK 17 und JDK 22 geben den Warnungstext für jedes Modul von Kitodo.Production aus. Dieser Pull Request korrigiert das, funktioniert aber auch ohne Einschränkung mit JDK 11.

solth (approved):
Although Java 17 isn't officially supported by Kitodo at the moment (which currently isn't documented properly, because our wiki still lacks a system requirements page), I see no problem in merging this pull request since it does seem to work with the officially supported and recommended Java 11 as well.

PR #6115, opened 2024-07-07
docs: Remove reference to non-existing file contentServerConfig.xml

stweil:


stweil:
The removed line was added in commit c55b130b40.

matthias-ronge (approved):

PR #6116, opened 2024-07-07
Remove `@Tag("isPositive(long)")` statements which cause build warnings

stweil:


stweil:
See [CI protocol](https://github.com/kitodo/kitodo-production/actions/runs/9812239427/job/27095910148#step:14:1538).

stweil:
The lines were added in pull request #5756.

henning-gerhardt (changes requested):
Instead of removing the useful tag annotation they should be fixed. See [tag usage in JUnit5](https://junit.org/junit5/docs/current/user-guide/#running-tests-tags).

stweil:
I agree, but I don't know how the right fix should look like. Therefore the current commit only removes the ignored tags to avoid warnings.

matthias-ronge:
My mistage. But I didn't know that either: These tags are a problem because they contain parentheses. But in general, we can use these tages.

Found here: https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/Tag.html

matthias-ronge (commented):
Suggested solution: See #6119

stweil:
Superseded by PR #6119.

PR #6119, opened 2024-07-08
Fix syntax in tags names

matthias-ronge:
See #6116

### DIFF
--- a/Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
+++ b/Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
@@ -112,21 +112,21 @@ class GuardTest {
     }
 
     @Test
-    @Tag("isPositive(double)")
+    @Tag("isPositive_double")
     void isPositiveDoubleShouldNotFailForPositiveValue() {
         double value = 42;
         Guard.isPositive("value", value);
     }
 
     @Test
-    @Tag("isPositive(double)")
+    @Tag("isPositive_double")
     void isPositiveDoubleShouldNotFailForPositiveBoundary() {
         double value = Double.MIN_VALUE;
         Guard.isPositive("value", value);
     }
 
     @Test
-    @Tag("isPositive(double)")
+    @Tag("isPositive_double")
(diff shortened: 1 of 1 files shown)


Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
```diff
@@ -112,21 +112,21 @@ void isNotNullShouldFailForNullObjekt() {
     }
 
     @Test
-    @Tag("isPositive(double)")
+    @Tag("isPositive_double")
```
stweil, line 115:
```suggestion
    @Tag("isPositiveDouble")
```

stweil: Technically your changes are fine, thank you! I only suggest this cosmetic change here (and similarly for the other lines below) because using camel-case might fit better to the remaining tags.

stweil (approved):

Kitodo-API/src/test/java/org/kitodo/utils/GuardTest.java
matthias-ronge: I chose the tag exactly as the function name is being tested. Here the camel case is part of the function name.

Unbenannt

Underscores are the arguments.

Unbenannt2


Two different methods are actually being tested here, which are different for different purposes, only the function name is the same, but the parameters are different. 

Unbenannt3

So I chose the name for the tage.

stweil: Thanks for explaining. I was more focused on the test function names like `isPositiveDoubleShouldNotFailForPositiveValue` which don't use an underscore. The suggested change is not a must for me – the pull request is also fine without it.

stweil:
@solth, please merge.

solth:
> @solth, please merge.

@stweil this pull request will be merged when it fits in my merge queue. Additionally, I will assign reviewers, so please refrain from requesting reviews for every pull request from me.

solth (approved):

PR #6120, opened 2024-07-10
Keep previous selection of new structure type if possible

### LINKED ISSUE #5933: Remember the last added structure element
In Kitodo.Production 2, the metadata editor "remembered" the last created structure element in the value list for one session. 
The advantage is, that during the structuring of several structure elements of the same type, as for example article or chapter, it is not necessary to select the element. 
Especially for processes with more then 10 element of the same type, this will reduce repetitive work.

In the list of available structure elements, the last selected one should be "remembered" and displayed. 

It could be possible to create several structure elements at once - but only in the structure tree. This has the disadvantage, that the image files have to be moved/assigned to each structure element afterwards. 
In the gallery view, it is not possible to create several structure elements at once. 

**Additional context**
List of structure element, in which the last chosen element should "remembered" (structure tree)
structureElementSelection01

List of structure element, in which the last chosen element should "remembered" (gallery view)
structureElementSelection02

oliver-stoehr:
Keep the previous selection in the dialog for adding new structures in the metadata editor. If the dialog is opened consecutively, the same value doesn't have to be reentered.
The selection is reset, when the previous value is not allowed on the current structure/position.
Resolves #5933.


andre-hohmann:
Thanks a lot! This is an important improvement for the ones who are working intensively in the metadata editor. 

In my tests, it works fine for structure elements and metadata of structure elements for one process. 
If the metadata editor is opened for a new process or is reopened for the same process, the value of the structure element or metadata has to be chosen again. 

@BartChris , @apiller , @subhhwendt , @matthiaskissler , @aetherfaerber, @westedt: what do you think?
- Should the metadata editor "remember" the last chosen value of a structure element and metadata for all processes until the user logs out? 
- Or is it the correct behavior, that the value is "remembered" only until the metadata editor is closed?

matthiaskissler:
I think it would be easiest if the editor remembers the value only as long as the process is open (it would be an improvement nevertheless).

andre-hohmann (approved):
This is a review from the users' perspective, not the technical one.
In my opinion, it works correctly - thanks a lot!

In case that further functionality is required (see https://github.com/kitodo/kitodo-production/pull/6120#issuecomment-2228504742), another issue should be created with the necessary requirements.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java
```diff
@@ -426,6 +424,10 @@ public void prepare() {
         }
         this.prepareDocStructTypes();
         prepareSelectPageOnAddNodeItems();
+        // reset type selection if previous selection is not allowed on current parent/position
+        if (getDocStructAddTypeSelectionItems().stream().noneMatch(a -> a.getValue().equals(docStructAddTypeSelectionSelectedItem))) {
```
henning-gerhardt, line 428: Is it possible to use a more meaningful variable name than `a` inside the expression? Maybe move this whole complex condition check outside the if condition itself like

```java 
bool dontRemberLastSelectedItem = getDocStructAddTypeSelectionItems()
   .stream()
   .noneMatch(item -> item.getValue().equals(docStructAddTypeSelectionSelectedItem));

if (dontRemberLastSelectedItem) {
   docStructAddTypeSelectionSelectedItem = "";
}
```

henning-gerhardt (changes requested):
Small suggestion only which can be resolved even on an other way.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddDocStrucTypeDialog.java
solth: I agree that `a` is not a very good variable name and that it should be renamed. 

Assigning the result of the stream expression to a new variable that is in turn immediately used just once in the next step does not really lower the complexity of the stream expression, though, and I wouldn't see the gain of introducing such a new variable. I would therefor suggest to keep the code as is and just rename the existing variable `a` within the stream to something more meaningful.

henning-gerhardt: I would agree with you if the condition itself is short and good readable. But this condition expression is long, complex and not good readable on the first look. I prefer to avoid this but if this is okay then it go.

solth: @henning-gerhardt I understand what you mean. You have a point because the expression is used in a condition check, which makes it a little less readable. If this was a case where the additional variable would just be returned in the next step, I would argue that it is indeed redundant, but in this case it comes down to personal preference, I guess. Would you mind re-reviewing the PR if you can live with this compromise?

henning-gerhardt (approved):

PR #6121, opened 2024-07-10
Make DocketTest an integration test

matthias-ronge:
We once defined an integration test as a test that reads or writes a file. But _DocketTest_ is defined as a unit test (file name `…Test.java`) and is executed on every Maven command. It doesn't just read one file, it reads very many—at least on Windows, it parses all installed fonts each time, produces a thousand lines of output and it takes minutes. So it should be an integration test instead (file name `…IT.java`).


solth (approved):
> We once defined an integration test as a test that reads or writes a file.

I do not remember that definition. IIRC an Integration test was described as a test that interacts with and depends on its surrounding systems, mainly the database and the search index. The filesystem should probably also be seen as such a surrounding system, though, so I can agree with the proposed change.

PR #6123, opened 2024-07-11
Fix mets file id check including an option for use of this check

henning-gerhardt:
It is a solution for an issue explained in discussion https://github.com/kitodo/kitodo-production/discussions/6087 . By default this check is disabled or not in use and can be enabled.

### DIFF
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
@@ -461,7 +461,7 @@ public class MetsXmlElementAccess implements MetsXmlElementAccessInterface {
                 for (FileType file : fileGrp.getFile()) {
                     if (strictCheck && usedFileId.contains(file.getID())) {
                         throw new IllegalArgumentException(
-                            "Corrupt file: file with id " + file.getID() + " is part of multiple groups"
+                            "Corrupt file: each METS file ID has to be unique but " + file.getID() + " is used multiple times!"
                         );
                     } else {
                         usedFileId.add(file.getID());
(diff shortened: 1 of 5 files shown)


Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
```diff
@@ -452,19 +453,21 @@ private StructLink createStructLink(LinkedList<Pair<String, String>> smLinkData)
...
+        Set<String> usedFileId = new HashSet<>();
+        boolean strictCheck = KitodoConfig.getConfig().getBoolean("useStrictMetsFileIdCheck", false);
         if (Objects.nonNull(fileSec)) {
             for (FileGrp fileGrp : fileSec.getFileGrp()) {
                 String use = fileGrp.getUSE();
                 for (FileType file : fileGrp.getFile()) {
-                    if (fileUseMap.containsKey(file)) {
+                    if (strictCheck && usedFileId.contains(file.getID())) {
                         throw new IllegalArgumentException(
                             "Corrupt file: file with id " + file.getID() + " is part of multiple groups"
```
solth, line 464: Is this still the correct error message for the new condition? With the updated check the exception would also be thrown if two files in the same file group use the same file ID, wouldn't it? Of course, that case is probably illegal as well, but does not match the exception message then.

Wouldn't it be better to have _two_ checks here, then? E.g. the original one that checks whether the current file is already in `fileUseMap` and a new one, that contains your added checks and throws an exception with a different message?

solth (commented):
Just one comment about the updated check. Maybe I misunderstood something, but it seems to me the original check should be _extended_ rather than being _replaced_.

Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
henning-gerhardt: 1. You are right the exception message could be updated. Did you have a proposal?
2. The old test `fileUseMap.containsKey(file)` did not work as the compared object `file` is different as it's internal hash value differ between same data. So this test can't be restored.

solth: > 1. You are right the exception message could be updated. Did you have a proposal?

How about something like
`"Corrupt file: each METS file ID has to be unique but " + file.getID() + " is used multiple times!"`
?

solth (approved):

PR #6124, opened 2024-07-12
Add relation between import configurations and clients

### LINKED ISSUE #6059: Import configurations not filtered by client
**Describe the bug**
Import configurations (and mapping files) are currently not associated to specific clients but instead available system wide to all users of all clients. This is in contrast to how templates, workflows, rulesets and dockets are managed in Kitodo.Production, which are all assigned to specific clients. The same should be the case for import configurations and mapping files. It is especially import for import configurations because they can contain sensitive information like passwords for access restricted ftp servers.

**To Reproduce**
1. Create an import configuration
2. Change the current client
3. See import configuration created with different client is still available for other client as well

**Expected behavior**
Import configurations and mapping files should be associated to specific clients, the same as templates, workflows and rulesets.

**Release**
All releases

solth:
Since import configurations can contain sensitive data like FTP credentials, they shouldn't be freely shared between clients. This pull request adds a relation between clients and import configurations similar to that between clients and projects, so that import configurations are explicitely mapped to specific lists of clients and users can only see and use those import configurations that are mapped to at least one of their assigned clients.

(since XSLT mapping files normally do _not_ contain such sensitive data and are - like rulesets - available to all users via the file system anyway, a relation between clients and mapping files does not seem necessary and therefore has not been added)

The mapping between import configurations and clients can be edited via a new third tab named "Mandanten" ("Clients") on the import configuration edit page:

Bildschirmfoto 2024-07-12 um 14 05 53

The SQL migration file in this pull request that adds the corresponding cross table also assigns _all_ existing import configurations to _all_ existing clients by default to maintain the current status quo. An admin can then make use of the new functionality instroduced in this pull request and restrict access to specific configurations using the import configuration edit formular shown above.

Fixes #6059

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/constants/StringConstants.java
+++ b/Kitodo/src/main/java/org/kitodo/constants/StringConstants.java
@@ -15,6 +15,6 @@ public class StringConstants {
 
     public static final String COMMA_DELIMITER = ", ";
     public static final String SEMICOLON_DELIMITER = "; ";
-    public static final String SAVE = "editForm:save";
+    public static final String EDIT_FORM_SAVE = "editForm:save";
 
 }
--- a/Kitodo/src/main/java/org/kitodo/constants/StringConstants.java
+++ b/Kitodo/src/main/java/org/kitodo/constants/StringConstants.java
@@ -15,5 +15,6 @@ public class StringConstants {
 
     public static final String COMMA_DELIMITER = ", ";
     public static final String SEMICOLON_DELIMITER = "; ";
+    public static final String SAVE = "editForm:save";
 
 }
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Client.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Client.java
@@ -40,6 +40,9 @@ public class Client extends BaseBean {
                     foreignKey = @ForeignKey(name = "FK_column_id"))})
     private List<ListColumn> listColumns;
 
+    @ManyToMany(mappedBy = "clients", cascade = CascadeType.PERSIST)
+    private List<ImportConfiguration> importConfigurations;
+
     /**
      * Gets name.
      *
@@ -98,5 +101,27 @@ public class Client extends BaseBean {
      */
     public void setListColumns(List<ListColumn> columns) {
         this.listColumns = columns;
+    }
+
+    /**
+     * Get import configuration.
+     *
+     * @return import configurations
+     */
+    public List<ImportConfiguration> getImportConfigurations() {
+        initialize(new ClientDAO(), this.importConfigurations);
+        if (Objects.isNull(this.importConfigurations)) {
+            this.importConfigurations = new ArrayList<>();
+        }
+        return importConfigurations;
+    }
+
+    /**
+     * Set import configurations.
+     *
+     * @param configurations import configurations
+     */
+    public void setImportConfigurations(List<ImportConfiguration> configurations) {
+        this.importConfigurations = configurations;
     }
 }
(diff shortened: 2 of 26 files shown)


henning-gerhardt (commented):
Code changes look good, but I have 2 functional questions:

1. It is possible to de-assign all clients from an import configuration. This import configuration becomes invisible but is still existing. Without manipulating the database table it is not possible anymore to interact with this hidden import configuration. In my opinion at least one client should be assigned to an import configuration.

2. If you have at least two clients Client_A and Client_B. You logged in in Client_A and de-assign an import configuration from Client_A so this import configuration should only assigned to Client_B, the import configuration is still visible while logged in in Client_A (even after re-log in). This behaviour is different to all the other client-depending relations like projects, process templates, workflows, ... Should this be the case?

solth:
@henning-gerhardt thanks a lot for the remarks!

> 1. It is possible to de-assign all clients from an import configuration. This import configuration becomes invisible but is still existing. Without manipulating the database table it is not possible anymore to interact with this hidden import configuration. In my opinion at least one client should be assigned to an import configuration.

Thanks for pointing this out. Indeed, it would become difficult to re-assign an import configuration that is not assigned to any client anymore. I see two ways to approach this:
- either add a specdial permission/authority that allows users to see _all_ import configurations, no matter to which client(s) they are assigned. This way an "System admin" could be defined that 
- or add some frontend logic that disables and item in the client list when it is the last selected item, to prevent the situation you described, where a import configuration isn't assigned to any client anymore. This approach would have to take into account that a user might only be assigned to _some_ clients (not all of them) and an import configuration might still be assigned to a client to which the user is _not_ assigned

What would be the better solution? Or do you have an alternative idea how to resolve this issue?

> 2. If you have at least two clients Client_A and Client_B. You logged in in Client_A and de-assign an import configuration from Client_A so this import configuration should only assigned to Client_B, the import configuration is still visible while logged in in Client_A (even after re-log in). This behaviour is different to all the other client-depending relations like projects, process templates, workflows, ... Should this be the case?

This indeed sounds like a bug, e.g. wasn't intended. I will try to find the cause for this issue and resolve it.

henning-gerhardt:
I would go for the first option (special authority to "see" all available clients) even this behaviour is different then an other places like projects, process templates, ... but I think that sharing import configurations is more common than sharing projects, process templates, ... between clients.

henning-gerhardt:
After a discussion between @solth  and @henning-gerhardt we come to the following solution:
- adding a new permission/authority to see all clients inside the import configuration independent of which your current client is
- adding a custon validator which ensures that at least one client is assigned to an import configuration

solth:
@henning-gerhardt I updated the pull request as discussed: a new special permission was added that allows assigning import configurations to clients globally. The corresponding and new third "Clients" tab in the "Import configuration edit" page is now only rendered when the user has this new permission. On this third tab _all_ clients are displayed and offered for assignement (not just those assigned to the current user/admin). Additionally, I added a new validator that ensures at least one client is assigned when saving the import configuration.

Kitodo/src/main/java/org/kitodo/constants/StringConstants.java
```diff
@@ -15,5 +15,6 @@ public class StringConstants {
 
     public static final String COMMA_DELIMITER = ", ";
     public static final String SEMICOLON_DELIMITER = "; ";
+    public static final String SAVE = "editForm:save";
```
henning-gerhardt, line 18: Is `SAVE` not to meaningless and may be used in wrong context? Maybe `EDITFORM_SAVE` can prevent this?

solth: Yes, you are right. The name of the constant should match the content. When it's called just "SAVE", I, too, would expect it to contain a simple String "Save". I will change that to your suggestion.

henning-gerhardt (approved):
Changes looks good and new feature is working as defined.

henning-gerhardt (approved):
Thank you for adding a test for the new feature.

PR #6129, opened 2024-07-18
Replace old javax.json by newer jakarta.json

stweil:
This fixes an error with Java 17 and OpenSearch:

javax.json.JsonException: Provider org.glassfish.json.JsonProviderImpl could not be instantiated: java.lang.ClassCastException: class org.glassfish.json.JsonProviderImpl cannot be cast to class javax.json.spi.JsonProvider (org.glassfish.json.JsonProviderImpl and javax.json.spi.JsonProvider are in unnamed module of loader 'app')
	at org.kitodo.production.forms.CommentFormIT.setUpClass(CommentFormIT.java:41)

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/search/Searcher.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/search/Searcher.java
@@ -18,10 +18,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-import javax.json.Json;
-import javax.json.JsonObject;
-import javax.json.JsonReader;
-
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
 import org.elasticsearch.search.SearchHit;
@@ -32,6 +28,10 @@ import org.elasticsearch.search.sort.SortBuilder;
 import org.kitodo.data.elasticsearch.Index;
 import org.kitodo.data.elasticsearch.exceptions.CustomResponseException;
 import org.kitodo.data.exceptions.DataException;
+
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonReader;
 
 /**
  * Implementation of ElasticSearch Searcher for Kitodo - Data Management
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -76,8 +76,12 @@
             <artifactId>elasticsearch-rest-high-level-client</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.glassfish</groupId>
-            <artifactId>javax.json</artifactId>
+            <groupId>jakarta.json</groupId>
+            <artifactId>jakarta.json-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.parsson</groupId>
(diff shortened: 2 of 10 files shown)


stweil:
Related PR: #6020.

Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/search/Searcher.java
```diff
@@ -33,6 +29,10 @@
 import org.kitodo.data.elasticsearch.exceptions.CustomResponseException;
 import org.kitodo.data.exceptions.DataException;
 
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonReader;
+
```
stweil, line 35: Sollten wir "jakarta" in den _config/checkstyle.xml_ ergänzen und als eigene Gruppe vor "java" einsortieren? Die aktuelle Regel kennt es nicht und verlangt daher diese import-Statements ganz am Schluss.

thomaslow:
@stweil Hello Stefan, thank you for your pull request. I was asked to review it by @solth. 

Unfortunately, I am a bit puzzled about the situation.

@solth Isn't it the current plan to migrate to `hibernate-search`? If I recall correctly, the motivation was to move away from ElasticSearch due to licensing concerns. That means, @matthias-ronge  has likely tested his `hibernate-search` migration with OpenSearch, and probably has already fixed related issues in his `hibernate-search`-branch. 

That means, this pull request will likely lead to merge conflicts.

In case you don't know, I was given the task to upgrade Kitodo.Production to Java 17 and Tomcat 10 as soon as the `hibernate search` and `PrimeFaces` migrations are implemented. There are plenty of other changes necessary to make that work. 

Although I appreciate your work, I would suggest pausing these changes until the `hibernate search` and `PrimeFaces` migrations are merged. In case you cannot wait, maybe an alternative would be to create and maintain a standalone fork `Kitodo-Production-OpenSearch` in the meantime?

If I am mistaken about the current development plan for Kitodo.Production, please let me know.

stweil:
@thomaslow, thank you for your feedback. I was not aware that this pull request is connected anyhow with the hibernate search activities.

Regarding newer versions of OpenJDK: my latest opensearch branch builds with all versions which I tested (11, 17 and 22).

thomaslow:
> @thomaslow, thank you for your feedback. I was not aware that this pull request is connected anyhow with the hibernate search activities.

Your pull request description says:

> This fixes an error with Java 17 and ***OpenSearch***: [...]

I order to review this pull request, I would need to verify your description and check that this problem is now resolved. This means I would have to set up OpenSearch and merge your OpenSearch-branch to reproduce the problem. Unfortunately, I do not really see this as a good use of my time, given that the plan is to migrate to `hibernate-search`.

Of course, the included changes from `javax` to `jakarta` are useful anyway. However, I would prefer to implement these changes all together instead of repeatedly reviewing small changes over the next couple of months. 

I would suggest to keep this pull request open for now. I will come back to it when I am working on the Java / Tomcat migration.

stweil:
The problem arises as soon as new software is integrated which uses the modern Jakarta while Kitodo still uses Javax, so any development will be confronted with it sooner or later.

Updating it now would not harm and avoid the failures in new developments. But of course it is your choice when to make the move.

stweil:
This pull request has been superseded by commit 59b37b105dcfba5b3340e7f5af1456a02c00c162.

Issue #6065, opened 2024-05-07
Automatically created images don't appear in `meta.xml`


mnscholz:
**Describe the bug**

I have a workflow script step that calls a shell script that generates derivative images of the main tiffs.
The directory where the images get stored is declared in the project settings as shown below:

Although it says that the images should get included in the METS, they will not appear as a `mets:fileGrp` in the
`meta.xml` during the whole workflow, only in the exported METS they are included correctly.

In the workflow script step, the properties for reading/writing/validating images are set.

**Expected behavior**
The created images should appear in the `meta.xml` directly after the step has finished.

**Screenshots**
![grafik](https://github.com/kitodo/kitodo-production/assets/1378632/ea478e7f-f4b9-482b-8cc1-c7ae54f928b0)

**Release**
3.5.0

**Desktop (please complete the following information):**

BartChris:
https://github.com/kitodo/kitodo-production/pull/4450

Would that functionality help you?

mnscholz:
Yes that helps, thanks a lot!
Is it also possible to trigger the action from outside Kitodo, ie. from the shell script that got called by Kitodo?

I leave this issue open as I still consider it a bug that the workflow step properties write/validate images won't update the `meta.xml`, which IMO is the expected behaviour.

BartChris:
> Yes that helps, thanks a lot! Is it also possible to trigger the action from outside Kitodo, ie. from the shell script that got called by Kitodo?
> 
Not yet, as far as i know. There is an open Pull Request for that functionality: https://github.com/kitodo/kitodo-production/pull/6013

solth:
The option to run Kitodo scripts via ActiveMQ and this trigger linking automatically created images from external scripts thus is now possible. I think the normal procedure for configuring workflows is now to add a second automatic tasks that call the `searchForMedia` script directly after the - also automatic - workflow step to create jpg or png copies of the scans. These two automatic steps now nearly always come in tandem.

@mnscholz does this work for you so that we can close this issue?

solth:
Closing this issue as completed, now that `searchForMedia` is a available as a KitodoScript that can also be triggered from external via ActiveMQ

PR #6131, opened 2024-07-18
Add support for OpenSearch

stweil:
This PR adds support for OpenSearch while still working with Elasticsearch.

It is for people who want to work with recent code, newer OpenJDK versions (17, 22, but 11 also still supported) and OpenSearch because they cannot use Elasticsearch.

It's also for people who want to use Elasticsearch with newer code, either for security reasons or because they want to work with newer OpenJDK versions.


stweil:
The latest code can be built with up to OpenJDK 22.

solth:
@stweil thanks for sharing this development. It is good to see that compatibilty with OpenSearch can be achieved this way. However, the current plan for Kitodo.Production is not to switch from ElasticSearch to the open source fork OpenSearch, but rather to incorporate the HibernateSearch framework to allow institutions to decide which search server variant they want to use (see #5760 and [#6065](https://github.com/kitodo/kitodo-production/discussions/6095) for details). 

Of course your development is of value anyway, but I do not think creating a pull request with these changes is the right way of giving users access to it and it might trigger wrong expections, because - as you said - it won't be merged. As mentioned above, _replacing_ the ElasticSearch client libraries in Kitodo.Production with corresponding OpenSearch libraries is currently not on the roadmap. Instead I would suggest to link the branch that this pull request is based (stweil:opensearch) in the discussion and issue mentioned above so that users interested in trying it out can access it that way.

stweil:
I rebased and minimized the pull request now. Based on PR #6215 there remains a single small commit.

With this pull request it is directly possible to use Elasticsearch as before (see CI).

The PR replaces old Elasticsearch module by newer OpenSearch modules, so OpenSearch can also be used instead of Elasticsearch (tested locally with the same commands as in CI test, but an OpenSearch server).

It's also possible to rebase the existing work for HibernateSearch on top of this PR with only a few merge conflicts which are easy to fix.

solth:
@stweil I just tested your branch and can confirm that it works direktly "out-of-the-box" with my existing ElasticSearch-Server, without the need for any other adjustments. Additionally, it seems all changes in the code - apart from the new dependencies in the `pom` files - amount to just the `elasticsearch` import statements being replaced with corresponding `opensearch` imports, withtout any changes to the usage in the actual business logic at all. I have to admit I am suprissed that facilitating OpenSearch compatibility while maintaining ElasticSearch compatibility was possible with so few changes.

That said, I do not see the advantage of having these changes in the code when they are replaced by HibernateSearch before the next larger release, e.g. 3.8. Therefore I think merging this pull request would only make sense if we shifted the HibernateSearch integration to a new release 3.9 and released a version 3.8 of Kitodo in the near future with the OpenSearch compatibility and the new functions for the metadata editor for which pull requests have already been opened.

Since the HibernateSearch integration has the highest priority right now, I wouldn't want to introduce any changes to the code that would impede that project, so I would like @matthias-ronge decide whether he agrees with these changes or whether he thinks they would further postpone the finalization of the HibernateSearch integration for him.

stweil:
Recent changes (0e43651ad5951dbcf3859d3b1fc83953e86a8eb8, 1f6a523d69c7cee14654b41646a950c084bd0c9d) caused a merge conflict which is now fixed.

matthias-ronge (approved):

pom.xml
```diff
@@ -65,7 +65,7 @@
         <saxon.version>9.9.1-8</saxon.version>
         <log4j.version>2.19.0</log4j.version>
         <junit.version>5.9.2</junit.version>
-        <elasticsearch.version>7.17.24</elasticsearch.version>
+        <opensearch.version>[2.15.0,)</opensearch.version>
```
henning-gerhardt, line 68: This kind of notation `[xxx,)` result in a non-reproducable builds as every build may or may not use the same version of this library/tool/...  It made it more complex to verify new functionality, reproduce bugs and may introduce changes which are hard to determinate. If an automatic way is needed to update the version tools like dependabot should be used.

solth: I support @henning-gerhardt's remark. 
@stweil please use one specific version number instead!

stweil: I also rebased the PR to git master.

stweil:
I had to rebase the PR and fix a merge conflict caused by commit 8f2015b848d4f2a1b928b652cf72adf9fd8d43e7.

It's now again ready for merging.

stweil:
Now rebased and fixed new merge conflict.

stweil:
Rebased and fixed merge conflict.

solth (approved):
Just tested this branch with ElasticSearch and OpenSearch servers and it indeed works with both as advertised 👍

PR #6133, opened 2024-07-18
Update version of flyway-maven-plugin (required for newer MariaDB)

stweil:
The latest version 10.15.2 cannot be used:
```
Error:  Failed to execute goal org.flywaydb:flyway-maven-plugin:10.15.2:baseline (default) on project kitodo-data-management: Execution default of goal org.flywaydb:flyway-maven-plugin:10.15.2:baseline failed: Unable to load the mojo 'baseline' in the plugin 'org.flywaydb:flyway-maven-plugin:10.15.2' due to an API incompatibility: org.codehaus.plexus.component.repository.exception.ComponentLookupException: org/flywaydb/maven/BaselineMojo has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

Therefore the update only updates to 9.22.3.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
         <properties-maven-plugin.version>1.2.1</properties-maven-plugin.version>
         <selenium.version>3.141.59</selenium.version>
         <spring.security.version>5.7.13</spring.security.version>
-        <flyway-maven-plugin.version>8.4.4</flyway-maven-plugin.version>
+        <flyway-maven-plugin.version>9.22.3</flyway-maven-plugin.version>
         <spotbugs-maven-plugin.version>4.3.0</spotbugs-maven-plugin.version>
         <pitest.version>1.4.10</pitest.version>
         <mariadb-java-client.version>2.7.12</mariadb-java-client.version>


henning-gerhardt:
I have on my local environment system even a more recent version of MariaDB running but I did not get this error. Maybe this error appear only on your system with a not yet official supported Java version for the application?

pom.xml
```diff
@@ -78,7 +78,8 @@
         <properties-maven-plugin.version>1.2.1</properties-maven-plugin.version>
         <selenium.version>3.141.59</selenium.version>
         <spring.security.version>5.7.11</spring.security.version>
-        <flyway-maven-plugin.version>8.4.4</flyway-maven-plugin.version>
+        <!-- https://mvnrepository.com/artifact/org.flywaydb/flyway-maven-plugin -->
```
henning-gerhardt, line 81: This comment is not needed.

henning-gerhardt (changes requested):
At least I can still access my local and inhouse used databases. I did not try to migrate any database itself.

pom.xml
stweil, line 81:
```suggestion
        <!-- Newer versions exist at https://mvnrepository.com/artifact/org.flywaydb/flyway-maven-plugin, but are incompatible with OpenJDK 11. -->
```

stweil: Is this comment better? Or would you prefer no comment at all?

henning-gerhardt: I would use no comment at all.

stweil, line 81:
```suggestion
```

stweil: Okay, I removed it now.

stweil:
I added a commit which fixes issue #6153 by supporting flyway commands from the root. See also the related discussion in #6127.

stweil:
> the changes here did not fix not the main issue from #6127

I had added the important commit which fixes #6127 only recently. Maybe it was still missing in your test.

Splitting the PR in several independent parts seems like a good idea. I'll do this.

stweil:
I removed the additional commits now and created new pull requests for them (see #6158 and #6159).

stweil:
Rebased + fixed merge conflict.

solth (approved):
Works for me.

PR #6135, opened 2024-07-18
Fix coding style issues and update maven-checkstyle-plugin and its dependencies

stweil:


stweil:
The CI build failed with this error which is not related to my changes:
```
Error:  Tests run: 5, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 47.778 s <<< FAILURE! - in org.kitodo.selenium.ImportingST
Error:  checkHierarchyImport  Time elapsed: 19.359 s  <<< ERROR!
org.openqa.selenium.StaleElementReferenceException: 
stale element reference: stale element not found in the current frame
  (Session info: chrome=126.0.6478.182)
For documentation on this error, please visit: https://www.seleniumhq.org/exceptions/stale_element_reference.html
```

I get this error from time to time even with unmodified code from Git master. So it looks like there is a general issue somewhere in the code which needs to be fixed.

pom.xml
```diff
@@ -891,20 +892,19 @@ from system library in Java 11+ -->
...
-                                <version>10.3.2</version>
+                                <version>10.17.0</version>
                             </dependency>
                         </dependencies>
                         <executions>
                             <execution>
                                 <id>checkstyle-check</id>
                                 <phase>validate</phase>
                                 <configuration>
-                                    <encoding>UTF-8</encoding>
```
solth, line 907: I think this should be changed to `<inputEncoding>` instead of being removed.

stweil: See https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html#inputEncoding. So you are right, `<inputEncoding>` would work, but it is not necessary because we already set the correct default encoding `<project.build.sourceEncoding>`.

solth: Right, thanks for pointing this out.

solth (approved):

PR #6136, opened 2024-07-19
Update mockito version (required for OpenJDK 22)

stweil:


stweil:
The CI failure looks like a random failure which can be seen in other CI builds, too, and which is unrelated to this pull request.

solth (approved):
This does work with current JDK versions so I do have no problem with it.

PR #6143, opened 2024-07-24
Enrich process with LABEL and ORDERLABEL during mass import

BartChris:
Because of the way Kitodo production works, it is important to have the LABEL and the ORDERLABEL before the data is exported and partially enriched via XSL. 
https://github.com/kitodo/kitodo-production/issues/5016#issuecomment-1378852324
This is especially a problem when providing LABELs during mass import, because an import XSL does not take into consideration fields provided via csv upload.

This PR explicitely adds LABEL and ORDERLABEL to the logical root, when they are provided during mass import.

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/ImportServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/ImportServiceIT.java
@@ -117,6 +117,8 @@ public class ImportServiceIT {
     private static final int RULESET_ID = 1;
     private static final String TITLE = "Title";
     private static final String PLACE = "Place";
+    private static final String LABEL = "LABEL";
+    private static final String ORDERLABEL = "ORDERLABEL";
     private static final int EXPECTED_NR_OF_CHILDREN = 23;
     private static final String PICA_XML = "picaxml";
     private static final String PICA_PPN = "pica.ppn";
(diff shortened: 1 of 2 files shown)


andre-hohmann:
Thanks a lot! 
If that PR works, it will solve a lot of problems! 
Especially if you want to want to modify the ORDER and ORDERLABEL for periodical volumes - which is what we do.
Good luck!

solth (approved):
Code looks good and works. Thanks a lot for the test.

solth:
@andre-hohmann did you have a chance to check whether the changes in this pull request do indeed solve some issues, as you anticipated?

Kitodo/src/test/java/org/kitodo/production/services/data/ImportServiceIT.java
```diff
@@ -199,6 +201,45 @@ public void testImportProcessWithAdditionalMetadata() throws DAOException, Impor
...
+        presetMetadata.put(ORDERLABEL, List.of("TEST-ORDERLABEL"));
+        Process processWithAdditionalMetadata = importProcessWithAdditionalMetadata(RECORD_ID,
+                MockDatabase.getK10PlusImportConfiguration(), presetMetadata);
+        Workpiece workpiece = ServiceManager.getMetsService()
+                .loadWorkpiece(processService.getMetadataFileUri(processWithAdditionalMetadata));
+        HashSet<Metadata> metadata = workpiece.getLogicalStructure().getMetadata();
+        String processLabel = workpiece.getLogicalStructure().getLabel();
+        String processOrderlabel = workpiece.getLogicalStructure().getOrderlabel();
+        try {
+            Assert.assertTrue("Process does not contain correct metadata",
```
henning-gerhardt, line 227: After PR #6093 is merged the old Junit4 based tests must be adjusted to be Junit5 based tests. In most cases the used parameters must be switched and  `Assert` must be replaced with `Assertions` or better don't use them directly in code and use them as static import.

solth:
@BartChris please rebase your branch against the current master and update your test to use Junit5 instead of Junit4, for which support has been removed in the master branch now.

BartChris:
@solth done

andre-hohmann:
> @andre-hohmann did you have a chance to check whether the changes in this pull request do indeed solve some issues, as you anticipated?

@solth : I am sorry, but i did not read the description of the pull request correctly. I thought, that the export is modified in the way that `<mets:structMap TYPE="LOGICAL">` of superordinate  processes is created after the export XSLT is applied to the subordinate processes. 
Thus, the modification of the mass import does not solve issues i anticipated.

PR #6144, opened 2024-07-24
Replace old commons-collections:commons-collections by org.apache.commons:commons-collections4

stweil:
commons-collections4 was already pulled in by other dependencies, and having two packages can cause conflicts and increases the size of the war file.


stweil:
My local build does not fail, and I still do not know why it fails in the CI GitHub action.

solth:
Superseded by #7053

PR #6145, opened 2024-07-24
Replace outdated org.codehaus.jackson:* by com.fasterxml.jackson.core:*

stweil:
[org.codehaus.jackson:*](https://github.com/codehaus/jackson) is archived (= unmaintained) legacy code which should be [replaced](https://www.google.com/search?q=replace+org.codehaus.jackson).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -416,10 +416,15 @@ from system library in Java 11+ -->
                 <artifactId>camunda-bpmn-model</artifactId>
                 <version>7.17.0</version>
             </dependency>
+	    <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-core</artifactId>
+                <version>2.17.1</version>
+            </dependency>
             <dependency>
-                <groupId>org.codehaus.jackson</groupId>
-                <artifactId>jackson-xc</artifactId>
-                <version>1.9.13</version>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>2.17.1</version>
             </dependency>
             <dependency>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -416,10 +416,15 @@ from system library in Java 11+ -->
                 <artifactId>camunda-bpmn-model</artifactId>
                 <version>7.17.0</version>
             </dependency>
+	    <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-core</artifactId>
+                <version>2.17.1</version>
```
stweil, line 422:
```suggestion
                <version>2.17.2</version>
```

pom.xml
```diff
@@ -416,10 +416,15 @@ from system library in Java 11+ -->
...
+            </dependency>
             <dependency>
-                <groupId>org.codehaus.jackson</groupId>
-                <artifactId>jackson-xc</artifactId>
-                <version>1.9.13</version>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>2.17.1</version>
```
stweil, line 427:
```suggestion
                <version>2.17.2</version>
```

stweil (commented):
Meanwhile a new version was released.

stweil:
@solth, I expect that the replacement might also be faster (see its name), so there seem to be several good reasons to consider this pull request.

stweil:
If you are interested why I was motivated to create this commit: I asked a KI to analyse `pom.xml` and to give hints for improvements. Of course this was not the only one, and I think most of them were reasonable.

solth (approved):

Issue #6139, opened 2024-07-22
Database migration fails partially on macOS (and other hosts with case insensitive filesystem)


stweil:
**Describe the bug**

The SQL statements in `Kitodo-DataManagement/src/main/resources/db/migration/V2_105__Fixes_#3998.sql` which were introduced in pull request #4412 don't work on hosts with a filesystem which is case insensitive (macOS, Windows).

An easy workaround is using an intermediate filename:

Replace
```
ALTER TABLE client_x_listColumn RENAME TO client_x_listcolumn;
```

by
```
ALTER TABLE client_x_listColumn RENAME TO renamed_table;
ALTER TABLE renamed_table RENAME TO client_x_listcolumn;
```

and similarly for all other renames.

**To Reproduce**
1. Run migration on macOS.
2. Check names of database tables.
3. See error

**Expected behavior**
Database migration must work on macOS and Windows (and Linux with a case insensitive filesystem), too.

**Release**
All recent releases up to git master.

stweil:
@matthias-ronge, @solth, should I prepare a PR with the suggested fix for the SQL file? Or do you have a different solution?

stweil:
If we patch the existing migration script, this will require a flyway:repair for at least some existing installations. Is this acceptable? See related discussion for typo fixes in pull request #6149.

henning-gerhardt:
My question is: why this is failing on your system? So far as I know this is working on Windows and if I remember me correct this was working on MacOS too.

stweil:
See https://mariadb.com/kb/en/identifier-case-sensitivity/.

PR #6149, opened 2024-07-25
Fix some typos in migration scripts

stweil:


henning-gerhardt:
This will break every installation which is maintaining their database with FlyWay:

```
[ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:8.4.4:validate (default-cli) on project kitodo-data-management: org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation
[ERROR] Migration description mismatch for migration version 2.19
[ERROR] -> Applied to database : Set properies to unique not null
[ERROR] -> Resolved locally    : Set properties to unique not null. Either revert the changes to the migration, or run repair to update the schema history.
[ERROR] Migration checksum mismatch for migration version 2.61
[ERROR] -> Applied to database : 1769151973
[ERROR] -> Resolved locally    : 1654358955. Either revert the changes to the migration, or run repair to update the schema history.
[ERROR] Migration checksum mismatch for migration version 2.68
[ERROR] -> Applied to database : 1266171681
[ERROR] -> Resolved locally    : -964385541. Either revert the changes to the migration, or run repair to update the schema history.
[ERROR] Migration checksum mismatch for migration version 2.71
[ERROR] -> Applied to database : -1061746512
[ERROR] -> Resolved locally    : 3664929. Either revert the changes to the migration, or run repair to update the schema history.
```

stweil:
For our own production database the existing scripts also produce checksum mismatches.

stweil:
Unrelated: I wonder why even new migration scripts are named V2_*.sql. Are they really still related to Kitodo.Production 2.x? Or what is the meaning of "V2"?

henning-gerhardt:
> Unrelated: I wonder why even new migration scripts are named V2_*.sql. Are they really still related to Kitodo.Production 2.x? Or what is the meaning of "V2"?

V2 has nothing to do with Kitodo.Production 2. 

V1 is the base for all upcoming migrations while developing for Kitodo.Production 3.x was ongoing until now and their version begins with V2 prefix.

henning-gerhardt:
> There were already renames and other changes made for the migration scripts in the past as far as I know. Therefore I thought that [running repair](https://documentation.red-gate.com/flyway/flyway-cli-and-api/commands/repair) is something which was accepted.

Yes this was done in the past  and I was not happy over this changes as I maintain a lot of databases which must all adjusted when the databases get updated to a new version usage. Now it is not only a typo correction inside the files now even a migration file get renamed. The typo's may be ugly but this should be discovered while the pull request is reviewed and not a lot of time later and even after the file is part of a release. In my opinion this typo's should stay and not get fixed.

solth:
I agree with @henning-gerhardt that fixing these typos in comments (!) is not worth creating additional hassle and workloads for administrators, especially when those typos do not really pose any problems. I will therefor not merge the pull request and close it instead.

stweil:
Is there additional hassle? I have to run flyway:repair anyway and see no additional workloads.

stweil:
@solth, issue #6139 might require changes which are not in comments and which also require flyway:repair. Therefore I think that fixing typos in comments can be done without any additional hassle and workloads.

henning-gerhardt:
> @solth, issue https://github.com/kitodo/kitodo-production/issues/6139 might require changes which are not in comments and which also require flyway:repair. Therefore I think that fixing typos in comments can be done without any additional hassle and workloads.

The difference is: in this pr only "comments" are changed which did not influence the migration itself. In Issue #6139 may the migration itself need be changed (if this is really needed on MacOS) which would be not nice but more understandable.

Using `mvn flyway:repair` did only "fixing" the entries inside the flyway database table without running the migration script itself which may or not lead to different database schema's if the changed migration file did something total different in its latest version.

stweil:
Meanwhile the flyway migration produces lots of warnings for migration scripts which seem to use SQL which will no longer be supported in future versions of MariaDB. So sooner or later there will be a hard requirement for updated migration scripts. Then minor fixes like the ones here can be applied without causing "additional hassle".

Issue #6138, opened 2024-07-21
CI GitHub action fails randomly


stweil:
**Describe the bug**
The CI GitHub action does not pass reliably but fails sometimes with one or both of these errors:
```
Error:  Errors: 
Error:    CalendarST.createProcessFromCalendar » StaleElementReference stale element ref...
Error:    ImportingST.checkHierarchyImport:180 » StaleElementReference stale element ref...
[INFO] 
Error:  Tests run: 79, Failures: 0, Errors: 2, Skipped: 2
```

As long as the reason for the random failures are unknown, we must assume that such random failures can also occur in production environments.

**To Reproduce**
1. Run GitHub CI action several times until it fails.

**Expected behavior**
The CI action should not fail randomly.

**Release**
Git master, observed there since at least several weeks now.

stweil:
Here is a recent example: https://github.com/kitodo/kitodo-production/actions/runs/10005589215/job/27656601360#step:14:27222.
See also the [list of failing CI on master branch](https://github.com/kitodo/kitodo-production/actions/workflows/main.yml?query=is%3Afailure+branch%3Amaster).

stweil:
Issue #5378 describes a similar bug which was solved.

stweil:
Another error which also seems to occur randomly is this one:
```
[INFO] Running org.kitodo.selenium.CalendarST
[INFO ] 2024-07-29 13:00:02.840 [main] ProcessService - No metadata file for indexing: 1/meta.xml
[INFO ] 2024-07-29 13:00:02.840 [main] ProcessService - No metadata file for indexing: 1/meta.xml
[INFO ] 2024-07-29 13:00:02.868 [main] MetsService - Reading 2/meta.xml
[INFO ] 2024-07-29 13:00:02.893 [main] ProcessService - No metadata file for indexing: 3/meta.xml
[INFO ] 2024-07-29 13:00:02.894 [main] ProcessService - No metadata file for indexing: 3/meta.xml
[INFO ] 2024-07-29 13:00:02.992 [main] ProcessService - No metadata file for indexing: 1/meta.xml
[INFO ] 2024-07-29 13:00:02.992 [main] ProcessService - No metadata file for indexing: 1/meta.xml
[INFO ] 2024-07-29 13:00:03.026 [main] MetsService - Reading 2/meta.xml
[INFO ] 2024-07-29 13:00:03.063 [main] ProcessService - No metadata file for indexing: 1/meta.xml
[...]

PR #6150, opened 2024-07-26
Add option to show last comment directly in comment column

oliver-stoehr:
This pull request adds the option `showLastComment` to the kitodo-config.properties file.
When set to true this will display the most recent comment in the  comments column in the task list and the process list.
This increases the informations the user can retrieve at a glance, without the need to hover over every comment icon.
Bildschirmfoto 2024-07-26 um 09 55 31
This change does not affect the functionality of the enhanced tooltip implemented in #5403. The new option is deactivated by default (the new parameter `showLastComment` defaults to `false` if not present).

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/CommentTooltipView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/CommentTooltipView.java
@@ -80,7 +80,7 @@ public class CommentTooltipView {
         if (comments.isEmpty()) {
             return "";
         }
-        return comments.get(0).getMessage();
+        return comments.get(comments.size() - 1).getMessage();
     }
 
     /**
--- a/Kitodo/src/main/java/org/kitodo/production/forms/CommentTooltipView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/CommentTooltipView.java
@@ -68,4 +68,28 @@ public class CommentTooltipView {
     public List<Comment> getComments(TaskDTO taskDTO) {
         return getComments(taskDTO.getProcess());
     }
+
+    /**
+     * Get the most recent comment of the given process.
+     *
+     * @param processDTO process as ProcessDTO
+     * @return message of the comment
+     */
+    public String getLastComment(ProcessDTO processDTO) {
+        List<Comment> comments = getComments(processDTO);
+        if (comments.isEmpty()) {
+            return "";
+        }
+        return comments.get(0).getMessage();
+    }
+
+    /**
+     * Get the most recent comment of process containing the given task.
+     *
+     * @param taskDTO task as TaskDTO
+     * @return message of the comment
+     */
+    public String getLastComment(TaskDTO taskDTO) {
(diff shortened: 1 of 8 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/CommentTooltipView.java
```diff
@@ -68,4 +68,28 @@ public List<Comment> getComments(ProcessDTO processDTO) {
...
+     *
+     * @param processDTO process as ProcessDTO
+     * @return message of the comment
+     */
+    public String getLastComment(ProcessDTO processDTO) {
+        List<Comment> comments = getComments(processDTO);
+        if (comments.isEmpty()) {
+            return "";
+        }
+        return comments.get(0).getMessage();
```
henning-gerhardt, line 83: You are sure that `get(0)` is returning the last created comment? [CommentDAO.getAllByProcess](https://github.com/kitodo/kitodo-production/blob/master/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/CommentDAO.java#L53-L56) is returning a list of comments ordered by ascending comment id and as the comment ids are increasing by adding new comment entries the `get(0)` call will return the oldest created entry.

oliver-stoehr: Thank you for pointing out! This was a copy paste error, since I developed this feature in a branch where the comment order in the tooltip is reversed (most recent first). I added another commit to retrieve the last entry from the list.

henning-gerhardt: As this whole class did not have any tests this could break again if your other branch with the reversed ordering get merged. Maybe you are remember to change it here again but if could even get forgotten.

stweil:
The CI failure is unrelated to this PR (see issue #6138).

henning-gerhardt (approved):
Code looks good in general. I tried the your changes without PR #6127 included and they are working for me. Only a small note: my IDE spell checking is telling me that "comments" in all your additions are written wrong. I think that the plural of "comment" is still "comment" and not "comments" but I'm not a language expert.

solth:
> I think that the plural of "comment" is still "comment" and not "comments" but I'm not a language expert.

No, the plural of "comment" is "comments" (see for example https://en.wiktionary.org/wiki/comments#English). I think the IDE incorrectly interprets "comments" as a the - wrong - genitive of "comment" and therefore marks this as an error, even though it's not intended as a genitive, but rather as the plural of "comment" in this context.

henning-gerhardt:
> > I think that the plural of "comment" is still "comment" and not "comments" but I'm not a language expert.
> 
> No, the plural of "comment" is "comments" (see for example https://en.wiktionary.org/wiki/comments#English). I think the IDE incorrectly interprets "comments" as a the - wrong - _genitive_ of "comment" and therefore marks this as an error, even though it's not intended as a genitive, but rather as the plural of "comment" in this context.

Good to know - so ignore my note from my IDE.

PR #6151, opened 2024-07-28
Remove unused dependency org.apache.pdfbox:pdfbox

stweil:
It's only used in Kitodo-Docket which has its own dependency entry.

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -145,18 +145,6 @@
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-web</artifactId>
         </dependency>
-        <!-- overwrite parent dependency -->
-        <dependency>
-            <groupId>org.apache.pdfbox</groupId>
-            <artifactId>pdfbox</artifactId>
-            <version>2.0.32</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-logging</groupId>
-                    <artifactId>commons-logging</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
         <dependency>
             <groupId>org.apache.poi</groupId>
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
I'm not sure that this is really working. The dependency is needed in the Kitodo-Docket module as at least a compile dependency but even needed as an execution / runtime dependency as the removed dependency is not part of the generated jar file of the module. I tried to create a PDF document (docket) which was working but maybe this is only working on my local development system with a  lot of different running Kitodo.Production versions.

The only change which I notice is that the used version of the `fontbox.jar` is changing but all used libraries in the packaged war file are the same.

markusweigelt:
I don't have a current development system for Kitodo.Production, as I am no longer developing for it. Therefore, I'll proceed with the change if the tests pass. The functional test of the PDF export within the application will then need to be checked by others.

stweil:
@solth, I now have several pull requests for dependencies (pom.xml, */pom.xml) in the queue, and several others are in my development branches. They fix issues reported by `mvn dependency:analyze`. Would you prefer single pull requests for independent dependency issues? Or should I combine all commits in a single pull request (or add them to one of the existing pull requests)?

solth:
@stweil preparing release 3.7.0 has priority for me at the moment. I will take a look at your pull requests or assign other reviewers after the release, but right now they have to wait.

stweil:
Thanks for explaining your reasons. Some of the dependency issues could be considered (minor) bug fixes, e.g. if unneeded dependencies unnecessarily increase the final war file size or if old versions are used, but since this are minor issues, they can wait until 3.7.0 is released. I'll merge my fixes into my opensearch branch and send more pull requests after 3.7.0.

stweil:
Rebased + fixed merge conflict.

PR #6155, opened 2024-07-30
Correction comment state

### LINKED ISSUE #5676: Marking a correction comment as "resolved" does not update the index properly

solth:
This pull request slightly adjusts the `WorkflowControllerService.solveProblem` function so that a comment's `corrected` flag with value `TRUE` is saved to the database _before_ `closeTaskByUser` is called. That method loads the comment from the database and requires the updated `corrected` state to work properly. 

The updated `corrected` state is then saved to the index field `correctionCommentStatus` which in turn is used to determine whether the grey or red exclamation mark icon should be displayed in the process and task lists. Thereby fixes #5676

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
@@ -405,6 +405,10 @@ public class WorkflowControllerService {
             taskService.replaceProcessingUser(currentTask, getCurrentUser());
             taskService.save(currentTask);
         }
+        // "currentTask" and "process" are re-added to the comment object without saving it to database, again, because
+        // comment is further used in calling method, hence we make use of a side effect here (without requiring to save
+        // the comment again, because process ID and current task ID remain the same)
+        // TODO: refactor to improve readability and maintainability
         comment.setCurrentTask(ServiceManager.getTaskService().getById(comment.getCurrentTask().getId()));
         if (Objects.nonNull(comment.getProcess())) {
             comment.setProcess(ServiceManager.getProcessService().getById(comment.getProcess().getId()));
--- a/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
@@ -386,6 +386,13 @@ public class WorkflowControllerService {
      */
     public void solveProblem(Comment comment, TaskEditType taskEditType)
             throws DataException, DAOException, IOException {
+        comment.setCorrected(Boolean.TRUE);
+        comment.setCorrectionDate(new Date());
+        try {
+            ServiceManager.getCommentService().saveToDatabase(comment);
+        } catch (DAOException e) {
+            Helper.setErrorMessage("errorSaving", new Object[] {"comment"}, logger, e);
+        }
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
```diff
@@ -399,12 +406,8 @@ public void solveProblem(Comment comment, TaskEditType taskEditType)
...
         }
         comment.setCurrentTask(ServiceManager.getTaskService().getById(comment.getCurrentTask().getId()));
-        comment.setCorrected(Boolean.TRUE);
-        comment.setCorrectionDate(new Date());
-        try {
-            ServiceManager.getCommentService().saveToDatabase(comment);
-        } catch (DAOException e) {
-            Helper.setErrorMessage("errorSaving", new Object[] {"comment"}, logger, e);
+        if (Objects.nonNull(comment.getProcess())) {
+            comment.setProcess(ServiceManager.getProcessService().getById(comment.getProcess().getId()));
```
BartChris, line 410: We are setting the currentTask and the process again but do not save the comment object. If calling those setters is necessary should we not persist the comment object again?

BartChris:
Functionality-wise everything works now, i just have a comment.

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
solth: Yes, you are right, this is quite confusing. The reason why we re-add the process (and task!) to the comment without saving it again is that we are making use of a side effect here. This `Comment` object is used again in the calling method afterwards, but with an updated `currentTask` and `process` property. It's not a _different_ process or task that is added here, but rather _updated_ objects, so re-setting these properties in the comment acts as a "refresh" of the object. Since the database entry of the comment only contains the _ID_ of the `process` and the `currentTask`, there is no need to save it again, because the IDs didn't change (e.g. its the same process and currentTask as before, like I mentioned above).

This is indeed implemented in a sub-optimal way and should be refactored more fundamentally when we find the time to do so but right now I do not have the resources for that, unfortunately. I think the bug should be fixed anyway, even if its just an incremental improvement over the status quo and not yet the optimal solution. I can add a comment or `TODO` as a reminder to the code that this code should be refactored and improved, if you like.

BartChris (approved):

Kitodo/src/main/java/org/kitodo/production/services/workflow/WorkflowControllerService.java
BartChris: @solth I see, thanks a lot for the explanation. In think you can go ahead and merge. Adding a TODO would be helpful.

PR #6157, opened 2024-07-30
Add missing `Override` annotation

solth:
CodeQL reported:

> Java enables you to annotate methods that are intended to override a method in a superclass. Compilers are required to generate an error if such an annotated method does not override a method in a superclass, which provides increased protection from potential defects. An annotated method also improves code readability.

This PR adds the missing annotation.


stweil (approved):
Is it required in SearchDatabaseService.java, too?

stweil:
The build failure is unrelated to this PR (see issue #6138).

solth:
> Is it required in SearchDatabaseService.java, too?

Yes, probably. Thanks for pointing this out. Since there seem to be _many_ more corresponding notes for missing `Override` annotations according to CodeQL however (at least 178 in the last report), I will not change them all in this PR but rather add a seperate PR to tackle the problem more comprehensively when I have time. This PR was just opened because this particular missing annotation was just reported recently.

PR #6160, opened 2024-07-30
Change groupId of Kitodo-Validation

stweil:
~No other module has a groupId, and there is no obvious reason for one here.~

The other modules have no explicit groupId, but use `org.kitodo` implicitly.

### DIFF
--- a/Kitodo-Validation/pom.xml
+++ b/Kitodo-Validation/pom.xml
@@ -23,7 +23,7 @@
     </parent>
 
     <name>Kitodo - Validation</name>
-    <groupId>org.kitodo.validation</groupId>
+    <groupId>org.kitodo</groupId>
     <artifactId>kitodo-validation</artifactId>
 
     <properties>


stweil:
Maybe using `<groupId>org.kitodo</groupId>` for all modules would be an even better solution.

stweil:
Some tests fail with this change. I still have to find the reason.

stweil:
As removing the groupId breaks the CI tests, I now changed it to the default value "org.kitodo" which seems to work.

stweil:
I noticed the strange setting for Kitodo-Validation while running `ls ~/.m2/repository/org/kitodo` for another issue:
```
kitodo			kitodo-data-format	kitodo-image-management			kitodo-query-url-import
kitodo-api		kitodo-data-management	kitodo-longtermpreservation-validation	kitodo-validation
kitodo-command		kitodo-docket		kitodo-persistentidentifier		kitodo-xml-schemaconverter
kitodo-data-editor	kitodo-file-management	kitodo-production			validation
```

The list shows the old directory `validation` which looks strange here. With the PR, it is replaced by `kitodo-validation` which uses the same naming convention as all other modules.

stweil:
The latest build failure is unrelated to this PR (see issue #6138).

solth (approved):

Issue #6161, opened 2024-07-31
Documentation cannot be created with Javadoc because of errors


stweil:
**Describe the bug**
`mvn javadoc:javadoc` fails with errors after processing the first module (Kitodo-API).
More errors exist in other modules.

**To Reproduce**
1. Run `mvn javadoc:javadoc
2. See error

**Expected behavior**
There must not be any errors because otherwise it is not possible to build the documentation.

**Release**
All recent releases including Git master.

PR #6162, opened 2024-07-31
Fix javadoc errors and warnings for Kitodo-API

stweil:
This partially fixes issue #6161.


stweil:
After these fixes, the next errors and warnings are reported for Kitodo-DataManagement.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
```diff
@@ -29,7 +29,7 @@ public class MediaPartial {
      *
      * @param begin
      *         The begin as formatted time in form of
-     *         {@link org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
+     *         ([0-1]\\d|2[0-3]):[0-5]\\d:[0-5]\\d\\.\\d{3}
```
henning-gerhardt, line 32: This is a bad change as changes to this static property won't get visible in the documentation. You should fix this by a working mechanism instead of copying the current value here.

stweil: I'm afraid this is not possible without a major refactoring because `FORMATTED_TIME_PATTERN` is defined in the Kitodo module which depends on Kitodo-API. Adding the required dependency here would create a circular dependency.

henning-gerhardt: You can use f.e. the `{@see }` annotation or no annotation at all to keep a readable information where to find the used pattern. Hard coding the pattern at this place has only one meaning: it is obsolete / wrong at time of change as no one will update the used pattern at this place if its changed in the original place. But this is only my opinion and I would only offer an other opinion or possibility to solve the issue. You can do what ever you want and even ignore my comments if they did not fit or not right.

stweil, line 32:
```suggestion
     *         HH:MM:SS.mmm (hour, minutes, seconds, milliseconds)
```

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
```diff
@@ -40,10 +40,10 @@ public MediaPartial(String begin) {
...
```
stweil, line 43:
```suggestion
     *         HH:MM:SS.mmm (hour, minutes, seconds, milliseconds)
```

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
```diff
@@ -40,10 +40,10 @@ public MediaPartial(String begin) {
...
      * @param extent
      *         The extent as formatted time in form of
-     *         {@link org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
+     *         ([0-1]\\d|2[0-3]):[0-5]\\d:[0-5]\\d\\.\\d{3}
```
stweil, line 46:
```suggestion
     *         HH:MM:SS.mmm (hour, minutes, seconds, milliseconds)
```

stweil (commented):
HH:MM:SS.mmm is a better description for the expected time format than showing the regular expression.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
stweil: Constructive comments are always welcome. I think that I found a good solution now. This solution assumes that changes of the time pattern HH:MM:SS.mmm are not likely.

stweil:
Can this PR be merged? Then I could work on the Javadoc warnings and errors for other modules of Kitodo.

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
solth: @stweil I think @henning-gerhardt's concern isn't addressed by using another time pattern that is more readable. Instead you should try the solution he suggested by replacing `@link` with `@see`. 

I just tried it. In contrast to `@link`, `@see` does not require adding new dependencies to the moduls in order to work. Instead, it resolves the Maven JavaDoc errors you documented in the original issue, and with the existing dependencies. 

At the same time it also takes into account that the pattern could change one day without the need to adjust the corresponding documentation in multiple places.

Therefor I would suggest to revert your changes and replace `@link` with `@see` in the code lines you changed.

solth: I cannot say if this time pattern changes in the future and neither can you.

Since the issue you try to fix is about errors when creating the documentation with Maven and JavaDoc, I suggest to relocate the discussion about how JavaDoc should be structured and move forward with the solution desribed above.

Please revert your changes in the lines in question and replace `@link` with `@see`, then I will merge this pull request and you can continue with your planned adjustments. That's in your intereset, isnt' it?

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
```diff
@@ -29,7 +29,7 @@ public class MediaPartial {
      *
      * @param begin
      *         The begin as formatted time in form of
-     *         {@link org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
+     *         HH:MM:SS.mmm (hour, minutes, seconds, milliseconds)
```
stweil, line 32:
```suggestion
     *         {@see org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
```

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
```diff
@@ -40,10 +40,10 @@ public MediaPartial(String begin) {
...
```
stweil, line 43:
```suggestion
     *         {@see org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
```

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
```diff
@@ -40,10 +40,10 @@ public MediaPartial(String begin) {
...
      * @param extent
      *         The extent as formatted time in form of
-     *         {@link org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
+     *         HH:MM:SS.mmm (hour, minutes, seconds, milliseconds)
```
stweil, line 46:
```suggestion
     *         {@see org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
```

Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java
stweil: I now replaced `@link` by `@see`. This does not fix the Javadoc errors, but creates different new ones. When I also remove `{` and `}` I get again the "reference not found" errors:
```
[ERROR] /Users/stweil/src/github/kitodo/kitodo-production/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java:32: error: reference not found
[ERROR]      *         @see org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN
[ERROR]                     ^
[ERROR] /Users/stweil/src/github/kitodo/kitodo-production/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java:43: error: no tag name after @
[ERROR]      *         {@see org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
[ERROR]                ^
[ERROR] /Users/stweil/src/github/kitodo/kitodo-production/Kitodo-API/src/main/java/org/kitodo/api/dataformat/MediaPartial.java:46: error: no tag name after @
[ERROR]      *         {@see org.kitodo.production.helper.metadata.MediaPartialHelper#FORMATTED_TIME_PATTERN}
[ERROR]                ^
```

stweil: I now removed the links to the time format (which fixes the errors for Kitodo-API) and rebased the PR.

stweil:
Many of the other javadoc errors are similar to this one:
```
[ERROR] /Users/stweil/src/github/kitodo/kitodo-production/Kitodo-DataManagement/src/main/java/org/kitodo/data/elasticsearch/api/TypeInterface.java:21: error: bad use of '>'
[ERROR]  * Note: MySQL -> Databases -> Tables -> Columns/Rows ElasticSearch -> Indices
```

Do you have a suggestion how this should be fixed?

stweil:
How can we proceed with this topic?

solth (approved):

PR #6163, opened 2024-07-31
Remove junit4 support

henning-gerhardt:
This is the last pull request to remove the old needed dependency to execute Junit 4 based tests inside a Junit 5 test system. If this pull request is merged then all tests should be written / created as Junit 5 based tests.

The pull request include an optional commit to remove the transitive dependency to an old junit 4  from a used dependency. If this transitive dependency is not removed it is still possible to create or write Junit 4 code but the execution of this code should fail. So far as I could I discovered no issues with removing this transitive dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -542,12 +542,6 @@ from system library in Java 11+ -->
                 <scope>test</scope>
             </dependency>
             <dependency>
-                <groupId>org.junit.vintage</groupId>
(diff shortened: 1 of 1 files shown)


solth (approved):
Looks good and works. I think the second, optional commit should definitely be included. Otherwise developers might accidentially still write tests with JUnit 4 which should be prevented going forward from now.

PR #6166, opened 2024-08-02
Fix NullPointerException if ruleset could not be loaded

### LINKED ISSUE #6165: Stack trace if ruleset cannot be loaded
If the ruleset cannot be loaded (i.e. if there is a typo in the XML, and it is syntactically invalid) a stack trace is shown, that also hides the error message.

Stack trace:

        at org.kitodo.production.forms.createprocess.AddMetadataDialog.prepareAddableMetadataForStructure(AddMetadataDialog.java:78)
        at org.kitodo.production.forms.createprocess.AddMetadataDialog.metadataAddableToStructureElement(AddMetadataDialog.java:67)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        ...

matthias-ronge:
Fixes #6165

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/AddMetadataDialog.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/AddMetadataDialog.java
@@ -12,6 +12,7 @@
 package org.kitodo.production.forms.createprocess;
 
 import java.util.List;
+import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


solth (changes requested):

solth (approved):
I think we should also display or at least log some sort of information about why the ruleset could not be loaded, but I can live with that being done in a separate pull request.

PR #6169, opened 2024-08-02
Kitodo.Production 3.7.0

solth:
Update version number to 3.7.0 in pom files and supported versions in `SECURITY.md`.


henning-gerhardt (approved):
Changes look good, files are created with right name schema, pom files contain right version information and the application itself is reporting back correct version information.

PR #6170, opened 2024-08-06
Update version to '3.8.0-SNAPSHOT'

solth:
Update version to `3.8.0-SNAPSHOT`.


henning-gerhardt (approved):
Changes looks good, right file names are created, content of pom files looks correct and the executed application is displaying even the new version.

PR #6173, opened 2024-08-07
Improve suggestions for properties in filter menu

oliver-stoehr:
Filtering for a property uses the following syntax: `property:KEY:VALUE`.
Currently the strings `property:` and `KEY` (dynamic value) can be taken from the suggestions. The user has to type the second colon by hand every time, before typing the actual, wanted value.

This pull request adds the colon to the suggestion, so the user only has to type the value manually.


BartChris (approved):
Looks good and works!

Issue #5934, opened 2024-02-21
Reduce the time to open superordinate processes with many subordinate processes


andre-hohmann:
**Is your feature request related to a problem? Please describe.**
Processes of superordinate processes as for example processes of newspaper years with more than 350 and up to 600 subordinate processes can take up to 20 minutes and more. 
Furthermore, corrections in ordering, ... is not possible as for example: 
- #5821 

**Describe the solution you'd like**
The time to open the superordinate process and the time to modify the structure tree, ... should be reduced. 

**Additional context**
There has been an issue before, but during the work in the productive system, the need open the processes faster rises again. 
- #3436

BartChris:
I suppose that one of the reasons is that, there are a lot of look ups needed to construct the list although the provided information is often not really useful. (See: https://github.com/kitodo/kitodo-production/discussions/4925#discussioncomment-7653084) 
If you have a lot of volumes you would e.g. be interested in the information which is configured as "displaySummary" in the ruleset. (https://github.com/kitodo/kitodo-production/wiki/Regelsatz-3.x#schl%C3%BCssel-mit-besonderer-verwendung), but this is not used here.
Regarding performance: Would it maybe be possible to use some of the mechanisms wich Hibernate Search offers to display this information way faster. It seems like we do lookups in hundreds of XML files to get minimal information.

PR #6185, opened 2024-08-19
Speed up display of parent processes

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/5934

As outlined in issue https://github.com/kitodo/kitodo-production/issues/5934, we currently face performance problems when working with parent processes in the editor, which contain a lot of children elements. In the issue discussion i thought that the problem lies in opening a lot of XML files for detection of the base type, but we are already using the index. I discovered that the problem seems to stem from accessing the index hundreds of times.
I tried to correct that and was able to bring the time to show a parent process with 600 files from more than 30 seconds to around 3 seconds.
My implementation does four things:
- Instead of retrieving the base types for the linked processes when the structure tree is constructed recursively, i fetch all the base types before the structure tree is constructed
- Instead of retrieving the documents from the Elasticsearch index one by one i fetch them in one go using the Mget operation. (https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-multi-get.html)
- We do not have to convert the document to a DTO since we only need the basetype, which we can fetch directly from the Elasticsearch response, which is reduced in size by specifiying the only source field we need: "baseType"
- We can use a view cache for already derived views

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -558,8 +558,26 @@ public class StructurePanel implements Serializable {
         invisibleRootNode.setExpanded(true);
         invisibleRootNode.setType(STRUCTURE_NODE_TYPE);
         addParentLinksRecursive(dataEditor.getProcess(), invisibleRootNode);
-        buildStructureTreeRecursively(structure, invisibleRootNode);
+        List<Integer> processIds = getAllLinkedProcessIds(structure);
+        Map<Integer, String> processTypeMap = processIds.isEmpty() ? Collections.emptyMap() : fetchProcessTypes(processIds);
+        buildStructureTreeRecursively(structure, invisibleRootNode, processTypeMap);
         return invisibleRootNode;
+    }
+
+    private List<Integer> getAllLinkedProcessIds(LogicalDivision structure) {
+        return structure.getAllChildren().stream()
+                .filter(division -> division.getLink() != null)
+                .map(division -> ServiceManager.getProcessService().processIdFromUri(division.getLink().getUri()))
+                .collect(Collectors.toList());
+    }
+
+    private Map<Integer, String> fetchProcessTypes(List<Integer> processIds) {
+        try {
+            return ServiceManager.getProcessService().getIdBaseTypeMap(processIds);
+        } catch (DataException e) {
+            Helper.setErrorMessage("metadataReadError", e.getMessage(), logger, e);
+            return Collections.emptyMap();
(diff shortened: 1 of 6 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -585,30 +604,30 @@ private String buildPageRangeFromLogicalDivision(LogicalDivision structure) {
      * @param structure the logical division
      * @return the StructureTreeNode instance
      */
-    private StructureTreeNode buildStructureTreeNode(LogicalDivision structure) {
+    private StructureTreeNode buildStructureTreeNode(LogicalDivision structure,  Map<Integer, String> idTypeMap,
```
henning-gerhardt, line 607: Please adjust the JavaDoc part as the method parameter list has changed.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -1730,6 +1730,10 @@ public String getBaseType(int processId) throws DataException {
         return "";
     }
 
+    public Map<Integer, String> getIdBaseTypeMap(List<Integer> processIds) throws DataException {
```
henning-gerhardt, line 1733: Please add a JavaDoc comment to this public method.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -620,10 +639,12 @@ private StructureTreeNode buildStructureTreeNode(LogicalDivision structure) {
      *
      * @param structure the current logical structure
      * @param result the current corresponding primefaces tree node
+     * @param processTypeMap the mapping of process id to basetype
      * @return a collection of views that contains all views of the full sub-tree
      */
-    private Collection<View> buildStructureTreeRecursively(LogicalDivision structure, TreeNode result) {
-        StructureTreeNode node = buildStructureTreeNode(structure);
+    private Collection<View> buildStructureTreeRecursively(LogicalDivision structure, TreeNode result, Map<Integer,
```
henning-gerhardt, line 645: Last method parameter is missing in the JavaDoc comment.

henning-gerhardt (approved):
Changes looks good and there is a big speed improvement. 

Two things:
1. I'm only not sure how this change will survive the hibernate-search integration as there is a total other access to this needed data.
2. This change should be back ported to the 3.7.x release branch as it is a big improvement for the 3.7.x releases which may be used a long time and it is currently not possible to change the metadata of hierarchy processes with a lot of child processes.

PR #6186, opened 2024-08-20
Translate metadata keys in mass import record table header

solth:
While writing the documentation for the [mass import](https://github.com/kitodo/kitodo-production/wiki/Massenimport) I noticed that the column headers in the record table only contain the keys, not the translated labels, of each metadata. This pull request changes that so that now the correct translation of the metadata keys are displayed in the column headers. (if the CSV file contains unknown metadata keys that are not defined in the current ruleset, the verbatim strings from the CSV are displayed instead)


henning-gerhardt (approved):
It is only a code reading review and the changes looks good and are similar to the changes in #6188.

PR #6187, opened 2024-08-22
Fix CID 431571 (PA: Public Attribute)

stweil:
[SEI CERT rule OBJ01-J](https://wiki.sei.cmu.edu/confluence/display/java/OBJ01-J.+Limit+accessibility+of+fields) requires that accessibility of fields must be limited. ([From SpotBugs description](https://scan4.scan.coverity.com/doc/en/spotbugs/sb_checker_ref.html#FB.PA_PUBLIC_MUTABLE_OBJECT_ATTRIBUTE))

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/file/BackupFileRotation.java
+++ b/Kitodo/src/main/java/org/kitodo/production/file/BackupFileRotation.java
@@ -49,7 +49,7 @@ public class BackupFileRotation {
     private String format;
     private Process process;
 
-    public final FileService fileService = ServiceManager.getFileService();
+    private final FileService fileService = ServiceManager.getFileService();
 
     /**
      * Start the configured backup.


stweil:
9 classes use `private static final`, 2 classes use `private final transient`, 1 class uses `private static`. 5 classes use `ServiceManager.getFileService()` without a local variable. Is there a good reason for so many different patterns?

stweil:
The CI build fails because of issue #6138.

henning-gerhardt:
> 9 classes use `private static final`, 2 classes use `private final transient`, 1 class uses `private static` and 5 classes use `ServiceManager.getFileService()` without a local variable. Is there a good reason for so many different patterns?

Because of `ServiceManager.getFileService()` is used in different context and classes and class inheritance which need different access levels. Or if the call is used only once in a class you not need to define it as a global class variable as it needs only to be a method variable.

stweil:
> I don't know what "CID 431571" or "PA" is meaning.

"CID" is the Coverity Scan ID of an issue which was reported there, "PA: Public Attribute" is the short form of the reported error or warning. We have a few handfuls of other commits which used exactly the same kind of description, for example commit 841c2aa76b208fd2a9fee49ff3ecb1b9bdc731a8.

stweil:
> Generally I would prefer if discussions like this would be moved to a proper GitHub discussion

See new discussion #6190.

solth (approved):

PR #6198, opened 2024-08-28
Submit filters with numpad enter

oliver-stoehr:
When a filter was entered and the enter key on the numpad is pressed, the filter is not submitted but the page will be reloaded.
This pr adds the code for the numpad enter key, so filters can be submitted correctly when pressing this key.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/filter.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/filter.js
@@ -131,6 +131,7 @@ function handleKeydown(event) {
             }
             break;
         case "Enter":
+        case "NumpadEnter":
             if ($(FILTER_OPTIONS_FORM_WRAPPER).is(':visible') && $(SELECTED_SUGGESTION_ITEM).length === 1) {
                 selectConfirm();
             } else {


solth (approved):
Works and code change looks correct. Would you mind openening a corresponding pull request against the `3.7.x` branch?

stweil:
`git cherry-pick` might also work for 3.7.x. Isn't that easier than having two pull requests for all changes?

solth:
> `git cherry-pick` might also work for 3.7.x. Isn't that easier than having two pull requests for all changes?

I know. That's how I create my separate pull requests for prior releases - in this case 3.7 - too, and that's how I would expect the author of a pull request to create their branches with corresponding patches for different releases as well. 

In most cases the changes are probably small enough to work without additional adjustments in separate branches, but that is not guaranteed and we already had numerous examples where verbatim patches couldn't be applied to different branches. If a change does _not_ work with the different code base the author is responsible for creating a corresponding adjustment to his patch.

Since cherry-picking the relevant commits is so easy it shouldn't put much more workload on the author of the changes anyway, so that is the way I follow for Kitodo. To create an established procedure I want to apply this policy to all bug fixes, independet of their complexity.

PR #6203, opened 2024-08-29
Update process when assigning task

oliver-stoehr:
When a task is assigned to a user the tasks status changes from "open" to "in work". This change is not saved properly in the related process object resulting in an outdated progress bar in the process list.
This change saves the tasks related objects to address this issue.


BartChris (approved):
Works for me.

PR #6213, opened 2024-09-02
Add permission to use mass import

### LINKED ISSUE #5722: Add different authority for mass import
Currently every user which had the authority to add a process has even the right to do a mass import. See at least https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/webapp/WEB-INF/templates/includes/projects/projectList.xhtml#L140 and maybe on some other places where the mass import feature can be started / used.

Mass import feature should have its own authority which should only added to users which need this.

If no authority is possible then made it possible to deactivate the mass import feature on client level (for every client independently different configurable).

solth:
Fixes #5722


henning-gerhardt (approved):
Changes looks good and the functionality with and without the right permission to access the mass import dialog is working. I did not check the mass import itself as this is not related to this changes or should not be related at least.

PR #6214, opened 2024-09-02
Don't run CodeQL if only minimized JavaScript was changed and fix some issues

stweil:
It's not necessary to run CodeQL (and spend related resources like energy and time) for all kinds of file changes. Here changes of minimized JavaScript are ignored.

In addition some missing `@Override` tags are added to fix related CodeQL issues.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -15,6 +15,8 @@ on:
   push:
   pull_request:
     branches: [ master ]
+    paths-ignore:
+      - '**/*_min.js'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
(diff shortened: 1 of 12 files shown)


.github/workflows/codeql.yml
```diff
@@ -15,6 +15,8 @@ on:
   push:
   pull_request:
     branches: [ master ]
+    paths-ignore:
+      - '**/*_min.js'
```
henning-gerhardt, line 19: This will hide any file ending of "_min.js" which must not be necessary a minimized JavaScript version of a bigger JavaScript file. Other minimized JavaScript files which have a other name schema for minimization will be checked still. If I would to ignore a file I would name it directly and not use a global big match pattern. But this is only my opinion and can be ignored if they did not right.

stweil: I think jquery-2.1.1.min.js should be removed because it is a ten years old security risk. Maybe Kitodo can use the JQuery package which is part of Debian / Ubuntu if this is required.

solth: I think @henning-gerhardt has a very good point here. The number of JavaScript files in Kitodo.Production is not great enough to warrant the use of wildcard filters when listing each file that should be ignored would be more accurate and correct. 

@stweil when you apply this change request and rebase your branch against the current `main` branch I think we can go forward and merge this pull request.

PR #6215, opened 2024-09-03
Use FileUtils.deleteDirectory instead of FileSystemUtils.deleteSubDirectories

stweil:
This eliminates an Elasticsearch dependency and replaces it by an already existing dependency.


stweil:
I wonder whether these private functions `removeOldDataDirectories` should be removed. The little bit of code which is only used once could also be used directly.

solth (approved):

PR #6211, opened 2024-08-30
Fix wrong checks of empty string in LdapUser context

gerlingsm:
In the following [commit](https://github.com/kitodo/kitodo-production/commit/b598388abce560b44a69f8aa0a76f43a3ec41cfa?diff=split&w=0) ".isEmpty()" was replaced with "StringUtils.isBlank".
But unfortunately the negation of the calls has been lost which is why they are now testing the opposite of what they should.
This merge request should fix the error.
fix #6210

PR #6217, opened 2024-09-04
[3.7] Fix wrong checks of empty string in LdapUser context

gerlingsm:
backport of #6211


solth:
@gerlingsm thanks a lot for this backport fix!

solth (approved):

PR #6229, opened 2024-09-17
Update ActiveMQ client to latest stable version

henning-gerhardt:
Updating to latest stable version of ActiveMQ with Java 11 support. ActiveMQ at version 6 needs at least Java 17 for running.

See current support table of ActiveMQ: https://activemq.apache.org/components/classic/download/


stweil (approved):
Looks good, thanks. The next update to 5.18.6 is expected for October, so maybe we can upgrade to this version soon.

PR #6230, opened 2024-09-19
[3.7] MySQL dependency update

solth:
Update MySQL dependency to fix security issue

Based on #6159 and #6205


henning-gerhardt (approved):
Changes look good and as the CI tests are not complaining about the MySQL usage it should work. I can test this changes only with MariaDB server in the background whih did not say anything about a MySQL server usage.

Issue #6231, opened 2024-09-23
Metadata elements of collapsed groups are lost when triggering a `preserve` action


solth:
**Describe the bug**
Metadata elements with multiple options that are part of metadata groups are not saved when the metadata group is collapsed in the metadata editors metadata panel and a `preserve` action is triggered by changing the value of another metadata followed by a focus change or `blur` event.

**To Reproduce**
1. Create a metadata group like a "Person" that contains at least one metadata that is rendered as a pulldown menu with options
2. Select one option of the metadata that is part of the metadata group created in step 1
3. Collapse the metadata group in the metadata panel
4. Click in the input text field of some other simple metadata and change its value
5. Leave the metadata field by clicking somewhere else, which triggers a `blur` event which in turn calls the `preserve` method of the metadata panel
6. Save and close process
7. Re-open process in metadata editor or check `meta.xml` and see that the metadata with it's selected option is missing

If step 3. is skipped and the metadata group remains exapanded before triggering the `preserve` action and saving the process, the metadata with its selected option is saved correctly.

**Expected behavior**
No metadata should be lost when saving collapsed metadata groups and the saved content should not depend on the visual state regarding collapsed and expanded metadata groups in the metadata table.

**Release**
3.7

**Additional context**
Might be related to #5932 because both the editor and the import mask use the same `metadataTreeTable` component.

Issue #5932, opened 2024-02-21
Selection of check boxes is deleted for collapsed metadata groups during creation of process


andre-hohmann:
**Describe the bug**
A process with many metadata groups for persons is created and the role code is captured by mandatory check boxes. 
For a better navigation, the metadata groups are collapsed. 
The process cannot be saved, because the selection of the mandatory check boxes is deleted and the following warn message is shown: 
- `Beziehungskennzeichnung (Code): Validation Error: Value is required.`

**To Reproduce**
1. Go to 'Create new process'
2. Import metadata of a catalog entry with metadata groups with checkboxes (_for example: PPN (K10plus) 1863817913_)
3. Collapse metadata groups
4. Click on 'Save'
5. See the warn message

**Expected behavior**
The selected values in the check boxes should never be deleted, when a process is saved.

**Release**
3.6.1

**Desktop (please complete the following information):**
 - Chrome

andre-hohmann:
The following GIF animation demonstrates the described behavior: 
![ImportBeziehungskennzeichnung](https://github.com/user-attachments/assets/3ca74615-8b1e-455c-9c0e-af4df54a0d44)

PR #6232, opened 2024-09-24
Fix the disappearance of collapsed metadata

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/6231
fixes https://github.com/kitodo/kitodo-production/issues/5932

Before collapsing or expanding elements in our metadata tree, we should make sure that the state is persisted to not get lost while collapsing/expanding the tree.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
@@ -42,6 +42,13 @@
                  styleClass="no-header"
                  widgetVar="metadataTable"
                  id="metadataTable">
+        <p:ajax event="expand"
+                process="@this"
+                update="@(.ui-tree, .stripe.selected)" />
+
+        <p:ajax event="collapse"
+                process="@this"
+                update="@(.ui-tree, .stripe.selected)"/>
         <p:column>
             <span class="input-wrapper">
                 <!-- field label


BartChris:
OK, it seems like i overlooked those remoteCommands:
```
<p:remoteCommand name="updateTitleMetadata" update="@(.ui-tree, .stripe.selected)"/>
<p:remoteCommand name="updateTitleMetadataWithTable"
                     actionListener="#{DataEditorForm.metadataPanel.update}"
                     update="@(.ui-tree, .stripe.selected) metadataTable"/>

    <p:remoteCommand name="updateProcessMetadata"
                     actionListener="#{CreateProcessForm.processMetadata.update(CreateProcessForm)}"
                     update="metadataTable"/>
```

I thought we are calling javascript functions here but appearantly we are calling Java code. So this needs more investigation. But removing those completed callbacks fixes the issue in question. So something has to be tweaked here.

solth:
This really does seem to fix the documented issue(s) - altough I am not quite sure I understand _why_. Does the `update` automatically trigger the `preserve` action? And why is the `.stripe.selected` in the update as well? What does the gallery have to do with collapsed and expanded metadata groups in the metadata panel? Is it related to the title of structures being shown in the gallery?

BartChris:
@solth To be honest, i am still a little confused by the logic in the template. E.g.: In the onComplete callback we call for every element type `preserveMetadata();` but this function does not exist, at least i do not find it.

The main reason why i used the above notation is, that this is what is used in the remoteCommand at the end of the file:

```
<p:remoteCommand name="updateTitleMetadata" update="@(.ui-tree, .stripe.selected)"/>
```

And this remoteCommand is called from the normal change event callback:
```
<p:ajax event="change"
 oncomplete="preserveMetadata(); #{item.leading ? (request.requestURI.contains('metadataEditor') ? 'updateTitleMetadataWithTable();' : 'updateProcessMetadata();') : (request.requestURI.contains('metadataEditor') ? 'updateTitleMetadata();' : '')}"/>
```
under the specific conditions that the item should not have the 'leading' attribute, because then we also have to update the medatatable which resets the view.

https://github.com/kitodo/kitodo-production/blob/eded39f924588f71d3adb5c2bc36abe0bc815623/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessDetail.java#L141-L151

So i just tried the same here as we do for normal change events in case of a collapse.

But the fix also works with less logic. All that is needed is adding 

```
 <p:ajax event="collapse"/>
```
to the tree table:
```
 <p:treeTable value="#{root}"
                 var="item"
                 nodeVar="node"
                 styleClass="no-header"
                 widgetVar="metadataTable"
                 id="metadataTable">
        <p:ajax event="collapse"/>
```
This will update the Backing Beans, so that the collapsed data is not lost.

As far as i can see this fixes the bugs, but unfortunately it introduces a new bug in the process creation form. The problem described in https://github.com/kitodo/kitodo-production/issues/5932 does no longer appear, but now, when a group is collapsed the data is not transfered to the validation logic.  

When the collapse menu is expanded i cannot save, because of validation failures:

image

But when the panel is collapsed, those fields seem to not be included in the Ajax request going to the server, which then does the validation (without checking the collapsed fields) by calling e.g.

https://github.com/kitodo/kitodo-production/blob/eded39f924588f71d3adb5c2bc36abe0bc815623/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java#L737

I can therefor store a person without a (required) role and the system is not complaining. Without my change, this validation works. But on "save" the data in the collapsed panel is there. The validation in the metadata editor still works after my change.

I am a little bit clueless here.

BartChris:
Another change which fixes the (current, but not the other) issue is to change the `oncomplete` callback from

```
<p:ajax event="change"
                                oncomplete="preserveMetadata(); #{item.leading ? (request.requestURI.contains('metadataEditor') ? 'updateTitleMetadataWithTable();' : 'updateProcessMetadata();') : (request.requestURI.contains('metadataEditor') ? 'updateTitleMetadata();' : '')}"/>
```

to
```
 <p:ajax event="change"
                                oncomplete="preserveMetadata(); #{item.leading ? (request.requestURI.contains('metadataEditor') ? 'updateTitleMetadataWithTable();' : 'updateProcessMetadata();') : ''}"/>
```
Since we are basically reconstructing the metadata tree every time preserveMetadata(); is called, the additional call to updateProcessMetadata(); seems to have no effect for me. And i did not notice any effect. But maybe i am overlooking something.

Then i noticed that the preserveMetadata command for the metadataeditor is wrapped in a form 
```
 <h:form id="preserveMetadataForm">
            <p:remoteCommand id="preserveMetadataCommand"
                             name="preserveMetadata"
                             actionListener="#{DataEditorForm.preserveMetadataPanel()}"/>
 </h:form>
```
while the one for the process creation interface is not:
```
<p:remoteCommand name="preserveMetadata"
                     update="@none"
                     actionListener="#{CreateProcessForm.processMetadata.preserve}"/>
```

All those subtle difference might effect how ajax requests are handled in case of collapsed fields....

BartChris:
I think what adds a lot to the confusing behaviour here is, that the template is reused in different contexts. In case of the process creation the in place validation by PrimeFaces is used and in the metadata editor the validation is done server side. 

The difference in context might explain different behavior. Another thing i noticed, that the behaviour in https://github.com/kitodo/kitodo-production/issues/5932 can be adressed by **not** calling preserveMetadata in the onComplete-callback. So it seems to me that the process creation form and the editor form show different behaviour.

I am not sure wether it would be easiere to duplicate the metadataTreeTable template to not having to deal with different behaviour in different contexts. 

Since my pull request is more like a collection of observations, i will close it for now. Maybe somebody can come up with another approach which does not introduce new bugs and takes care of all the complexities in the JSF lifecycle.

solth:
> <p:remoteCommand name="updateTitleMetadata" update="@(.ui-tree, .stripe.selected)"/>

Yes, in this case the `.selected.stripe` needs to be updated because the gallery stripe titles are created from certain special metadata and any change to the metadata will therefore potentially alter the stripe titles as well.

> Then i noticed that the preserveMetadata command for the metadataeditor (metadataEditor.xhtml) is wrapped in a form while the one for the process creation interface (dataEdit.xhtml) is not

That's probably because the `dataEdit.xhtml` file is already included within a form (specifically inside the `editForm` in `baseEditView.xhtml`, on which `processFromTemplate.xhtml` is based, which in turn includes `dataEdit.xhtml`), and nested forms are not allowed.

I do think we should tackle both mentioned issues separately though, if possible. It's true that we re-use the `metadataTreeTable.xhtml` for the metadata editor as well as the "new process" form. Until now it had advantages because we basically need the same component in both forms, but the more complex the component gets, the more difficult it becomes to manage the different usages in both contexts, so maybe the limit has been reached from where it doesn't make sense anymore to maintain just one component that is reused for both pages.

Perhaps it would be best to just go for a minimal approach to just fix #6231 though, that should only contain the smallest change possible required for keeping metadata in collapsed metadata groups in the editor, while being ignored in the process creation form altogether. Unfortunately it's not possible to wrap the `<p:ajax event="collapse"/>` tag inside a conditionally rendered output panel, for example, but perhaps there is another way to only register the event handler for the `collapse` event conditionally in the metadata editor.

PR #6237, opened 2024-09-27
Fix the disappearance of collapsed metadata 

BartChris:
Enhanced version of https://github.com/kitodo/kitodo-production/pull/6232
Fixes https://github.com/kitodo/kitodo-production/issues/6231

@solth I followed your suggestion to only "activate" the ajax tag in the metadata editor by using the onstart functionality for checking a passed in context value, which makes it possible to decide what the rendering context is.

Edit: I can probably inspect the URI directly to not having to introduce an addional renderContext.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
@@ -42,6 +42,13 @@
                  styleClass="no-header"
                  widgetVar="metadataTable"
                  id="metadataTable">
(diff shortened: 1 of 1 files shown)


solth:
@BartChris Please rebase against current master to avoid failing builds caused by broken Selenium tests!

solth (approved):
@BartChris it seems you found the best variation of the solution, now. It does indeed fix the nasty bug in the editor without having any side effects on the import mask. Thanks a lot for this - evolved! - fix!

solth:
@BartChris would you mind opening a corresponding fix pull request for the 3.7 branch?

Issue #5220, opened 2022-07-06
Setting the ORDERLABEL to excluded breaks the pagination


BartChris:
**New issue description:**

**Describe the bug**

When setting the `ORDERLABEL`-key to excluded in the ruleset

```xml
 <setting key="ORDERLABEL" excluded="true" editable="false"/>
```

a lot of pagination bugs occur in the metadata editor. See below. The only way to fix that is to not exclude the `ORDERLABEL`.
```xml
 <setting key="ORDERLABEL" editable="false"/>
```

**Expected behavior**
It should be possible to exclude the `ORDERLABEL` and still have a working pagination. A possible cause of this error is described below:

https://github.com/kitodo/kitodo-production/issues/5220#issuecomment-1208031116

----
**Old description:**

When a object is already paginated in the editor

![grafik](https://user-images.githubusercontent.com/7516499/177601877-f5d8d1b8-5238-452f-9f53-aae3f371e4f7.png)

and you want to add a new logical element to one or multiple pages

![grafik](https://user-images.githubusercontent.com/7516499/177602035-38fbca66-4431-434d-b602-2d2f576ff5cb.png)

The pagination of one or multiple pages is reset to "uncounted" and has to be redone. Example with one page.

![grafik](https://user-images.githubusercontent.com/7516499/177602152-15cc59ba-68b1-4042-94be-502c18134096.png)

If you associate a logical element to multiple pages the  pagination is distorted as well:

![grafik](https://user-images.githubusercontent.com/7516499/177602931-7640b00c-c107-45b5-b89e-5c7fe0fec745.png)

illipsum:
This really is a killer bug, tbh. Makes the metadata editor basically unusable at this moment. 
![grafik](https://user-images.githubusercontent.com/51130057/179489788-fa2f75a2-df63-4122-8b78-eafdb0486fe3.png)
After adding several structure elements pagination was a total mess with randomly added "uncounted" pages.

IkramMaalej:
Unfortunately I could not reproduce this problem with the current master version.
Can you please give us more details on the steps you took to cause this behavior?

![pagination](https://user-images.githubusercontent.com/34158824/179731597-cae72257-fd89-44f5-87f4-f3f7fdef047d.gif)

illipsum:
Hi Ikram, thanks for this. Here's how I did it:

[...]

PR #6238, opened 2024-09-27
Do not reset the ORDERLABEL if division is a page

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/5220
alternative: https://github.com/kitodo/kitodo-production/pull/6244
We should not reset the ORDERLABEL of a page during metadata preservation since this breaks the pagination when the ORDERLABEL is set to excluded.
While this fixes the pagination issue there are other potential problems. I did an alternative pull request which tries to adress them all:
https://github.com/kitodo/kitodo-production/pull/6244

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -44,6 +44,7 @@ import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.SimpleMetadataViewInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.StructuralElementViewInterface;
 import org.kitodo.api.dataformat.Division;
+import org.kitodo.api.dataformat.PhysicalDivision;
 import org.kitodo.exceptions.InvalidMetadataValueException;
 import org.kitodo.exceptions.NoSuchMetadataFieldException;
 import org.kitodo.production.services.dataeditor.DataEditorService;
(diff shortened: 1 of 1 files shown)


BartChris:
Although i did not yet notice any problems when not resetting the ORDERLABEL for pages, we can also extend  the original logic. The cause of the linked issue is, that the necessary logic for setting the LABEL and ORDERLABEL is not called for hidden metatdata.

The necessary setters are called as a side effect of the (IMHO very complicated) BiConsumer logic, which does some transformations to the values from the tree, but mainly defines the setter for different division values, which is then applied.

 https://github.com/kitodo/kitodo-production/blob/eded39f924588f71d3adb5c2bc36abe0bc815623/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessTextMetadata.java#L108-L118

https://github.com/kitodo/kitodo-production/blob/eded39f924588f71d3adb5c2bc36abe0bc815623/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessSimpleMetadata.java#L59-L73

For hiddenMetadata we can probably call the setters directly

```
private void processHiddenMetadata() {
        for (Metadata metadatum : hiddenMetadata) {
            if (metadatum instanceof MetadataEntry) {
                MetadataEntry metadataEntry = (MetadataEntry) metadatum;
                String key = metadataEntry.getKey().toUpperCase();
                String value = metadataEntry.getValue();
                switch (key) {
                    case "LABEL":
                        division.setLabel(value);
                        break;
                    case "ORDERLABEL":
                        division.setOrderlabel(value);
                        break;
                    case "CONTENTIDS":
                        division.getContentIds().add(URI.create(value));
                        break;
                }
            }
        }
        metadata.addAll(hiddenMetadata);
    }
```

solth:
@BartChris Please rebase against current master to avoid failing builds caused by broken Selenium tests!

PR #6240, opened 2024-09-30
Disable faulty Selenium tests until they can be fixed

solth:
This PR temporarily disables three Selenium tests (partly mentioned in #6138) that have been very unreliable during CI builds. Since I haven't been able to find the cause for the problem, it seems to me deactiving them until someone fixes them is the best way to go forward since currently the problems they cause far outweigh their use.


stweil (approved):
For the moment this is a reasonable workaround, thanks.

PR #6244, opened 2024-10-01
Refactor preservation logic and include hidden metadata

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/5220
alternative to https://github.com/kitodo/kitodo-production/pull/6238

This pull request addresses an issue where the special metadata fields "LABEL", "ORDERLABEL", and "CONTENTIDS" are lost when they are set as excluded metadata. When opening a process in the metadata editor and saving it, this loss can occur. A special problem is, that an excluded "ORDERLABEL" will result in broken pagination (https://github.com/kitodo/kitodo-production/issues/5220).

Pull request [#6238](https://github.com/kitodo/kitodo-production/pull/6238) specifically resolves the issue of preserving pagination even when the ORDERLABEL is excluded by not resetting the ORDERLABEL in case of a page. However, in order to fully address the broader problem of losing the special metada fields when they are excluded, we must also ensure that the division data is updated based on hidden/excluded metadata.

This pull request not only resolves the issue of losing excluded metadata but also refactors the existing logic. The goal of the refactoring is to make the logic more explicit, improving both readability and maintainability. The current BiConsumer logic is difficult to follow, so I’ve aimed to preserve the functionality while making the code more understandable.

If this approach is rejected or introduces new problems, we should seek an alternative solution to properly handle hidden metadata.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -686,12 +686,12 @@ public class ProcessFieldedMetadata extends ProcessDetail implements Serializabl
                     metadata.addAll(row.getMetadataWithFilledValues());
                 }
             }
-            if (Objects.nonNull(hiddenMetadata)) {
-                if (!hiddenMetadata.isEmpty()) {
-                    for (Metadata hiddenmetadatum : hiddenMetadata) {
-                        if (hiddenmetadatum instanceof MetadataEntry
-                                && specialFields.contains(hiddenmetadatum.getKey())) {
-                            updateDivision(hiddenmetadatum.getKey(), ((MetadataEntry) hiddenmetadatum).getValue());
+            if (!Objects.isNull(hiddenMetadata) && !hiddenMetadata.isEmpty()) {
+                for (Metadata hidden : hiddenMetadata) {
+                    if (hidden instanceof MetadataEntry) {
+                        MetadataEntry entry = (MetadataEntry) hidden;
+                        if (specialFields.contains(entry.getKey())) {
+                            updateDivision(entry.getKey(), entry.getValue());
                         }
                     }
                 }
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -686,7 +686,7 @@ public class ProcessFieldedMetadata extends ProcessDetail implements Serializabl
                     metadata.addAll(row.getMetadataWithFilledValues());
                 }
             }
-            if (!Objects.isNull(hiddenMetadata) && !hiddenMetadata.isEmpty()) {
+            if (Objects.nonNull(hiddenMetadata) && !hiddenMetadata.isEmpty()) {
                 for (Metadata hidden : hiddenMetadata) {
                     if (hidden instanceof MetadataEntry) {
                         MetadataEntry entry = (MetadataEntry) hidden;
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -24,11 +24,10 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.function.BiConsumer;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections4.list.UnmodifiableList;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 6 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
```diff
@@ -676,16 +671,29 @@ public void preserve() throws InvalidMetadataValueException, NoSuchMetadataField
                 division.setLabel(null);
             }
             metadata.clear();
+            Set<String> specialFields = new HashSet<>(Set.of(METADATA_KEY_LABEL, METADATA_KEY_ORDERLABEL,
```
matthias-ronge, line 674: Wouldn’t it be enough to assign directly? Further more, I would suggest to make this a constant

     private static final Set<String> SPECIAL_FIELDS = Set.of( ...

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
```diff
@@ -705,6 +713,42 @@ public void preserve() throws InvalidMetadataValueException, NoSuchMetadataField
...
+        }
+        updateDivision(key, simpleValue);
+    }
+
+    private String extractSimpleValue(ProcessDetail value) {
+        if (value instanceof ProcessTextMetadata) {
+            return ((ProcessTextMetadata) value).getValue();
+        } else if (value instanceof ProcessSelectMetadata) {
+            return String.join(" ", ((ProcessSelectMetadata) value).getSelectedItems());
+        }
```
matthias-ronge, line 732: Missing to handle `ProcessBooleanMetadata`

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
```diff
@@ -705,6 +713,42 @@ public void preserve() throws InvalidMetadataValueException, NoSuchMetadataField
...
+        if (simpleValue == null) {
+            return;
+        }
+        updateDivision(key, simpleValue);
+    }
+
+    private String extractSimpleValue(ProcessDetail value) {
+        if (value instanceof ProcessTextMetadata) {
+            return ((ProcessTextMetadata) value).getValue();
+        } else if (value instanceof ProcessSelectMetadata) {
```
matthias-ronge, line 730: In general, this refactoring is less object-oriented. It is not good practise to have lists of `if (x instanceof A) { … } else if (x instanceof B) { … } else if (x instanceof C) { … }` ...

matthias-ronge (changes requested):
Please add missing case for `ProcessBooleanMetadata`.

In general, I am only half-convineced about this refactoring, because is less object-oriented then. It is not good practise to have lists of `if (x instanceof A) { … } else if (x instanceof B) { … } else if (x instanceof C) { … }`, if that can be avoided. But I understand as well that the other variant is clumsy to read and understand.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
BartChris: I tried to adress that remark by putting the necessary logic in the specific class and calling it using polymorphism.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
BartChris: I added boolean metadata.

matthias-ronge (approved):
From looking at it, it should be OK. (I haven't tested it.)

solth:
@BartChris please rebase agains current `main` branch!

BartChris:
@solth i have rebased.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
```diff
@@ -677,16 +674,27 @@ public void preserve() throws InvalidMetadataValueException, NoSuchMetadataField
...
+                if (row instanceof ProcessSimpleMetadata && specialFields.contains(id)
+                        && ((ProcessSimpleMetadata) row).getSettings()
+                        .getDomain().orElse(Domain.DESCRIPTION).equals(Domain.METS_DIV)) {
+                    updateDivisionFromProcessDetail(id, (ProcessSimpleMetadata) row);
                 } else {
                     metadata.addAll(row.getMetadataWithFilledValues());
                 }
             }
-            if (Objects.nonNull(hiddenMetadata)) {
+            if (!Objects.isNull(hiddenMetadata) && !hiddenMetadata.isEmpty()) {
```
solth, line 689: Just one very small nitpick: is there a reason why you replaced `Objects.nonNull` with `!Objects.isNull`? The later reads like a double negation to me and is just less readable than the dedicated `nonNull` check, IMHO.

BartChris: Good catch; i do not know. I replaced it with nonNull.

solth:
@BartChris since this is pull request is a bugfix, would you mind providing a corresponding bugfix for the 3.8 branch of Kitodo.Production?

PR #6255, opened 2024-10-08
Allow to link page to next element independently of hierarchy level

### LINKED ISSUE #5457: Linking pages to the following structural element not possible if the elements are not on the same hierarchical level
**Describe the bug**
Not entirely sure whether this is a matter of configuration or an issue of Kitdo itself. Here's the scenario: If you create more than one level of structural data, linking a page at the end of a structure element to the following one does not work if the following element is not on the same hierarchical level.

**To Reproduce**
To illustrate:
grafik
On the last page of the last section (Abschnitt) both the chapter and section end and the following chapter (Kapitel) begins. The page looks like this: 
grafik
While it worked for all the sections on the same structural level, creating a link to add this page to the following chapter is not a given option in the context menu:
grafik

**Expected behavior**
It should be possible to create a link that adds the current page to the following structural element

**Release**
3.4.4

 - OS: Windows 10
 - Browser Firefox
 - Version 102.5.0

thomaslow:
Fixes #5457 

This pull request allows to add a new link (context menu entry "Assign to next element") for the currently selected page to the following folder independently of its hierarchy level. The selection logic works as follows:

1. find a parent folder whose next sibling is also a folder (continue with parent of parent, if not found)
2. select the deepest child folder of the next sibling only if they are the respective first children

This pull request also checks if a link has already been added before such that users cannot accidentally add the same link twice (the menu entry "Assign to next element" disappears).

Demo

https://github.com/user-attachments/assets/d45130f4-405e-4270-9642-02a8e0653037


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -1765,43 +1775,114 @@ public boolean isAssignedSeveralTimes() {
...
+                }
+                return nextLogical;
+            }
+
+            // node is last amongst siblings
+            // iterate to parent node
+            current = current.getParent();
+        }
+        return null;
+    }
```
BartChris, line 1848: Method is very long and could probably be refactored in smaller methods, but as the logic is quite complex it is also good to have it in one place.

PR #6259, opened 2024-10-09
Fix German translations

### LINKED ISSUE #6258: Wrong spelling in German message after applying pagination
**Describe the bug**

The spelling of the message after applying the pagination is wrong: 
- `Paginierung wurde erfolgreich übernehmen.`

**To Reproduce**
1. Go to 'Paginierung' and do some stuff
2. Click on 'Übernehmen'
3. See message: `Paginierung wurde erfolgreich übernehmen.`

**Expected behavior**
The message should be spelled: 
- `Paginierung wurde erfolgreich übernommen.`

**Screenshots**
FehlermeldungPaginierung01

**Release**
KITODO.PRODUCTION Version 3.7.0

 - Windows
 - Chrome

**Additional context**
Probably, the following line should be changed: 
- https://github.com/kitodo/kitodo-production/blob/5efafdb7cb78848ba86ca9fc877c5f3f60ec48e2/Kitodo/src/main/resources/messages/messages_de.properties#L891

stweil:
Fix #6258.

Reported-by: André Hohmann


Kitodo/src/main/resources/messages/messages_de.properties
```diff
@@ -888,7 +888,7 @@ paginationRectoUseHint=Nur auf der Vorderseite (nach Erh\u00F6hung bei Halbschri
 paginationVersoUse=[\u00BD]
 paginationVersoUseHint=Nur auf der R\u00FCckseite (nach Gleichbleiben bei Halbschritten)
 paginationNoPagesSelected=Keine Seiten f\u00FCr Paginierung ausgew\u00E4hlt
-paginationSaved=Paginierung wurde erfolgreich \u00FCbernehmen.
+paginationSaved=Paginierung wurde erfolgreich \u00FCbernommen.
```
stweil, line 891: Could we also write "übernommen"? Other texts simply use the German umlauts.

solth: We had some problems with using actual German umlauts in the message files in the past, so I would prefer to stick to the escaped codepoints for now (at least until we find the reason and a corresponding solution for the problems that arose with special characters in the message files albeit the correct encoding)

stweil: Okay, there is no need to do such a change soon. Maybe `project.build.sourceEncoding` and `project.reporting.outputEncoding` where missing in the past which might explain the problems.

Maybe you noticed that I fixed a wrong encoding which resulted in "Záhler" instead of "Zähler".

henning-gerhardt: This has nothing to do with project properties from the pom.xml file. Wrong character encodings are created by different editors on different operation systems with different user settings for character encoding. Plain text files like here are the most painfull chase to detect and fix character encodings. As long as all editor persons use the same character encoding while they editing the files and ensuring that storing did not break anything any but the same encoding can be used. This was not the case in the past, not now and will be never in the future as long as plain text files are used.

stweil: That's not correct. UTF-8 encoding is supported by all current editors and used in a huge number of other projects. And it is also already used in Kitodo – obviously without causing problems.

henning-gerhardt: I know that UTF-8 is supported in a wide way but as you can see in this project it is not used in a correct way in the past and now and until not every contributor to this project is doing it right, this issue will still exists and shown again in the future.

solth (approved):

PR #6152, opened 2024-07-28
Fix some smaller issues in dependencies

stweil:
The updates were selected from this list:

    mvn versions:display-dependency-updates | grep -- "->" | sort | uniq

PR #6266, opened 2024-10-11
Update some dependencies to latest bugfix versions

stweil:
The updates were selected from this list:

    mvn versions:display-dependency-updates | grep -- "->" | sort | uniq

Updates were the major or minor version changed were not included.


stweil:
This pull request includes all commits from #6152 which only change versions, so hopefully it is easier to review.

solth (approved):

PR #6268, opened 2024-10-14
Fix property not found exception in metadata editor with physical structure tree

thomaslow:
When opening the metadata editor in "extended structure"-mode (with both a logical and physical tree), the following exception occurs:

Screenshot from 2024-10-14 17-33-03

This pull request fixes the exception (without any detailed investigation what exactly is going on). With this fix, the metadata editor loads again for me.


solth (approved):
@thomaslow thanks for the fix. Would you mind opening a corresponding pull request for the 3.7 branch?

PR #6269, opened 2024-10-21
Open media in separate browser window

### LINKED ISSUE #5887: Add option to open select pages in the metadata editor gallery in a separate browser window
It can be difficult to apply pagination to a set of pages in the metadata editor if the information for the pagination has to be obtained from a table of contents in the same process. Since selecting the scans for pagination moves the page with the table of contents out of view, it becomes cumbersome and difficult for the user to move back and forth between those pages.

A new option should be added to the context menu of individual pages in the metadata editor (gallery and structure tree) to open a selected page in a separate window. Some users have reported that they actually print out a table of contents and use it next to the computer to help with pagination, so this approach with a separate browser window should fulfil the requirement in a similar way (albeit without the need for a printer and paper).

Instead of opening a selected page in a separate browser window, it was initially proposed to add the option to "pin" the page to keep it in view even when the user subsequently selects different pages for the purpose of pagination. This would also resolve the issue described above, but would presumably be much more expensive in the implementation (since the whole layout and structure of the metadata would have to be refactored to take into account the pinned page) without adding much more value as a solution for the described problem.

**Additional context**
This topic has been discussed at the Kitodo community meeting 2023 and has been documented at #5881 and https://github.com/kitodo/kitodo-production/wiki/Barcamp-2023:-Usability-Metadateneditor

thomaslow:
Fixes #5887.

This pull request adds another menu item to the context menu of the metadata editor. It allows to open a page in a new browser window (depending on browser settings, usually a new tab). The media is shown similar to the media detail view in the meta data editor, meaning, images can be zoomed using OpenLayers, videos and audio is shown with a standard html5 video tag. There is no communication between the new window and the metadata editor. It is a simple static view of the media.

Demo video:

https://github.com/user-attachments/assets/51e0c936-b2e5-4818-95ed-de53d6d4ad51


BartChris:
It would be very nice if we can also achieve the same behaviour, with a Ctrl/Str+Click on the node or the thumbnail.

BartChris:
> > It would be very nice if we can also achieve the same behaviour, with a Ctrl/Str+Click on the node or the thumbnail.
> 
> I think `CTRL` + `Click` is currently reserved for multi-selecting pages so that would probably produce a conflict in the shortcuts.

Right... Maybe Alt+Click then?

thomaslow:
@BartChris I tried to implement your idea to open a new window with an `alt+click`. Unfortunately, there are multiple problems that are difficult to solve. 

When using the `mouse down` event of a thumbnail, jQuery gets confused and thinks that the users want to drag the thumbnail. So, when the user returns to the metadata editor, suddenly the thumbnail is being dragged around:

https://github.com/user-attachments/assets/875e0f3d-14c4-4143-b70d-8729cef2d84b

Unfortunately, the `mouse up` event can also not be used, because Primefaces prevents it from being triggered consistently. There is a Primefaces `click` handler that stops the `mouse up` event if the mouse is clicked quickly. The window is only opened when clickling "slowly" (like waiting 1 second between mouse down and mouse up).

Then there is the issue that the correct URL to open the window is current only updated when a thumbnail is selected.

All in all, I can only offer a `alt+(double)-click`. Meaning, if the thumbnail is not yet selected, there needs to be two clicks (one for selecting the thumbnail, and one for opening the window). If the thumbnail is already selected, a single `alt+click` is sufficient. Here is a demo:

https://github.com/user-attachments/assets/04c0829f-a320-4bb9-9e7e-f0cc7ab58c78

Personally, I wouldn't recommend to include this inconsistent mouse click behavior, which is probably confusing to the user. Unfortunately, I do not see another solution at the moment.

BartChris:
@thomaslow Thanks a lot for your investigation. I agee, that given the circumstances it would be of little value to add that. Maybe we can revisit this at a later point in time. The right-click menu is already a big improvement.

BartChris (approved):
Everything works as expected. Code looks good!

I would only suggest to maybe move the menu entry "View page in new browser windows" to the second position in the menu, do not have it directly next to "Remove element".

thomaslow:
@BartChris Thank you for your feedback. I agree with your idea to move the context menu entry to the second position. I updated the pull request accordingly.

PR #6271, opened 2024-10-23
Change scope of class `MediaProvider` to `SessionScope`

solth:
This PR changes the scope of the `MediaProvider` class to `SessionScope` to fix a potential security risk.

@markusweigelt even though you are not actively involved in the Kitodo development at the moment, would you mind checking whether you see any problems with changing the scope of the `MediaProvider` as proposed, since you are the author of this class?

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/MediaProvider.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/MediaProvider.java
@@ -11,11 +11,12 @@
 
 package org.kitodo.production.forms.dataeditor;
 
+import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 
-import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.SessionScoped;
 import javax.faces.context.FacesContext;
 import javax.faces.event.PhaseId;
 import javax.inject.Named;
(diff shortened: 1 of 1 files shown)


markusweigelt:
I can't see a problem at the moment. However, it would have to be tested with the supported MimeTypes in [`kitodo_fileFormats.xml`](kitodo_fileFormats.xml`https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/resources/kitodo_fileFormats.xml)  If it works, I see no reason not to change the scope.

solth:
> However, it would have to be tested with the supported MimeTypes in `kitodo_fileFormats.xml`

@markusweigelt I just tested playing a video and adding media partials with this branch, and that worked. Or did you mean something else would have to be tested?

solth:
> I think if the main types are checked, it is sufficient.

So with "testing" you just mean whether all media variants are still displayed correctly in the editor or should something else, some other functionality also be checked?

markusweigelt:
No, a complete test is not necessary; only a visibility check for image, audio, and video types should be enough

solth:
@markusweigelt I tested all multi media files listed in `kitodo_fileFormats.xml` and couldn't find a difference in the behaviour when using the `SessionScoped` `MediaProvider`. So if you do not see any other problems I would go ahead and merge this pull request.

markusweigelt (approved):

henning-gerhardt:
As this is a security related issue: should this not be back ported to the `3.7.x` release branch as there is this issue too?

solth:
> As this is a security related issue: should this not be back ported to the `3.7.x` release branch as there is this issue too?

Yes, will do.

PR #6274, opened 2024-10-25
Move default date format to constant

### LINKED ISSUE #6223: Put date format string in central place
The date fromat string `"yyyy-MM-dd HH:mm:ss"` is used in several places. It should be put in one global constant and be used from there. Maybe even, put in `kitodo_config.properties`?

solth:
Fixes #6223

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/Helper.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/Helper.java
@@ -423,7 +423,7 @@ public class Helper {
             Instant instant = localDate.toInstant(ZoneOffset.UTC);
             return Date.from(instant);
         } catch (DateTimeParseException e) {
-            logger.info("invalid date format ('{}') for date string: '{}'", DEFAULT_DATE_FORMAT, date);
+            logger.info("invalid date format '{}' for date string: '{}'", DEFAULT_DATE_FORMAT, date);
             return null;
         }
     }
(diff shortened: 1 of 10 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/Helper.java
```diff
@@ -416,12 +418,12 @@ public static Date parseDateFromFormattedString(String date) {
...
         }
         try {
-            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern(DEFAULT_DATE_FORMAT);
             LocalDateTime localDate = LocalDateTime.parse(date, formatter);
             Instant instant = localDate.toInstant(ZoneOffset.UTC);
             return Date.from(instant);
         } catch (DateTimeParseException e) {
-            logger.info("invalid date format (yyyy-MM-dd HH:mm:ss) for date string: '" + date + "'");
+            logger.info("invalid date format ('{}') for date string: '{}'", DEFAULT_DATE_FORMAT, date);
```
stweil, line 426: Here you change the output. I think that either () or '' is sufficient for the date format.

stweil (approved):
I only have a small nit-picking comment. Otherwise the changes look fine.

PR #6276, opened 2024-10-28
Update all dependencies to latest bugfix versions

stweil:
All dependencies were updated to the latest bugfix versions with
```
mvn versions:use-latest-releases -DallowMajorUpdates=false -DallowMinorUpdates=false.
```

The properties were updated with
```
mvn versions:update-properties --no-snapshot-updates -DallowMajorUpdates=false -DallowMinorUpdates=false
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,7 @@
             <dependency>
                 <groupId>com.xebialabs.restito</groupId>
                 <artifactId>restito</artifactId>
-                <version>0.9.3</version>
+                <version>0.9.5</version>
                 <exclusions>
                     <exclusion>
                         <groupId>junit</groupId>
@@ -205,7 +205,7 @@
             <dependency>
                 <groupId>javax.el</groupId>
                 <artifactId>javax.el-api</artifactId>
-                <version>3.0.0</version>
+                <version>3.0.1-b06</version>
                 <scope>provided</scope>
             </dependency>
             <dependency>
@@ -265,7 +265,7 @@ from system library in Java 11+ -->
             <dependency>
                 <groupId>org.apache.activemq</groupId>
                 <artifactId>activemq-client</artifactId>
-                <version>5.18.5</version>
+                <version>5.18.6</version>
                 <exclusions>
                     <exclusion>
                         <groupId>org.slf4j</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -205,7 +205,7 @@
             <dependency>
                 <groupId>javax.el</groupId>
                 <artifactId>javax.el-api</artifactId>
-                <version>3.0.0</version>
+                <version>3.0.1-b06</version>
```
henning-gerhardt, line 208: `3.0.1-b06` looks for me like Beta 6 version for release 3.0.1. Should we really use such version even it is the latest version - newer versions are available under `jakarta.el/jakarta.el-api` (see https://mvnrepository.com/artifact/jakarta.el/jakarta.el-api).

stweil:
I wonder why it did not find #6275. Nevertheless, 9 updates are better than none.

pom.xml
stweil: 3.0.1-b06 is rather old, but 5 years newer than 3.0.0. It looks like there never was a 3.0.1.

Newer versions would be nice, but I don't see them in the current repository. Where can 3.0.2 or 3.0.3 be downloaded?

henning-gerhardt: Group `javax.el`  got relocated to ` jakarta.el` including the artifact from ` javax.el-api` to ` jakarta.el-api`. So newer versions are found under ` jakarta.el/jakarta.el-api` like my link shows. Mvnrepository.com is a good source for finding like this (and many others).

stweil:
> I wonder why it did not find #6275. Nevertheless, 9 updates are better than none.

`mvn versions:use-latest-releases` does not find newer releases if the version is defined in a property.
But I could update the properties with another command.

pom.xml
stweil: I suggest to handle the migration from javax.el to jakarta.el separately and either use version 3.0.1-b06 here (my preferred solution) or keep it at 3.0.0.

solth: @stweil I would prefer to leave this dependency at 3.0.0 in this case.

stweil, line 208:
```suggestion
                <version>3.0.0</version>
```

stweil: Done. If you want, you can suggest (and commit) the desired change directly in GitHub. That might be easier.

henning-gerhardt (approved):
Changes looks good and so far as I can test it nothing is broken but I can not test everything right now and it would take a lot of time to do this.

PR #6218, opened 2024-09-04
[hibernate-search] Introduce Hibernate Search framework and implement indexing page

matthias-ronge:
Issue #5760 2a) and 2b)

Follow-up pull request to #6209 ([immediate diff](/matthias-ronge/kitodo-production/compare/5760_1...5760_2a+b))

Recording

The three numbers before the slash in “Indexed entries” represent the number of objects that Hibernate has already loaded from the database, the number of objects that have been prepared as indexable documents (JSONs), and finally the number of indexed documents.

**Basic experience:** Hibernate Search and lazy loading don't mix. It looks like we have to accept that. As a result, I have deactivated lazy loading wherever the number of members of a set is typically small (< 25). This affects most sets, e.g. projects of a template, tasks, users or properties of a template or a process, etc. If the set can typically be large (> 1000), the elements of the set are not indexed. Example: Processes of a batch. Consideration: If the number of subelements to be indexed in an object is very large, the findability of the object approaches infinity (it becomes increasingly likely that it will be found with _any_ search query). Such indexing also makes the index enormously large. Therefore, it can be considered justifiable not to index these fields.

Issue #4378, opened 2021-05-03
Slowly displaying list of users


henning-gerhardt:
If you have a big list of users, f.e. 300 active and inactive users, displaying this list is significant slower in Kitodo.Production 3.3 in compare to 3.2.1. Maybe adding the possibility to remove assigned tasks of a user did slow down this displaying. As this slow speed effects even other tabs (f.e. roles, ldap group, ldap server) in the "user page" site you need a lot of time if you want to change / adjust f.e many roles.

solth:
I am not sure how the number of users in the system can have any effect on the loading times of that list - unless you have configured your user settings to show all 300 active users on one page.

henning-gerhardt:
Table listing is configured to 100 shown rows as this was the maximum in 2.x and a good value for any large admin task. Why should I reduce this value? I did not understand why this "removing task from user" are displayed in any case as this is only useful in certain scenarios like deleting user or if an user in holidays and is blocking some work. For the last scenario you can - at least in 2.x - use the administration task to set down a task from "in work" to "open" on the process list.

solth:
If you do not want this action to be shown on default because you do not need it, you can always configure the corresponding roles to exclude the authority `Aufgaben authority`. That should skip the rendering of the action icon and the related checks.

henning-gerhardt:
This is good to know but the lack of documentation where which authority is used and / or how the interaction is where is executed make it hard to decide which authority is really useful and should be used in which role. 

But the problem remains: displaying this icon slow down the interaction of this whole page on all tabs of this page.

henning-gerhardt:
Even after removing this authority from my list of authorities and log out and log in back, the displaying of users is slow :-(

solth:
Can you be more specific what "slow" means? Seconds? Minutes? And how much of a difference is there between 3.2.1 and 3.3 in your opinion? (I guess you can't provide numbers to support your claim, right?)

henning-gerhardt:
Loading a user list page in 3.3 with 100 entries takes
- with the assigned right `Aufgaben zurücksetzen` between 25 and 45 seconds as this depends on how many users has assigned tasks
- without assigned right `Aufgaben zurücksetzen` displaying the same user page takes between 12 and 17 seconds

I did not have access to a 3.2.x system anymore but there the loading time of this page was between 5 and 10 seconds at least. In comparing with 2.x loading the user list with 100 entries per page takes less then 2 seconds.

solth:
[...]

Issue #4266, opened 2021-03-12
Simplify ElasticSearch mapping of metadata


solth:
Currently, process metadata is saved in the following JSON structure in the ElasticSearch index:
<img width="473" alt="Bildschirmfoto 2021-03-12 um 10 49 49" src="https://user-images.githubusercontent.com/19183925/110923108-cf987780-8320-11eb-9373-ad29043ee31c.png">
Since Kitodo.Production internally always uses METS as the container format for metadata, the nodes `mdWrap`, `xmlData` and `kitodo` are not really required and unnecessarily complicate dealing with metadata retrieved from the search index. 

I would therefore propose to remove the nodes mentioned above and directly save all metadata for a given structure element as a JSON array associated directly with the structure elements UUID.

matthias-ronge:
We should think twice about this when redesigning the search engine integration. In any case, the user should be able to search for metadata, also specifically, i.e. that a certain metadata has a certain value. As I would understand, a separate search field would have to be written in the index for each searchable metadata key (except ElasticSearch offers an additional functionality here that I don't know about).

solth:
I am sorry, perhaps there was a misunderstanding. My idea was to remove just the `mdWrap`, `xmlData` and `kitodo` nodes, not the `metadata` node or its content. So all key values pairs concerning metadata entries and groups would be retained.
For example:
```
meta:
    0:
        mdWrap:
            xmlData:
                kitodo:
                    metadata:
        ID:
    1:
        mdWrap:
            xmlData:
                kitodo:
                    metadata:
        ID:
...
```
would become:
```
meta:
    0:
        metadata:
        ID:
    1:
        metadata:
        ID:
...
```

matthias-ronge:
My question is: Is the way in which metadata is indexed suitable that you can search using individual metadata fields? So, I want to find "cook", but only in the title, not in the author. Is this already possible, or does your suggested change make it possible?

I am just thinking that we need a change here, but I it should be a change to make the above possible, if if isn’t yet. This can go along with mapping simplification.

solth:
I totally agree that searching individual metadata fields in the way you described is important and should be possible - actually that and loading metadata from the index into `ProcessDTO` objects to make them usable there was the incentive for this issue. But in my opinion that should be discussed in another topic. If anything, the proposed changes should make it _easier_ to work with the metadata stored in the index.

[...]

PR #6283, opened 2024-10-29
[hibernate-search] Implement indexing and search

matthias-ronge:
Issue #5760 2c), 2d), 2e) and 2f)

Follow-up pull request to #6218 ([immediate diff](/matthias-ronge/kitodo-production/compare/5760_2a+b...5760_2c))

The filters work. Metadata is indexed and searchable. The metadata search syntax works as documented [in the wiki](/kitodo/kitodo-production/wiki/Suche-und-Filter).


BartChris:
Great work! Can you elaborate how the metadata is indexed now in general? (For reference: https://github.com/kitodo/kitodo-production/issues/4266)

Kitodo/src/main/java/org/kitodo/production/filters/FilterMenu.java
```diff
@@ -296,6 +297,12 @@ public List<ParsedFilter> getParsedFilters() {
      * @return value of filterInEditMode
      */
     public String getFilterInEditMode() {
+        Map<String, String> parameters = FacesContext.getCurrentInstance().getExternalContext()
+                .getRequestParameterMap();
+        if (parameters.containsKey("input") && StringUtils.isBlank(filterInEditMode)) {
+            filterInEditMode = parameters.get("input");
```
henning-gerhardt, line 303: You are taking over any argument from the outside without validating that the given value is correct or at least is in range of the expectations?

matthias-ronge: Yes, the user can enter any string into the search slot. The input is not “checked”. But of course, if the user has entered nonsense, the search will not find any results.

henning-gerhardt: Thank you for this clarification but I expected something different as I read the method, method content  and the used written class variable `filterInEditMode`. Handling of this class variable `filterInEditMode` is strange in this class but this is a different story.

matthias-ronge:
> Can you elaborate (short high level overview) how the metadata is indexed now in general?

The metadata is stored in a text field that is indexed. The search is later carried out on the text field.

"Pseudo words" are used to search on certain metadata fields, i.e. if the word _Berlin_ is in the _TitleDocMain_ field in the metadata, a pseudo word "titledocmainqberlin" is also indexed. If a user later enters "TitleDocMain:Berlin" in the search field, this is converted into the pseudo word and searched for.

Pseudo words are generated for metadata keys, translated metadata keys according to the rule set, and domains according to the ruleset. The user can therefore also search for "Haupttitel:Berlin" (if the translation in the ruleset is this).

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Keyworder.java
```diff
@@ -0,0 +1,512 @@
...
+                    for (int i = 0; i < word.length(); i++) {
+                        for (int j = i + 1; j <= word.length(); j++) {
+                            tokens.add(word.substring(i, j));
+                        }
+                    }
+                }
+            }
+        }
+        return tokens;
+    }
```
BartChris, line 367: Can you explain the reasoning here? Do i read this correctly and you are indexing every possible substing of all the words in a comment? Is this really necessary? Normal indexing should make it possible to search by word? Is this not enough?

matthias-ronge:
> Do i understand you correctly that you make the fields searchable by their german and english label as well? So for example "Dokumenttyp:Monograph" as a possible search with the german label?

Yes.

> What happens if somebody changes the german label of a metadata key in the ruleset?

The search will only find the process with the old label, until the process is saved again, or all processes are re-indexed. Since the index is now only used for filters, re-indexing (or a missing index) won’t “break” the application any more, as it did in the past, it should be possible to run re-indexing in an environment where people work on without breaking something (but it will still make it slow meanwhile!)

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Keyworder.java
matthias-ronge: The [wiki page](/kitodo/kitodo-production/wiki/Suche-und-Filter) requires:

> Eine Suche nach `Meldung` findet alle Vorgänge, die einen Kommentar mit einer Nachricht wie `Das ist eine Korrekturmeldung!` haben.

To achieve this, we need both-side subword search. I only follow the expected behaviour here. I would prefer a simple-word search, too, but this is what the requirement actually looks like. Maybe we can discuss this, if it is really needed, as it inflates the index. I don’t know a clever way of splitting a German compound-word to its single words, to index just these. (This is possible, but it would be a sophisticated task of computational linguistics, typically involving a dictionary file, too, that I don’t have.)

BartChris:
> > What happens if somebody changes the german label of a metadata key in the ruleset?
> 
> The search will only find the process with the old label, until the process is saved again, or all processes are re-indexed. Since the index is now only used for filters, re-indexing (or a missing index) won’t “break” the application any more, as it did in the past, it should be possible to run re-indexing in an environment where people work on without breaking something (but it will still make it slow meanwhile!)

I understand. I fear a situation, where correcting a typo in a label would require a reindex of everything, otherwise old data is only found with with the misspelled label. I do not want to increase your workload too much, but woul it be the option to allow the LABEL-filters in the UI, but only index and search under the common key in the index? By e.g. having a mapping function, which maps the labels to the actual key and searches for that.

BartChris:
> Since the index is now only used for filters, re-indexing (or a missing index) won’t “break” the application any more, as it did in the past, it should be possible to run re-indexing in an environment where people work on without breaking something (but it will still make it slow meanwhile!)

A more general remark here. I have not really tested anything so far, but if it seems to me almost everything gets indexed. My assumption would be that functionality like displaying list of processes (with all the necessary icons, which require lookups) is faster from the index since we can store data there in denormalized form. It is nice, that we can run Kitodo without a running index for certain functionality but my intuition would be to always use the index if that gives us an edge over database-based retrieval. Maybe i am overlooking something, but it might be worth considering to rely more on the index to have better performance when displaying 100 processes at once, which is very slow in Kitodo 3.7.

This is not necessarily adressed at the current PR, but more a general remark which can be considered.

matthias-ronge:
> I fear a situation, where correcting a typo in a label would require a reindex of everything, otherwise old data is only found with the misspelled label.

In my experience, rulesets are rarely changed at all during a running project, except perhaps the addition of a field; and I know librarians are very good at not making typos.

> would it be the option to allow the label-based-filters in the UI, but only index and search under the common key in the index? By e.g. having a mapping function, which maps the labels to the actual key and searches for that.

1. The metadata is also indexed under the key string defined in the rule set, so you can always use the key name for the search.
2. Yes, of course, that would be possible, but I think it goes against the purpose of using search engines. It would be possible to implement this, but it would significantly slow down search queries because all relevant rulesets would have to be determined each time and checked to see whether a field name needs to be translated. I don't think that would be useful.

Perhaps it would be interesting to discuss this case among users. Such a change can be introduced later.

matthias-ronge:
> it appears to me that almost everything gets indexed.

Yes, a lot of indexing is still going on at the moment and I don't think that's necessary. But the task here was not to make more than necessary changes, but to keep the application as most as it was before. But I think that in a separate development, it will be possible to clean this up so that in the end only processes and tasks are indexed, as only these contain metadata. The other objects are never searched, so they don’t do something sensible in the index by now.

> My assumption would be that functionality like displaying list of processes […] is faster from the index since we can store data there in denormalized form. […] Maybe i am overlooking something, but it might be worth considering to rely more on the index to have better performance when displaying 100 processes at once, which is very slow in Kitodo 3.7. (which probably uses a mixture of index and database-based retrieval)

No, it's exactly the other way round: In Production 3.7, the display is taken from the index alone, as you describe here as a wish. And it is slow. In this version, this behavior was exactly inverted and now everything except the keyword search is taken from the database. On my developer laptop, the application has become significantly faster as a result.

Don't ask me why it behaves like this. This is another example of how performance is less about well-intentioned programming than about actual measurements. I suspect that both database queries and restoring Java in-memory objects from database rows are so much more sophisticated, having been researched for several decades longer, that it beats out indexing.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/IndexingKeyworder.java
```diff
@@ -0,0 +1,502 @@
...
+    public IndexingKeyworder(Process process) {
+        this.titleKeywords = filterMinLength(initTitleKeywords(process.getTitle()));
+        String projectTitle = Objects.nonNull(process.getProject()) ? process.getProject().getTitle() : "";
+        this.projectKeywords = filterMinLength(initSimpleKeywords(projectTitle));
+        this.batchKeywords = filterMinLength(initBatchKeywords(process.getBatches()));
+        var taskKeywords = initTaskKeywords(process.getTasksUnmodified());
+        this.taskKeywords = filterMinLength(taskKeywords.getLeft());
+        this.taskPseudoKeywords = filterMinLength(taskKeywords.getRight());
+        String place = null;
+        if (logger.isDebugEnabled()) {
```
henning-gerhardt, line 85: Why is this following only executed if log level is on debug or lower?

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
```diff
@@ -505,4 +518,62 @@ private static List<String> getDatesFromConfig() {
...
+                Object parameter = parameters.get(matchResult.group(1));
+                if (Objects.isNull(parameter)) {
+                    return matchResult.group();
+                }
+                if (parameter instanceof String) {
+                    return '\'' + ((String) parameter) + '\'';
+                }
+                if (parameter instanceof Collection) {
+                    int size = ((Collection<?>) parameter).size();
+                    if (size > 12) {
```
henning-gerhardt, line 562: Why is a value greater than `12` bad? Should not all parameters be used?

Kitodo-DataManagement/src/test/resources/log4j2.xml
```diff
@@ -18,6 +18,12 @@
         </Console>
     </Appenders>
     <Loggers>
+        <Logger name="org.kitodo" level="debug" additivity="false">
+            <AppenderRef ref="STDOUT"/>
+        </Logger>
+        <Logger name="org.opensearch.node" level="info" additivity="false">
+            <AppenderRef ref="STDOUT"/>
+        </Logger>
```
henning-gerhardt, line 26: Is this really needed? Log output is already really big and CI executions got aborted if log output is to big.

Kitodo/src/test/java/org/kitodo/production/services/command/CommandServiceTest.java
```diff
@@ -158,7 +158,7 @@ public void runNotExistingScriptAsync() throws InterruptedException {
         String commandString = scriptPath + "not_existing_script" + scriptExtension;
         CommandService service = new CommandService();
         service.runCommandAsync(commandString);
-        Thread.sleep(1000); // wait for async thread to finish;
+        Thread.sleep(2000); // wait for async thread to finish;
```
henning-gerhardt, line 161: This change and similar changes increase the test execution time for everyone :-( and the test execution itself is already really long.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/IndexingKeyworder.java
```diff
@@ -0,0 +1,502 @@
...
+     * separated into groups of consecutive characters and numbers, and these
+     * are generated in their entirety and starting from the front or from the
+     * back. The latter is more common, as the last four digits of a PPN are
+     * used for the search and the hit rate is excellent.
+     * 
+     * @param processTitle
+     *            the title of the process
+     * @return keywords
+     */
+    private static Set<String> initTitleKeywords(String processTitle) {
```
henning-gerhardt, line 109: Is this method and similar methods are really needed? Should this not be done by ElasticSearch/Opensearch in a internal way without bothering every one who is using it? This will increase the execution time and the data size on ElasticSearch/OpenSearch server without knowing that a internal handling is better.

Maybe the submitted data must provided in an other way to get this handled by ElasticSearch/Opensearch itself instead of implementing and maintaining it on our side?

henning-gerhardt (changes requested):
It is a code only reading review. I'm not sure if the word splitting for any kind of indexing type and there content is really needed to be done on our side - this should be done on an internal way in ES/OS itself.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/IndexingKeyworder.java
matthias-ronge: The string is only created to be used in logging when DEBUG logging is enabled. So, in production, it would not need to be created since it is not used.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
matthias-ronge: I have based this on the written language: Numbers up to twelve are written out, from 13 onwards the number is written. Yes, a list of 500,000 IDs in a query makes the logging unreadable, which is why only the number of elements is output in the debug output for longer lists. I think 80% of users will be happy with this logging format, so no parameterization is required for log output.

Kitodo/src/test/java/org/kitodo/production/services/command/CommandServiceTest.java
matthias-ronge: We don't need to discuss the one second. This is just a single test being run. Yes, on a slower laptop, 1 second sometimes doesn't work and the build aborts.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/IndexingKeyworder.java
matthias-ronge: I think it's possible that you can build this into OpenSearch, but then you'll have to compile your own OpenSearch binaries and you won't be able to just install OpenSearch from the distribution. Maybe that makes sense if we have several of them, but for administrator convenience, it would be more convenient to install the web application and have it just work.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/IndexingKeyworder.java
henning-gerhardt: Thank you for the explanation. Maybe this should be added as a comment as I think that would cause questions in future.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
henning-gerhardt: Thank you for the explanation. Can you add a comment with this explanation as this can lead to questions in future.

Kitodo/src/test/java/org/kitodo/production/services/command/CommandServiceTest.java
henning-gerhardt: One second at one point, an other second an other point. Yes, this is not much at a single place but in common this could be extend the execution time. But for now I'm fine with this change.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/IndexingKeyworder.java
henning-gerhardt: As @solth already written: I thought that this could be configured on the ES / OS side or through the type of data so no change is needed explicit on ES / OS side itself instead of implementing this on our side. Maybe my imagination is wrong but this would blow up the data size of the ES / OS index a lot if you have many processes and big meta data (task data, comment data, data from the meta data files itself, ...) entries per process. But we will see how the handling and outcome is.

henning-gerhardt: I'm sorry I have not the knowledge to do this as this is not my part of my work. I should look into the changes and I should comment them if I'm unsure about the changes. I don't know which exact requirements are made as I'm not involved in this discussions. The solutions looks complex and bad to maintain and I still thinking that this is not really necessary to do this required work every time a ES / OS should be used. Maybe this is limited because of the way of hibernate-search or through other limitations. Maybe I'm overrating what should be done by ES / OS on default and this stuff must be done by everyone how is using ES / OS. Maybe using the [binding and brigdes](https://docs.jboss.org/hibernate/search/6.2/reference/en-US/html_single/index.html#binding) could be a start of writing self defined bridges in a more clean way if they are a possible solution.

BartChris:
All the filters i enter are always duplicated in the filter menu, which also duplicates the conditions in the database query.
image

The SQL query generated has two IN-conditions with the same process Ids.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -470,6 +472,7 @@ public List<Process> loadData(int offset, int limit, String sortField, SortOrder
                 .collect(Collectors.toList());
         query.restrictToProjects(projectIDs);
         query.defineSorting(SORT_FIELD_MAPPING.get(sortField), sortOrder);
+        query.performIndexSearches();
```
BartChris, line 475: The strategy here is to retrieve the process IDs for which a specific condition e.g. `stepopen:Archivierung` is met from the Search Index. Those IDs are used to extend the sql query with a IN-statement containing all the IDs retrieved from the search index. This can lead to REALLY large queries when you have 10.000 processes in your database and 9000 of them are in the state "Strukturierung". If you search for that stage the SQL IN clause will contain 9000 process IDs. With more processes in Kitodo the queries can become really large, maybe exhausting MySQL here.

henning-gerhardt: > Just an observation: The strategy here is to retrieve the process IDs for which a specific condition e.g. stepopen:Archivierung is met from the Search Index. Those IDs are used to extend the sql query with a IN-statement containing all the IDs retrieved from the search index

This was necessary in the past or with ES / OS as ES / OS can only retrieve JSON based data which are no objects and are not so good accessible. For this reason the ES / OS result was given to Hibernate to retrieve the objects from the database which could contain more information and accessing them was a lot easier. As the used ES / OS API could never result more than 10.000 hits there was a limit to the Hibernate query itself.

This is not needed anymore and should be removed. As after the query is send to Hibernate / Hibernate-Search the results are objects which are easy accessible and hold all information which are needed.

BartChris: We have a situation, where some information (metadata, but also filter settings like "stepopen") can only be queried using the index. So we need the index to answer the query we have to query the index somewhere in the process. In my review i will try to concentrate on functionality so that we can focus on performance later, but the general remark i have is this.

We have seen in different places (e.g. https://github.com/kitodo/kitodo-production/issues/4378) that raw SQL / a dedicated index queries often outperforms Hibernate standard queries, which do somethines hundred of subsequent queries to fetch data which could be fetched in one go. (If Hibernate supports the optimized SQL query)
 
The process list search is one of the most used features and it should be optimized as much as possible (withing the constraints of Hibernate) in my oppinion. When scrolling through the process list i still see A LOT of SQL queries. And queries like here

https://github.com/kitodo/kitodo-production/blob/b6fb9bac0836e9677d5d22b3693c229e79eac429/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/processActionsColumn.xhtml#L35

are executed for every single process leading to the classical n+1 query.

I am wondering wether we can not construct something like a ProcessListDTO Object (or ProcessListObject, this is not about the DTO concept), which contains only the information needed to construct the list. (and the statistics). The objects in the list already have the necessary properties like `hasChildren` precomputed, so there is no reason to do additional database queries to calculate them hundred of times. 

And since we are already using the index, why not **store** the necessary properties in the index. We do not have to store all projects associated with a process in the index, but we should store the information which is needed for display like `hasChildren`. If we have this, we ideally only need one query insted of hundreds to get all the data for the list and have good performance. This will also make issues like https://github.com/kitodo/kitodo-production/issues/6348 less annoying, although we can do something on usability there, too. 

The necessary features of Hibernate Search have already been referenced in other places

https://docs.jboss.org/hibernate/search/7.1/reference/en-US/html_single/#projections
https://docs.jboss.org/hibernate/search/7.1/reference/en-US/html_single/index.html#search-mapping-associated
https://docs.jboss.org/hibernate/search/7.1/reference/en-US/html_single/index.html#mapping-reindexing-derivedfrom

I think it is fantastic that we now reduce indexing time by only index the processes and using the Hibernate abstractions simplifies a lot of logic. But when necessary for performance it would be great if we could use dedicated custom logic to achieve the perfomance in places where it is mostly needed.

BartChris:
Requirement checklist; as discussed in the Community Board:

1. Blue search box in the top left jumps to the normal process list and not a special list with less selectable actions

Works in general. Some problems

a) Searching in the top left box does not reset the process list filter, but appends

image

BartChris:
> ~I get an exception when navigating to the processes list (after rebuilding the index via the updated "indexing" page):~
> 
> **Edit**: I tested on an outdated version of this branch. With the latest commits the exception on the process list page disappears.
> 
> Instead, I now get an exception when trying to edit a project, though:
> 
> Bildschirmfoto 2024-12-02 um 13 20 43

I get th

> ~I get an exception when navigating to the processes list (after rebuilding the index via the updated "indexing" page):~
> 
> **Edit**: I tested on an outdated version of this branch. With the latest commits the exception on the process list page disappears.
> 
> Instead, I now get an exception when trying to edit a project, though:
> 
> Bildschirmfoto 2024-12-02 um 13 20 43

I get the same exception when trying to add a new process through the project list.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/IndexingKeyworder.java
BartChris: I might need to revisit the requirements, but from my understanding, the main source of complexity seems to be supporting multi-term searches like "digi abc" that would match titles such as digi_1234_abc, ignoring separators like underscores or spaces. This feels like overkill and might go beyond what’s necessary.

A simpler approach could be focusing on searches based on three or more consecutive characters/tokens, which would still allow matching relevant titles.

For example, searching for "abc" or "1234" should match the title digi_1234_abc if the search logic considers tokens generated by an ngram tokenizer with min_gram: 3. This would generate tokens such as:


"digi_1234_abc" → ["dig", "digi", "igi", "gi_", "i_1", "123", "234", "abc"]
With this setup:

- A search for "abc" would match because "abc" exists as a token.
- A search for "igi" would match because "igi" is also generated as a token.

henning-gerhardt: Similar to the programming way in Java, it is also possible to define the same in [custom index settings](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#backend-elasticsearch-configuration-index-settings) which could even used for other general or index specific adjustments.

matthias-ronge: The problem here is that tokenization does not take place in Hibernate Search, but in the respective search engine backend. It is configured in the index profile, whereby you can only use the configuration options that are available, otherwise you would have to make custom implementations in the search engine itself, if that is even possible. The possible options here are also backend-specific, if we configure special OpenSearch processors here, it may no longer work with Lucene. This is a huge field that would require a test environment, test procedure to check whether the right thing is working, and in the end it might not work exactly as intended.

The implementation provided here

- is very performant,
- does exactly what we want,
- is under our control,
- is guaranteed not to change when the search engine version changes, and
- is independent of the search backend used.

Moreover, changing this now would make the task enormously larger, which is not feasible within the scope of the funded project effort.

BartChris: I understand your point. If it’s feasible to revisit this later, my main considerations would be:

- Is custom tokenization truly necessary? I’d prefer to stick with the default approach unless there's a proven need for a custom solution.

- Defining requirements is challenging. I’ve tried to gather input from colleagues on the Community Board, but identifying the essential needs purely in theory has proven difficult.
- A standard n-gram tokenizer should suffice. Given that n-gram tokenization has been a longstanding feature in both Elasticsearch and OpenSearch, I doubt it will be deprecated anytime soon. That said, there’s always some level of uncertainty.
- That being said, I agree that progressing with the current implementation is important. However, I plan to revisit this with my colleagues to explore opportunities for simplification. Ideally, we can rely on the search engine’s default settings and make more informed decisions in future iterations based on real-world needs.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
matthias-ronge: The basic design decision was that all data is loaded from the database, and the index is only used when necessary. This is the right decision for a business system with live data and a few users. For a system with thousands of users and less time-critical (for example a travel portal where it doesn't matter if newly added offers only appear after a few minutes), a completely different design would make sense.

Secondly, with this design we wanted to ensure that the application always works, even without an available index. There is no downtime in productivity due to a broken index. The only exception is that you cannot create child processes under existing parent processes, and most search queries will remain empty.

I would like to assure you at this point that I have thought this through very carefully and made this decision to the best of my knowledge and belief.

Technically, only the IDs are queried here, the objects are _not_ loaded. I was very disappointed to find that such functionality is not available in Hibernate Search in the Hibernate Query Language. I had expected that it would be possible to search on indexed fields using a specific HQL keyword or syntax, but that is not the case, I have not found anything like that. So here the IDs of the hits from the index query are passed in HQL; yes, this can possibly cause queries that are several KB in size in some cases, but should not be a problem on a server with GBs of memory. Comparing integers with IDs for the database engine should not be a big deal either.

BartChris: I can get on board with your design decision and it is very good that Kitodo is basically usable without an index. If we go own this road we probably have to investigte the performance of the database retrieval. It seems to me that the retrieval of the IDs from the index is fast but the Hibernate query tries to load more data than what is necessary.
It takes me almost 30 seconds to go to the next page when i have 30 entries per page on the process list and this is definitely WAY faster on current Master. I would investigate wether it is really necessary to load full (process) entities here which seem to trigger hundreds of queries and if we can not directly load a lightweight DTO (loaded from the Database) containing only the data we need.  In my tests i can even do the calculation of the number of child processes for each process on the fly using one SQL query.

BartChris:
When trying to create a process i still get

```
org.hibernate.LazyInitializationException: could not initialize proxy [org.kitodo.data.database.beans.ImportConfiguration#3] - no Session
	at org.hibernate.proxy.AbstractLazyInitializer.initialize(AbstractLazyInitializer.java:176)
	at org.hibernate.proxy.AbstractLazyInitializer.getImplementation(AbstractLazyInitializer.java:322)
	at org.hibernate.proxy.pojo.bytebuddy.ByteBuddyInterceptor.intercept(ByteBuddyInterceptor.java:45)
	at org.hibernate.proxy.ProxyConfiguration$InterceptorDispatcher.intercept(ProxyConfiguration.java:95)
	at org.kitodo.data.database.beans.ImportConfiguration$HibernateProxy$fWH8cidi.getConfigurationType(Unknown Source)
	at org.kitodo.production.forms.createprocess.CreateProcessForm.setDefaultImportConfiguration(CreateProcessForm.java:476)
	at org.kitodo.production.forms.createprocess.CreateProcessForm.prepareProcess(CreateProcessForm.java:431)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```
Do you consider the PR to be in `draft` state?

matthias-ronge:
The basic functionality should work. I was able to start the branch myself. Maybe not everything works, there can always be bugs that I need to fix. For a review (proofreading the source code) I think it's already good enough for you to look at it and many features can work and most of it works. For now I managed to fix all the integration tests. And the Selenium tests worked for me today, too. So far it's going well. After @solth it's just the final approval on the merge to the "main" branch, the conflicting commits still need to be merged and fixed. This can be done separately afterwards. I always check out the branch locally and let it run as far as possible.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
matthias-ronge: On my very old laptop, running everything locally in an IDE and debug mode, flipping through the process list pages ist quick (a second) with 10 processes p.P. and I don’t see more queries happen. Maybe it is a special combination of processes, rulesets and maybe other circumstances lead to this.

matthias-ronge:
Hi @solth, is that OK so far? Can you merge the PR? Then I could start merging up to `main`. I would like to finish the project. Or do you see any need for changes here? I don't like rebasing a branch with lots of merges later, this is why I ask. This is not an exclusion to fix any issues that arise before the final merge, just that I would then fix them on the updated branch.

solth:
@matthias-ronge Even though there are still some issues that need to be addressed (for example potential performance problems) I think those can be discussed and resolved in the final pull request from the `hibernate-search` branch against the `main` branch. Since we need to get this project finished I will merge this pull request so that you can work on updating the `hibernate-search` branch with the changes introduced into the `main` branch. 

_Important_: this does **_not_** mean that all review remarks from this pull request can be discarded. I want to be clear that it will still be legitimate to request conceptional changes in the final pull request, for example regarding the retrieval of many objects via huge SQL queries with thousands of IDs. I am not convinced that we cannot find a better way in that regard and expect we will still need changes at some other points as well.

@henning-gerhardt, @BartChris, @thomaslow, @oliver-stoehr please apply any further review remarks to the upcoming pull request from the `hibernate-search` branch against the `main` branch once @matthias-ronge deems it ready for review!

PR #6286, opened 2024-10-31
Require global role editing permission to see global authorities

### LINKED ISSUE #5837: User with only client authority to edit roles can still add global authorities to role
**Describe the bug**
We have "client-power-users" who are able to create and edit users and roles for the client but not globally. 
We assign these users the client authorities 

- `deleteRole_clientAssignable`
- `editRole_clientAssignable`  
- `addRole_clientAssignable`

but not the corresponding global authorities. 
However, these users are still able to assign global authorities to roles.
Not sure if this should be classified as a bug, but to me it does not seem to be the desired behaviour.

**To Reproduce**
1. Create a User with client authortiy to edit roles.
2. Log in as this user.
3. Edit a role.
4. You are able to add global authorities to the user.

**Expected behavior**
A user with only the client authority for editing roles should not be able to add global authorities to a role. As a solution a check for the global edit role authority could be added to `roleEdit/details.xhtml`. If the user only has the client authority, the global authorties should not be shown.

**Release**
3.5.0

solth:
This pull request adds the check suggested in #5837 and hides the "Global authorities" on the "Role edit" page when the user does not have the permission to globally edit a role:

Bildschirmfoto 2024-10-31 um 08 38 25

@pontus-osterdahl does this work for you?

Fixes #5837

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/controller/SecurityAccessController.java
+++ b/Kitodo/src/main/java/org/kitodo/production/controller/SecurityAccessController.java
@@ -957,6 +957,24 @@ public class SecurityAccessController {
     }
 
     /**
+     * Checks if current user has global authority for editing roles.
+     *
+     * @return true if current user has global authority for editing roles
+     */
+    public boolean hasAuthorityGlobalToEditRole() {
+        return securityAccessService.hasAuthorityGlobalToEditRole();
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


henning-gerhardt (approved):
Code changes looks good and a simple functional test was positive. If possible the original author of the issue should look into this too.

pontus-osterdahl:
Looks good!

PR #6287, opened 2024-10-31
Calculate the default timeout correctly

matthias-ronge:
Due to a unit mix-up, the default timeout was only set to 1/60 of the intended time period, which occasionally causes errors in large images that you have to find first.

### DIFF
--- a/Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ConvertRunner.java
+++ b/Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ConvertRunner.java
@@ -51,7 +51,7 @@ class ConvertRunner {
     /**
      * Default timeout.
      */
-    private static final int DEFAULT_TIMEOUT_MINS = (int) TimeUnit.MINUTES.convert(2, TimeUnit.HOURS);
+    private static final int DEFAULT_TIMEOUT_SECS = (int) TimeUnit.SECONDS.convert(2, TimeUnit.HOURS);
 
     /**
      * {@code convert} command, optionally with full path.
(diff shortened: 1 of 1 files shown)


Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ConvertRunner.java
```diff
@@ -51,7 +51,7 @@ class ConvertRunner {
     /**
      * Default timeout.
      */
-    private static final int DEFAULT_TIMEOUT_MINS = (int) TimeUnit.MINUTES.convert(2, TimeUnit.HOURS);
+    private static final int DEFAULT_TIMEOUT_SECS = (int) TimeUnit.SECONDS.convert(2, TimeUnit.HOURS);
```
stweil, line 54: I wonder whether `7200`, `2 * 3600` or `2 * 60 * 60` would not be easier to read. At least it would avoid the type cast.

Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ConvertRunner.java
```diff
@@ -51,7 +51,7 @@ class ConvertRunner {
     /**
      * Default timeout.
```
stweil, line 52: Even if we have a speaking variable name, "Default timeout (seconds)" or a similar description might help additionally.

stweil (approved):
Thanks, the code changes fix a bug (so might also be backported), and my two comments are only suggestions for optional small improvements.

Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ConvertRunner.java
stweil: The type cast could also be avoided by using a `long DEFAULT_TIMEOUT_SECS` and `KitodoConfig.getLongParameter` below.

PR #6292, opened 2024-11-05
Add missing declaration of 'groupDisplayLabel' to test ruleset.xsd

solth:
This PR adds `groupDisplayLabel` to the ruleset schema definition used for tests.


thomaslow (approved):
Checked with `diff` that both files are the same again. Looks good.

PR #6295, opened 2024-11-07
Add option to set default client for each user

### LINKED ISSUE #6262: Default client
When a user has multiple clients assigned to him but mainly uses one of those clients most of the time, it can quickly become cumbersome to always select the same client when logging into the system.

It should be possible to configure a default client that is automatically selected when a user logs into the system and has multiple clients. Configuring a default client should be optional, though, and the system should work as it currently does when no default client is configured.

It may be worth considering a slight variation of the solution proposed above where the default client is automatically selected in the "Select client" popup dialog upon login, but the user still has to confirm the selection. He might thus still select a different client when required. 

**Additional context**
Since it is possible to change the client using the user menu in the upper right corner of the screen at any time after login I would prefer the first approach, where an optional default client is not only pre-selected, but the selection is actually committed so that the user can skip the whole client selection entirely during login if he opts to use the option.

solth:
This pull request adds the option to set a default client in the user settings. This default client will be selected automatically upon login, skipping the "Select client" dialog for users with multiple clients. 

This functionality is optional which means the system will work exactly the same as before when no default client is set.

Fixes #6262

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/User.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/User.java
@@ -117,6 +117,10 @@ public class User extends BaseBean {
     @Column(name = "show_physical_page_number_below_thumbnail")
     private boolean showPhysicalPageNumberBelowThumbnail;
 
+    @ManyToOne
+    @JoinColumn(name = "default_client_id", foreignKey = @ForeignKey(name = "FK_user_default_client_id"))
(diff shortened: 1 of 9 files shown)


henning-gerhardt (changes requested):
Code changes looks good but one improvement should be made: if a user has only one client assigned than this user settings dialog should not appear. Or in other words: the user settings dialog which client should be used in general should only be visible if a user as access to more then one client.

henning-gerhardt (approved):

PR #6297, opened 2024-11-07
Add EAD collection import

### LINKED ISSUE #5984: Import von Archiv-Metadaten aus EAD-DDB-XML um hierarchische Vorgänge automatisch zu erstellen
### Discussed in https://github.com/kitodo/kitodo-production/discussions/5829

<div type='discussions-op-text'>

<sup>Originally posted by **solth** November  2, 2023</sup>
Kitodo.Production 3.x unterstützt verschiedene Formen des Metadaten-Imports. Diese werden über sogenannte "Importkonfigurationen" konfiguriert (siehe https://github.com/kitodo/kitodo-production/wiki/Projektauswahl-und-Katalogsuche#import-konfigurationen), die die alte Datei `kitodo_opac.xml` ersetzen:

1. **Katalogabfrage** - Metadaten einzelner Datensätze können via ID (oder andere Suchparameter) von konfigurierten Suchschnittstellen abgefragt und daraus Digitalisierungsvorgänge erstellt werden. Wenn die Suchschnittstelle dies unterstützt, sind verschiedene Arten des **hierarchischem Imports** möglich, d.h. die Katalogschnittstelle wird für den Import von potentiellen Über- oder Unterordnungen erneut abgefragt, wenn deren IDs in den Metadaten des ursprünglich importierten Datensatzes enthalten sind, und die zurückgegebenen Datensätze als hierarchisch verknüpfte Vorgänge importiert. (siehe auch https://github.com/kitodo/kitodo-production/wiki/Kalliope-SRU-deutsch). Wenn der Datensatz die ID einer Überordnung enthält, ein Vorgang mit dieser ID aber bereits in Kitodo existiert, wird die Überordnung nicht erneut importiert, sondern der importierte Datensatz mit der bereits in Kitodo bestehenden Überordnung hierarchisch verknüpft.

2. **Datei-Upload** - XML-Dateien mit Metadaten können direkt vom lokalen Rechner hochgeladen werden. Es ist möglich, aus der gleichen XML-Datei zusätzlich die Überordnung des Datensatzes zu importieren, wenn eine separate Abbildungsdatei für die Extraktion der Metadaten der Überordnung in der verwendeten Importkonfiguration hinterlegt ist. Es ist also auch hier ein (auf die Überordnung) eingeschränkter, hierarchischer Import möglich. 

3. **Vorgangsvorlagen** - Metadaten können von entsprechend ausgezeichnten Vorgängen übernommen werden, z.B. für das Anlegen multipler Zeitschriftenbände mit größtenteils gleichen Metadaten.

4. **Massenimport** - Über eine zusätzliche Maske können Listen von IDs an eine konfigurierte Katalogschnittstelle (siehe 1.) übergeben werden, die dann den Import sequentiell für jede einzelne ID automatisch durchführt und multiple Vorgänge erstellt. Wie beim regulären Katalogimport werden importierte Datensätze (z.B. Bände eines mehrbändigen Werkes) als Unterordnungen an existierende Überordnungen (das mehrbändige Werk) gehängt, wenn ein entsprechender Vorgang der Überordnung in Kitodo bereits existiert oder die ID der Überordnung in der Liste der zu importierenden IDs _vor_ den IDs der Unterordnungen steht und damit zuerst importiert wird.

---

Konzept Weiterentwicklung: 

**Hierarchischer Massenimport (e.g. "Archivdatenimport")**:
Aus mehreren Gesprächen mit Archiven ist hervorgegangen, dass die bestehenden und oben beschriebenen Möglichkeiten des Metadatenimports für den Einsatz von Kitodo.Production in Archiven unzureichend sind. Es besteht der Wunsch, mehrere Vorgänge inkl. einer Überordnung (Bestand) und potentiell vieler Unterordnungen (Verzeichniseinheiten) aus _EINER_ manuell hochgeladenen oder von einer Schnittstelle zurückgegebenen XML-Datei (z.B. EAD) automatisch als einzelne, hierarchisch miteinander verknüpfte Vorgänge zu importieren.

Dies könnte über eine Erweiterung der Massenimport-Seite funktionieren. Wenn hierbei von einer festen Struktur einer solchen, hochgeladenen Archivmetadaten-Datei ausgegangen werden kann, würde der Massenimport aus **einer** Datei automatisch einen Vorgang für den Bestand ("collection") und n Vorgänge für die Verzeichniseinheiten ("item") erstellen und diese hierarchisch miteinander verknüpfen. 

@vernst @aetherfaerber @rchr @PeterJunger @Aigeus erfüllt dieses Konzept aus Sicht der Archive die prinzipielle Anforderung an eine Erweiterung des Imports? Was müsste noch ergänzt werden bzw. was habe ich nicht bedacht oder vielleicht falsch verstanden? 

@kitodo/kitodo-community-board könnte dieses Thema in Eure Liste aufgenommen werden?</div>

## Description

Bei uns wären genau die oben beschriebenen Funktionalität nützlich. Unser Archivinformationssystem hat keine Schnittstelle. Wir würden uns aber wünschen, dass wir hierarchische Exportdateien (EAD-DDB-XML) in Kitodo importieren können und, dass so ganze Vorgänge mit den jeweiligen Metadaten der Bestände, Unterbestände und einzelner Archivalieneinheiten als verknüpfte hierarchische Vorgänge importiert werden können. Momentan müssen sämtliche Metadaten manuell eingegeben werden, was einen erheblichen Zeitaufwand bei der Erstellung vieler hierarchischer Vorgänge bedeutet. 

Zeitersparnis bei der Erfassung von Metadaten hierarchischer Vorgänge, die sich aus der Tektonik eines Archivinformationssystems ergeben. 

low to medium

e.g.

* **low** ~ less than 5 working days
* **medium** ~ less than 10 working days
* **high** ~ more than 10 working days

solth:
This pull request adds the option to import EAD collections from uploaded XML files. The import creates one (parent) process for the collection and one (child) for _each_ "file" contained in the XML file, thus allowing complete collections ("Bestände) into individual processes from a single file upload.

To be able to use this EAD import, first a ruleset and an Import configuration of type "File Upload" has to be configured with metadata format `EAD`:
Bildschirmfoto 2024-11-08 um 00 34 29

Accordingly, the mapping file used in the import configuration has to map EAD to the Kitodo internal metadata format:
Bildschirmfoto 2024-11-08 um 00 39 34

A separate mapping file can be configured to map the collection level of the imported EAD XML file, in the field "Mapping file for parent record":
Bildschirmfoto 2024-11-08 um 00 41 17

This pull request adds example mapping files for the parent and child processes (`eadParent2kitodo.xsl` and `ead2kitodo.xsl`) as well as a corresponding ruleset (`ruleset_ead.xml`) that can be used for this purpose and that of cause can also extended when required.

The user can then use this import configuration to upload an EAD XML file containing the EAD collection. The upload dialog also allows to select the EAD level of the element that is to be transformed into the parent process from `COLLECTION`, `CLASS` and `SERIES` and the EAD level of items that are transformed into child processes from `FILE` and `ITEM`:
Bildschirmfoto 2024-11-08 um 00 30 33

Additionally, a limit for the maximum number or child processes that can be processed interactively in the import mask can now be configured in `kitodo_config.properties` using the parameter `maxNumberOfProcessesForImportMask`. (this value defaults to 5). If the number of child elements in the uploaded file exceeds this limit, the user is offered with a choice to relocate the import of all processes to a background task:
Bildschirmfoto 2024-11-08 um 00 51 50

If the user has the corresponding permission, the progress of this background import can then be monitored on the "Task manager" page of Kitodo:
Bildschirmfoto 2024-11-08 um 00 53 15

Fixes #5984

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CatalogImportDialog.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CatalogImportDialog.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 
 import javax.faces.context.FacesContext;
 import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.stream.XMLStreamException;
 import javax.xml.transform.TransformerException;
 import javax.xml.xpath.XPathExpressionException;
 
@@ -29,7 +30,9 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.kitodo.api.dataeditor.rulesetmanagement.FunctionalMetadata;
 import org.kitodo.api.externaldatamanagement.SingleHit;
+import org.kitodo.api.schemaconverter.DataRecord;
 import org.kitodo.api.schemaconverter.ExemplarRecord;
+import org.kitodo.api.schemaconverter.MetadataFormat;
 import org.kitodo.data.database.beans.ImportConfiguration;
 import org.kitodo.data.database.exceptions.DAOException;
 import org.kitodo.exceptions.CatalogException;
@@ -53,7 +56,7 @@ import org.xml.sax.SAXException;
 
 public class CatalogImportDialog  extends MetadataImportDialog implements Serializable {
     private static final Logger logger = LogManager.getLogger(CatalogImportDialog.class);
-    private final LazyHitModel hitModel = new LazyHitModel();
+    private final LazyHitModel hitModel;
 
     private static final String ID_PARAMETER_NAME = "ID";
     private static final String HITSTABLE_NAME = "hitlistDialogForm:hitlistDialogTable";
@@ -64,6 +67,10 @@ public class CatalogImportDialog  extends MetadataImportDialog implements Serial
     private int numberOfChildren = 0;
     private String opacErrorMessage = "";
     private boolean additionalImport = false;
+    private String selectedField = "";
+    private String searchTerm = "";
+    private int importDepth = 2;
+
 
     /**
      * Standard constructor.
@@ -72,6 +79,7 @@ public class CatalogImportDialog  extends MetadataImportDialog implements Serial
      */
     CatalogImportDialog(CreateProcessForm createProcessForm) {
         super(createProcessForm);
+        this.hitModel = new LazyHitModel(this);
     }
 
     /**
@@ -87,11 +95,11 @@ public class CatalogImportDialog  extends MetadataImportDialog implements Serial
      * @return list of search fields
      */
     public List<String> getSearchFields() {
-        if (Objects.isNull(hitModel.getImportConfiguration())) {
+        if (Objects.isNull(createProcessForm.getCurrentImportConfiguration())) {
             return new LinkedList<>();
         } else {
             try {
-                return ServiceManager.getImportService().getAvailableSearchFields(hitModel.getImportConfiguration());
+                return ServiceManager.getImportService().getAvailableSearchFields(createProcessForm.getCurrentImportConfiguration());
             } catch (IllegalArgumentException e) {
                 Helper.setErrorMessage(e.getLocalizedMessage(), logger, e);
                 return new LinkedList<>();
@@ -104,8 +112,8 @@ public class CatalogImportDialog  extends MetadataImportDialog implements Serial
      */
     public void search() {
         try {
-            if (skipHitList(hitModel.getImportConfiguration(), hitModel.getSelectedField())) {
-                getRecordById(hitModel.getSearchTerm());
+            if (skipHitList(createProcessForm.getCurrentImportConfiguration(), getSelectedField())) {
+                getRecordById(getSearchTerm());
             } else {
                 List<?> hits = hitModel.load(0, 10, null, SortOrder.ASCENDING, Collections.EMPTY_MAP);
                 if (hits.size() == 1) {
@@ -175,16 +183,21 @@ public class CatalogImportDialog  extends MetadataImportDialog implements Serial
                 createProcessForm.setChildProcesses(new LinkedList<>());
                 int projectId = this.createProcessForm.getProject().getId();
                 int templateId = this.createProcessForm.getTemplate().getId();
-                ImportConfiguration importConfiguration = this.hitModel.getImportConfiguration();
+                ImportConfiguration importConfiguration = createProcessForm.getCurrentImportConfiguration();
 
-                // import current and ancestors
-                LinkedList<TempProcess> processes = ServiceManager.getImportService().importProcessHierarchy(
-                        currentRecordId, importConfiguration, projectId, templateId, hitModel.getImportDepth(),
-                        createProcessForm.getRulesetManagement().getFunctionalKeys(
-                                FunctionalMetadata.HIGHERLEVEL_IDENTIFIER));
-                // import children
-                if (this.importChildren) {
-                    importChildren(projectId, templateId, importConfiguration, processes);
+                LinkedList<TempProcess> processes;
+                if (MetadataFormat.EAD.name().equals(importConfiguration.getMetadataFormat())) {
+                    processes = createEadProcesses(importConfiguration);
+                } else {
+                    // import current and ancestors
+                    processes = ServiceManager.getImportService().importProcessHierarchy(currentRecordId,
+                            importConfiguration, projectId, templateId, getImportDepth(),
+                            createProcessForm.getRulesetManagement().getFunctionalKeys(
+                                    FunctionalMetadata.HIGHERLEVEL_IDENTIFIER));
+                    // import children
+                    if (this.importChildren) {
+                        importChildren(projectId, templateId, importConfiguration, processes);
+                    }
                 }
 
                 if (!createProcessForm.getProcesses().isEmpty() && additionalImport) {
@@ -196,13 +209,37 @@ public class CatalogImportDialog  extends MetadataImportDialog implements Serial
                     attachToExistingParentAndGenerateAtstslIfNotExist(currentTempProcess);
                     showMessageAndRecord(importConfiguration, processes);
                 }
+
             } catch (IOException | ProcessGenerationException | XPathExpressionException | URISyntaxException
-                    | ParserConfigurationException | UnsupportedFormatException | SAXException | DAOException
-                    | ConfigException | TransformerException | NoRecordFoundException | InvalidMetadataValueException
-                    | NoSuchMetadataFieldException e) {
+                     | ParserConfigurationException | UnsupportedFormatException | SAXException | DAOException
+                     | ConfigException | TransformerException | NoRecordFoundException | InvalidMetadataValueException
+                     | NoSuchMetadataFieldException | XMLStreamException e) {
                 throw new CatalogException(e.getLocalizedMessage());
             }
         }
+    }
+
+    private LinkedList<TempProcess> createEadProcesses(ImportConfiguration importConfiguration) throws NoRecordFoundException,
+            XPathExpressionException, IOException, ParserConfigurationException, SAXException, XMLStreamException,
+            UnsupportedFormatException, ProcessGenerationException, URISyntaxException, InvalidMetadataValueException,
+            TransformerException, NoSuchMetadataFieldException {
+        LinkedList<TempProcess> processes = new LinkedList<>();
+        DataRecord externalRecord = ServiceManager.getImportService()
+                .importExternalDataRecord(importConfiguration, this.currentRecordId, false);
+        createProcessForm.setXmlString(externalRecord.getOriginalData().toString());
+        if (createProcessForm.limitExceeded(externalRecord.getOriginalData().toString())) {
+            createProcessForm.calculateNumberOfEadElements();
+            Ajax.update("maxNumberOfRecordsExceededDialog");
+            PrimeFaces.current().executeScript("PF('maxNumberOfRecordsExceededDialog').show();");
+        } else {
+            LinkedList<TempProcess> eadProcesses = ServiceManager.getImportService()
+                    .parseImportedEADCollection(externalRecord, importConfiguration,
+                            createProcessForm.getProject().getId(), createProcessForm.getTemplate().getId(),
+                            createProcessForm.getSelectedEadLevel(), createProcessForm.getSelectedParentEadLevel());
+            createProcessForm.setChildProcesses(new LinkedList<>(eadProcesses.subList(1, eadProcesses.size() - 1)));
+            processes = new LinkedList<>(Collections.singletonList(eadProcesses.get(0)));
+        }
+        return processes;
     }
 
     private void showMessageAndRecord(ImportConfiguration importConfiguration, LinkedList<TempProcess> processes) {
@@ -238,7 +275,7 @@ public class CatalogImportDialog  extends MetadataImportDialog implements Serial
         try {
             if (this.importChildren) {
                 this.numberOfChildren = ServiceManager.getImportService().getNumberOfChildren(
-                        this.hitModel.getImportConfiguration(), this.currentRecordId);
+                        createProcessForm.getCurrentImportConfiguration(), this.currentRecordId);
             }
             if (this.importChildren && this.numberOfChildren > NUMBER_OF_CHILDREN_WARNING_THRESHOLD) {
                 Ajax.update("manyChildrenWarningDialog");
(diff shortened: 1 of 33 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -108,6 +122,10 @@ public CreateProcessForm() {
         this.priorityList = priorityList;
     }
 
+    public void calculateNumberOfEadElements() throws XMLStreamException {
```
henning-gerhardt, line 125: Missing JavaDoc

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CatalogImportDialog.java
```diff
@@ -360,4 +397,59 @@ public void setAdditionalImport(boolean additionalImport) {
...
+     * @return value of selectedField
+     */
+    public String getSelectedField() {
+        return this.selectedField;
+    }
+
+    /**
+     * Set selectedField.
+     *
+     * @param field as java.lang.String
```
henning-gerhardt, line 431: see above.

Kitodo/src/main/java/org/kitodo/production/model/LazyHitModel.java
```diff
@@ -19,25 +19,21 @@
...
-    private int importDepth = 2;
-
+    private final CatalogImportDialog catalogImportDialog;
     private SearchResult searchResult = null;
 
     /**
      * Empty default constructor. Sets default catalog and search field, if configured.
      */
-    public LazyHitModel() {
+    public LazyHitModel(CatalogImportDialog dialog) {
```
henning-gerhardt, line 35: Javadoc should be adjusted.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -589,21 +613,46 @@ private void importParents(String recordId, ImportConfiguration importConfigurat
     /**
      * Check if there already is a parent process in Database.
      */
-    public void checkForParent(String parentID, Ruleset ruleset, int projectID)
-            throws DAOException, IOException, ProcessGenerationException {
-        if (Objects.isNull(parentID)) {
-            this.parentTempProcess = null;
-            return;
+    public void checkForParent(String parentID, Ruleset ruleset, int projectID) {
```
henning-gerhardt, line 616: JavaDoc should be adjusted.

Kitodo/src/main/java/org/kitodo/production/thread/ImportEadProcessesThread.java
```diff
@@ -0,0 +1,348 @@
...
+                 | URISyntaxException | InvalidMetadataValueException | TransformerException
+                 | NoSuchMetadataFieldException | DataException | CommandException e) {
+            logger.error(e.getMessage(), e);
+            cleanUpProcesses(newProcessIds, newParentId);
+            throw new RuntimeException(e);
+        }
+    }
+
+    private void cleanUpProcesses(List<Integer> processIds, int newParentId) {
+        logger.error("Deleting processes created until this point to resolve erroneous intermediate state.");
```
henning-gerhardt, line 233: Is `error` log level correct? Should not log level `info` or lower be used here? It did not look like that an error occur and get logged.

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CatalogImportDialog.java
```diff
@@ -360,4 +397,59 @@ public void setAdditionalImport(boolean additionalImport) {
...
+     * @return value of searchTerm
+     */
+    public String getSearchTerm() {
+        return this.searchTerm;
+    }
+
+    /**
+     * Set searchTerm.
+     *
+     * @param searchTerm as java.lang.String
```
henning-gerhardt, line 413: `java.lang.String` could be simplified to ` string` or the whole wording should be adjusted.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -679,19 +728,116 @@ public LinkedList<TempProcess> getChildProcesses(ImportConfiguration importConfi
...
+                importConfiguration, isParent);
+        return createTempProcessFromDocument(importConfiguration, internalEadCollectionDocument, templateId, projectId);
+    }
+
+    private List<Element> getEADElements(Document document, String level) {
+        return XMLUtils.getElementsByTagNameAndAttributeValue(document, StringConstants.C_TAG_NAME,
+                StringConstants.LEVEL, level);
+    }
+
+    public boolean isMaxNumberOfRecordsExceeded(String xmlString, String tagName) throws XMLStreamException {
```
henning-gerhardt, line 838: Missing JavaDoc.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -409,8 +412,7 @@ public String getParentID(Document document, String higherLevelIdentifier, Strin
      */
     public TempProcess createTempProcessFromDocument(ImportConfiguration importConfiguration, Document document,
                                                      int templateID, int projectID)
-            throws ProcessGenerationException, IOException, TransformerException, InvalidMetadataValueException,
-            NoSuchMetadataFieldException {
+            throws ProcessGenerationException, IOException, TransformerException {
```
henning-gerhardt, line 415: JavaDoc should be adjusted.

henning-gerhardt (commented):
Some notes from reading the changes.

BartChris:
While testing this i first made a dumb configuration error. I associated the wrong XSL-file with the EAD-Mapping-File.  
This brought up the following exception message:

KITODO_EAD_ERROR

This error is produced by the fact that the `internalRecord` does only contain the XML header and no real content at that point. 

https://github.com/kitodo/kitodo-production/blob/c947be25295c3e7a3cc09ef7764b8037b32b55b9/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L894

It is a dumb mistake on my side, but maybe we can assist by also check if we also check if the importDocument is empty and the importer has to check the mapping file configuration.

BartChris:
I tested with the given test ead file. One problem is, when i try to import the file two times. I then get a Null pointer for the  `createProcessLocation` call of the child processes and they are not created.
The problem is that when `processTempProcess` is called for a child process

https://github.com/kitodo/kitodo-production/blob/c947be25295c3e7a3cc09ef7764b8037b32b55b9/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L1227

the process titles are considered invalid because they do already exist. 
https://github.com/kitodo/kitodo-production/blob/c947be25295c3e7a3cc09ef7764b8037b32b55b9/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L1342-L1344

This should throw a `ProcessGenerationException`, but it seems this exeption does not stop the execution  in the CreateProcessForm. The execution continues and we have processes without an ID. 

The the Nullpointer is thrown here, because the file location is based on the missing ID.

https://github.com/kitodo/kitodo-production/blob/c947be25295c3e7a3cc09ef7764b8037b32b55b9/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java#L242-L246

When processing the list in the background (because the number of imported processes is greater than the configured limit) the process stops with a ProcessGenerationException.

The behaviour should probably be more like in the mass import, where items with uplicate titles are skipped.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
```diff
@@ -679,19 +736,125 @@ public LinkedList<TempProcess> getChildProcesses(ImportConfiguration importConfi
...
+        LinkedList<TempProcess> eadCollectionProcesses = new LinkedList<>();
+
+        Document eadCollectionDocument = XMLUtils.parseXMLString((String) importedEADRecord.getOriginalData());
+
+        List<Element> parentElements = getEADElements(eadCollectionDocument, eadParentProcessLevel);
+        List<Element> childElements = getEADElements(eadCollectionDocument, eadChildProcessLevel);
+
+        if (parentElements.size() != 1) {
+            throw new ProcessGenerationException(
+                    String.format("EAD XML does not contain exactly one element of parent level '%s'!",
```
BartChris, line 790: This should be translated.

BartChris:
While inspecting the Pull request i noticed a general problem:
https://github.com/kitodo/kitodo-production/issues/6303

Kitodo/src/main/java/org/kitodo/production/helper/XMLUtils.java
```diff
@@ -179,4 +195,97 @@ public static void validateXPathSyntax(String xpathString) throws XPathExpressio
...
+    /**
+     * Retrieve and return number of EAD elements with given level "eadLevel", e.g. "<c level='file'/>" from given
+     * String "xmlString".
+     *
+     * @param xmlString String containing XML to be parsed for elements
+     * @param eadLevel EAD level of elements to be counted
+     * @return number of EAD elements with given level
+     * @throws XMLStreamException when parsing XML string fails
+     */
+    public static int getNumberOfElements(String xmlString, String eadLevel) throws XMLStreamException {
```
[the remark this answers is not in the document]
solth: @BartChris yes, you are right. This was intended to be used in more general imports, but at the moment only works for the EAD import. I have renamed the method as you suggested.

Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java
solth: Done.

solth:
> When processing the list in the background (because the number of imported processes is greater than the configured limit) the process stops with a ProcessGenerationException.
> 
> For the background processing the behaviour should probably be more like in the mass import, where items with duplicate titles are skipped. In the form Kitodo should catch invalid titles and force the user to correct them.

@BartChris thanks for your analysis. Actually I implemented it this way deliberately. The goal here was to restore the original state - from _before_ the import started - when an error occurs so that the user can first check the uploaded file for problems, fix those problems and then re-try the upload with the updated file. Therefore, all processes that were already created up to the point where the exception is thrown are "cleaned up", e.g. removed.

I do understand, though, that a differnt behavior could be preferable, where importing collections does not fail with individual problematic EAD data records, but instead skips those processes and proceeds with the next "items" from the EAD file. If there is no consense on which option is better, we could ideally introduce a parameter in `kitodo_config.properties` to control the error handling that way.

BartChris:
> I tested with the given test ead file. A problem appears, when i try to import the file two times. I then get a Null pointer for the `createProcessLocation` call of the child processes and they are not created. The problem is that when `processTempProcess` is called for a child process (with an already existing title)
> 

I still have the problem, that when using the standard import way (without background task) i get a NullPointer exception, when i try to import the same EAD collection multiple times.

solth:
> > I tested with the given test ead file. A problem appears, when i try to import the file two times. I then get a Null pointer for the `createProcessLocation` call of the child processes and they are not created. The problem is that when `processTempProcess` is called for a child process (with an already existing title)
> 
> I still have the problem, that when using the standard import way (without background task) i get a NullPointer exception, when i try to import the same EAD collection multiple times.

This seems to be a problem in the hierarchical import in general that also occurs in the master branch and is not related to the changes in this pull request. There (in the master branch), I just tried to import a process hierarchy from Kalliope (ID `DE-2060-BE-13-C35`, import depth = 1, import children = true) _twice_ and on the second try got the same `NullPointerException` you documented above. 

@BartChris If this is easy to resolve I can try to include a fix in this pull request, otherwise I would suggest to handle that issue separately.

solth:
> I still have the problem, that when using the standard import way (without background task) i get a NullPointer exception, when i try to import the same EAD collection multiple times.

@BartChris I think this will take a little more time to properly fix. I would therefor suggest to deal with this problem separately. I will create a ticket for this problem.

BartChris (approved):

BartChris:
@solth I agree and approve the Pull request. I think the real usage in the Archives have to prove, wether we need additional adjustments, but it works good for me.

 Some additional ideas for the future:

- Also allow the background processing for the "normal" mass import
- Think about a way to provide a report after a background processed mass import (EAD and normal) to report the result of the processing  (Skipped processes? Reason?) in a way which is usable for non-admins

solth:
> Some additional ideas for the future:
> 
>     * Also allow the background processing for the "normal" mass import
> 
>     * Think about a way to provide a report after a background processed mass import (EAD and normal) to report the result of the processing  (Skipped processes? Reason?) in a way which is usable for non-admins

Yes, these are definitely two incremental steps that we should try to achieve and that I already had in mind as well. Thanks a lot for the review!

PR #6300, opened 2024-11-13
Fix height of metadata accordion to accommodate new header

### LINKED ISSUE #6291: Sticky header in metadata column breaks scrolling
The metadata column of the metadata editor cannot be scrolled correctly any more after #6233. This breaks #6285.

A similar problem is also present in the structure tree column (when using tab to navigate to the "add comment" button).

**To Reproduce**
1. Open metadata editor
2. Try to scroll all the way down in the metadata column

**Expected behavior**
Scrolling should work correctly (all the way down).

**Screenshots**

https://github.com/user-attachments/assets/68c2c9a9-9fb5-43bc-94b3-37b174255c87

**Release**
Master

oliver-stoehr:
Resolves #6291

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -2711,7 +2711,7 @@ Column content
 
 #structureAccordion .scroll-wrapper {
     box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


solth:
The build of this pull request fails because of a failing test introduced in #6255 . This also affects the current master.

thomaslow:
@solth I will check the selenium test today.

PR #6312, opened 2024-11-19
Update hit count when removing filter on process list page

### LINKED ISSUE #6222: Hit counter is wrong after deleting search phrase
**Describe the bug**
You are entering a search phrase in the process list. On getting the results of the search the hit counter shows the number of hits. On deleting the search phrase a new search over all processes are done and results are displayed but the hit counter shows still the number of hits from the search before

**To Reproduce**
1. Go to process list
2. Enter a search phrase and execute the search. Note the hit counter value
3. Deleting the existing search phrase which is resulting in a new search over all processes.
4. Look at the hit counter value and still see number of hits from last search instead of the hit number of the "new" search

**Expected behavior**
The number of hits should be correct even after deleting the existing search phrase.

**Release**
- 3.7.0 (was working before at least on 3.6.0-SNAPSHOT from 2023-03-01, so maybe even with 3.6.0 release).

 - OS: Linux
 - Browser: Firefox
 - Version: 130

solth:
Fixes #6222

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/filterMenu.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/filterMenu.xhtml
@@ -19,7 +19,9 @@
     <p:remoteCommand name="removeFilter"
                      action="#{ProcessForm.filterMenu.removeFilter}"
                      oncomplete="setFilterInputPadding()"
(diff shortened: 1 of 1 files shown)


henning-gerhardt (commented):
Changes are the same like in #6315 and should work in master branch too. I can not verify this as I'm currently limited to the 3.7.x release branch.

PR #6314, opened 2024-11-19
Update hit count when removing filter on process list page

solth:
Backport of #6312 for Kitodo.Production 3.8


henning-gerhardt (commented):
Changes are the same like in #6315 and should work in 3.8.x too. I can not verify this as I'm currently limited to the 3.7.x release branch.

PR #6315, opened 2024-11-19
[3.7] Update hit count when removing filter on process list page

solth:
Backport of #6312 for Kitodo.Production 3.7


henning-gerhardt (approved):
Changes look good and are working for me.

Issue #6303, opened 2024-11-14
No check for duplicate titles when creating the process using the process form 


BartChris:
**Describe the bug**
There seems to be no validation for a duplicate process title happening when i use the standard way to create a new process using the CreateProcess form.

Kitodo configuration seems to be correct

```
# Configure whether each process needs to have a unique title or not
uniqueProcessTitles=true
```

When checking the code in `importProcessHierarchy` i see no call of `processTempProcess` where this validation would take place.

https://github.com/kitodo/kitodo-production/blob/74b76be07a672701faec2da329932fd4283e6af1/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L506

https://github.com/kitodo/kitodo-production/blob/74b76be07a672701faec2da329932fd4283e6af1/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L1171

**To Reproduce**
1. Create a process with title 'ABC' using the normal create process form
2. Try to create another process with the same title 'ABC'
3. The duplicate title is not prevented 

**Expected behavior**
It should not be possible to create processes with duplicate titles, when the configuration disallows that.

**Release**
Kitodo 3.7.1, Master

solth:
I think this describes the same problem as #6055

stweil:
Is it necessary to avoid titles which only differ in upper/lower case? Such titles would not work on NTFS (Windows), macOS filesystems and also on some case insensitive Linux filesystem.

henning-gerhardt:
Kitodo.Production is using case sensitive title generation since years (I think 1.8 oder 1.9 of the old Goobi Community edition removed the case insensitive process titles) and until now no one issue was reported from productive environments where this made a problem. The problems reported here was only from a few developers on there systems which are not able to use case sensitive names like database table names as their development OS limited them (instead of fixing or adjusting it). Please don't add the old way back as this introduce new trouble and you can not adjust already existing titles without creating more trouble everywhere.

solth:
Fixed by #6326

Issue #6055, opened 2024-04-29
Process title rules not validated during process creation


solth:
**Describe the bug**
Process titles in Kitodo.Production are not allowed to contain white spaces or other non-alphanumerical special characters apart from underscore ("`_`") and hyphen ("`-`") because the process title is used as a folder name during export. These constraints are validated when editing a process (see first screenshot below). However, the same rules are _not_ enforced when _creating_ a process. This leads to processes with invalid titles that can (and will) produce errors during export. (second screenshot)

**To Reproduce**
1. Create a process using the "New process" form
2. Manually enter a process title containing spaces
3. Save the new process
4. See the invalid process title

**Expected behavior**
The process title rules should be checked during process creation as well.

**Screenshots**
Invalid process titles are prevented when _editing_ a process:
![Bildschirmfoto 2024-04-29 um 14 05 47](https://github.com/kitodo/kitodo-production/assets/19183925/f8d5fce2-a226-4888-8d35-477f75464dab)

Invalid process titles are NOT prevented when _creating_ a process:
![Bildschirmfoto 2024-04-29 um 14 06 59](https://github.com/kitodo/kitodo-production/assets/19183925/e77d6fad-6fed-4e81-ba0f-ccb206582577)

**Release**
Current master (3.7.0-SNAPSHOT) and all versions of 3.6.x

solth:
Fixed by #6326

PR #6326, opened 2024-11-22
Prevent duplicate and invalid titles

BartChris:
Fixes 
https://github.com/kitodo/kitodo-production/issues/6303
https://github.com/kitodo/kitodo-production/issues/6055

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
+++ b/Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
@@ -367,7 +367,7 @@ public enum ParameterCore implements ParameterInterface {
     /**
      * Validation of process title via regular expression.
      */
-    VALIDATE_PROCESS_TITLE_REGEX(new Parameter<>("validateProzessTitelRegex", "^[a-zA-Z0-9_-]+$")),
+    VALIDATE_PROCESS_TITLE_REGEX(new Parameter<>("validateProzessTitelRegex", "^[\\w-]+$")),
 
     /**
      * Validation of the identifier via regular expression.
--- a/Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
+++ b/Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
@@ -367,7 +367,7 @@ public enum ParameterCore implements ParameterInterface {
     /**
      * Validation of process title via regular expression.
      */
-    VALIDATE_PROCESS_TITLE_REGEX(new Parameter<>("validateProzessTitelRegex", "[\\w-]*")),
+    VALIDATE_PROCESS_TITLE_REGEX(new Parameter<>("validateProzessTitelRegex", "^[a-zA-Z0-9_-]+$")),
 
     /**
      * Validation of the identifier via regular expression.
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
@@ -514,7 +514,7 @@ public class CreateProcessForm extends BaseForm implements MetadataTreeTableInte
     /**
      * Create process hierarchy.
      */
-    private void createProcessHierarchy()
+    public void createProcessHierarchy()
             throws DataException, ProcessGenerationException, IOException {
         // discard all processes in hierarchy except the first if parent process in
         // title record link tab is selected!
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
@@ -523,7 +523,7 @@ public class CreateProcessForm extends BaseForm implements MetadataTreeTableInte
                 && !this.titleRecordLinkTab.getSelectedInsertionPosition().isEmpty()) {
             this.processes = new LinkedList<>(Collections.singletonList(this.processes.get(0)));
         }
-        ProcessService.checkTasks(this.getMainProcess(), processDataTab.getDocType());
+        processTempProcess(this.processes.get(0));
         processAncestors();
         processChildren();
         // main process and it's ancestors need to be saved, so they have IDs before creating their process directories
@@ -612,18 +612,22 @@ public class CreateProcessForm extends BaseForm implements MetadataTreeTableInte
                 ProcessService.setParentRelations(processes.get(index + 1).getProcess(), process);
             }
             if (Objects.nonNull(tempProcess.getMetadataNodes())) {
-                try {
-                    tempProcess.getProcessMetadata().preserve();
-                    ImportService.processTempProcess(tempProcess, rulesetManagement, acquisitionStage, priorityList, null);
-                } catch (InvalidMetadataValueException | NoSuchMetadataFieldException e) {
-                    throw new ProcessGenerationException("Error creating process hierarchy: invalid metadata found!");
-                } catch (RulesetNotFoundException e) {
-                    throw new ProcessGenerationException(
-                            "Ruleset not found:" + tempProcess.getProcess().getRuleset().getTitle());
-                } catch (IOException e) {
-                    throw new ProcessGenerationException("Error reading Ruleset: " + tempProcess.getProcess().getRuleset().getTitle());
-                }
+                processTempProcess(tempProcess);
             }
+        }
+    }
+
+    private void processTempProcess(TempProcess tempProcess) throws ProcessGenerationException {
+        try {
+            tempProcess.getProcessMetadata().preserve();
+            ImportService.processTempProcess(tempProcess, rulesetManagement, acquisitionStage, priorityList, null);
+        } catch (InvalidMetadataValueException | NoSuchMetadataFieldException e) {
+            throw new ProcessGenerationException("Error creating process hierarchy: invalid metadata found!");
+        } catch (RulesetNotFoundException e) {
+            throw new ProcessGenerationException(
+                    "Ruleset not found:" + tempProcess.getProcess().getRuleset().getTitle());
+        } catch (IOException e) {
+            throw new ProcessGenerationException("Error reading Ruleset: " + tempProcess.getProcess().getRuleset().getTitle());
         }
     }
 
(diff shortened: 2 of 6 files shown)


Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
```diff
@@ -367,7 +367,7 @@ public enum ParameterCore implements ParameterInterface {
     /**
      * Validation of process title via regular expression.
      */
-    VALIDATE_PROCESS_TITLE_REGEX(new Parameter<>("validateProzessTitelRegex", "[\\w-]*")),
+    VALIDATE_PROCESS_TITLE_REGEX(new Parameter<>("validateProzessTitelRegex", "^[a-zA-Z0-9_-]+$")),
```
henning-gerhardt, line 370: Should the new default value not even updated in the `kitodo_config.properties` files?

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java
```diff
@@ -588,7 +588,7 @@ private void saveChildProcessLinks() throws IOException, DataException {
             }
         }
         if (Objects.nonNull(PrimeFaces.current()) && Objects.nonNull(FacesContext.getCurrentInstance())) {
-            PrimeFaces.current().executeScript("PF('progressDialog')");
+            PrimeFaces.current().executeScript("PF('proqgressDialog')");
```
henning-gerhardt, line 591: Is this change correct? Looks like a mistake.

Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
BartChris: I am wondering: This is a hard limitation of the Kitodo software, some titles lead to problems later. Should we allow to change that in the config.properties at all? Or should we hardcode that in the application? Additionally one can think of allowing **additional** regexes which might be checked on top of the minimal requirements. See https://github.com/kitodo/kitodo-production/issues/6055

cc @solth

henning-gerhardt: At SLUB we allow even the `.` inside the process title as for historical reasons this was needed to create the processes. Maybe this can be changed now but this should be discussed.

henning-gerhardt: One more note: the `w` meta character in regular expressions stands for `a-z, A-Z, 0-9, including _ (underscore)` . So the expression can be simplified to `^[\\w-]+$`?

stweil: What about chosing a new parameter name "validateProcessTitle" which avoids the current "denglisch" (English-German mix)?

BartChris: > One more note: the `w` meta character in regular expressions stands for `a-z, A-Z, 0-9, including _ (underscore)` . So the expression can be simplified to `^[\\w-]+$`?

Thanks i changed that. It turned out that `[\\w-]+` in combination with Java's `matches` als works, but i think the changed regex is more expressive.  

>What about chosing a new parameter name "validateProcessTitle" which avoids the current "denglisch" (English-German mix)?

I also have sympathy for that, but that would require institutions having to adapt their config. I think the Pull Request should probably be backported which maybe would probably be easier, when we do not change configuration.

BartChris:
I think, for now, it would be less disruptive if institutions are allowed to override the regex in the configuration. This approach might introduce the possibility of configurations that Kitodo cannot handle (e.g., titles containing whitespace). However, one could consider introducing a baseline regex in the future—one that enforces basic constraints, such as disallowing whitespace.

The configuration value could then be redefined and renamed to build on top of this baseline regex, allowing institutions to apply additional rules. For example, institutions could enforce identifiers that always start with a specific prefix, such as "digi_".

henning-gerhardt (commented):
I'm tested this change and let a few code comments inside.

The changes are working if you have access to all projects or at least to the same projects where a process is already existing. It did not work if the already process is an other project where the user has no access. 

As in both referenced issues there is no mention if process duplication check should work over all projects for a client independent of the user as project access or not I'm can not judge if this is solution is in current state correct or not. In my opinion the duplication check should work over all projects of a client and not only over the projects where the user has access.

henning-gerhardt:
For an other issue I switched to release 3.7.1 and tried by accident to create a new monographic process from a catalogue but I got the message, that this process is already existing after I clicked on the save button in the create process form. So at least some kind of title duplication check is done even without this change.

Kitodo/src/test/java/org/kitodo/production/forms/copyprocess/CreateProcessFormIT.java
```diff
@@ -75,65 +77,110 @@ public static void cleanDatabase() throws Exception {
...
+            if (enable) {
+                ExecutionPermission.setExecutePermission(script);
+            } else {
+                ExecutionPermission.setNoExecutePermission(script);
+            }
         }
+    }
+
+    // Helper to clean up database and file system
+    private void cleanUpProcess(Process process) throws Exception {
```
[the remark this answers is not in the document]
BartChris: Thank you, i tried to adapt the tests.

Kitodo/src/test/java/org/kitodo/production/forms/copyprocess/CreateProcessFormIT.java
```diff
@@ -75,65 +81,138 @@ public static void cleanDatabase() throws Exception {
...
+        setScriptPermissions(false);
+
         long after = processService.count();
         assertEquals(before + 1, after, "No process was created!");
 
-        // clean up database, index and file system
-        Integer processId = newProcess.getId();
-        processService.remove(processId);
-        fileService.delete(URI.create(processId.toString()));
+        createdProcess = underTest.getMainProcess();
```
henning-gerhardt, line 135: The determination of the created process could be moved before the first assertion is executed so you have anytime the right process information before cleaning up. If adjusted here it should be adjusted on the other places too.

henning-gerhardt (approved):
In general the changes are right. My two notes (adjustment of the other existing `kitodo_config.properties` files and the determination of the created process inside the tests) can be done and even not they do nothing wrong.

Kitodo/src/test/java/org/kitodo/production/forms/copyprocess/CreateProcessFormIT.java
BartChris: Thanks, good idea

Kitodo/src/test/java/org/kitodo/production/forms/copyprocess/CreateProcessFormIT.java
```diff
@@ -75,65 +80,132 @@ public static void cleanDatabase() throws Exception {
...
+            // Assert that the user 2 is NOT associated with project 2
+            assertFalse(ServiceManager.getProjectService()
+                    .getById(firstProjectId)
+                    .getUsers()
+                    .contains(userTwo), "User 2 should not have access to project 2");
+        }
+        // Second process creation with duplicate title
+        CreateProcessForm underTestTwo = setupCreateProcessForm("Monograph");
+        underTestTwo.getMainProcess().setTitle("title");
+        underTestTwo.getMainProcess().setProject(ServiceManager.getProjectService().getById(secondProjectId));
```
BartChris, line 203: Before i used project 2 in both process creations during the `shouldNotAllowProcessTitlesInProjectsTheUserDoesNotBelongTo`-test. I changed that. The second process creation should happen in project 1, where user 2 has access to. The process creation should fail because of a duplicate title not because the user has not access to the project. It worked before but this makes the intent clearer.

solth:
> However, one could consider introducing a baseline regex in the future—one that enforces basic constraints, such as disallowing whitespace (defined in the application and not changeable).

I think this is a very good idea. This would prevent technically problematic titles while still keeping some flexibility for institutions that want or need slight variations from the default in their process titles!

BartChris:
> > However, one could consider introducing a baseline regex in the future—one that enforces basic constraints, such as disallowing whitespace (defined in the application and not changeable).
> 
> I think this is a very good idea. This would prevent technically problematic titles while still keeping some flexibility for institutions that want or need slight variations from the default in their process titles!

If we go down this road, we should maybe make this configurable in the project(?) settings and remove the value from the Kitodo config, to allow for more flexibility.

BartChris:
> Read I this correct that this value is moved from a global / general place to a project specific place? How should this value be transferred if you have a lot of projects which need the same value? Did we really need different values for the projects?

Good point. Maybe a 'layererd' architecture
1. Global general setting in the **application**
2. Instance specific setting in the config.properties
3. Project specific setting in the project 
(4. Maybe even in the tenant (Mandant), to have tenant-specific settings)

henning-gerhardt:
> > Read I this correct that this value is moved from a global / general place to a project specific place? How should this value be transferred if you have a lot of projects which need the same value? Did we really need different values for the projects?
> 
> Good point. Maybe a 'layererd' architecture
> 
>     1. Global general setting in the **application**
> 
>     2. **Instance specific** setting in the config.properties
> 
>     3. **Project specific** setting in the project
>        (4. Maybe even in the tenant (Mandant), to have tenant-specific settings)

If we use a layered architecture, I would use

1. global / instance setting in kitodo_config.properties with an application coded value as fallback (like in many other places already used)
2. optional tenant / client setting
3. project specific setting

solth:
> Good point. Maybe a 'layererd' architecture

Yes, I just had the same idea. Not sure, if we need _all_ these layers, though - perhaps just a global parameter in `kitodo_config.properties` that can be overwritten for each project in the projects settings in the GUI would be enough. Not sure if a client wide setting is really required. 

Tbh, I also think it is not asked to much to update one string manually in multiple projects, even if an institution has more than just a handful of projects - this is just a simple copy & paste task, after all.

solth:
> I agree, the check is useless if it does not check every title of every project (in the same client); i adapted the logic and the tests to ensure that duplicate titles are not allowed on the same client. (No matter if the user has access to the projects or not)

I think we have to keep in mind that in special circumstances a user might see processes of projects that do not belong to one of the clients he is assigned to - specifically when he has the special permission to link new processes to processes of unassigned projects, see #5869 - and if and how a duplicate check would have to take this into account.

solth:
@BartChris please rebase against the current `main` branch!

Issue #6325, opened 2024-11-22
Focus login field when login page is loaded


aetherfaerber:
**Is your feature request related to a problem? Please describe.**
When a user loads the login page to Kitodo they want to enter their login data. To do so, they must first select the field ”Benutzername” to do so.

**Describe the solution you'd like**
As the login page is rendered, the ”username” input field should be selected automatically so users can type away their username without switching between keyboard and mouse.

**Describe alternatives you've considered**
On page load, you cannot even tab into the username field but need to click anywhere on the page.

**Additional context**
There is not much to do on the login page anyway besides entering user credentials and logging in. So why not start doing it right away?

PR #6329, opened 2024-11-26
Autofocus login input on page load

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6325


solth (approved):

aetherfaerber:
Okay, sometimes just having a look at the code will actually save you time you otherwise invest in writing an issue. After the recent exchange about primefaces, I did not expect this to be so concise.

Thank you for fixing it right away!

solth:
@BartChris would you mind opening a corresponding PR against the `3.7.x` and `3.8.x` branches?

PR #6332, opened 2024-11-29
Add test to check encoding of message properties

### LINKED ISSUE #6324: Source files with wrong encoding
The latest code contains two property files with ISO-8859-1 encoding:

- Kitodo/src/main/resources/messages/password_es.properties
- Kitodo/src/main/resources/messages/errors_de.properties

They were found with `find * -type f | xargs file --mime | grep iso-8859-1`.

Here is a complete list of all encodings in the current directory:
```
% find * -type f | xargs file --mime | sed 's/.*charset/charset/' | sort | uniq -c
 115 charset=binary
   2 charset=iso-8859-1
1511 charset=us-ascii
 195 charset=utf-8
```

So there are already nearly 200 files with UTF-8 encoding (which is fine), and the two files mentioned above are the only ones with wrong encoding.

### LINKED ISSUE #6331: Unit test for character encodings in property files
The character encoding of property files like `messages_de.properties` and `messages_es.properties` containing the German and Spanish translations is sometimes changed accidentally in the repository. This leads to special characters not being displayed correctly with the assumed character encoding of `UTF-8`.

Taking a cue from https://github.com/kitodo/kitodo-production/pull/6293#issuecomment-2493223269 a unit test should be added that automatically checks whether the message files are commited with the correct encoding by reading messages with special characters from them, using the `UTF-8` for decoding the file contents and comparing the results with expected strings containing those special characters. This could perhaps be achieved by adding the `src/test/resources/messages` directory to the test ressources, as described [here](https://stackoverflow.com/a/19220729/5467214). 

**Additional context**
The problem with the wrong character encodings is described in #6324

solth:
Resolves #6331 

This PR adds a test that loads the German and Spanish message property files used in Kitodo.Production and checks if they their contents are encoded with `UTF-8`. This is done by loading specific messages known to contain unicode special characters and decode them with `UTF-8`. The results are then compared with strings containing those special characters. The tests fail if the decoded message does not match the expected strings.

This PR also sets the encoding of `Kitodo/src/main/resources/messages/password_es.properties` and `Kitodo/src/main/resources/messages/errors_de.properties` to `UTF-8` and therefore fixes #6324 

**Note**: this PR does _not_ yet replace all the codepoints like `\u00FC` with the actual special characters like `ü` in the message files. That should be done in a follow-up pull request.

### DIFF
--- a/Kitodo/src/main/resources/messages/errors_de.properties
+++ b/Kitodo/src/main/resources/messages/errors_de.properties
@@ -13,15 +13,15 @@ error=Fehler...
 errors=Fehler
 
 # A
-errorAdding=Fehler beim Hinzuf\u00FCgen von ''{0}''
+errorAdding=Fehler beim Hinzufügen von ''{0}''
 authorityAssignedError=Berechtigung kann nicht gel\u00F6scht werden, da dieser noch Rollen zugewiesen sind.
 
 # B
-batchPropertyEmpty=Die Eigenschaft {0} im Vorgang {1} ist ein Pflichtfeld, enth\u00E4lt aber keinen Wert.
+batchPropertyEmpty=Die Eigenschaft {0} im Vorgang {1} ist ein Pflichtfeld, enthält aber keinen Wert.
 
 # C
 calendar.upload.error=Fehler beim Laden der Datei\:
-calendar.upload.isEmpty=Es wurde keine Datei \u00FCbertragen.
+calendar.upload.isEmpty=Es wurde keine Datei übertragen.
 calendar.upload.missingMandatoryElement=Ein erforderliches XML-Element konnte nicht gefunden werden.
 calendar.upload.missingMandatoryValue=Ein erforderlicher Wert konnte nicht gefunden werden.
 calendar.upload.overlappingDateRanges=Der Erscheinungsverlauf konnte nicht importiert werden, da sich Datumsbereiche von Bl\u00F6cken \u00FCberlappen\:
@@ -98,7 +98,7 @@ kitodoScript.importProcesses.indir.isNull=Parameter ''indir'' fehlt!
 kitodoScript.importProcesses.indir.isNoDirectory=''indir'' ist kein existierendes Verzeichnis!
 kitodoScript.importProcesses.indir.cannotExecute=Der Inhalt des Importverzeichnisses kann nicht abgerufen werden (fehlende Berechtigung \u201Cexecute\u201D)
 kitodoScript.importProcesses.missingChildren=Vorgang {0} enth\u004E Verweise auf fehlendene Kindvorg\u004Enge {1}
-kitodoScript.importProcesses.noProcessTitleRule=Regelsatz gibt keine Bildungsvorschrift f�r den ''processTitle'' f�r Division {0} an!
+kitodoScript.importProcesses.noProcessTitleRule=Regelsatz gibt keine Bildungsvorschrift für den ''processTitle'' für Division {0} an!
 kitodoScript.importProcesses.project.isNull=Parameter ''project'' fehlt!
 kitodoScript.importProcesses.project.isNoProjectID=''project'' ist keine g\u00FCltige Projekt-ID
 kitodoScript.importProcesses.project.noProjectWithID=Ein solches Projekt existiert nicht!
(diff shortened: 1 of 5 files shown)


thomaslow:
@solth Looks good. Another idea would be to check the whole file with a regular expression containing only allowed characters for the respective language?

stweil (commented):
The changes for the property files look good. I suggest to apply them in a separate commit.

But what is the purpose of the new Java test code? Don't we want to catch any new contribution which uses a wrong encoding (!= UTF-8), no matter whether it is a property or any other non binary file? It's also sufficient to run the encoding test for files which are affected by a push or a pull request.

stweil:
Would this [workflow which checks all files in pull requests or push operations](https://github.com/stweil/kitodo-production/actions/workflows/check_file_encodings.yml) be better?

henning-gerhardt:
> Would this [workflow which checks all files in pull requests or push operations](https://github.com/stweil/kitodo-production/actions/workflows/check_file_encodings.yml) be better?

You can use `file` to detect the file content encoding but you must extend the usage of used `file` parameters. Quoting form the manual page of [file](https://www.man7.org/linux/man-pages/man1/file.1.html):

```
file tests each argument in an attempt to classify it.  There are three sets of tests, performed in this order: filesystem tests, magic tests, and language tests.  The  first test that succeeds causes the file type to be printed.
```

So, you must at least use the parameter `-k` or `--keep-going` to not to stop on the first result.  The sub results should / must be evalated too as after one encoding an other encoding in the same file can occur. Quoting the man page:

```
     -k, --keep-going
             Don't stop at the first match, keep going.  Subsequent matches will be have the string ‘\012- ’ prepended.  
             (If you want a newline, see the -r option.)  The magic pattern with the highest strength (see the -l option) 
             comes first.
```

You should even know how many bytes on default are read by file to determinate the encoding and file type. Quoting:

```
     -P, --parameter name=value
             Set various parameter limits.

                   Name         Default    Explanation
                   bytes        1048576    max number of bytes to read from file
                   elf_notes    256        max ELF notes processed
                   elf_phnum    2048       max ELF program sections processed
                   elf_shnum    32768      max ELF sections processed
                   encoding     65536      max number of bytes to scan for encoding evaluation
                   indir        50         recursion limit for indirect magic
                   name         50         use count limit for name/use magic
                   regex        8192       length limit for regex searches
```

Only the first `65.536` bytes of a file are used for encoding detecting. Our files - even only the message resource files - are larger than this limit, so you must extend the limits to get the whole file and not only the first 65.536 bytes.

solth:
> But what is the purpose of the new Java test code?

The main intention was to provide an automatic way to detect when the encoding of the property files has been changed to something that results in broken special characters being displayed in the frontend and thus would resolve #6331. The purpose was _not_ to detect encoding changes to _any_ source code files, as in my memory that didn't cause any otherwise undetected problems in the past (if the source files contain non-ASCII special characters - outside of String literals, for example in variable names - the whole compilation of the code will probably fail if the encoding is changed to something incompatible with UTF-8). 

> @solth Looks good. Another idea would be to check the whole file with a regular expression containing only allowed characters for the respective language?

This is a good suggestion and would probably be a better alternative to my implementation, since it would cover the whole files and not just sample messages from them. Do you know if there are predefined regular expressions covering all valid characters for specific languages?

> Would this [workflow which checks all files in pull requests or push operations](https://github.com/stweil/kitodo-production/actions/workflows/check_file_encodings.yml) be better?

Yes, that is probably a better way to detect file encoding changes in _any_ source file, and thus would cover #6324.

thomaslow:
> This is a good suggestion and would probably be a better alternative to my implementation, since it would cover the whole files and not just sample messages from them. Do you know if there are predefined regular expressions covering all valid characters for specific languages?

Not really, no. I googled a bit and found this: [LocaleData.getExemplarSet](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/util/LocaleData.html#getExemplarSet-int-int-) of the ICU4J library, which provides all valid characters of a language. But I have not used this library before. For German, English and Spanish it is probably not that difficult to specify regular expressions manually.

stweil:
PR #6341 adds a CI action which checks the whole file (all which are affected by a push or pull request).

PR #6341, opened 2024-12-06
Add new GitHub workflow which checks file encodings

stweil:
It checks whether files related to a push or pull request contain encodings which are not UTF-8.

The workflow was created with Claude.AI.

### DIFF
--- a/.github/workflows/check_file_encodings.yml
+++ b/.github/workflows/check_file_encodings.yml
@@ -1,0 +1,48 @@
+name: Check File Encodings
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - '**'
+
+jobs:
+  check-encoding:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+
+      - name: Check file encodings
+        run: |
+          #!/bin/bash
+          set -euo pipefail
+          
+          EXIT_CODE=0
+          
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            # Skip if file is binary or doesn't exist (was deleted)
+            if [[ ! -f "$file" ]] || [[ -z "$(grep -Il '.' "$file")" ]]; then
+              continue
+            fi
+            
+            # Try to detect encoding using file command
+            encoding=$(file -i "$file" | grep -oP "charset=\K.*")
+            
+            if [ "$encoding" != "utf-8" ] && [ "$encoding" != "us-ascii" ]; then
+              echo "::error file=${file}::File is not UTF-8 encoded (detected: ${encoding})"
+              EXIT_CODE=1
+            else
+              # Double-check with iconv
+              if ! iconv -f utf-8 -t utf-8 "$file" > /dev/null 2>&1; then
+                echo "::error file=${file}::File contains invalid UTF-8 sequences"
+                EXIT_CODE=1
+              fi
+            fi
+          done
+          
+          exit $EXIT_CODE


stweil:
Examples: https://github.com/stweil/kitodo-production/actions/workflows/check_file_encodings.yml.

stweil:
A manual test which checks all files in the current directory structure can be done with similar bash code:
```bash
for file in $(find * -type f); do
  # Skip if file is binary or doesn't exist (was deleted)
  if [[ ! -f "$file" ]] || [[ -z "$(grep -Il '.' "$file")" ]]; then
    continue
  fi

  # Try to detect encoding using file command
  encoding=$(file -i "$file" | grep -oP "charset=\K.*")

  if [ "$encoding" != "utf-8" ] && [ "$encoding" != "us-ascii" ]; then
    echo "::error file=${file}::File is not UTF-8 encoded (detected: ${encoding})"
    EXIT_CODE=1
  else
    # Double-check with iconv
    if ! iconv -f utf-8 -t utf-8 "$file" > /dev/null 2>&1; then
      echo "::error file=${file}::File contains invalid UTF-8 sequences"
      EXIT_CODE=1
    fi
  fi
done
```

solth:
This does indeed seem to work. If I understand the script correctly, it is supposed to output the _name_ of the file with the wrong encoding, though, which would be very helpful to resolve the issue. That doesn't seem to work however, as the only output I get in the log of the workflow is
`Error: File is not UTF-8 encoded (detected: ebcdic)`

stweil:
In your test case https://github.com/solth/kitodo-production/actions/runs/14310759081 I see this:
```
check-encoding: Kitodo/src/test/java/org/kitodo/selenium/MetadataST.java#L0
File is not UTF-8 encoded (detected: ebcdic)
```

.github/workflows/check_file_encodings.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  check-encoding:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
```
stweil, line 18: Meanwhile v46.0.4 was released. I did not test this newer version.

solth:
> In your test case https://github.com/solth/kitodo-production/actions/runs/14310759081 I see this:
> 
> ```
> check-encoding: Kitodo/src/test/java/org/kitodo/selenium/MetadataST.java#L0
> File is not UTF-8 encoded (detected: ebcdic)
> ```
> 
> There is even a direct link to the buggy file.

Yes, thank you, I found the message after your response.

solth:
This looks good and seems to work as intended. I am just wondering if we perhaps need a disclaimer in the file itself that it was created by ClaudeAI, since we do not have an official stance about the usage of code generating AI in the project, yet.
@kitodo/kitodo-board would you mind commenting on this?

stweil:
Question:
```
May I reuse code which was produced by Claude.AI or another AI in my code? Do I have to add an attribution?
```

ClaudeAI gives this information:
```
Generally, code produced by AI systems like Claude can be reused in your projects. Here's what you should know:

For Anthropic's Claude specifically:
- There are no explicit requirements to attribute code generated by Claude
- Anthropic does not claim copyright on outputs generated by Claude

However, a few considerations:

1. **Best practice**: Adding a comment indicating the code was AI-generated is considered good practice, though not strictly required

2. **Legal landscape**: The copyright status of AI-generated code is still evolving legally in many jurisdictions

3. **License considerations**: If you're integrating AI-generated code into a project with a specific open-source license, check if that license has attribution requirements

4. **Quality assurance**: Remember that AI-generated code should be reviewed before implementation, as it may contain errors or security issues

5. **Terms of service**: Other AI providers may have different terms regarding the outputs of their systems

If you're using the code in a commercial or public project, you might want to consult legal advice specific to your situation, especially as laws regarding AI-generated content continue to develop.
```

stweil:
I don't think that we should give credits in the code base for AI contributions as long as they are short and more or less trivial. Most software developers reuse code snippets from discussion forums like Stack Overflow and others ever since such forums exist, and they also contribute their own code snippets there. There is currently no AI which invents completely new code, so it's like an improved web search.

stefanCCS:
I agree with @stweil :
I think, that we should not mentioned anything about any help of any AI tool in general.
I also think, this is "just" a better web search (as @stweil  mentioned).

solth:
Merged as @kitodo/kitodo-board agrees that no special disclaimer about the usage of AI generated code is required.

PR #6360, opened 2025-01-02
Do not update results while editing filter

### LINKED ISSUE #6348: Background search requests while editing search filter
**Describe the bug**
While editing / changing the current used filter a lot of search requests are started in the background and may display results which are not wanted or need.

1) First background search is started when the editing of the current used filter is started. If this search is done you can start editing the used filter. 

2) Even while you type the new filter or change the filter searches are started and results are displayed or even an error message that the generated query results in a search error. 

3) If you are finish the editing nothing is happening if you end the edit with pressing the enter or tab key. You must lost the focus of the filter search that the new filter search is started and the results are displayed.

**Expected behavior**
Changing the used filter should start a search if you finish the edit of the filter. Finishing should be may enter key and / or tab key or when the filter input field lost its focus.

**Release**
3.7.1, 3.8.0-RC1, master branch

oliver-stoehr:
With this change the result list is no longer updated while the user clicks on a filter to edit it. The result list is updated when the user hits enter or leaves the input field.

Resolves #6348.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/filter.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/filter.js
@@ -8,7 +8,7 @@
  * For the full copyright and license information, please read the
  * GPL3-License.txt file that was distributed with this source code.
  */
-/* globals removeFilter, updateSuggestions, submitFilters, filterKeydownEvents */
+/* globals removeFilter, removeFilterForEdit, updateSuggestions, submitFilters, filterKeydownEvents */
 
 /* Define identifiers used to select elements */
 const FILTER_INPUT_FORM = "#filterInputForm";
@@ -192,7 +192,7 @@ $(document).ready(function () {
             let filter = target.siblings(".plainFilter").length ? target.siblings(".plainFilter").first().text()
                 : target.find(".plainFilter").first().text();
             $(FILTER_INPUT).val(filter);
-            removeFilter([{name: "plainFilter", value: filter}]);
+            removeFilterForEdit([{name: "plainFilter", value: filter}]);
             $(FILTER_INPUT).focus();
         }
     });
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/filterMenu.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/filterMenu.xhtml
@@ -22,6 +22,10 @@
                      update="processesTabView:processesForm:processesTable
                              parsedFiltersForm:parsedFilters
                              processCount"/>
+    <p:remoteCommand name="removeFilterForEdit"
+                     action="#{ProcessForm.filterMenu.removeFilter}"
+                     oncomplete="setFilterInputPadding()"
+                     update="parsedFiltersForm:parsedFilters"/>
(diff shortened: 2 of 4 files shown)


solth (approved):
Thanks for this fix. Works as advertised and code looks good! Would you mind opening a correspoding backported pull request against the [3.8 branch](https://github.com/kitodo/kitodo-production/tree/3.8.x)?

henning-gerhardt (commented):
Thanks for your changes. It is now better but I have still one issue:

If I start editing the filter 3 seconds after I entered / changed something a search is executed in the background too but without displaying the search results.

I will try to capture this in a video.

henning-gerhardt:
https://github.com/user-attachments/assets/df360597-eae7-4f36-909a-1985f4404f3a

In top you can see the navigation through the used filter and on the bottom the access to the ES through the Apache. There are 3 search requests until I changed the task / step status from "inwork" to "open" to get new results. After I left the focus of the filter the "open" search is executed (again) and the results got shown.

solth:
@henning-gerhardt can you verify that these additional three requests are related to the filter being manipulated? If they also occur on the normal process list page, without involving the filter, that is a different issue and isn't related to the changes in this pull request or the filter in general, I think.

henning-gerhardt:
I can confirm that this 3 requests are done while editing the filter. There was no other user logged in or other process in background running. Maybe this behaviour is existing a longer time already but I noticed it now after the changes was introduced.

henning-gerhardt:
I just testet this "new" behaviour on 3.7.2: beside the issues there which should be fixed with this pull requests on 3.7.2 is a search even executed after 3 seconds after the last keyboard interaction. So this new found behaviour is existing even without this change.

oliver-stoehr:
@henning-gerhardt The request you observed is correct: When typing in the filter input, the value is sent with a 1000 ms delay to server. This is not a full search but updates the suggestions displayed below the filter input.

henning-gerhardt:
> @henning-gerhardt The request you observed is correct: When typing in the filter input, the value is sent with a 1000 ms delay to server. This is not a full search but updates the suggestions displayed below the filter input.

I did not know that but would explain it. I did not use the suggestion until now. So it is not a bug it is more a feature.

solth:
> This is not a full search but updates the suggestions displayed below the filter input.

@oliver-stoehr Do you think it would be easy to suppress the loading screen for these partial requests? In that case perhaps that little improvemet could be added to this PR. If it prooves to be more complex, it should be resolved in a different PR, in my opinion.

oliver-stoehr:
@solth I cannot reproduce the loading screen on my local dev system but that might be because my system is quite empty (and fast).
Currently the loading screen is activated for all requests on the processes page and I'm not sure if I can change that easily.
However the display of the loading screen is (on all pages) delayed for 500 ms. This helps to prevent a "flickering" for very fast requests. If the whole request is finished within 500 ms the loading screen is not shown at all. 
I could easily increase the global delay that the loading screen is also not displayed on longer requests, but this would also affect _all_ requests in Kitodo.

solth:
Thanks for the explanation. In that case I think we should tackle the loading screen topic separately.

henning-gerhardt (approved):

PR #6174, opened 2024-08-07
Create xlsx Excel file

BartChris:
resolves https://github.com/kitodo/kitodo-production/issues/6172
This PR changes the used classes in order to create xlsx files when exporting an Excel file to be compatible with newer Excel versions.  

I am using SXSSFWorkbook to generate Excel and PDF files, which has a lower memory footprint by flushing intermediary files to disk. See: https://www.baeldung.com/java-apache-poi-workbook-evaluation 

This will hopefully enable exporting even longer lists of processes to Excel in the future.

The dependency issues described in this PR were resolved by https://github.com/kitodo/kitodo-production/pull/6362

PR #6362, opened 2025-01-06
Update plexus-archiver and poi dependency

BartChris:
While working on https://github.com/kitodo/kitodo-production/pull/6174 i encountered the problem that the current versions of plexus-archiver and apache-poi require different versions of Apache commons-compress. (https://commons.apache.org/proper/commons-compress/) 
Via plexus-archiver a older version (1.23.0) of commons-compress is installed as transitive dependency which seems to be incompatible with the installed version of apache poi. In addition to that there are  vulnerabilities for this version of commons-compress (https://mvnrepository.com/artifact/org.apache.commons/commons-compress)

To adress both the incompatbility as well as the security issues, this PR updates apache-poi and plexus-archiver to their latest version. I have not yet discovered any issues by doing so. PDF and Excel file generation are still working.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
         <myfaces.version>2.3.10</myfaces.version>
         <mysql.version>8.2.0</mysql.version>
         <pdfbox.version>3.0.3</pdfbox.version>
-        <poi.version>5.2.5</poi.version>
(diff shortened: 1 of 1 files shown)


stweil (approved):
Thanks for this update to the latest releases.

PR #6364, opened 2025-01-09
Update bytebuddy dependency for Mockito

matthias-ronge:
Mockito needs a newer version of ByteBuddy if there is a recent JVM.

See https://github.com/mockito/mockito/issues/3328

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,7 @@
         <maven-jar-plugin.version>3.3.0</maven-jar-plugin.version>
         <maven-war-plugin.version>3.4.0</maven-war-plugin.version>
         <mockito.version>5.9.0</mockito.version>
+        <byte-buddy.version>1.14.17</byte-buddy.version>
         <properties-maven-plugin.version>1.2.1</properties-maven-plugin.version>
         <selenium.version>3.141.59</selenium.version>
         <spring.security.version>5.7.13</spring.security.version>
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
We are still on Java 11 and want to switch to Java 17. The mentioned issue is for Java 21+ which will be used in a further future. Why should we now introduce a "bugfix" which ist needed in the future and may only relevant for now? If we are using a most recent version of Java then this would be a current real solution but as we are miles away from this I would not recommend this change.

solth:
I agree with @henning-gerhardt. Kitodo currently only supports Java 11, so that is the Java version that should be used for development environments as well. Introducing a new dependency for an as of yet unsupported newer version of Java might have unwanted side effects for the current version.

stweil:
Citing the link mentioned above: "The code was successfully run on JDK 11,17".

Isn't Mockito only relevant for the unit tests, so unwanted side effects would be directly visible? That's why I think the risk is low (and reverting this change if required would be easy). Sooner or later the change will be needed, so why not now?

henning-gerhardt:
> Isn't Mockito only relevant for the unit tests,

Mockito can be used in all kind of tests.

> so unwanted side effects would be directly visible?

Depending on how good or a bad a test is written this could be visible

> That's why I think the risk is low (and reverting this change if required would be easy). Sooner or later the change will be needed, so why not now?

If Mockito have some issues with Java 21+ which would be solved in the Mockito project (maybe by rising the dependency versions) and we still miles away from Java 21+ why we should add a bugfix which is not needed now?

matthias-ronge:
I am setting up a new computer and installed the latest software of everything, Java, Maven, Eclipse, Firefox, MariaDB, …

I am unhappy to run any of my software on old Java version only because of a fixable bug. And I only have _one_ `JAVA_HOME` (at least I don’t know otherwise) to set the Java location for _everything_ on the PC. And Java _should_ be backward compatible.

matthias-ronge:
Solution (for documentation):

Create file `%userprofile%\mavenrc_pre.bat` with the line:
`SET JAVA_HOME=...`

PR #6366, opened 2025-01-10
Fix "path must include project and resource name" in Eclipse

matthias-ronge:
Eclipse is unwilling to resolve the  directory '../config-local' because it is located outside the module's sub-folder. Using the Maven variable instead fixes it.

### DIFF
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -145,6 +145,9 @@
               that means first entry will be added as last resource and may override other resources -->
             <resource>
                 <!-- ATTENTION! we need the config-local declaration at first cause it shall be placed as last resource in the classpath -->
+                <!-- See also:
+                       • https://github.com/apache/maven/pull/840
+                       • https://github.com/apache/maven/pull/1061          -->
                 <directory>${maven.multiModuleProjectDirectory}/config-local</directory>
             </resource>
             <resource>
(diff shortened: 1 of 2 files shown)


solth:
I read through a long thread discussing the `multiModuleProjectDirectory` property (see https://github.com/apache/maven/pull/840 and https://github.com/apache/maven/pull/1061). 

It seems the property `maven.multiModuleProjectDirectory` is only meant for internal purpose and not intended to be used in other projects. It seems Maven 4 will contain a new property that should be used instead, but for now I can live with using `multiModuleProjectDirectory` if that resolves the IDE problems for you, @matthias-ronge 
I tested the change locally and Kitodo compiles and runs correctly on my system as well. I would perhaps prefer that you add a comment to the variable pointing to the discussion mentioned above, though. 

Before merging this pull request I would also like to know if it works in other environments as well. (@henning-gerhardt & @stweil , any thoughts about this?)

henning-gerhardt:
The change did not introduce a regression in my usage scenarios: building the application (development and non-development version) on console with mvn and executing the application (development version) inside the IDE.

Like @solth already mentioned: on both places should be added a hint that this solution is maybe only working with Maven 3.x and may not work with Maven 4.x. Maybe the hint should even include a link to this issue / pull request so it is understandable in the future why this change was made in the past.

solth (approved):

PR #6369, opened 2025-01-16
Prevent potential NullPointerExceptions

solth:
Prevent some potential `NullPointerException`s reported by CodeQL by adding corresponding null checks.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -538,12 +538,14 @@ public class StructurePanel implements Serializable {
 
     private void restoreSelection(String rowKey, TreeNode parentNode) {
         for (TreeNode childNode : parentNode.getChildren()) {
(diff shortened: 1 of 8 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -538,12 +538,14 @@ public void show() {
 
     private void restoreSelection(String rowKey, TreeNode parentNode) {
         for (TreeNode childNode : parentNode.getChildren()) {
```
stweil, line 540: Can `parentNode` have children which are Null?

solth: Potentially, yes, because `children` is just a list of `DefaultTreeNode`,  itself an implementation of the interface `TreeNode`, and instances of every non-primitive type in Java can be `null`. Whether this can actually be the case in the Kitodo context is a different story, but shouldn't be the deciding factor for adding null checks, IMHO.

stweil (approved):
I wonder whether all these null checks are really needed. The CodeQL reports could also mean that there were too many of them in the old code. But at least the changes won't harm.

Issue #5336, opened 2022-08-31
Slow process template page access


henning-gerhardt:
**Describe the bug**
The first access of the process template page is slow and there are only 10 process templates to be shown - i guess the time for this is 30 seconds. I don't know if the time depends on how many projects and indirect processes are assigned to a process template or which other reason is responsible to this slow displaying. 
If you later - or in the same login session - open the process template page again, then you get the results a lot faster. After you logged out and logged in again the display speed is on the first access slow again.

**To Reproduce**
1. Create some process templates and assign them to processes
2. Logout / Login
3. Open the process template page
4. Wait a long time to see all process templates

**Expected behavior**
Displaying of process templates should not take "ages" to get displayed.

**Release**
3.4.4-SNAPSHOT

BartChris:
It seems like the function `isTemplatUsed` is called four times

https://github.com/kitodo/kitodo-production/blob/f58592bfe58d58bfc6aa72399574cbcfd3102e65/Kitodo/src/main/webapp/WEB-INF/templates/includes/projects/templateList.xhtml#L84

https://github.com/kitodo/kitodo-production/blob/f58592bfe58d58bfc6aa72399574cbcfd3102e65/Kitodo/src/main/webapp/WEB-INF/templates/includes/projects/templateList.xhtml#L107

for every template in the list. Which would probably mean 40 database queries in your case:

https://github.com/kitodo/kitodo-production/blob/f58592bfe58d58bfc6aa72399574cbcfd3102e65/Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java#L414

BartChris:
I am wondering if this flag ("is_used_by_processes") should just be a property of the templatet, set to NULL for existing templates. Then this has to be only calculated once and then stored in the database. Whenever a process is deleted it could be checked, if the template is still referenced by a process and the property could be changed if necessary. 
This would duplicate information which already contained in the database, but it would speed up the use case.

henning-gerhardt:
If getting this information from the database (template is used by processes or not) instead of ElasticSearch you did not need such a field as this information can be retrieved through a good query (left or right join, depending on which side you start the query) to the database - which is possible through Hibernate as used database layer too.

matthias-ronge:
I just came over this again and saw that it is loading _all_ processes for each comment or batch. But I don’t understand why.
```
[...]

PR #6377, opened 2025-01-23
Speed up process template page

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/5336

Using a custom HQL-query should speed up the retrieval of the process templates significantly. We only have to compute the template usage once and reuse it in the view.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
@@ -15,12 +15,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.text.MessageFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
```diff
@@ -46,7 +49,7 @@
 import org.kitodo.production.workflow.model.Converter;
 
 @Named("TemplateForm")
-@SessionScoped
+@ViewScoped
```
BartChris, line 52: Switch to Viewscoped to that changes to the templates are reflected in the process template list.

henning-gerhardt (approved):

BartChris:
Backport: https://github.com/kitodo/kitodo-production/pull/6401

PR #6379, opened 2025-01-24
Add WebListener annotation to SessionService

matthias-ronge:
Issue #6372

PR #6047, opened 2024-04-22
Improve handle sessions

matthias-ronge:
- move logout handling from version management to session management (clean-up)
- log warnings if logout fails

PR #6380, opened 2025-01-24
Logout idle user automatically when HTTP session expires

### LINKED ISSUE #4480: Metadata editor does not notice logout
After lunch, I came back, had left the metadata editor open. I can click around but nothing really happens:

Animation

Judging by the outputs in the browser console, the Tomcat seems to want to perform a redirect on the login screen, but that does not happen:

Screenshot

**Goal:** When the session is gone, the metadata editor should bring the user to the start page.

### LINKED ISSUE #6372: Handling of expired sessions bugged
**Describe the bug**
The changes of #6047 have broken the handling of expired http sessions. This seems to stem from the `sessionDestroyed` method being moved from `KitodoVersionListener` to `SessionService`. The later is not annotated as `WebListener` (as `KitodoVersionListener` is). Adding the annotation alone does not resolve the issue, but moving the `sessionDestroyed` method back to `KitodoVersionListener` does, so we probably need a dedicated `HttpSessionListener` extension to handle expiring sessions, instead of (mis-)using the service class `SessionService` or the `KitodoVersionListener` for this purpose.

**To Reproduce**
1. Log into Kitodo with one user account in one browser (for example Chrome)
2. Close the browser (without logging out of Kitodo)
3. Log into Kitodo with a different user account in a different browser (for example Firefox)
4. After http session expires (check `web.xml` for session timeout value) check the user list in the second browser
5. See error that first user - whose http session is expired - is still displayed as logged into the system on the user page - which also results in his metadata locked processes not being released!

**Expected behavior**
User sessions should expire on http session timeout even if the user does not have an open Kitodo window in the browser.

**Release**
3.8, current main branch

solth:
This pull request restores the correct handling of HTTP sessions being destroyed and thus fixes #6372 . 

Expired HTTP sessions were previously handled by the `KitodoVersionListener`, which was the wrong place for this particular event handling and was then supposed to be moved to `SessionService` by #6047, but that pull request was missing a required annotation and therefore broke the handling of expired sessions. 

#6379 adds the missing annotation to `SessionService`, but a service class is the wrong place for this in my opinion, so I opted to add a dedicated `CustomHttpSessionListener` that extends `HttpSessionListener` and does this job instead.

This pull request also adds a message shortly before the user is logged out automatically, to inform him about the pending, automatic logout, that contains a timer with the remaining time until the automatic logout:

Bildschirmfoto 2025-01-24 um 13 25 56

This makes use of the PrimeFaces [IdleMonitor](http://www.primefaces.org/showcase/ui/misc/idleMonitor.xhtml) component and the [Timer](https://www.primefaces.org/showcase-ext/views/timer.jsf) component of PrimeFaces Extensions.

By also clearing the metadata locks of processes opened by a user in the metadata editor upon automatic logout, this pull request resolves #4480 as well.


henning-gerhardt (approved):
Changes looking good and are working. Maybe the small issues (missing semicolon) raised by Codacy could be fixed.

henning-gerhardt (approved):

PR #6382, opened 2025-01-25
[3.8] Logout idle user automatically when HTTP session expires

solth:
Backport of #6380 for Kitodo 3.8


henning-gerhardt (approved):
Changes identical to #6380 and are working.

PR #6389, opened 2025-01-29
Remove unnecessary string conversion

solth:
Cleanup some code by removing unnecessary explicit String conversions in methods were objects are converted to strings automatically.


thomaslow (approved):
Looks good to me.

PR #6396, opened 2025-02-03
Make use of centralized String constants

solth:
Extend usage of  constants defined in centralized String constants class for some common Strings.


BartChris (approved):
Application works and changes look fine.

PR #6400, opened 2025-02-06
Render panelGroup only if user has the right authority

henning-gerhardt:
Render the `h:panelGroup` for unassign tasks of users only if the current user has the authority  to do this. This condition is already existing on the `p:commandLink` but the check if a user has assigned tasks or not will be executed independent of the authority is available or not. 

With this change the user overview will be displayed a lot faster if the current user did not have this authority.

This change can be back ported to `3.8.x` if wanted.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/users/userList.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/users/userList.xhtml
@@ -137,13 +137,13 @@
                 </h:link>
 
                 <h:panelGroup styleClass="action"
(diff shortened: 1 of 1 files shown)


solth (approved):
The changes make sense and the checks for tasks assigned to individual users are skipped when the current user lacks the permission to reset assigned tasks, so works as advertised. I couldn't detect any significat boost to page load times, though, but that is probably down to the limited number of users and other database objects in my test system.

solth:
> This change can be back ported to `3.8.x` if wanted.

Please provide a backport if you don't mind.

PR #6401, opened 2025-02-06
[3.8] Speed up process template page

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6377


henning-gerhardt (approved):
Code is identical to #6377 and works in 3.8.x.

PR #6403, opened 2025-02-06
Keep pagination scope when changing pagination type

solth:
Fixes #6044

PR #6404, opened 2025-02-06
[3.8] Keep pagination scope when changing pagination type

solth:
Backport of #6403 for Kitodo.Production 3.8.


andre-hohmann (approved):
Thanks a lot! From the users' perspective, it works fine.
All tested combinations are possible without resetting the value in "Pagination scope" or other elements.

henning-gerhardt (approved):

PR #6406, opened 2025-02-07
Replace Matadaten with Metadaten

henning-gerhardt:
Found while using and testing 3.8.x


thomaslow (approved):
Sorry. My mistake. Looks good.

PR #6355, opened 2024-12-19
[3.8] Display error message when saving workflow with existing title

solth:
Backport of #6354 for Kitodo.Production 3.8.x

PR #6408, opened 2025-02-07
Fix error with empty workflow svg and error when saving empty workflow

### LINKED ISSUE #6290: Empty .svg worfklow file, if workflow saved without changes
**Describe the bug**
If an existing workflow is opened for editing via Kitodo and then saved again without making any changes to it, the corresponding workflow .svg will end up being 0 bytes in size and therefore empty/broken.
This only affects the svg image file, the workflow itself remains unchanged. (The file is only rewritten, which can be recognised by the modification date of the file)

**To Reproduce**
1. Go to 'Project' --> 'Workflows'
2. Open an existing workflow
3. Do NOT make any changes to the workflow and save it
4. Workflow is saved without an error message
5. Determine that the associated .svg file is empty / 0 bytes in size
5. 1. Apart from the filesystem, you will also notice this in the corresponding process template when you unfold it

**Expected behavior**
Saving workflow without changes should also result in a functioning .svg file

**Workaround**
1. Open the Workflow
2. CHANGE anything in the workflow (for example remove and readd "end")
3. Save the workflow
4. A new correct svg file should be generated

**Screenshots**

**Release**
Tested with 3.7.1

 - OS: Kubuntu 22.04
 - Browser Firefox 
 - Version 132.0 (64-bit)

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6290
Adresses https://github.com/kitodo/kitodo-production/pull/6355#issuecomment-2627106319

Additional changes
- Switch to ViewScoped to prevent having old workflows in the session which migh overwrite another workflow
- Deletion of old workflow files when the title of a workflow get changed

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowForm.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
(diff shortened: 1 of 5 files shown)


thomaslow:
@BartChris Are you planning to continue working on this? I am working on refactoring all @SessionScoped form classes, which includes the WorkflowForm, see #6456.

BartChris:
Sounds good i would say on first take, especially if it simplifies the logic on form setup. I think this check for the flash context is a little bit cumbersome.

BartChris:
I will maybe keep it as draft, so this does not get lost, but let you do your work uninterrupted.

PR #6409, opened 2025-02-10
`Data corrupted` message fix

### LINKED ISSUE #6041: `Data corrupted` message when moving a page to an unsaved structure element in gallery
**Describe the bug**
When moving a page via drag'n'drop in the gallery from "unstructured media" to a new structure element which has not been saved, yet, a popup dialog about corrupted data appears (see screenshot below). The same does _not_ happen when moving a page to an unsaved structure element in the structure tree, however. Saving the structure before the drag'n'drop action prevents the error.
Additionally, the popup dialog cannot be closed by clicking on the "Ok" button. The metadata editor page needs to be reloaded to remove the dialog.

**To Reproduce**
1. Open process containing at least one page in the metadata editor
2. Create a new structure element
3. Do _not_ save the process after creating the element!
4. Try to move the page via drag'n'drop to the newly created, unsaved structure element in the **gallery** 
5. See error

**Expected behavior**
It should be possible to move pages to an unsaved structure element in the gallery without an error. In situations where it is warranted that the dialog appears, it should be possible to close it with the "Ok" button.

**Screenshots**
Bildschirmfoto 2024-04-19 um 08 30 07

**Release**
current master, 3.7.0-SNAPSHOT

solth:
Fixes #6041 by adding missing `preserve()` call in structure panel after adding a structure.


andre-hohmann (approved):
From the users' perspective, it works fine. 
It is possible to assign unstructured pages to unsaved structure elements.

henning-gerhardt (approved):
Identical to changes in #6410

PR #6421, opened 2025-02-17
Sort newspaper years by 'ORDERLABEL'

solth:
Fixes #5821 by sorting years in a newspaper process by `ORDERLABEL` and inserting new year processes accordingly. 

**_Note:_** Does **not** guarantee year divisions of existing newspaper processes actually _have_ an `ORDERLABEL`. This has to be ensured separately.

PR #6422, opened 2025-02-17
[3.8] Sort newspaper years by 'ORDERLABEL'

solth:
Backport of #6421 for Kitodo.Production 3.8 (note that the same caveats apply!)


andre-hohmann (approved):
See: 
- https://github.com/kitodo/kitodo-production/pull/6421

henning-gerhardt (approved):

PR #6426, opened 2025-02-17
[3.8] Fix newspaper issue being displayed as 'unknown' in year process

solth:
Backport of #6425 for Kitodo.Production 3.8


BartChris (approved):
Works for 3.8 as well

PR #6432, opened 2025-02-19
Speed up process list by faster check for empty directory

BartChris:
Whily trying to use a S3 storage for the Kitodo files i noticed that the process list became very slow. S3 storage is slower than NFS or local storage of course, but we can sigificantly speed up the process lf we refactor file existence checks.

The problem is that for every process in the process list a check is triggered wether the document can be exported.

https://github.com/kitodo/kitodo-production/blob/84b049749e7c8d8f3f773e18d560f768b2d37cf4/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java#L2830-2847

This calls a function which checks if the process has images, which checks wether the directory with the original images is empty

https://github.com/kitodo/kitodo-production/blob/84b049749e7c8d8f3f773e18d560f768b2d37cf4/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java#L1451-1461

This in turn calls a complex function which does way more than returning the contents of the dir to check if the dir is empty.

https://github.com/kitodo/kitodo-production/blob/84b049749e7c8d8f3f773e18d560f768b2d37cf4/Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java#L346-357

I replaced this with an optimized file existence check, which speeds up performance significantly. (On S3 from 15 seconds to one second.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
+++ b/Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
@@ -13,8 +13,11 @@ package org.kitodo.production.model;
 
 import java.io.File;
 import java.io.FilenameFilter;
+import java.io.IOException;
 import java.lang.reflect.UndeclaredThrowableException;
 import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
@@ -302,6 +305,23 @@ public class Subfolder {
         URI uri = getUri(canonical);
         return new File(uri.getPath()).isFile() ? Optional.of(uri) : Optional.empty();
     }
+
+    /**
+     * Checks if the folder determined by determineDirectoryAndFileNamePattern() is empty.
+     * Retrieve the directory URI and lists its contents.
+     * It stops checking as soon as the first file is found.
+     *
+     * @return true if the folder is empty, false if it contains at least one file
+     */
+    public boolean isFolderEmpty() {
+        URI dir = determineDirectoryAndFileNamePattern().getLeft();
+        try (Stream<Path> entries = Files.list(Paths.get(dir))) {
+            return entries.findFirst().isEmpty(); // Stop after first file is found
+        } catch (IOException e) {
+            return false;
+        }
+    }
+
 
     /**
      * Returns a map of canonical file name parts to URIs with all files
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -1453,11 +1453,8 @@ public class FileService {
      * @return whether given URI points to empty directory or not
      */
     public static boolean hasImages(Process process, Folder generatorSource) {
-        if (Objects.nonNull(generatorSource)) {
-            Subfolder sourceFolder = new Subfolder(process, generatorSource);
-            return !sourceFolder.listContents().isEmpty();
-        }
-        return false;
+        Subfolder sourceFolder = new Subfolder(process, generatorSource);
+        return !sourceFolder.isFolderEmpty();
     }
 
     /**


henning-gerhardt:
In common I'm unsure if your way of change is the right way. The file management module was implemented for accessing files on local storage. For other access types like S3 the file management interface should be implemented for S3 access and must be replaced on runtime with the version of the local storage as you can not have two implementations of the same interface (well known but never solved issue). This was the plan while development the 3.x version. As I see now that this whole DFG project documentation was not got public published  - at least not here: https://github.com/kitodo/kitodo-production/wiki/Developer-Documentation-Kitodo.Production-3.x

So your change can maybe break this general attempt to add a separation layer between but it if even possible that this layer was already broken without your change.

I will try your changes and report back if there are issues or anything else.

BartChris:
> In common I'm unsure if your way of change is the right way. The file management module was implemented for accessing files on local storage. For other access types like S3 the file management interface should be implemented for S3 access and must be replaced on runtime with the version of the local storage as you can not have two implementations of the same interface (well known but never solved issue). This was the plan while development the 3.x version. As I see now that this whole DFG project documentation was not got public published - at least not here: https://github.com/kitodo/kitodo-production/wiki/Developer-Documentation-Kitodo.Production-3.x
> 
> So your change can maybe break this general attempt to add a separation layer between but it if even possible that this layer was already broken without your change.
> 
> I will try your changes and report back if there are issues or anything else.

Sorry for creating confusion here. I am not using S3 natively but mounting it using rclone and FUSE mount. So my S3 storage acts like a normal file system. So the new code should work with all form of local or mounted Network storage and speed things up for those storages as well.

henning-gerhardt (changes requested):
I tried this changes but I can not confirm that is a speed up on local storage - I had the impression that displaying needs more time. But then major issue is that now processes are displayed as exportable which are not exportable.

Without your changes its looks like this for a few processes:

Screenshot_2025-02-20_09-55-21

With your changes the processes in the middle are now exportable but they did not contain any images nor other media files

Screenshot_2025-02-20_09-58-13


Both processes from the audio and video test project did not contain any files but shown before and after as exportable. But this is an other issue.

BartChris:
Thanks for your review. I will inspect that.

BartChris:
I tried to extend the logic to maybe fix the wrong result in your case. (Folders are indicated as non empty also they are empty)
In what folders have you put the originals? Maybe my logic did not cover different cases of folder structures. I can imagine that on local storage the speed difference is small or non noticable. On my S3 based network mount the difference when navigating the process list is from "unusable slow" to "quite fast".

> I tried this changes but I can not confirm that is a speed up on local storage - I had the impression that displaying needs more time. But then major issue is that now processes are displayed as exportable which are not exportable.
> 
> With your changes the processes in the middle are now exportable but they did not contain any images nor other media files

BartChris:
> Both processes from the audio and video test project did not contain any files but shown before and after as exportable. But this is an other issue.

Only files that are used to generate **image** erivatives are considered here.

https://github.com/kitodo/kitodo-production/blob/6e9e69bbb424892bd828f904d1cb069f0a12fd85/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java#L2841-L2842

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
```diff
@@ -303,6 +306,31 @@ public Optional<URI> getURIIfExists(String canonical) {
...
+        File dir = fileService.getFile(query.getLeft());
+        if (!dir.exists() || !dir.isDirectory()) {
+            return false;  // Ensures we only check valid directories
+        }
+        try (Stream<Path> entries = Files.list(dir.toPath())) {
+            // Apply regex filtering to check only for files matching the pattern
+            return !entries
+                    .map(Path::getFileName)
+                    .map(Path::toString)
+                    .anyMatch(name -> query.getRight().matcher(name).matches()); // Stop after first match
```
henning-gerhardt, line 327: IntelliJ IDEA suggest here instead of `!entries.anyMatch()` to use `entries.noneMatch()` to use.

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
```diff
@@ -303,6 +306,31 @@ public Optional<URI> getURIIfExists(String canonical) {
...
+     * Checks if the folder determined by determineDirectoryAndFileNamePattern() is empty.
+     * Retrieve the directory URI and lists its contents.
+     * It stops checking as soon as the first file is found.
+     *
+     * @return true if the folder is empty, false if it contains at least one file
+     */
+    public boolean isFolderEmpty() {
+        Pair<URI, Pattern> query = determineDirectoryAndFileNamePattern();
+        File dir = fileService.getFile(query.getLeft());
+        if (!dir.exists() || !dir.isDirectory()) {
```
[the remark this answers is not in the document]
BartChris: I suppose you are right. I think i forget that there might be processes which do not have a folder autogenerated. 

image

I will change this to true.

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
```diff
@@ -303,6 +306,31 @@ public Optional<URI> getURIIfExists(String canonical) {
...
+            return false;  // Ensures we only check valid directories
+        }
+        try (Stream<Path> entries = Files.list(dir.toPath())) {
+            // Apply regex filtering to check only for files matching the pattern
+            return !entries
+                    .map(Path::getFileName)
+                    .map(Path::toString)
+                    .anyMatch(name -> query.getRight().matcher(name).matches()); // Stop after first match
+        } catch (IOException e) {
+            return false;
```
henning-gerhardt, line 329: Should an rised IOException interpreted as and directory with content? I'm unsure about this but it could be okay. It could be helpful if this exception got logged in trace or debug log level.

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
BartChris: I got the same recommendation. I was just wondering wether entries.noneMatch() is slower here. But maybe we should trust the IDE more.

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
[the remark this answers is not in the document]
henning-gerhardt: I changed the return statement to `true` and for my test cases it is working. Yes, we are not creating the directories through the application as this is done when the process get linked into the users home directory as we must adjust the owner and access rights to the user which is accessing this directories.

henning-gerhardt:
> > Both processes from the audio and video test project did not contain any files but shown before and after as exportable. But this is an other issue.
> 
> This seems to be intended by the current code. Only files that are used to generate **image** derivatives are considered here.

Good to know. It is different for non-image media files as for image media files but if this must be changed then not in this pull request.

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
henning-gerhardt: I noticed no differences on my local development system. Maybe you can time it on your system with an S3 in background? Maybe the first solution was faster but maybe there is no difference.

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
BartChris: IOException now marks the folder has having no images.

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
BartChris: There seem to be no real performance differences
https://stackoverflow.com/a/57779846

Kitodo/src/main/java/org/kitodo/production/model/Subfolder.java
```diff
@@ -303,6 +306,28 @@ public Optional<URI> getURIIfExists(String canonical) {
         return new File(uri.getPath()).isFile() ? Optional.of(uri) : Optional.empty();
     }
 
+    /**
+     * Checks if the folder determined by determineDirectoryAndFileNamePattern() is empty.
+     * Retrieve the directory URI and lists its contents.
+     * It stops checking as soon as the first file is found.
+     *
+     * @return true if the folder is empty, false if it contains at least one file
+     */
```
henning-gerhardt, line 315: In JavaDoc thrown exception is missing.

henning-gerhardt (approved):
Code changes looks good and are working. I did not get any speed boost but on local storage the benefits maybe to low.

BartChris:
Backport: https://github.com/kitodo/kitodo-production/pull/6435

Issue #6438, opened 2025-02-25
Metadata frame not fully available after adding new metadata in metadata editor


andre-hohmann:
**Describe the bug**
If a new metadata is added via the `+`-icon in the metadata-frame, the frame shifts, so that 
* at the bottom a white space is shown
* at top the functional buttons to add or re-import metadata is not available anymore 

It seems that it is Chrome-specific. 
Maybe a result of: 
- #5929

**To Reproduce**
1. Go to Metadata frame / `+`-icon
2. Click on `+`-icon and add a metadata which is place at bottom of the frame
3. See white space at the bottom of the frame
4. Scroll up
5. Try to find the `+`-icon

**Expected behavior**
The metadata frame should be completely visible.

**Screenshots**
Top of the frame 
(compare functional buttons in the structure-element frame)
![Image](https://github.com/user-attachments/assets/5c49c401-cb6a-4d5f-bcfe-b3863d1bbaab)

Bottom of the frame
(scrollbar is visible)
![Image](https://github.com/user-attachments/assets/88b0b615-eb29-4aba-9f13-4ff010028169)

**Release**
Version 3.8.2-SNAPSHOT

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome

PR #6439, opened 2025-02-25
Use JQuery animate for row focus

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/6438

I am not sure why this issue exists in Chrome, but since we are already using JQuery, we can replace the functionality with JQuery.animate(), which works across browsers in my tests. I am also adding a timeout, to account for UI rendering delays and some checks.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_table.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_table.js
@@ -62,20 +62,26 @@ class FocusMetadataRow {
      */
     focus() {
         this.#removeHighlights();
+        setTimeout(() => {
+            // find last metadata row matching currently selected metadata type
+            let row = $(
+                "div[id$='metadataTable'] " + // metadata table selector
+                "tr[data-prk='" + this.#lastSelectedRowKey + "'] " + // remembered metadata group
+                "label[data-metadataid='" + this.#getSelectedMetadataType() + "']" // selected metadata type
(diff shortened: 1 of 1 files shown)


thomaslow:
@BartChris I had a look at the issue and was curious what was going on exactly. I came up with another solution that also prevents the same issue in other scenarios (e.g. on the create process page). See #6441.

I can still have a look at your jQuery animate solution if you think it is improving the scroll behavior even more. 

> I am also adding a timeout, to account for UI rendering delays

I am not sure what problem is solved by adding a `setTimeout`. What kind of rendering delays did you observe?

BartChris:
> @BartChris I had a look at the issue and was curious what was going on exactly. I came up with another solution that also prevents the same issue in other scenarios (e.g. on the create process page). See #6441.
> 
> I can still have a look at your jQuery animate solution if you think it is improving the scroll behavior even more.
> 
> > I am also adding a timeout, to account for UI rendering delays
> 
> I am not sure what problem is solved by adding a `setTimeout`. What kind of rendering delays did you observe?

@thomaslow Great! I will have a look. The setTimeout change was more a "just in case" implementation, i did not notice any problems in your implementation regarding that, so we can drop it. 

Have you also seen the error message i quoted above? ("Blocked aria-hidden on an element because its descendant retained focus....")

BartChris:
Closed in favour of 
https://github.com/kitodo/kitodo-production/pull/6441

PR #6442, opened 2025-02-26
distinguish placeholder from actual values

### LINKED ISSUE #6434: Input placeholder should be distinguishable from actual content
Many input elements in Kitodo.Production display placeholders if no value is present.
These placeholders are styled in the same way as actual values.

Image
The first input "number of pages" contains no value and displays its placeholder. The second input "number of volumes" contains the actual value "0".
It is not possible to know if an input field contains a value or just displays its placeholder without clicking into the input.

The placeholder should be easily distinguishable from an actual value, e.g. by a different text color or style like italics.

clemensmach:
Fixes #6434

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
@@ -619,7 +619,7 @@ button.ui-datepicker-trigger.ui-button-icon-only::before {
 
 
 .ui-inputfield::placeholder {
-    color: var(--blue);
+    color: var(--light-blue);
 }
 
 .ui-inputfield.ui-state-hover::placeholder {


solth:
Hi @clemensmach, thanks for this pull request. If you want it to be reviewed please set it "Ready for review" with the corresponding button!

oliver-stoehr (changes requested):
Thank you @clemensmach, I like this simple solution. 
However it might be difficult to distinguish the two colors from each other, especially for users with vision impairments. Maybe you could also add a color-independent difference like
`font-style: italic`
in addition to the light blue color.

clemensmach:
>  `font-style: italic` in addition to the light blue color.

that looks even better

image

oliver-stoehr (approved):

solth:
@clemensmach would you mind opening a corresponding pull request for the 3.8 branch of Kitodo.Production?

clemensmach:
sure: #6452

PR #6443, opened 2025-02-27
Import catalog metadata from any record in the metadata editor

### LINKED ISSUE #5915: Import of structural data elements from the catalog
In Kitodo 2 importieren wir Metadaten für Strukturdatenelemente aus dem Katalog, genauer gesagt: Wir importieren Metadaten anhängender Werke aus dem K10Plus im Metadateneditor ("Strukturelemente aus Opac hinzufügen"). Dabei wird derselbe Weg für den Import gewählt wie beim Vorgang anlegen, heißt: der Import funktioniert äquivalent zur sonstigen Importkonfiguration. Diese Funktion benötigen wir auch in Kitodo 3, da wir mit Os-Stufen, also Katalogaufnahmen für Aufsätze, aus dem CBS arbeiten, und deren Metadaten PPN-basiert in den übergeordneten Kitodovorgang importieren. 

Zunächst legt ein:e Bearbeiter:in händisch das Strukturelement „Anhängendes Werk“ an und weist die Images zu (äquivalent zu einem Kapitel o.Ä.). Anschließend wird das angelegte Element angewählt, und als Suchbegriff für das Strukturelement aus dem Opac die PPN des gewünschten Katalog-Datensatzes eingegeben. Per Klick auf „Nur Metadaten“ werden die Metadaten aus dem CBS importiert. 

image

In diesem Beispiel sehen die Metadaten für das anhängende Werk nun wie folgt aus:
 
image

Unsere SBB-Kolleg:innen vermissen dieses Feature bei der Strukturdatenerfassung in der neuen Kitodoversion. Wir wünschen uns eine Implementierung in der 3er-Version.

-----

English: In Kitodo 2, we import structural data elements from the catalog, or more precisely: we import attached works from the CBS in the metadata editor ("Add structural elements from Opac"). We also need this function in Kitodo 3, as we work with Os-records from the K10Plus. My SBB colleagues sorely miss this feature in the new Kitodo version. We would like to see it implemented in the 3 version.

thomaslow:
Fixes #5915

Todo
- [ ] Discuss user interface design with community
- [ ] Improve code style
- [ ] Add selenium tests

Demonstration video:

https://github.com/user-attachments/assets/9bd2835b-8cae-41c6-bbe9-362b7356b253

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/DataEditorForm.java
@@ -49,6 +49,7 @@ import org.apache.logging.log4j.Logger;
 import org.kitodo.api.MetadataGroup;
 import org.kitodo.api.dataeditor.rulesetmanagement.FunctionalMetadata;
 import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
+import org.kitodo.api.dataformat.Division;
 import org.kitodo.api.dataformat.LogicalDivision;
 import org.kitodo.api.dataformat.PhysicalDivision;
 import org.kitodo.api.dataformat.View;
@@ -1328,8 +1329,13 @@ public class DataEditorForm implements MetadataTreeTableInterface, RulesetSetupI
      * Perform metadata update for current process.
      */
     public void applyMetadataUpdate() {
-        DataEditorService.updateMetadataWithNewValues(workpiece, updateMetadataDialog.getMetadataComparisons());
-        metadataPanel.update();
+        Division<?> division = metadataPanel.getLogicalMetadataTable().getDivision();
+        if (Objects.nonNull(division) && division instanceof LogicalDivision) {
+            DataEditorService.updateMetadataWithNewValues((LogicalDivision) division, updateMetadataDialog.getMetadataComparisons());
+            metadataPanel.update();
+        } else {
+            Helper.setErrorMessage("cannot update metadata of non-logical division");
+        }        
     }
 
     /**
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/ImportMetadataDialogSequence.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/ImportMetadataDialogSequence.java
@@ -1,0 +1,162 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.production.forms.dataeditor;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.net.URISyntaxException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+
+import javax.faces.view.ViewScoped;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+import javax.xml.xpath.XPathExpressionException;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.kitodo.api.Metadata;
+import org.kitodo.api.dataeditor.rulesetmanagement.FunctionalMetadata;
+import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
+import org.kitodo.constants.StringConstants;
+import org.kitodo.data.database.beans.Process;
+import org.kitodo.data.database.exceptions.DAOException;
+import org.kitodo.exceptions.ConfigException;
+import org.kitodo.exceptions.InvalidMetadataValueException;
+import org.kitodo.exceptions.NoRecordFoundException;
+import org.kitodo.exceptions.NoSuchMetadataFieldException;
+import org.kitodo.exceptions.ProcessGenerationException;
+import org.kitodo.exceptions.UnsupportedFormatException;
+import org.kitodo.production.forms.createprocess.CreateProcessForm;
+import org.kitodo.production.forms.createprocess.ProcessDetail;
+import org.kitodo.production.helper.Helper;
+import org.kitodo.production.helper.MetadataComparison;
+import org.kitodo.production.helper.ProcessHelper;
+import org.kitodo.production.helper.TempProcess;
+import org.kitodo.production.services.ServiceManager;
+import org.kitodo.production.services.dataeditor.DataEditorService;
+import org.primefaces.PrimeFaces;
+import org.primefaces.model.TreeNode;
+import org.xml.sax.SAXException;
+
+@Named("ImportMetadataDialogSequence")
+@ViewScoped
+public class ImportMetadataDialogSequence implements Serializable {
+
+    private static final Logger logger = LogManager.getLogger(ImportMetadataDialogSequence.class);
+
+    @Inject
+    private CreateProcessForm createProcessForm;
+
+    @Inject
+    private DataEditorForm dataEditorForm;
+
+    /**
+     * Open catalog search dialog when the user requests to import metadata from a catalog record.
+     */
+    public void importMetadata() {
+        int templateId = dataEditorForm.getProcess().getTemplate().getId();
+        int projectId = dataEditorForm.getProcess().getProject().getId();
+
+        createProcessForm.prepareProcess(templateId, projectId, null, null);
+
+        PrimeFaces.current().executeScript("PF('catalogSearchDialog').show();");
+    }
+
+    /**
+     * Perform a catalog search and open either the histlist dialog (if there are multiple search results) 
+     * or the metadata comparison dialog, if there is only a single search result.
+     */
+    public void catalogSearch() {
+        createProcessForm.getProcesses().clear();
+        createProcessForm.getCatalogImportDialog().search();
+        showMetadataComparisonDialog();
+    }
+
+    /**
+     * Open the metadata comparison dialog after a user selects a record from the hitlist dialog.
+     */
+    public void selectRecord() {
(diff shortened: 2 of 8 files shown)


thomaslow:
I merged both buttons to a single update metadata button. Depending on the current selection and information known about the process, different things happen:

- if the root structure element is selected, and a record identifier and import configuration is know for the current process, the catalog metadata is loaded without an additional search dialog (exactly like the previous behavior)

- if no record identifier is known for the process, or any other structure element is selected, the catalog search dialog opens, such that the user can search for the appropriate catalog record

- if a page or parent process is selected, the button is disabled

The updated implementation is shown in the updated demonstration video at the top.

solth:
@thomaslow Thank you very much for this pull request. This seems to be a _very_ valuable addition to Kitodo and something that has been asked for many times. I wonder why you haven't got more feedback about this. Perhaps the @kitodo/kitodo-community-board would like to comment on this proposal?

oliver-stoehr (changes requested):
This is a great addition for the metadata editor!

I would like to suggest one change:
Updating existing metadata and adding new metadata (potentially from a completely different source) are different features and it could be confusing when they are accessible through the same button.
In my opinion it would be helpful to split it into two separate buttons and use [fa-book](https://fontawesome.com/v4/icon/book) as icon for importing catalog metadata.
This is the icon we already use for the "catalog search" button (and the "data transfer" in general.)
The header of the metadata panel offers enough space for two buttons and it would be clearer if both buttons were available side by side and just disabled when not applicable.

thomaslow:
@oliver-stoehr Thank you for your review.

I understand your reasoning for having two buttons. Actually, the pull request was adding a separate button in the beginning. 

I changed it to one combined button, because the update-metadata-based-on-known-import-configuration-button was disabled most of the time anyway, and it seemed wasteful to add another button that does almost exactly the same thing (also in the backend). If you look at it "in reverse" and interpret the update-metadata-based-on-known-import-configuration function as a "convenience feature" of the new button (not asking the user to type in a PPN that is already known), the combined button makes sense to me.

The current solution was already tested and accepted by users of the SPK-Berlin. They were aware of the two slightly different functions, and there were no reports of any confusion or misunderstanding. However, I don't think they had a strong opinion about this one button vs. two buttons situation.

Personally, I think both solutions would be valid.

I can change the pull request to two buttons. However, I would like to avoid having to revert these changes again in two weeks in case someone else voices their opinion about liking the combined button. I don't know how something like this is decided usually in Kitodo. Maybe someone else can comment on this so we have a clear democratic majority decision?

@Westedt What do you think?

solth (commented):
I am an favor of separated buttons as well. The reason is that at least for the logical root element, both actions - _updating_ metadata from a linked import source and _augmenting_ existing metadata with metadata from a differenct source/catalog - are semantically different and therefor it should be possible to trigger them individually. _Updating_ metadata should stay possible for users without the knowledge of where the metadata was originally retrieved from and requiring them to select the correct catalog manually.

Also, in my opinion importing metadata into the root element of a process that does *not* have a linked import confguration, yet, should actually set the import configuration of that process to the catalog source the user actively selected, so that next time the metadata can be updated directly from the stored catalog/source via the first, dedicated "Update" button (or augmented with metadata from a different source via the separate, dedicated "Import"/"Augment" button.)

thomaslow:
> I am an favor of separated buttons as well.

Ok. I'll separate the buttons again.

> Also, in my opinion importing metadata into the root element of a process that does not have a linked import configuration, yet, should actually set the import configuration of that process to the catalog source the user actively selected, so that next time the metadata can be updated directly from the stored catalog/source via the first, dedicated "Update" button (or augmented with metadata from a different source via the separate, dedicated "Import"/"Augment" button.)

Maybe, but that would mix the functions of both buttons again. For example, if there is no import configuration set for a process yet, and the users clicks the "augment" button, but accidentally or on-purpose types in a "wrong" PPN to augment the metadata from a record that is not really the source-record, the wrong default import configuration is remembered for this process. This makes everything more complicated. There would need to be a separate option such that the user can decide whether to remember this import configuration as the default configuration. Unfortunately, this is not covered by #5915.

Kitodo/src/test/java/org/kitodo/production/services/security/SecurityAccessServiceIT.java
```diff
@@ -53,7 +53,7 @@ public void shouldGetAuthorities() throws DAOException {
         SecurityTestUtils.addUserDataToSecurityContext(user, 1);
         Collection<? extends GrantedAuthority> authorities = SecurityContextHolder.getContext().getAuthentication()
                 .getAuthorities();
-        assertEquals(174, authorities.size(), "Security context holder does not hold the corresponding authorities");
+        assertEquals(176, authorities.size(), "Security context holder does not hold the corresponding authorities");
```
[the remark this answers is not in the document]
thomaslow: I don't know exactly either. Maybe the user has multiple roles or multiple clients, and authorities are duplicated for each client or role. I haven't checked. It has been like this for while.

Kitodo/src/main/webapp/WEB-INF/templates/includes/processFromTemplate/dialogs/errorPopup.xhtml
```diff
@@ -25,7 +25,7 @@
               appendTo="@(body)"
               resizable="false"
               closable="false">
-        <h:panelGroup id="messageWrapper">
```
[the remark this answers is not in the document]
thomaslow: Yes, the id collides with another dialog, but is not referenced anywhere, so I removed it.

thomaslow:
I updated the demonstration video to show that there are now 2 buttons:
- the previous update button that only works if the process was created from an import configuration
- the new import button that can be used for any logical structure element (including the root node)

solth:
Right now the button "OPAC durchsuchen" remains disabled after entering a search ID:

Bildschirmfoto 2025-05-05 um 08 46 16

I made sure to reload the page including its cache to prevent problems stemming from outdated javascript or style resources. Did I miss something?

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/UpdateMetadataDialog.java
```diff
@@ -24,15 +35,25 @@
 import org.primefaces.PrimeFaces;
 import org.primefaces.model.TreeNode;
 
-public class UpdateMetadataDialog {
+/**
+ * Manages the dialog when a user clicks on the update metadata button.
+ * 
+ * <p>Re-imports the catalog record for a known record identifier and import configuration
+ * and viusalizes the differences between the current metadata and re-imported metadata of the
```
solth, line 42:
```suggestion
 * and visualizes the differences between the current metadata and re-imported metadata of the
```

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/UpdateMetadataImportDialogSequence.java
```diff
@@ -0,0 +1,216 @@
...
+                );
+
+                if (metadataComparisons.isEmpty()) {
+                    PrimeFaces.current().executeScript("PF('metadataUnchangedDialog').show();");
+                } else {
+                    PrimeFaces.current().ajax().update("updateMetadataDialog");
+                    PrimeFaces.current().executeScript("PF('updateMetadataDialog').show();");
+                }
+
+            };
```
solth, line 203:
```suggestion
            }
```

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/UpdateMetadataImportDialogSequence.java
```diff
@@ -0,0 +1,216 @@
...
+        createProcessForm.getCatalogImportDialog().setSearchTerm("");
+
+        PrimeFaces.current().executeScript("PF('catalogSearchDialog').show();");
+    }
+
+    /**
+     * Perform a catalog search and open either the histlist dialog (if there are multiple search results) 
+     * or the metadata comparison dialog, if there is only a single search result.
+     */
+    public void catalogSearch() {
```
solth, line 132: My IDE falsely marks this method and `selectRecord` as unused. Perhaps it would make things clearer for developers in the future if we add a comment that this method and `selectRecord` are indeed not unused, but that their method names are passed to in the `metadataEditor` and `processFromTemplate` templates?

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/UpdateMetadataImportDialogSequence.java
```diff
@@ -0,0 +1,216 @@
...
+
+                ProcessHelper.generateAtstslFields(tempProcess, Collections.emptyList(), StringConstants.EDIT, false);
+
+                metadataComparisons.addAll(
+                    DataEditorService.initializeMetadataComparisons(
+                        process, 
+                        existingMetadata, 
+                        tempProcess.getWorkpiece().getLogicalStructure().getMetadata(), 
+                        dataEditorForm.getPriorityList(),
+                        dataEditorForm.getSelectedStructure().get().getType()
```
solth, line 192: Since `dataEditorForm.getSelectedStructure()` returns an `Optional<LogicalDivision>`, should we not first check whether the division `isPresent` before calling `get()` on the optional object?

Westedt:
We actually liked the previous version with one shared button, but we are also fine with the new solution. Thanks @thomaslow !

solth:
> I made sure to reload the page including its cache to prevent problems stemming from outdated javascript or style resources. Did I miss something?

I just tried again and know it works. Perhaps my last build was not somehow corrupted or I forgot to replace some modules or reload my browser cache. Anyway: now the "OPAC durchsuchen" button is activated when "Wert" field is filled in. 👍

solth (approved):
Works very well for me and I have no further change requests for the code.

solth:
@thomaslow if you want you can accept my small suggestions regarding code changes, but they are not condition for merging this pull request. @oliver-stoehr do you approve this pull request in its current state?

oliver-stoehr (approved):

thomaslow:
@solth Thank you for your feedback. I included your code suggestions. Unfortunately, I currently cannot investigate, test or implement any changes this week and most of next week.

solth:
> Unfortunately, I currently cannot investigate, test or implement any changes this week and most of next week.

No worries, the failing build was caused by an unreliable test unrelated to your changes. I re-started the build and now it went through successfully.

Issue #6449, opened 2025-03-04
Display output of non automatic script tasks to the assigned user


lutzhelm:
## Description

Output of a non automatic script task of a process should be displayed to the assigned user of the task in the task detail tab under `pages/currentTasksEdit.jsf`.

## Related Issues

The necessary change has previously been filed as a bug in #6096.

## Expected Benefits of this Development

Expected output of script tasks will be directly visible in the Kitodo UI for users who would otherwise have to retrieve the necessary information through other means / other systems.

Users who switched to Kitodo 3 from Kitodo 2 will finally be able to use a feature they have been used to in the previous version.

## Estimated Costs and Complexity

**low** ~ less than 5 working days - I'd assume that a person familiar with the code will probably be able to implement this in a few hours, even if it would require changes e.g. to the database scheme.

BartChris:
It seems like the necessary behavior is already implemented but with a bug and with questionable settings.

1. The referenced lines

https://github.com/kitodo/kitodo-production/blob/71d3053f7698241b758857ba0f18eb901386912f/Kitodo/src/main/java/org/kitodo/production/services/command/CommandService.java#L152-L157

are only called from `runCommandAsync`, but this function is never called in Kitodo

https://github.com/kitodo/kitodo-production/blob/174e6bde7d050a73ec8f0afd19a5e3599377ea63/Kitodo/src/main/java/org/kitodo/production/services/command/CommandService.java#L140-L144

2. The relevant error handling is in the `runCommand` method

https://github.com/kitodo/kitodo-production/blob/174e6bde7d050a73ec8f0afd19a5e3599377ea63/Kitodo/src/main/java/org/kitodo/production/services/command/CommandService.java#L50-L60

3. There are problems with the method

```
 if (!commandResultMessages.isEmpty() && commandResultMessages.get(0).contains("IOException")) {
            throw new IOException(commandResultMessages.get(1));
        }
```
we are checking for a commandResultMessage at postion 0, but then access it at position on. This throws an `IndexOutOfBoundsException` when the error is thrown. In my opinion this is correct:

```
 if (!commandResultMessages.isEmpty() && commandResultMessages.get(0).contains("IOException")) {
            throw new IOException(commandResultMessages.get(0));
        }
```

We are then checking wether the exception contains "IOException". What that means is that errors are only propagated to the UI if they contain the string IOException. To test that, i changed the above method and wrote the following script

```
#!/bin/sh
[...]

PR #6450, opened 2025-03-04
Fix script output

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6449

This Pull Request tries to apply a minimal fix for the specific problem in https://github.com/kitodo/kitodo-production/issues/6449, by printing the sucess messages line by line to the task screen on script success. 
The user sees only the success or the error messages. Error messages are joined with " | ". The application log tracks all  messages combined without specific order. 
The printed messages should not take up too much space since the notification area is limited in size and scrollable.

image

All further ideas as discssed in https://github.com/kitodo/kitodo-production/discussions/6461 are not considered here.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/CommandService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/CommandService.java
@@ -20,18 +20,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.kitodo.api.command.CommandInterface;
 import org.kitodo.api.command.CommandResult;
 import org.kitodo.production.helper.Helper;
-import org.kitodo.production.services.data.TaskService;
 import org.kitodo.serviceloader.KitodoServiceLoader;
 
 public class CommandService {
     private final CommandInterface commandModule;
     private final ArrayList<CommandResult> finishedCommandResults = new ArrayList<>();
-    private static final Logger logger = LogManager.getLogger(CommandService.class);
 
     /**
      * Initialize Command Service.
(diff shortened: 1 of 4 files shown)


solth:
@lutzhelm does this fix resolve the issue you reported (#6449) sufficiently for you?

oliver-stoehr (approved):
Code changes are fine and the script output is displayed as intended.
Is there a reason why error messages are joined in one line, but normal messages are printed on separate lines?

BartChris:
> > @lutzhelm does this fix resolve the issue you reported (#6449) sufficiently for you?
> 
> I can't answer this question conclusively as I ran into several problems trying to follow the [demo setup](https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Getting-started-Demo).
> 
> The message itself looks good. I'd say this solution meets our use case if the message text can be marked and copied.

Text can be marked and copied.

BartChris:
@oliver-stoehr - I now print the error messsages line by line. This can lead to duplications in the UI, because we are not differentiating between _CommandExceptions_ and _real_ _IOExceptions_. But maybe is it not too bad.
image

Kitodo/src/main/java/org/kitodo/production/services/command/CommandService.java
```diff
@@ -20,13 +20,18 @@
...
 import org.kitodo.api.command.CommandInterface;
 import org.kitodo.api.command.CommandResult;
+import org.kitodo.production.helper.Helper;
+import org.kitodo.production.services.data.TaskService;
 import org.kitodo.serviceloader.KitodoServiceLoader;
 
 public class CommandService {
     private final CommandInterface commandModule;
     private final ArrayList<CommandResult> finishedCommandResults = new ArrayList<>();
+    private static final Logger logger = LogManager.getLogger(TaskService.class);
```
henning-gerhardt, line 34: `logger` is not used and is using wrong log class `TaskService.class` instead of `CommandService.class` where the logger is declared.

BartChris: Yes, it is not used, thanks...will fix

oliver-stoehr (approved):
I think the error messages are more readable this way. 👍

PR #6468, opened 2025-03-21
Add missing foreign key relations

### LINKED ISSUE #6467: Missing foreign key relation for import configuration in process table
**Describe the bug**
While analysing #6466 I found out that import configuration is stored in the process table but as a plain integer value and not as a foreign key relation like all the other used places inside the database. The main functionality (updating the meta data of a process) is working but the used solution is in an other way like other key relations.

**Expected behavior**
The relation between import configuration and process table should be realised as a foreign key relation.

**Release**
3.8.2-SNAPSHOT

henning-gerhardt:
Fixes #6467 

While adjusting the one missing foreign key relationship I found a few more missing foreign key relationships. It could be possible that there are more missing but these was easy to found.

If needed / wanted this can be back ported to 3.8.x but so far I know no database changes should be made there.

### DIFF
--- a/Kitodo-DataManagement/src/main/resources/db/migration/V2_137__Add_missing_forein_key_relations.sql
+++ b/Kitodo-DataManagement/src/main/resources/db/migration/V2_137__Add_missing_forein_key_relations.sql
@@ -1,0 +1,32 @@
+--
+-- (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+--
+-- This file is part of the Kitodo project.
+--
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
Without this change it is possible that references are entered which reference to non existing data which will throw unexpected issues in the application itself. This may not happen on a regular usage but on issues in the application, database re-imports or if database changes are needed or must be done.
This change is helping to made the stored data in the database more reliable and is helping to find issues easier.

solth (approved):
Thanks for adding these missing relations

PR #6475, opened 2025-03-25
Fix MARC XPATHs

BartChris:
The default MARC XPath expressions provided by Kitodo for use in the Import Configuration are incorrect. The XPath intended for the ID is mistakenly used to retrieve the MARC title field, and vice versa. This PR corrects the XPath mappings.


solth (approved):
@BartChris thanks a lot for this pull request. 👍 I noticed this error some time ago, but forgot to fix it!

Issue #6474, opened 2025-03-25
Catalog import mapping file (xslt) does no longer work as expected: Xpaths do not match


mnscholz:
**Describe the bug**
For importing a process from catalog, we configured Kitodo to do a custom import search and apply a xslt mapping file to convert from MARC to KITODO format.

This config worked fine with previous Kitodo versions, including 3.6.*.
We recently upgraded to 3.8.1.

The xslt now no longer works as expected:
Static metadata returned from the xslt is read and displayed by Kitodo, however, all Xpaths in the xslt seem to fail now. 
**My suspicion is that with 3.7 or 3.8, Kitodo changed how the search result is input into to xslt.**
Unfortunately, the import process generates no logs for the xslt transformation, even when setting loglevel TRACE for the `org.kitodo` logger.

When run outside Kitodo with Saxon, the transformation works fine.

**Expected behavior**
Xslt works as it did before.

**Release**
3.8.1

**Desktop (please complete the following information):**
 - OS: Ubuntu 24.04
 - Browser: FF

BartChris:
What does your import configuration look like? Are you using SRU? What i noticed is that in the import configuration menu the configuration fields for ID and title seem to be confused. The xpath expression for the ID uses the MARC title field 245.

![Image](https://github.com/user-attachments/assets/f2df2919-d0b4-42ef-87ac-bf52b1074c5b)

But i am not sure if that was not the case with 3.6 as well and what might have caused the problems NOW when it worked in 3.6.

BartChris:
I made a Pull Request for the problem adressed in my comment. I do not know wether this has anything to do with your problems: https://github.com/kitodo/kitodo-production/pull/6475

mnscholz:
We do not use SRU but CUSTOM method as our SRU endpoint doesn't quite adhere to the standard and requires extra query parameters... :/ 

Effectively the response from the endpoint is a SRU xml document with MARCXML inside.

I now could drill it down to the CUSTOM method apparently not building the query URL as expected: When run in Kitodo, the endpoint returns an empty SRU result set for a given catalog ID. If I build the URL manually from the URL parts given in the Kitodo import config page and feed it into curl, the endpoint correctly returns a MARC record.

Unfortunately, again, I cannot see in the logs what URL Kitodo uses...

mnscholz:
OK, I found the log entry with the URL Kitodo uses (it slipped my eyes before...)

The problem is that the query param field must contain a `=` in our case:
[...]

PR #6477, opened 2025-03-26
Fix Query URL encoding

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6474

This PR changes the behavior of the query URL encoding. Encoding the query parameter keys breaks some search interfaces. We should therefor remove it. See: https://github.com/kitodo/kitodo-production/pull/6007#issuecomment-2751962304

SRU interfaces are still handled different than custom interfaces, since we need to Encode the equals sign there as well.

### DIFF
--- a/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
+++ b/Kitodo-Query-URL-Import/src/main/java/org/kitodo/queryurlimport/QueryURLImport.java
@@ -89,6 +89,7 @@ public class QueryURLImport implements ExternalDataImportInterface {
     private static final String HTTPS_PROTOCOL = "https";
     private static final String FTP_PROTOCOL = "ftp";
     private static final String EQUALS_OPERAND = "=";
+    private static final String EQUALS_OPERAND_ENCODED = "%3D";
     private static final String AND = "&";
     private static final String OAI_IDENTIFIER = "identifier";
     private final Charset encoding = StandardCharsets.UTF_8;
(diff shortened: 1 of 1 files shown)


solth:
@BartChris thanks for providing this fix!
@mnscholz & @gerlingsm would you mind checking if both issues reported by you are indeed resolved with this bugfix?

mnscholz:
I didn't have the time to go into the details, but I think the code will work for CUSTOM (my interest :) )

However, just out of interest: I wonder if createSearchFieldString() can be called with multiple search fields (map count > 1) for type other than SRU. As only SRU has a non-null queryString property, in this case the method would produce ill-formed URLs containing ANDs...

BartChris:
> However, just out of interest: I wonder if createSearchFieldString() can be called with multiple search fields (map count > 1) for type other than SRU. As only SRU has a non-null queryString property, in this case the method would produce ill-formed URLs containing ANDs...

I just adapted the logic to fix yout use case, but you have a point: Right now, the CUSTOM endpoints do not allow search at all, so this part of the code will never be triggered for non SRU-interfaces which probably also join their queries different than with " AND ". If we ever going to allow search for non SRU-endpoints, this has to be reconceptualized.

mnscholz:
Again, I don't have the big picture, so I may be mistaken:
createSearchFieldString() is called from getMultipleFullRecordsFromQuery() and performHTTPRequest() and both don't have an explicit check on interfaceType being SRU in the surrounding code. So from this limited point of view it is possible to for createSearchFieldString() to be applied on other types than SRU.
If the method should not be called on non-SRU types, maybe in that case better throw an NotImplementedException or the like?

solth:
I agree that the part concerning the evaluation of given search parameters needs to be refactored to become clearer, but as @BartChris pointed out, currently that part is only called for SRU interfaces so the problem of malformed URLs caused by errornous `AND` fragments shouldn't occur at the moment. In any case, I would prefer to tackle that (potential) problem in a separate issue/pull request so that the fix in _this_ pull request isn't delayed and can be added to the main branch as soon as possible. 

Speacking of branches: @BartChris would you mind opening a corresponding pull request against the `3.8.x` branch once @gerlingsm has confirmed that the proposed change does work for him as well?

BartChris:
3.8. backport is there: https://github.com/kitodo/kitodo-production/pull/6478

solth:
> 3.8. backport is there: #6478

oh, right, thanks! Must have missed that :see_no_evil:

gerlingsm:
> I agree that the part concerning the evaluation of given search parameters needs to be refactored to become clearer, but as @BartChris pointed out, currently that part is only called for SRU interfaces so the problem of malformed URLs caused by errornous `AND` fragments shouldn't occur at the moment. In any case, I would prefer to tackle that (potential) problem in a separate issue/pull request so that the fix in _this_ pull request isn't delayed and can be added to the main branch as soon as possible.
> 
> Speacking of branches: @BartChris would you mind opening a corresponding pull request against the `3.8.x` branch once @gerlingsm has confirmed that the proposed change does work for him as well?

I have tested it and can say that it works as expected.👍

solth (approved):
Nice clean code, works as advertised! @BartChris Thanks for the fix!

PR #6480, opened 2025-03-28
Fix blur event that requires a second click

### LINKED ISSUE #3346: Allow to save changes to the project immediately
If you change something in a single field on the project, the button for saving is not activated. You first have to click out of the field for the button to be activated.

**Goal:** The button should be activated as soon as something is changed.

### LINKED ISSUE #3472: Focused PF components block other PF components
When a PrimeFaces component is focused a click on another PrimeFaces component will not focus this component or call its action. A second click is necessary while the first click only dispatches the `blur` or `change` event of the originally focused component.

### LINKED ISSUE #6100: Avoid second click after changing the content of a metadata field
If in the metadata editor the content of a metadata field is changed (add, delete, ...) it is not possible to choose directly another structure element or to apply the buttons "Validate", "Save", ...
A similar behavior occurs during the creation of newspaper processes wit the calendar. 

Probably this is necessary to avoid the loss of the changes of the content of the metadata. 

For users, this is annoying and in a few cases it lead to loss of data, because the buttons "Save" and "Exit" has been applied too quickly. It hasn't been regarded, that the "Save"-function was not been triggered and the metadata editor was closed without saving the changes in the metadata field. 

After changing the content in a metadata field, it should be possible to select directly other structure elements or the buttons "Validate", "Save", ...

If the implementation of the requirement causes too far-reaching adjustments to the software, the current status should be retained.

The users can get into the habit of clicking on somewhere in the “Metadata” frame after changing the content in a metadata field and before before clicking on an object outside of the metadata frame.

oliver-stoehr:
Resolves #3346
Resolves #3472
Resolves #6100

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/pattern-library.css
@@ -1473,7 +1473,7 @@ Popup dialogs
 }
 
 #loadingScreen {
-    animation: overlay-delayed .3s .5s forwards;
+    animation: overlay-delayed .3s .7s forwards;
     position: fixed;
     width: 100vw;
     height: 100vh;
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/defaultScript.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/defaultScript.js
@@ -11,6 +11,18 @@
 
 $(document).ready(function() {
     $('#loadingScreen').hide();
+
+    // Polyfill for 'paste' event in PrimeFaces 8
+    $(document).on('paste', 'input, textarea', function (e) {
+        /*
+         * Handle paste event and dispatch a 'keyup' event for the original target element.
+         * Polyfill for PrimeFaces 8: Since PrimeFaces 8 does not support the 'paste' event for components like 'inputText'
+         * or 'inputTextarea' we have to listen to 'paste' events separately and dispatch an event
+         * supported by these components in PrimeFaces 8.
+         * A listener for the 'keyup' event must be defined for the respective component for this polyfill to work.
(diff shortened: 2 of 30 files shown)


andre-hohmann (changes requested):
@oliver-stoehr : Thank you very much!

The changes fixes the problems in the metadata editor, creation of newspaper and projects settings. 
I have two remarks/questions: 

_1 Metadata in newspaper-processes_
During the creation of newspaper-processes, the input of correct metadata in text-fields is not possible anymore, as the text-fields is "refreshed" during typing. 
After the refresh, the cursor is set to the beginning of the text-field. 

<details>
<summary>Klick here to see the GIF. </summary>

EingabeTextZeitungsausgabe
</details>

_2 Catalogue Search_
My colleagues have pointed out to me that the “Search OPAC” button still has to be clicked twice to search in the catalog. 
This requirement is not described in the issues and therefore i can understand, if this would not be fixed in this PR. 

<details>
<summary>Klick here to see the GIF. </summary>

SucheOPACImport
</details>

oliver-stoehr:
@andre-hohmann thank you for testing!
I added another Commit to fix both problems.

andre-hohmann:
@oliver-stoehr : Thanks a lot for the quick response. 

_1 Metadata in newspaper-processes_
This is solved. 

_2 Catalogue Search_
Unfortunately, this does still occur. 
Could you have a look on it?

oliver-stoehr:
@andre-hohmann I applied my fix now also for multi-line inputs in addition to single-line inputs. Can you please check again if that solves the problem for you?

andre-hohmann (approved):
@oliver-stoehr : thanks a lot!
Now everything works fine.

henning-gerhardt (approved):

PR #6440, opened 2025-02-25
[3.8] Drag & drop multiple tree nodes in structure tree

thomaslow:
Backport of #6340 for 3.8.x

- [x] Test backport

PR #6482, opened 2025-03-29
Enable linking pages via gallery while media is hidden in structure tree

### LINKED ISSUE #6098: Allow always "Assign to next element"
If "Show only structure elements" is activated in "customize display" for the structure, it is not possible anymore to apply "Assign to next element" for the pages in the structured view of the gallery.
Therefore, it is necessary to switch in between the two states of the structure tree. That is annoying for the users.

Probably, this is connected to the reason, that "Assign to next element" is not possible in the "Extended structure tree".
- #4380

It should be possible to assign pages to the next element in the structured view of the gallery, even if only structure elements are shown in the structure tree.

If the implementation of the requirement causes too far-reaching adjustments to the software, the current status should be retained. 
These use cases probably do not occur so frequently and the resources can be invested in other developments.

**Additional context**
_Activate "Show only structure elements"_
assignToNextElement01

_Example: "Assign to next element" is not possible_
assignToNextElement02

_Example: "Assign to next element" is possible_
assignToNextElement03

solth:
Fixes #6098 

To fix the linked issue, some logic had to be adapted. The node relationships in the structure tree are used to determine whether a page can be linked to the next structure. Since the structure tree was rebuild _skipping_ the leaves representing pages or other media entirely when toggling the "hide media" switch, though, those relations could not be evalutated correctly for linking when the "show structures only" option was selected. 

I now changed this behavior so that nodes/leaves representing media in the structure tree are never omitted, but only _hidden_ when the option mentioned above is activated.

@thomaslow do you remember why you chose to actually exclude those page nodes from the structure tree instead of just not rendering them when you implemented this feature? I analysed the code but couldn't find a reason why the pages would have to be discarded entirely from the tree when the "show structures only" option was selected, but perhaps you remember some special constellations I missed.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -482,11 +482,7 @@ public class StructurePanel implements Serializable {
             this.preserveLogical();
             this.preservePhysical();
         } else {
-            if (isHideMediaInLogicalTree()) {
-                this.preserveLogical();
-            } else {
-                this.preserveLogicalAndPhysical();
-            }
+            this.preserveLogicalAndPhysical();
         }
     }
 
@@ -1270,10 +1266,10 @@ public class StructurePanel implements Serializable {
      */
     private void preserveAfterDragDrop(boolean logicalMoved, boolean pageMoved, boolean physicalMoved) {
         if (logicalMoved || pageMoved) {
-            if (logicalStructureTreeContainsMedia()) {
-                preserveLogicalAndPhysical();
-            } else {
+            if (this.isSeparateMedia()) {
                 preserveLogical();
+            } else {
+                preserveLogicalAndPhysical();
             }
             this.dataEditor.getGalleryPanel().updateStripes();
             this.dataEditor.getPaginationPanel().show();
@@ -2040,15 +2036,6 @@ public class StructurePanel implements Serializable {
             throw new IllegalArgumentException("node label option must be either type, title or type+title");
         }
         this.nodeLabelOption = nodeLabelOption;
-    }
-
-    /**
-     * Returns true if the logical structure tree is a combined tree of structure nodes and view nodes (media).
-     *
-     * @return true if logical structure tree contains media
-     */
-    public boolean logicalStructureTreeContainsMedia() {
-        return !this.isSeparateMedia() && !this.isHideMediaInLogicalTree();
     }
 
     /**
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -1825,6 +1825,8 @@ public class StructurePanel implements Serializable {
     /**
      * Check if the given TreeNode's PhysicalDivision is assigned to several LogicalDivisions.
      *
+     * @param treeNode TreeNode to be checked for assignment to multiple LogicalDivisions
+     *
      * @return {@code true} when the PhysicalDivision is assigned to more than one logical element
      */
     private boolean isAssignedSeveralTimes(TreeNode treeNode) {
@@ -1933,6 +1935,8 @@ public class StructurePanel implements Serializable {
      * Check if the given TreeNode's PhysicalDivision can be assigned to the next logical element in addition to the
      * current assignment.
      *
+     * @param treeNode TreeNode representing View to be checked
+     *
      * @return {@code true} if the PhysicalDivision can be assigned to the next LogicalDivision
      */
     private boolean isAssignableSeveralTimes(TreeNode treeNode) {
@@ -1973,6 +1977,8 @@ public class StructurePanel implements Serializable {
 
     /**
      * Assign given TreeNode's PhysicalDivision to the next LogicalDivision.
+     *
+     * @param treeNode TreeNode representing View to be assigned to next LogicalDivision
      */
     private void assign(TreeNode treeNode) {
         TreeNode nextLogical = findNextLogicalNodeForViewAssignment(treeNode);
@@ -1992,6 +1998,7 @@ public class StructurePanel implements Serializable {
 
     /**
      * Unassign the selected Node's PhysicalDivision from the LogicalDivision parent at the selected position.
+     * If media is hidden in the structure tree the selected node is retrieved from the selected logical structure first.
      * This does not remove it from other LogicalDivisions.
      */
     public void unassign() {
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -2017,6 +2017,8 @@ public class StructurePanel implements Serializable {
     /**
      * Unassign the givens TreeNode's PhysicalDivision from the LogicalDivision parent at the selected position.
      * This does not remove it from other LogicalDivisions.
+     *
+     * @param treeNode TreeNode representing View to be unassigned from the LogicalDivision parent at the selected position
      */
     private void unassign(TreeNode treeNode) {
         if (isAssignedSeveralTimes()) {
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -1814,12 +1814,7 @@ public class StructurePanel implements Serializable {
      * @return {@code true} when the PhysicalDivision is assigned to more than one logical element
      */
     public boolean isAssignedSeveralTimes() {
-        TreeNode selected = getSelectedLogicalNodeIfSingle();
-        if (isHideMediaInLogicalTree()) {
-            return isAssignedSeveralTimes(getSelectedView(selected));
-        } else {
-            return isAssignedSeveralTimes(selected);
-        }
+        return isAssignedSeveralTimes(getSelectedLogicalNodeIfSingle());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


andre-hohmann (commented):
Thanks a lot!
It works for elements of the same hierarchy level. 
It does not work for elements of different hierarchy levels. 

In the issue it has not been mentioned that it should work for elements of different hierarchy levels. Additionally, this has been generally implemented not long ago: 
- https://github.com/kitodo/kitodo-production/issues/5457

I would therefore accept the current state, if it was complicated to apply the change for different hierarchy levels. I cannot assess the aditional effort. 

@solth: I think you are the person wh can decide it.

thomaslow:
> @thomaslow do you remember why you chose to actually exclude those page nodes from the structure tree instead of just not rendering them when you implemented this feature? I analysed the code but couldn't find a reason why the pages would have to be discarded entirely from the tree when the "show structures only" option was selected, but perhaps you remember some special constellations I missed.

@solth I don't remember exactly, sorry. I guess my thought was that this exact variation of the tree was already implemented in the "separate media" version of the metadata editor, and it would make sense to just re-use that code instead of assuming the burden of implementing a new variation of the tree that might introduce new issues.

Since PrimeFaces has direct access to the `TreeNode` tree and modifies it in-place based on various user interactions (selection, drag & drop), maybe there could be issues if the backend tree does not exactly match the rendered tree. I don't know.

If your approach (not rendering page nodes) works fine, maybe it would make sense to get rid of the different backend `TreeNode` tree variations altogether and also use this approach in the "separate media" version of the metadata editor as well?

solth:
@thomaslow thank you for the response. PrimeFaces nodes not being rendered shouldn't be a problem since they do not behave differently from other components that can be "hidden" in the frontend while still having a backing bean property associated to them. Actually, that is the very reason why I think hidding the media nodes instead of re-building the tree and skipping the nodes is the correct solution.

I think the idea to adopt this approach for the "Extended tree" is interesting, albeit out of scope of this pull request.

solth:
> It does not work for elements of different hierarchy levels.

I might have missed a detail, but I just tested this again and it does indeed seem to work:

Bildschirmfoto 2025-04-02 um 16 39 38
(never mind the nested "books" here!)

Or are we takling about a different use case?

andre-hohmann:
No, that's the use case i meant. 
See GIF below. 

Maybe @henning-gerhardt can have a look, if something (another PR, ...) is missing. 

VerknuepfungHierarchie

henning-gerhardt:
Changes of #6483 are applied to the 3.8.x test version.

andre-hohmann:
It seems to be a caching problem.
The option is available, but only belated and/or after some actions.
It is a general behavior for both expanded and collapsed structure tree and difficult to explain. Thus, i created GIF-files.

**1 expanded structure tree**
<details>
  <summary>click here</summary>

VerknuepfungHierarchie01
</details>

**2 collapsed structure tree**
<details>
  <summary>click here</summary>

VerknuepfungHierarchie02
</details>

andre-hohmann:
This is not a caching problem, but very weird. 

- It seems in Firefox, it is more confusing then in Chrome
- It seems the reason is the check, if a link is allowed 
  - links to/from unstructured elements are not allowed
  - links to subordinate objects are not (always) allowed
- It seems to depend, if the selected object is displayed in the first row of the gallery

Nevertheless, this behavior does not depend on the collapsed structrure tree and should be examined in a new issue.

andre-hohmann (approved):

henning-gerhardt (changes requested):
There are incomplete JavaDoc comments.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -1901,11 +1939,23 @@ public void assign() {
...
     public void unassign() {
+        if (isHideMediaInLogicalTree()) {
+            unassign(getSelectedView(selectedLogicalNode));
+        } else {
+            unassign(selectedLogicalNode);
+        }
+    }
+
+    /**
+     * Unassign the givens TreeNode's PhysicalDivision from the LogicalDivision parent at the selected position.
```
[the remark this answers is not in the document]
henning-gerhardt: JavaDoc for method parameter `treeNode`  is still missing.

henning-gerhardt (changes requested):
JavaDoc for method parameter `treeNode`  for method `unassign` in class `StructurePanel` is still missing.

henning-gerhardt (approved):

solth:
@andre-hohmann & @henning-gerhardt after merging #6340 I needed to update this branch to resolve some merge conflicts. If you don't mind I think it would be prudent that you take another look and check one more time if everything still works as expected (I also checked all cased discussed above, of course, but in case something slipped my eye you should verify the current state independently). The updated branch has _less_ changes than before, so I hope you agree it is an approvement over the previous code changes you already approved (given all functionality is retained, of course!)

 If you would rather make this test/review on the base of Kitodo 3.8 you can wait for @thomaslow to review the changes first and for me to apply the same updates to #6483 after merging #6440

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -1924,11 +1919,7 @@ public boolean isAssignableSeveralTimes() {
         if (Objects.isNull(selectedLogicalNode)) {
             return false;
         }
-        if (isHideMediaInLogicalTree()) {
-            return isAssignableSeveralTimes(getSelectedView(selectedLogicalNode));
-        } else {
-            return isAssignableSeveralTimes(selectedLogicalNode);
-        }
+        return isAssignableSeveralTimes(selectedLogicalNode);
```
henning-gerhardt, line 1922: This is a recursive call and the only one exit statement is false when the node is null. Is this really correct or missed I something?

solth: This is not a recource call. It calls an different method with the same name, but with a parameter (so an overloaded function).

henning-gerhardt:
@solth: Its look like that applying 6440 and then 6482 on top of the 3.8.x branch is possible with only small glitches which can patch solve by it self. Deed we need something more on the current 3.8.x branch?

solth:
> @solth: Its look like that applying 6440 and then 6482 on top of the 3.8.x branch is possible with only small glitches which can patch solve by it self. Deed we need something more on the current 3.8.x branch?

The changes in #6483 will probably be identical to those in this pull request after merging #6440 and then resolving the resulting conflicts.

henning-gerhardt:
> > @solth: Its look like that applying 6440 and then 6482 on top of the 3.8.x branch is possible with only small glitches which can patch solve by it self. Deed we need something more on the current 3.8.x branch?
> 
> The changes in #6483 will probably be identical to those in this pull request after merging #6440 and then resolving the resulting conflicts, since the code base doesn't differ between `main` and `3.8.x` branches in the affected files.

Thanks. I will do it in this way and provide @andre-hohmann a new version in our testing system so he can check this changes.

solth:
> Thanks. I will do it in this way and provide @andre-hohmann a new version in our testing system so he can check this changes.

@henning-gerhardt I just reviewed #6440 and will merge it into the `3.8.x` branch now, and afterwards will be updating #6483 accordingly, so you can wait for that updated branch/pull request as well, if you want.

thomaslow:
I'm sorry to be the "buzzkill", but these changes seem to cause a drag & drop issue that is not present in the current main branch (including the new drag and drop changes). When dropping a folder into another folder that contains both folders and hidden pages, the drop-point seems to be calculated incorrectly. 

@solth I'm not sure what the problem is exactly. When you asked me a few days ago, I was mostly worried whether the tree nodes are correctly labelled (`id="logicalTree:3_2_0_2"`) if some nodes are not rendered by PrimeFaces. However, it seems that all nodes are still labelled consecutively as if all nodes (including hidden ones) are rendered. Hidden nodes are simply not added to the DOM. The ids are correctly referencing the backend tree node order. So, I have no idea what the problem is.

Demo for this pull request:

https://github.com/user-attachments/assets/098feda4-caf4-40f0-893a-4f68c431afdc

Demo for current main branch:

https://github.com/user-attachments/assets/5cc8215b-84da-4b0a-8868-2b6a874b5355

thomaslow:
@solth I think the issue is this line:

https://github.com/primefaces/primefaces/blob/cd4fbdf1d9d4ae054da19b9a84001d0c34d142eb/src/main/resources/META-INF/resources/primefaces/tree/tree.vertical.js#L599

PrimeFaces calculates the drop-position based on the parent node and and index within the list of children of that node. On the frontend side, the index is determined by counting the `li.ui-treenode` elements that are listed before the drop point. Since hidden tree nodes are not in the DOM, the index does not match the backend tree. PrimeFaces will insert the dragged node at the wrong child index further up the tree.

In theory it should be possible to correct the order in the drag and drop event handler knowing that PrimeFaces made a mistake here. 

Updating to PrimeFaces 15 probably will not resolve the issue. The calculation is the same in PrimeFaces 15:

https://github.com/primefaces/primefaces/blob/9f631c984e3e3b83f650542897634061f843d047/primefaces/src/main/resources/META-INF/resources/primefaces/tree/tree.vertical.js#L766

solth:
> In theory it should be possible to correct the order in the drag and drop event handler knowing that PrimeFaces made a mistake here.

@thomaslow thanks for investigating this issue. Hopefully I can find a way to adjust the drop index manually in the drop event handler 🤞.

solth:
@thomaslow I think the correct solution here is to simply hide the TreeNodes representing `View`s using CSS instead of JSFs 'rendered' attribute. This way we don't need to fiddle with the PrimeFaces row index calculation and potentially disrupt the intricit JSF/PrimeFaces logic while still being able to hide the pages in the frontend. I wasn't able to reproduce the issue you described when applying this change. Do you see any problem with this - in my opinion cleaner - low-tech fix?

andre-hohmann (approved):

henning-gerhardt (approved):

thomaslow:
@solth Yes, your solution to hide tree nodes via CSS seems reasonable. I tested drag & drop again and it work fine now.

thomaslow (approved):

Issue #6486, opened 2025-03-31
Duplicate process template results in an exception while entering the new process template title


henning-gerhardt:
**Describe the bug**
While making a video for #6484 I discovered an other issue - which could be happened through the changed used scope from `SessionScoped` to `ViewScoped`.

```
javax.el.PropertyNotFoundException: Target Unreachable, [template] returned null
	at org.apache.el.parser.AstValue.getTarget(AstValue.java:123)
	at org.apache.el.parser.AstValue.getType(AstValue.java:57)
	at org.apache.el.ValueExpressionImpl.getType(ValueExpressionImpl.java:173)
	at org.jboss.weld.el.WeldValueExpression.getType(WeldValueExpression.java:93)
	at org.jboss.weld.el.WeldValueExpression.getType(WeldValueExpression.java:93)
	at org.apache.myfaces.view.facelets.el.ContextAwareTagValueExpression.getType(ContextAwareTagValueExpression.java:76)
	at org.primefaces.util.ComponentUtils.getConverter(ComponentUtils.java:160)
	at org.primefaces.renderkit.InputRenderer.getConvertedValue(InputRenderer.java:45)
	at javax.faces.component.UIInput.getConvertedValue(UIInput.java:789)
	at javax.faces.component.UIInput.validate(UIInput.java:708)
	at javax.faces.component.UIInput.processValidators(UIInput.java:293)
	at org.apache.myfaces.context.servlet.PartialViewContextImpl$PhaseAwareVisitCallback.visit(PartialViewContextImpl.java:779)
	at org.apache.myfaces.component.visit.PartialVisitContext.invokeVisitCallback(PartialVisitContext.java:213)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1019)
	at javax.faces.component.UIComponentBase.visitTree(UIComponentBase.java:1192)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1047)
	at javax.faces.component.UIComponentBase.visitTree(UIComponentBase.java:1192)
...
```

**To Reproduce**
1. Go to Project settings
2. Go to Process templates tab
3. Duplicate an existing process template
4. Try to enter a new process template title (as non is generated which is or could even an issue)
5. Change the focus to an other input element (even without entering a new process template title)
6. See the error

**Expected behavior**
Loosing input focus should not crash the dialog / application

**Screenshots**
See the video in https://github.com/kitodo/kitodo-production/issues/6484#issuecomment-2766268819

[...]

Issue #6485, opened 2025-03-31
Duplicate process templates results in new process template


henning-gerhardt:
**Describe the bug**
While try to investigate #6484 I found an other issue may introduced by #6377: duplicated process templates acts like new process templates. The selected projects are not checked / selected like in the original process template, all list entries (like workflow) point to the first entry instead of the original / selected one. Its look like that now duplicating acts like new with one difference: new process templates are be default as active while duplicated process templates are not active be default.

**To Reproduce**
1. Go to Project settings
2. Go to Process templates tab
3. Click on duplicate button of existing process template
4. See the not selected / first entries of all the options like in new process template

**Expected behavior**
Duplicated process templates should have the same settings like the original process template.

**Release**
3.8.x from 2025-03-31, main from 2025-03-31

henning-gerhardt:
Changing back from `ViewScoped` to `SessionScoped` for the `TemplateForm.java` class solve the issue for me.

Issue #6484, opened 2025-03-31
NPE when duplicating process template


thomaslow:
**Describe the bug**
When clicking on the "duplicate" button of an existing process template, there is a Null Pointer Exception.

```
java.lang.NullPointerException
	at org.kitodo.production.forms.TemplateForm.getOcrdWorkflow(TemplateForm.java:294)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at javax.el.BeanELResolver.getValue(BeanELResolver.java:94)
...
```

**To Reproduce**
1. Go to Project settings
2. Go to Process templates tab
3. Click on duplicate button of existing process template
4. See NPE

**Expected behavior**
There should be no NPE.

**Release**
Current main branch

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox

**Additional context**
The problem seems to be related to pull request #6377. The change from Session scope to View scope will cause the `template` variable of `TemplateForm.java` to become null after the duplication is triggered, because the bean is re-initialized in the view scope.

henning-gerhardt:
I can not confirm the NPE but an different behaviour on clicking the duplicate button. I will open an issue for this.

henning-gerhardt:
I did the same and I got the following screen (process template) without any error. I try to make a video/gif animation of it.

henning-gerhardt:
While I made the video: I found other issue: `javax.el.PropertyNotFoundException: Target Unreachable, [template] returned null` - I will open an issue for this too.

But now the current working behaviour:

https://github.com/user-attachments/assets/8f827eab-49e3-468d-8992-ff461398f529

PR #6487, opened 2025-03-31
Revert scope change from SessionScoped to ViewScoped back

henning-gerhardt:
Fixes #6486 
Fixes #6485 

Maybe fixes #6484 too

Should / can be back ported to 3.8.x branch too.

PR #6489, opened 2025-04-01
Use FlashStorage for Template Duplication

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6484
Fixes https://github.com/kitodo/kitodo-production/issues/6485
Fixes https://github.com/kitodo/kitodo-production/issues/6486
Alternative to https://github.com/kitodo/kitodo-production/pull/6487

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
@@ -354,7 +354,7 @@ public class TemplateForm extends TemplateBaseForm {
         try {
             Template flashTemplate = (Template) FacesContext.getCurrentInstance()
                     .getExternalContext().getFlash().get("duplicatedTemplate");
-            if (flashTemplate != null) {
+            if (Objects.nonNull(flashTemplate)) {
                 setTemplate(flashTemplate);
             } else if (id != 0) {
                 setTemplate(ServiceManager.getTemplateService().getById(id));
--- a/Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Objects;
 
 import javax.annotation.PostConstruct;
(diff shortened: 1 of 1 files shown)


BartChris:
@henning-gerhardt Unfortunately i overlooked the problems with duplicating templates when implementing https://github.com/kitodo/kitodo-production/pull/6377

Especially considering the plan of deprecating all `Session Scopes` (https://github.com/kitodo/kitodo-production/issues/6456), it would be good if we can stick with `ViewScoped`. I tried to fix the duplication of templates by using the JSF Flash storage which preserves the data over redirects. Can you check wether i  overlooked something and there are still problems?

henning-gerhardt:
I did not know this JSF Flash storage technology and I must read and understand this first to understand and rate your change. This will take a while.

thomaslow:
I tested your changes, and it resolves my issue (#6484). I do not have a NPE anymore.

I didn't know about the JSF flash API either. It seems to me that this API was specifically designed to solve this kind of issue. 

In my opinion, this is a nice solution.

solth:
Flash storage is the correct solution here, IMO, since we do not want to go back to SessionScoped beans where ViewScoped beans are appropriate.

henning-gerhardt:
JSF Flash storage is an interesting technology. I was not aware of it. It look like that it is fixing the issues while duplicating process templates.

Kitodo/src/main/java/org/kitodo/production/forms/TemplateForm.java
```diff
@@ -349,15 +352,18 @@ public void setTemplateById(int id) {
      */
     public void loadTemplate(int id) {
         try {
-            if (id != 0) {
+            Template flashTemplate = (Template) FacesContext.getCurrentInstance()
+                    .getExternalContext().getFlash().get("duplicatedTemplate");
+            if (flashTemplate != null) {
```
henning-gerhardt, line 357: Instead of `flashTempltate != null` use `Objects.nonNull(flashTemplate)` as on many places `Objects.nonNull()` checks are already used.

henning-gerhardt (approved):

solth (approved):
Thanks for this fix!

Issue #6473, opened 2025-03-24
Problem with reimport into metadata groups with empty fields


BartChris:
**Describe the bug**
As outlined in 
https://github.com/kitodo/kitodo-production/discussions/6436#discussioncomment-12310932
the reimport of metadata does currently not work correctly if metadata groups in the editor contain empty fields. The same person from the catalogue is duplicated.

**To Reproduce**
1. Have metadata groups with empty fields (e.g. a person's role) in the metadata of the current document
2. Reimport the medadata from the catalogue
3. Metadata Groups with empty fields are duplicated 

**Expected behavior**

The reimport should also work for metadata groups with empty fields.

**Release**
Kitodo 3.8.x, master

solth:
@BartChris unless you have a fix for this problem, I do not think it's feasible to include its solution in 3.8.2, which is overdue anyway. But since 3.9.x will take some time to be released, I see a real possibility of a release 3.8.3 happening, which should then include the fix for this problem.

solth:
@BartChris IIRC I included empty fields in order to maintain a comparable field table in the comparison view, so that it becomes more obvious which fields are empty at one side (for example in the local record) vs filled with a value remotely (in the OPAC record). For that reason the fields in a group would also normally be sorted by metadata key, so that corresponding fields are lined up. Why the fields of the metadata groups in the screenshots you provided seem to be sorted differently I cannot say, that confuses me a little bit.

PR #6495, opened 2025-04-03
Skip empty metadata fields when doing a Re-import

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6473


solth:
I think empty metadata entries in groups were kept intentionally to make it easier to understand which fields in the comparison are filled on one side but empty on the other (see my comment here https://github.com/kitodo/kitodo-production/issues/6473#issuecomment-2782420656)

BartChris:
@solth I found a different way to prevent the duplication of groups while keeping empty fields in the comparison dialog.

solth:
@BartChris would you mind providing backports of this fix for `3.9.x` (and potentially `3.8.x` as well)?

PR #6471, opened 2025-03-24
Preserve metadata in collapsed metadata group panels during process creation

solth:
This pull request fixes #5932 by replacing the client side validation of HTML form fields with server side ruleset based validation, which is also used in the metadata editor, albeit in a stricter mode.

PR #6497, opened 2025-04-03
Optional validation on process creation

solth:
This pull request adds the _optional_ parameter `metadataValidationOptionalDuringProcessCreation` to the `kitodo_config.properties` file proposed in https://github.com/kitodo/kitodo-production/pull/6471#issuecomment-2774876764 to make metadata validation during process creation skippable.

If `metadataValidationOptionalDuringProcessCreation` is set to `true`, a small checkbox is displayed in the "Create process form" with which the user can control metadata validation during process creation and thus whether creating a process with invalid or incomplete metadata should be possible or not:

Bildschirmfoto 2025-04-03 um 23 09 06

Setting `metadataValidationOptionalDuringProcessCreation` to `false` or not adding the parameter to `kitodo_config.properties` at all results in the checkbox not being shown and thus ensures the current behaviour, where metadata validation is always mandatory. Therefore, using this new option/checkbox is completely optional, e.g. the new checkbox is only offered when actively adjusting `kitodo_config.properties` and adding the new parameter mentioned above.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
+++ b/Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java
@@ -667,7 +667,14 @@ public enum ParameterCore implements ParameterInterface {
      * Optional parameter controlling whether the import of all elements from an uploaded EAD XML file should be
      * canceled when an exception occurs or not. Defaults to 'false'.
      */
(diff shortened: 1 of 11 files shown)


BartChris (approved):
Looks good and works!

PR #6499, opened 2025-04-04
Fix new structure element is not inserted at the current position

thomaslow:
This pull request fixes an issue related to the new multi-select feature of the logical structure tree of the metadata editor introduced by #6340. 

When adding a structure element while selecting one or multiple images, the structure element is not inserted at the current position of the selection. 

Demo of the problem:

https://github.com/user-attachments/assets/22cdca07-0915-48db-8e7e-6c260dd2b467

When opening the AddDocStrucDialog, the dialog would sneakily change the node selection to the parent logical divison of the current selection. Due to the refactoring in #6340, this would affect the check that determines whether the user has selected consecutive pages such that the new structure element can be inserted at the correct position.

This pull request removes this internal tree node selection change and determines the target parent logical division directly.

Demo of the solution:

https://github.com/user-attachments/assets/cd73bcca-056d-4ad5-9e14-b559740545f5


oliver-stoehr (approved):
Code is fine, fixes the described bug.

Issue #6378, opened 2025-01-24
Release 3.8.0 - Wrong Process Template is used, when creating a Process from Dashboard View.


stefanCCS:
**Describe the bug**
It looks like that, if I create a process from Dashboard view, a wrong Process Template will be  used.
If I do the same from Project view it is ok.

**_1st try: Create process via Project View._**
You can see that I have two projects, one named with "KITODO-only". Each of these two Projects have an according Process Template assigned (again one named "KITODO-only".

![Image](https://github.com/user-attachments/assets/b9ee036c-b9af-4b1f-9a73-602b41915d49)

If I create the process using this marked "+" button, an according Window is opened with correct used Process template:

![Image](https://github.com/user-attachments/assets/ed7c2ae9-bf0a-4a52-b118-49e96e76915f)

**_2nd try: Create process via Dashboard View:_**
Now, I try to create a process using same Project "KITODO-only":

![Image](https://github.com/user-attachments/assets/c5d7ea73-6c8f-47e8-a3aa-aee1ae5db40d)

But I get a Window opened which uses the wrong Process template (it fact it is the Process template from the other Project):

![Image](https://github.com/user-attachments/assets/0cb1138e-fd2f-43e6-acb8-7add77cf1666)

stefanCCS:
Another hint: On a different KITODO instance, it works ok.
So, there might be some random effect behind ...

solth:
Perhaps the source of the problem is similar to that of #6345 ?

henning-gerhardt:
I think this issue is similar to #6345 as here even the desktop widget is in use.

solth:
Fixed by #6526

Issue #6345, opened 2024-12-10
Desktop project widget use wrong project on process creation


henning-gerhardt:
**Describe the bug**
If you use the project widget on the desktop page to create a new process then a wrong project is used which results that the created project in inside the wrong project and may have a wrong workflow assigned.

For example I used the "TG_16/17" project to create a new process inside this project which is using our default process template "PV-SLUB-Standard" from the desktop project widget:

![Screenshot_2024-12-10_15-56-59](https://github.com/user-attachments/assets/1c32b435-b37b-4c81-8c11-8df277534034)

But in the following dialog this was visibile (I skipped the process search dialog as this dialog is not relevant):

![Screenshot_2024-12-10_15-57-45](https://github.com/user-attachments/assets/99f3ea04-b2b8-4fe3-8b9a-f4c1e20e1157)

It is mentioned the process template "PV-SLUB-Dingler" which is from an other project which has different settings and even more a different workflow. Even the used `templateId` with value `3` and the `projectId` with value `13` in the URI pointing to the wrong process template "PV-SLUB-Dingler" (template id 3) and  project "Dingler" (project Id 13) instead of the correct process template "PV-SLUB-Standard" (template id 1) and project "TG_16/17" (project id 2) which is working if I try to create a new process from the project list overview page:

![Screenshot_2024-12-10_15-58-20](https://github.com/user-attachments/assets/18f94bf5-75a8-4586-b710-132d070f827a)

**To Reproduce**
1. Click on the "+" sign of the desktop project widget to create a new process from the selected project
2. See in the creating dialog a wrong content for the used process template and if you save the new process the process is assigned to an other project.

**Expected behavior**
Creating a new process from the desktop project widget should use the right project and process template on creating the process.
**Release**
3.6.0-SNAPSHOT, 3.7.1, 3.8.0-RC1, master branch

**Additional context**
This has something to do with the sorting of the projects inside the desktop project widget. If the sorting in https://github.com/kitodo/kitodo-production/blob/master/Kitodo/src/main/webapp/WEB-INF/templates/includes/desktop/projectsWidget.xhtml#L22 is removed than the process creating is working as expected but than the list of projects inside the project widget is not sorted anymore.

solth:
Somehow related to the changes in #5802

solth:
Fixed by #6526

henning-gerhardt:
[...]

PR #6501, opened 2025-04-04
Sort Projects on Desktop Widget by Custom Server Side function

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6345
Fixes https://github.com/kitodo/kitodo-production/issues/6378

This PR tries to adress the above issues by doing the sort server side. I cannot directly reproduce the bug in my test system, so maybe @henning-gerhardt or @stefanCCS can check if it fixes the issue.


henning-gerhardt:
@BartChris : I tried your changes but they did not work. I have the same issue like already reported. I think it has something to do with the parameters itself which are incorrect or did not fit to the sorted and selected value on clicking the + button. The creation of a new process is called in a total different way then the other commands in the widget.

BartChris:
@henning-gerhardt I gave it another try.

henning-gerhardt:
@BartChris : sorry, no change at all :-( Maybe helpful or maybe not: the displayed project "TG16/17" is one of the oldest projects (id is 2) and should not  displayed at all as one of the first 10 projects as there are many more projects in a ordered list by title before "TG16/17". Maybe the data itself is the cause for this trouble - but I just guessing.

BartChris:
@henning-gerhardt One problem i discovered is that we are sorting on the tokenized title field, which can lead to problems. We get more consistent results by sorting on title.keyword  by setting

```
private static final String SORT_TITLE = "title.keyword";
```

instead of 
```
private static final String SORT_TITLE = "title.keyword"; 
```
in `DesktopForm.java`.

```
 public List<ProjectDTO> getProjects() {
        try {
            if (ServiceManager.getSecurityAccessService().hasAuthorityToViewProjectList() && projectList.isEmpty()) {
                projectList = ServiceManager.getProjectService().loadData(0, 10, SORT_TITLE, SortOrder.ASCENDING, null);
            }
        }
```

But that still does not solve all problems since lowercase letters are sorted after Uppercase letters:

image

So  to solve that we would have to change the index defintion to something like this:

```json
"title": {
  "type": "text",
  "fields": {
    "keyword": {
      "type": "keyword",
      "ignore_above": 256
    },
    "sort": {
      "type": "keyword",
      "normalizer": "lowercase_normalizer"
    }
  },
  "fielddata": true
}
```

Maybe it would be better to not use Elasticsearch here but the database.

henning-gerhardt:
We at SLUB did not used mixed up project titles - all our projects starts with a capitalized letter. But in general this could be an issue. Changing the search index could be difficult and must be documented for all possible use cases so that everyone can use it. Switching to a database only search can be the way to go especially as we change to a Hibernate-Search based in any way but I don't know if this change is "small" enough for a 3.8.x release and for any later 3.9.x release this is not or should not be needed. Maybe @solth can decide that before an implementation is done or tried?

BartChris:
By using `title.keyword` we at least bring the Desktop form in line with the full project list which also sorts on` title.keyword`, which means that the sorting on the full project list is also not working correctly right now. I am not sure if we change more in 3.8. When doing that we maybe do not even have to sort the list any more.

BartChris:
In general: When using capital letters sorting is not really working.

image

But since projects are not indexed anymore, when using Hibernate Search, the fix would be 3.8 only.

henning-gerhardt:
> In general: When using capital letters sorting is not really working.

Sorting works in this way: capital letters followed by normal letters:

Screenshot_2025-04-07_15-08-52

henning-gerhardt:
I tried your latest changes and it is now working for me. It could be minimized to the change in

* DesktopForm.java: `SORT_TITLE = "title"`  -> `SORT_TITLE = "title.keyword"` 

No other change is needed. 

This change even add a more understandable sorting in the task list in the desktop view as both lists are sorted by the same property value (`SORT_TITLE`).

BartChris:
I did a new 3.8-only PR: https://github.com/kitodo/kitodo-production/pull/6502

PR #6532, opened 2025-05-09
Fix order of 'actual' and 'expected' values of 'assert' method in ImportService test

solth:
Fixes order of parameters of some assert statements made necessary by switching from Junit version 4 to version 5.


henning-gerhardt (approved):
Changes looks correct.

henning-gerhardt:
@solth: I run with my IDEA a check for this kind of issue "wrong order of arguments in assertEquals()". Should I fix them in a separate request or should I put this list here?

solth:
> I run with my IDEA a check for this kind of issue "wrong order of arguments in assertEquals()". Should I fix them in a separate request (maybe against your branch?) or should I put this list here?

Please feel free to open a pull request with the remaining fixes!

PR #6539, opened 2025-05-15
Expand textareas on page load

BartChris:
This PR tries to make a quality of life improvement for disabled textareas. Those textareas might contain text which spans over multiple lines. If the textarea has editing disabled, the box cannot be expanded and the editor does not see the full content. I am using Javascript to expand the textareas to the necessary size.

With this change the editor always sees the full content.

image

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataTreeTable.xhtml
@@ -85,6 +85,15 @@
                                      rows="2"
                                      disabled="#{not item.editable or readOnly}"
                                      styleClass="#{not item.editable or readOnly ? 'read-only disabled' : ''}">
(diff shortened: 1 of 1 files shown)


oliver-stoehr (commented):
The script works fine and displays the textAreas with full height.
Is resetting the height really necessary? I also tested the script without `textarea.style.height = "auto"; // Reset height first` which worked in all browsers.
Is there a special scenario where this line is necessary?

BartChris:
I removed that line, which might be necessary, if the box needs to be shrinked if an initial size is larger than the actual content. I think we do not have this scenario and therefor removed it. 

I added another check to ensure the listener is only registered once when having multiple textareas on the screen.

oliver-stoehr (approved):

solth:
@BartChris perhaps it would make sense to backport this QoL improvement for Kitodo.Production 3.8 as well?

PR #6542, opened 2025-05-23
Filter role list by client

### LINKED ISSUE #6261: Role list not filtered by client
**Describe the bug**
Nearly all lists in Kitodo.Production are filtered by the users currently selected client. Two exceptions are the "Clients" list itself and the "Roles" list on the "Users" page. While it can be argued that the "Clients" list is a justified exception, the "Roles" list should definitely be filtered by client. Instead it is actually _sortable_ by clients, but that is of little use and discloses information that shouldn't be visible to the user in this situation, since there is always a specific client selected in the system that all displayed information should be based upon.

**To Reproduce**
1. Go to "Users" page
2. Click on tab "Roles"
3. See error

**Expected behavior**
The role list should be filtered by client, like all other lists in the system.

solth:
Fixes #6261 bei filtering the role list by current client. 

As a compromise to not make the work for administrators more difficult (see https://github.com/kitodo/kitodo-production/issues/6261#issuecomment-2401746464) a switch has been added to allow returning to the unfiltered role list:

Bildschirmfoto 2025-05-23 um 14 48 53

This switch is not displayed when the current user is assigned to only one client. 

I hope this sufficiently addresses the valid concerns you were raising, @henning-gerhardt !

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/UIInteractionsST.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/UIInteractionsST.java
@@ -97,6 +97,7 @@ public class UIInteractionsST extends BaseTestSelenium {
         assertEquals(rolesInDatabase, rolesDisplayedFiltered, "Displayed wrong number of roles filtered");
 
         Browser.getDriver().findElement(By.cssSelector(FILTER_ROLES_SWITCH_SELECTOR)).click();
+        Thread.sleep(Browser.getDelayAfterPickListClick());
 
         // verify that switch to show other clients roles is off after clicking
         WebElement roleSwitchClicked = Browser.getDriver().findElement(By.cssSelector(FILTER_ROLES_SWITCH_SELECTOR));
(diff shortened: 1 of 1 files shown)


henning-gerhardt (changes requested):
Code changes looks good, functionality works in most but not all cases:
- Sorting by role title is possible for all clients and only current client
- Sorting by client name is possible for all clients but throws an exception on current client only

Thrown exception on current client limited by sorting roles by client:

```
Exception message: org.hibernate.hql.internal.ast.QuerySyntaxException: Invalid path: 'client.name' [SELECT r FROM org.kitodo.data.database.beans.Role AS r INNER JOIN r.client AS c WITH c.id = :clientId ORDER BY client.name ASC]
```

Stack trace:

```
java.lang.IllegalArgumentException: org.hibernate.hql.internal.ast.QuerySyntaxException: Invalid path: 'client.name' [SELECT r FROM org.kitodo.data.database.beans.Role AS r INNER JOIN r.client AS c WITH c.id = :clientId ORDER BY client.name ASC]
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:138)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:181)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:188)
	at org.hibernate.internal.AbstractSharedSessionContract.createQuery(AbstractSharedSessionContract.java:757)
	at org.hibernate.internal.AbstractSharedSessionContract.createQuery(AbstractSharedSessionContract.java:114)
	at org.kitodo.data.database.persistence.BaseDAO.getByQuery(BaseDAO.java:182)
	at org.kitodo.production.services.data.RoleService.loadData(RoleService.java:98)
	at org.kitodo.production.model.LazyRoleModel.load(LazyRoleModel.java:58)
	at org.primefaces.component.datatable.DataTable.loadLazyData(DataTable.java:567)
	...
```

Question: Why is sorting by client name possible or better: client column still displayed if not all possible roles of all clients are shown and only the roles from the current client are shown?

Edit: an other question: until now the current behaviour should be made by default and not switching the new behaviour. If this is still valid why are the roles are limited by current client by default instead of showing all roles of all clients?

PR #6558, opened 2025-06-03
Provide default hibernate.properties file

### LINKED ISSUE #6552: [Regression Hibernate-Search] No default hibernate.properties file for Hibernate Search usage
**Describe**
Before merging Hibernate Search it was possible to define different options for the search index usage like on which URI the search index is available or on which port or how to authenticate on the search instance. As this and other options can only be defined in a `hibernate.properties` file a default file with default values should be provided in any case as this file with some content was added to different source code directories but not as a file in `Kitodo/src/main/resources` directory with working configuration values. 

This values are not needed if you run everything on a local host instance but you need to know that this file must exists and contain correct key-value entries to get in on non local host environments working. Even more - like already described in #6551 - there are left overs which may indicate that this values still work but this is not the case!

henning-gerhardt:
Fixes #6552 

Even Kitodo.Production is working without this file as this all default values the file is needed for some other in future implemented changes.

### DIFF
--- a/Kitodo/src/main/resources/hibernate.properties
+++ b/Kitodo/src/main/resources/hibernate.properties
@@ -1,0 +1,30 @@
+#
+# (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+#
+# This file is part of the Kitodo project.
+#
+# It is licensed under GNU General Public License version 3 or later.
+#
+# For the full copyright and license information, please read the
+# GPL3-License.txt file that was distributed with this source code.
+
+# The Hibernate Search integration into Hibernate ORM will start automatically,
+# at the same time as Hibernate ORM, as soon as it is present in the classpath.
+# If for some reason you need to prevent Hibernate Search from starting, set
+# the boolean property hibernate.search.enabled to false.
+#
+# Kidodo.Production works when this value is set to True.
+
+hibernate.search.enabled=true
+
+
+# the hostname and ports of the Elasticsearch servers to connect to
+
+hibernate.search.backend.hosts=localhost:9205
+
+
+# You may change the protocol used to communicate with the hosts using this
+# configuration property. The default for this property is http. This property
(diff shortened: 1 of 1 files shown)


BartChris:
After reading the docs: Would it maybe also be possible to add this options to the hibernate.cfg.xml (I have not tested it)? Or would you rather use the hibernate.properties?

https://docs.jboss.org/hibernate/stable/search/getting-started/orm/en-US/html_single/#getting-started-configuration

<property name="hibernate.search.enabled">true</property>
<property name="hibernate.search.backend.type">elasticsearch</property>
<property name="hibernate.search.backend.hosts">localhost:9200</property>
<property name="hibernate.search.backend.protocol">http</property>

henning-gerhardt:
It is new to me that this is working and I can confirm this itself. I tested it while the Hibernate-Search development was on-going and I asked Matthias why he added the `hibernate.properties` file additional to the `hibernate.cfg.xml` file. The answer was that HibernateSearch is only working with `hibernate.properties` file and the options belong to this file. At this time in the past I tried to move this HibernateSearch properties to `hibernate.cfg.xml` file but was not successful. So I thought that we need this file and tried it not any more until now. But now I see that the application is working with only a `hibernate.cfg.xml` file which is containing the HibernateSearch options.

So my question in return is: Did we need the `hibernate.properties` file at all and should we not return to a single Hibernate configuration file?

henning-gerhardt:
Opposite pull request is #6571.

BartChris:
I prefer going with the single config file (so https://github.com/kitodo/kitodo-production/pull/6571), as it simplifies things. If it works for you i think we are good to go with that.

solth:
I agree. If both variants indeed achieve the same, I would prefer adding some parameters to the already existing `hibernate.cfg.xml` files (#6571) over having additional, identical `hibernate.properties` files (#6558)

Issue #6531, opened 2025-05-09
wrong automatic value for LDAP login on User creation


domerz:
**Describe the bug**
In some configurations when creating a new User the text field "LDAP login" is automatically filled with the current user's login name. This is  wrong in any case when admins create those users. Even worse: when you delete the entry, change the tab (e.g. to specify roles) and come back to the first tab, the default (wrong) value will reappear.
This is most probably due to the browser mistaking "LDAP login" for a standard login field and auto-filling login data for the instance. It won't happen when saving login data is switched off in the browser.

**Expected behavior**
Do not automatically fill the "LDAP login"-field. Or at least: Do not make the browser do that.

**Release**
3.8.1

**Tested Browser**
Microsoft Edge

PR #6559, opened 2025-06-03
Prevent LDAP autocomplete

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6531

One might call it a dirty hack, but injecting two (by using extreme absolute values) invisible boxes was the only reliable way for me to prevent the behaviours of browsers like chrome to autofill the ldap login and the password field.

Using `pt:autocomplete="off"` for injecting autocomplete=off directly in the browser does not work. (using xmlns:pt="http://xmlns.jcp.org/jsf/passthrough" namespace declaration)

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/userEdit/details.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/userEdit/details.xhtml
@@ -51,7 +51,7 @@
                              value="#{UserForm.userObject.location}"/>
             </div>
             <!--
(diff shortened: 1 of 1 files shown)


BartChris:
I tried to prevent this using Javascript and other markup based techniques, without success so far. The agressive autofill behaviour combined with JSF speficis is hard to tackle...

oliver-stoehr (approved):
Since chrome seems to ignore all `autocomplete` values for password inputs, this might be the only way to fix it.
This solution does indeed feel like a hack, but I think it's fine with the comment.

Issue #6544, opened 2025-05-26
No synchronisation between database and search index values


henning-gerhardt:
**Describe the bug**
For certain fields in the database and used search index is no synchronisation of its values. This will become important after hibernate search has replaced the own developed search integration. It is not the fault of the hibernate search development (#6504) as this synchronisation should have been correct even with the before used search index integration.

This an issue placeholder for the result and way of discussion #6543.

**To Reproduce**
See some examples in the discussion #6543.

**Expected behavior**
Values which should be synchron should be syncron.

**Release**
Maybe since first offical 3.x release but sure with 3.6.0

PR #6572, opened 2025-06-18
[3.8] Fix amount of internal meta information

henning-gerhardt:
Partial fix for #6543 (background information and discussion) and #6544 (general placeholder issue for the issue) based on the 3.8.x branch.

With this changes all updates to the process object should reflect current amount values from the internal meta data file to the database. It is similar to the way how this data is stored in the search index. 

Only "big" change it that only referenced files from the meta data file are counted and not like before all files in the defined directories.

I tried to minimize reading of the needed meta information from the file system the amount of updates to the database but it could be that there a few more than before.

I will port this changes to the main branch after this pull request is merged as the issue must be fixed in the main branch too.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -2896,7 +2896,7 @@ public class ProcessService extends ProjectSearchService<Process, ProcessDTO, Pr
     }
 
     /**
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
Workflow CI runs are failing because of new OS image builds in GitHub. Fixing paths in used shell scripts should fix this. I will open pull requests in 3.8.x branch and main branch when they are working.

henning-gerhardt:
I opened #6580 and #6581 to fix this build issue on 3.8.x and main branch.

Issue #6191, opened 2024-08-26
Mass import: Switching CSV separator character discards manually added metadata columns


solth:
**Describe the bug**
After uploading a CSV file in the mass import form, any additionally added metadata columns added manually using the "+" button are discarded when switching the CSV separator character.

**To Reproduce**
1. Go to 'Mass import'
2. Upload an appropriate CSV file 
3. Manually add metadata column using the "+" button in the table header
4. Change the "CSV separator character"
5. See error

**Expected behavior**
Changing the CSV separator character should not discard other new metadata added manually via the "+" button.

**Screenshots**
![Bildschirmaufnahme 2024-08-26 um 11 24 48](https://github.com/user-attachments/assets/b014391e-a588-4a13-b3d7-6800012006cb)

**Release**
3.7 and prior

**Additional context**
see https://github.com/kitodo/kitodo-production/wiki/Massenimport for details on how to use the mass import feature in Kitodo.Production

PR #6574, opened 2025-06-20
Implement process creation from data contained in uploaded CSV file without catalog search

### LINKED ISSUE #6391: Mass import without catalog search
The metadata mass import of Kitodo.Production allows uploading a CSV file that contains a list of IDs, which are used to query a configured OPAC search interface. Each data record queried by ID from the catalog results in one Kitodo process. Optionally, the CSV file can contains metadata which are used to augment the catalog metadata of the individual records. It is currently _not_ possible, though, to create processes without querying a search interface and just with the data contained in the CSV file. 

The mass import should allow creating multiple processes from a CSV file without the need to query a catalog search interface, e.g. all metadata to be added to a process is already included in the uploaded CSV file.

**Additional context**
_This development is funded by the [WLB Stuttgart](https://github.com/WLB-IT)_

solth:
This pull request enhances the mass import to allow creating processes directly from data contained in uploaded CSV files, foregoing the requirement to query catalog search interfaces with record identifiers to import metadata. 

To use this functionality, the first column in the first row of an uploade CSV file has to contain the key of a metadata configured as functional metadata `docType` and the first column of each following row - each representing an individual data record to be transformed into a Kitodo process - should contain the key of a division from the current ruleset ("Art des Inhalts" is configured as `docType` in the ruleset used for the following screenshot). In this case, the label in the header of the first column will contain a special icon to show that a metadata of type `docType` has been found. Note that the "Catalog" selection is deactivated when the first column contains document type information:

Bildschirmfoto 2025-06-20 um 13 51 25

The previous functionality of importing metadata from catalog search interfaces still exists, though, and can be triggered by the first row of the first column in the uploaded CSV file containing the key of a metadata set up as a functional metadata of type `recordIdentifier` instead of `docType`. The header of the first column will contain a different symbol to signal a valid functional metadata of type `recordIdentifier` has been found. The catalog selection will only be activated when this is the case:

Bildschirmfoto 2025-06-20 um 14 08 31

If the first column of the uploaded CSV file contains neither functional metadata of type `docType` nor of type `recordIdentifier`, the header of the first column in the record table will be marked in red to signal a wrong configuration and the "Mass import" button will be deactivated:

Bildschirmfoto 2025-06-20 um 14 00 27

A popup dialog will be displayed when opening the mass import to explain the usage described above. This dialog can be hidden for the rest of the session by marking the checkbox in the lower left of the dialog.

Bildschirmfoto 2025-06-20 um 14 10 57

Fixes #6391


Kitodo/src/main/java/org/kitodo/production/forms/LoginForm.java
```diff
@@ -36,6 +36,27 @@ public class LoginForm implements Serializable {
...
+
+    /**
+     * Set value of 'hideMassImportExplanationDialog', indicating whether the explanation dialog for the mass import
+     * should be shown during this session or not.
+     *
+     * @param hideMassImportExplanationDialog whether explanation dialog should be shown or not
+     */
+    public void setHideMassImportExplanationDialog(boolean hideMassImportExplanationDialog) {
+        this.hideMassImportExplanationDialog = hideMassImportExplanationDialog;
+    }
```
henning-gerhardt, line 59: Should this not go to the `HelperForm` class? There are - in my opinion - already to many things it this class which does not belong in general to the login itself.

Kitodo/src/main/resources/messages/messages_en.properties
```diff
@@ -738,8 +738,13 @@ mappingFile.outputFormat=Metadata output format
 mappingFile.selectFormats=Please select input and output file formats for mapping file
 massDownload=Bulk downloads
 massImport=Mass import
+massImport.explanation=You can either query metadata from a search interface for each record in the datatable below or directly create processes just from the data in the datatable.
+massImport.explanation.recordIdentifier=If you want to import metadata from a search interface, the first column in the table has to contain functional metadata configured as '<b>recordIdentifier</b>'.
+massImport.explanation.docType=Directly converting the data from the table into processes requires the first column to contain function metadata configured as '<b>docType</b>' instead.
```
BartChris, line 743:
```suggestion
massImport.explanation.docType=Directly converting the data from the table into processes requires the first column to contain functional metadata configured as '<b>docType</b>' instead.
```

Kitodo/src/main/java/org/kitodo/production/services/data/MassImportService.java
```diff
@@ -99,10 +107,17 @@ public List<CsvRecord> parseLines(List<String> lines, String separator) throws I
                      .withSkipLines(0)
                      .withCSVParser(parser)
                      .build()) {
+            int targetNumberOfColumns = -1;
             for (String[] entries : csvReader.readAll()) {
                 if (isSingleEmptyEntry(entries)) {
```
BartChris, line 112: This is already in the old code, but i noticed that lines like 
```
;;;;;;;;;;;;;;;;;;
```
get included in the table as empty lines, maybe we should adapt the check to exclude all empty lines:

```java
 for (String[] entries : csvReader.readAll()) {
                if (isLineCompletelyEmpty(entries)) {
                ..
                }
}

private boolean isLineCompletelyEmpty(String[] entries) {
    return entries.length == 0 || 
           Arrays.stream(entries).allMatch(StringUtils::isBlank);
}
```

Kitodo/src/main/webapp/pages/massImport.xhtml
```diff
@@ -39,7 +39,7 @@
                   styleClass="secondary"/>
         <p:commandButton id="importButton"
                          value="#{msgs.massImport}"
-                         disabled="#{empty MassImportForm.records or not MassImportForm.getRulesetConfigurationForOpacImportComplete()}"
+                         disabled="#{empty MassImportForm.records or not MassImportForm.getRulesetConfigurationForOpacImportComplete() or not MassImportForm.firstColumnContainsRecordIdentifierOrDocType()}"
```
BartChris, line 42: I stumbled over this already, as i - after first put 'ID' in the column - put 'recordIdentifier' in the first column. After reading the description again i got that you have the name of the actual ruleset field which has the functional key applied. Probably just a matter of documentation, but it would be great if there was feedback, why exactly the button is disabled, some message like:

"The metadatum in the first column is not marked as a functional metadata of type `docType` or `recordIdentifier`.

BartChris:
Switching to an non existing seperator

image

shows an error message, but only for the split of a second.

BartChris:
The ajax call for checking the progress of the mass import gets triggered even when the import is not actually started. When uploading csv records and clicking on mass import without selecting a catalogue, no import is started but, Kitodo starts polling indefinitely. 

image

BartChris:
When adding a new row and filling in values by hand the whole row gets discarded when "Skip empty colums is selected"-
image

Kitodo/src/main/java/org/kitodo/production/services/data/MassImportService.java
```diff
@@ -99,10 +107,17 @@ public List<CsvRecord> parseLines(List<String> lines, String separator) throws I
...
                     continue; // Skip processing this line
                 }
+                // throw exception different number of metadata entries were parsed for different CSV lines
+                if (targetNumberOfColumns >= 0 && entries.length != targetNumberOfColumns) {
+                    throw new KitodoCsvImportException(Helper.getTranslation("massImport.separatorCountMismatchEntries",
+                            Pattern.quote(separator)));
+                }
+                targetNumberOfColumns = entries.length;
```
BartChris, line 120: Why is the targetNumberOfColumns determined for every row? Is it not enough to calculate it once - for the first row and then compatre every following row to it?

Kitodo/src/main/resources/messages/messages_en.properties
solth: Done

Kitodo/src/main/java/org/kitodo/production/services/data/MassImportService.java
solth: You are right. I added the additional check to discard completely empty lines.

solth:
> Switching to an non existing seperator
> 
> image
> 
> shows an error message, but only for the split of a second.

This was caused by the import progress being updated every second, which you also observed. I now prevent this by keeping the "Start Mass import" button deactivated until a catalog has been selected (in case the first column contains `recordIdentifier` for opac metadata import.

Kitodo/src/main/java/org/kitodo/production/forms/massimport/MassImportForm.java
```diff
@@ -113,17 +121,75 @@ public void handleFileUpload(FileUploadEvent event) {
...
+            updateMetadataKeys();
+            records = massImportService.parseLines(importedCsvLines, csvSeparator.getSeparator());
+            boolean success = true;
+            if (!records.isEmpty() && !records.get(0).getCsvCells().isEmpty()
+                    && records.get(0).getCsvCells().size() != metadataKeys.size()) {
+                Helper.setErrorMessage(Helper.getTranslation("massImport.separatorCountMismatchHeader", csvSeparator.toString()));
+                records = new LinkedList<>();
+                success = false;
+            }
+            if (metadataKeys.size() == 1) {
```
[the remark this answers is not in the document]
solth: I agree. I do not remember if this code fragment was perhaps some kind of "residue" I used during development and which I forget to remove before opening the pull request, but I think it should be totally valid to upload a CSV file with just one column (if this column contains the required functional metadata of type `recordIdentifier` or `docType`), so I removed this check.

Kitodo/src/main/java/org/kitodo/production/forms/massimport/MassImportForm.java
```diff
@@ -113,17 +121,75 @@ public void handleFileUpload(FileUploadEvent event) {
...
+                Helper.setWarnMessage(Helper.getTranslation("massImport.separatorMissing", csvSeparator.toString()));
+                success = false;
+            }
+            if (success && skipEmptyColumns) {
+                List<Integer> skipIndices = ServiceManager.getMassImportService().getColumnSkipIndices(records, metadataKeys.size());
+                if (!skipIndices.isEmpty()) {
+                    metadataKeys = discardMetadataKeysOfEmptyColumns(metadataKeys, skipIndices);
+                    records = discardEmptyColumns(records, skipIndices);
+                }
+            }
```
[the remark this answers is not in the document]
solth: Yes, I think this would work, but than we need the whole business logic in the backing bean, which I would like to avoid. That's the reason I moved the corresponding methods to the `MassImportService` and therefore couldn't facilitate side effects directly changing values of bean properties in a private method. 

I am unsure if this change would have a real effect on performance and whether we chould forego coding guidelines for it. @BartChris what do you think?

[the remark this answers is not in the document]
BartChris: You are raising good points. I am unsure about performance. Most of the time people will not import more than 2000 records, which should be very fast process. 
Would it been an option to use a static Service method in the MassImportService which just receives the beans and returns a new list and does not mutate the properties directly?

BartChris:
> > When adding a new row and filling in values by hand the whole new inserted row gets discarded when "Skip empty colums" is selected. The row is also discarded when switching the seperator.
> > image
> 
> Yes, this is a known issue - and not restricted to the newly introduced "Skip empty columns" switch. I documented the same problem for changing the the separator character in #6191 before.
> 
> Perhaps I can come up with a fix, but if this turns out to be too complex, I would prefer to deal with that issue in a separate pull request. @BartChris Would you be ok with that?

Yeah,totally fine with that. When thinking about that, i am not even sure if the "Add record" button is a required features -this is a csv import in the end.Especially if it adds much complexity.

Kitodo/src/main/java/org/kitodo/production/forms/massimport/MassImportForm.java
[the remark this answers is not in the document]
solth: > Would it been an option to use a static Service method in the MassImportService which just receives the beans and returns a new list and does not mutate the properties directly?

I am not sure this would work, as we need to update not one, but _two_ lists (`records` and `metadataKeys`) with the corresponding indices, so we can't just work with a simple list as return value here. Passing the whole bean (or a clone of it), just to access and modify those two lists and then retrieving the updated lists from the cloned form instance seems a little awkward to me.

Kitodo/src/main/webapp/pages/massImport.xhtml
BartChris: nevermind, sorry, it works.

solth:
> When using the CSV import it is not easy to see from the result list, which entry failed. image
> 
> Maybe make a "identifying" column selectable in case of a csv data only import, which could be put as Identifier in the statusmap.

I forgot about this comment, sorry. 

I am not sure adding _another_ new configuration or functional metadata just to be shown in this result list isn't a little oversized of an solution, but perhaps we could include metadata from potentially provided `recordIdentifier` metadata (which can be included when creating processes only from the uploaded CSV data without catalog search, albeit then not in the first column!), if a corresponding column exists. That should be a good indicator to identify records.
@BartChris What do you think?

BartChris:
> I am not sure adding _another_ new configuration or functional metadata just to be shown in this result list isn't a little oversized of an solution, but perhaps we could include metadata from potentially provided `recordIdentifier` metadata (which can be included when creating processes only from the uploaded CSV data without catalog search, albeit then not in the first column!), if a corresponding column exists. That should be a good indicator to identify records. @BartChris What do you think?

Sounds very good to me!

solth:
> Sounds very good to me!

@BartChris I now incorporated available record identifiers into the result list, which should be more informative now:

Successful imports:
Bildschirmfoto 2025-07-09 um 14 55 31

Failed imports:
Bildschirmfoto 2025-07-09 um 14 45 09

solth:
@BartChris does this solution work for you? One potential small change could be the counter I added to the first column (that you can see in the two screenshots about). As you can see, the counter does not reference the rows in the two datatables "failed imports" and "successful imports", but rather denotes the number of the records in the original datatable of records to be imports (thus those number missing from one result table are contained in the other).

After thinking about this a little more, this might not be as helpful as I intended and perhaps is even a little confusing. Perhaps it would make more sense to just count each result table indidivually, thus resulting with the following tables (identical situation as described above):

Successful imports:
Bildschirmfoto 2025-07-09 um 16 11 45


Failed imports:
Bildschirmfoto 2025-07-09 um 16 12 10

BartChris:
Did your changes make it mandatory to have a key in the pure csv import, which is marked as `recordIdentifer` in the ruleset. 
I get the following exception, when providing only columns with keys which are not a `recordIdentifier`.

image

I think that should still be allowed - the `recordIndentifier` is only a **optional** info to display something in the result list and not should not be mandatory for a csv only import.

solth:
> Thanks a lot for pointing this out. This change was not intended. I fixed the import so that `recordIdentifier` is not mandatory anymore during process creation solely from uploaded CSV files.

@BartChris I think the quick solution I just added - just catch the `ConfigException` caused by the missing `recordIdentifier` - my not be optimal since it potentially adds thousand of unnecessary entries to the logs files. I will change this by adding another parameter to `getRecordId` to control whether an exception should be thrown or not, if no record identifier was found.

solth:
@BartChris I hope the current solution works for you. Thanks again for your review

BartChris:
If there are no records in the successMap or in the failureMap we always get an english message:
`No records found.`, which does not get translated.

The message can be localized by adding a localized string emptyMessage fallback to both tables (Hardcoded in german for demonstration):

 <p:dataTable value="#{MassImportForm.failedImports}"
                                 rowIndexVar="index"
                                 var="importResult"
                                 emptyMessage="Keine Datensätze gefunden">

<p:dataTable value="#{MassImportForm.successfulImports}"
                                 rowIndexVar="index"
                                 var="importResult"
                                 emptyMessage="Keine Datensätze gefunden">

BartChris (approved):
As far as i can see, it looks very good!

Issue #6551, opened 2025-06-02
Leftovers from ES / OS usage


henning-gerhardt:
**Describe the bug**
With merging the Hibernate Search pull request there are some left overs from former ES / OS usage. This are

- configuration entries in different `kitodo_config.properties` files in the source code
- enumeration values defined in `Kitodo/src/main/java/org/kitodo/config/enums/ParameterCore.java` and used from there in different classes from the defined configuration values in `kitodo_config.properties`
- maybe direct used configuration entries from `kitodo_config.properties` files in classes

PR #6577, opened 2025-06-24
Add optional configurable search index prefix

### LINKED ISSUE #6550: [Regression Hibernate-Search] Reenable search index prefix

Before merging the Hibernate Search pull request it was possible to define a own search index prefix to distinguish between different search indexes on one search instance. Maybe this was only used on development stage but even there really helpful. The entry `elasticsearch.index` in `kitodo_config.properties` still exists but is not working any more.

Hibernate Search offer such a feature not in a direct way but you can intercept the default usage in a good way by providing a class which made this name schema handling editable.

henning-gerhardt:
Fixes #6550 

Without enabling this option the access / usage behaviour to current used search index is not changed as the implementation is identical except that is possible to define an prefix in `kitodo_config.properties`.

Until #6551 is not solved there are now two options to define an prefix (`elasticsearch.index` and `searchindex.prefix`) but only `searchindex.prefix` is working if this pull request is merged.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/index/KitodoCustomLayoutStrategy.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/index/KitodoCustomLayoutStrategy.java
@@ -50,7 +50,7 @@ public class KitodoCustomLayoutStrategy implements IndexLayoutStrategy {
 
     @Override
     public String extractUniqueKeyFromElasticsearchIndexName(String elasticsearchIndexName) {
-        String patternString = getSearchIndexPrefix() + "(.*)-\\d{6}";
+        String patternString = Pattern.quote(getSearchIndexPrefix()) + "(.*)-\\d{6}";
         Pattern uniqueKeyExtractionPattern = Pattern.compile(patternString);
         Matcher matcher = uniqueKeyExtractionPattern.matcher(elasticsearchIndexName);
         if (!matcher.matches()) {


Kitodo-DataManagement/src/main/java/org/kitodo/data/index/KitodoCustomLayoutStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class KitodoCustomLayoutStrategy implements IndexLayoutStrategy {
+
+    private String getSearchIndexPrefix() {
+        return ConfigMain.getParameter("searchindex.prefix", "");
+    }
+
+    @Override
+    public String createInitialElasticsearchIndexName(String hibernateSearchIndexName) {
+        return getSearchIndexPrefix() + hibernateSearchIndexName + "-000001";
```
thomaslow, line 33: Adding `-000001` seems to be related to the "Index Lifecycle Management" module of ElasticSearch, in case the index is configured to "rollover" if it has reached a certain limit (e.g., the number of indexed documents, size in GB), similar to Linux log files can be configured to rollover every day or after a certain size is reached. 

Do we need that? I investigated this feature a bit and I don't think this works out-of-the-box at all. It has to be specifically configured in [ElasticSearch](https://www.elastic.co/docs/reference/elasticsearch/curator/rollover) and there does not seem to be full support for it in [HibernateSearch](https://hibernate.atlassian.net/browse/HSEARCH-3499) yet. Meaning, rewriting aliases to different indexes would need to be implemented manually.

If there is no reason for it (other than it was implemented like this in the example class provided by HibernateSearch), I think it would be better to remove `-000001` to not confuse some admins that might think Kitodo.Production uses this particular feature of ElasticSearch.

```suggestion
        return getSearchIndexPrefix() + hibernateSearchIndexName;
```

Kitodo-DataManagement/src/main/java/org/kitodo/data/index/KitodoCustomLayoutStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public String extractUniqueKeyFromHibernateSearchIndexName(String hibernateSearchIndexName) {
+        return getSearchIndexPrefix() + hibernateSearchIndexName;
+    }
+
+    @Override
+    public String extractUniqueKeyFromElasticsearchIndexName(String elasticsearchIndexName) {
+        String patternString = getSearchIndexPrefix() + "(.*)-\\d{6}";
```
thomaslow, line 53: Including the prefix in the pattern string could cause issues if the prefix contains characters that are interpreted as part of the regular expression.

```suggestion
        String patternString = Pattern.quote(getSearchIndexPrefix()) + "(.*)(-\\d{6})?";
```

Kitodo-DataManagement/src/test/java/org/kitodo/data/index/KitodoCustomLayoutStrategyTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.kitodo.config.ConfigMain;
+import org.mockito.MockedStatic;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class KitodoCustomLayoutStrategyTest {
+
+    private static final String HIBERNATE_SEARCH_INDEX_NAME = "kitodo-process";
+    private static final String SEARCH_INDEX_PREFIX = "testprefix-";
+    private static final String SEARCH_INDEX_NAME = "kitodo-process-000001";
```
thomaslow, line 29:
```suggestion
    private static final String SEARCH_INDEX_NAME = "kitodo-process";
```

Kitodo-DataManagement/src/main/java/org/kitodo/data/index/KitodoCustomLayoutStrategy.java
thomaslow: I didn't realize this was the default naming strategy of hibernate search. I guess this is a good enough reason to keep it that way.

Issue #6573, opened 2025-06-19
Using the catalog import while creating a new process results in workpiece and process property duplication


lutzhelm:
**Describe the bug**
When a new process is created and data is imported via catalog search, after saving the process, workpiece and process properties are duplicated.

**To Reproduce**
1. Got to "Projects", look for correct project (it needs to have an import configuration)
2. Click "Create new project"
3. Search catalog for required entry
4. After successful import, fill out additional required fields and save
5. Go to process, check "Workpiece properties" and "Properties"

**Expected behavior**
No duplication of properties.

**Screenshots**
The error is obvious in the process property tab; it can also be seen in the workpiece property tab.
![Image](https://github.com/user-attachments/assets/bfcb47a2-d7dd-4cc2-ac6c-01a9b2f07ec2)

SQL result for process properties:
```
MariaDB [kitodo]> select p.id, p.title, p.value, p.creationDate from process_x_property wp join property p on wp.property_id=p.id where wp.process_id=26051 order by p.title;
+--------+------------+-------+---------------------+
| id     | title      | value | creationDate        |
+--------+------------+-------+---------------------+
| 497488 | Template   | VD16  | 2025-06-19 14:51:23 |
| 497490 | Template   | VD16  | 2025-06-19 14:51:23 |
| 497489 | TemplateID | 4     | 2025-06-19 14:51:23 |
| 497491 | TemplateID | 4     | 2025-06-19 14:51:23 |
+--------+------------+-------+---------------------+
4 rows in set (0.001 sec)
```

SQL result for workpiece properties:
```
MariaDB [kitodo]> select p.title, p.value from workpiece_x_property wp join property p on wp.property_id=p.id where wp.process_id=26051 order by p.title;
+-------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| title                                     | value                                                                                                                                                                |
+-------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Alternativtitel                           | IOANNIS REVCHLIN PHORCENSIS LL. DOCTORIS COmoediae duae, Scenica progymnasmata, hoc est, ludicra praeexercitamenta. Et Sergius uel Capitis caput.                    |
| Alternativtitel                           | Sergius                                                                                                                                                              |
[...]

PR #6582, opened 2025-06-30
Prevent duplicate creation of process properties

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6573


henning-gerhardt (approved):
After I had some issues to reproduce this issue in general (main data did not have duplicated process properties but after I created new processes I had duplicated process properties) I tested your changes with a monograph and a multi volume work with one volume as a child process. With your changes there are no duplicated process properties in new created processes.

It should be possible to back port this solution to the 3.8.x branch so that in a new 3.8.x version this issue is fixed there too and not more processes with duplicated process properties are created.

PR #6583, opened 2025-07-01
Fix adding physical divisions

oliver-stoehr:
The handling of physical metadata in the separated structure tree is buggy:

- When adding a new physical division or changing the selection, physical metadata that was changed beforehand is not preserved.
- When adding a new physical division with default metadata values (configured in the ruleset via `min-occurs` or `always-showing` and `preset`) these metadata values are not preserved when left unchanged.  
(Changing one value would preserve all metadata of that division. But if none is changed, none of the new division's metadata is preserved, leading to missing metadata after saving.)

This pull request fixes both issues.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddPhysicalDivisionDialog.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/AddPhysicalDivisionDialog.java
@@ -22,9 +22,14 @@ import java.util.Optional;
 
 import javax.faces.model.SelectItem;
 
-import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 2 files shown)


oliver-stoehr:
Hi Thomas, 

you are right regarding the first point. I cannot reproduce the error on the main branch. I must have mixed up something with another branch where I first observed this problem. Something else in that branch must cause this problem.

The second point is reproducible with these steps:
1. Open the process in the metadata editor with separate structure trees (Use a ruleset where a division has a key with a preset and is configured with `minOccurs="1"` or `"alwaysShowing="true"`.)
2. Add the new physical division (e.g. via context menu on the root node)
3. Directly click the "Save" button. **Do not** actively select the new division beforehand.
4. Look into the meta.xml. The new division does not have any metadata nor a DMD section.

(When you now actively select the new division in the editor you can see the metadata. Since you just clicked "Save" you would assume that this metadata was saved, since you did not change anything.)

thomaslow:
I think it should be fine, but I'll have to test it again.

PR #6586, opened 2025-07-04
[3.8] Update internal meta information through user interface

henning-gerhardt:
Partial fix for https://github.com/kitodo/kitodo-production/discussions/6543 (background information and discussion) and https://github.com/kitodo/kitodo-production/issues/6544 (general placeholder issue for the issue) based on the 3.8.x branch.

With #6572 only new and changed process internal meta information get updated in the database. But there are a lot more of processes before pull request #6572 got merged and the issue hopefully fixed. This remaining processes - which could be a lot of processes - must be corrected in some way. With this pull request I added a way through the "migration page" where you can select the projects which should go updated and the execution itself is done in a task inside  the task manager execution.

UI code is copied and pasted from similar places of the migration page and should work. Maybe someone with more UI knowledge look into this deeper. I did not add selenium based tests as they are to confusing for me and the main logic is covert by tests as integration tests.

Maybe I add in a separate pull request this kind of updating as a KitodoScript call in a follow up pull request.

This pull request will be back-/forward ported to main branch after this pull request is merged / approved.


Kitodo/src/main/java/org/kitodo/production/helper/tasks/UpdateInternalMetaInformationTask.java
```diff
@@ -0,0 +1,89 @@
...
+public class UpdateInternalMetaInformationTask extends EmptyTask {
+
+    private static final Logger logger = LogManager.getLogger(UpdateInternalMetaInformationTask.class);
+    private final Project project;
+    private final ProcessService processService;
+
+    /**
+     * Default constructor.
+     *
+     * @param project Chosen projet which process meta information should be updated.
```
solth, line 35:
```suggestion
     * @param project Chosen project which process meta information should be updated.
```

solth (commented):
Functionality wise it works, found just one small typo. 

I did observed some very strange UI behavior on the migration tab, however: when you click the individual buttons like "Migrate LDAP manager passwords" and then "Update internal meta information", the UI isn't updated correctly, but contains _both_ formulars in a mixed up, broken state. I don't think this is caused by this pull request, but I wanted to point it out anyway.

solth:
> I did observe some very strange UI behavior on the migration tab, however: when you click the individual buttons like "Migrate LDAP manager passwords" and then "Update internal meta information", the UI isn't updated correctly, but contains _both_ formulars in a mixed up, broken state. I don't think this is caused by this pull request, but I wanted to point it out anyway.

I just checked and this happens in `main` as well. Basically, the `Migrate LDAP manager passwords` part of the form is not removed correctly when switching to a different tab/function, the other options (like `Migrate metadata`, `Migrate workflows` etc.) seem to update the UI correctly.

Kitodo/src/main/java/org/kitodo/production/helper/tasks/UpdateInternalMetaInformationTask.java
henning-gerhardt: Same typo exists in main branch pull request. Should I fixed it in the other pull request too?

PR #6587, opened 2025-07-04
Changes to search

matthias-ronge:
**Changes to the project search**

The project search is now only performed in the database. The search index field `searchProject` has been deleted. Searching with the "project:" search field is now also possible using the SQL `LIKE` operator. If a "*" is entered as the symbol, a left-truncated search is also possible, for example, "project:*eitungen." The "project_loose:" field always searches with a right-truncated search.

**Changes to task search**

The task search is now performed only on the database. The search index field `searchTask` has been deleted.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/Process.java
@@ -979,18 +979,6 @@ public class Process extends BaseTemplateBean {
     }
 
     /**
-     * When indexing, outputs the index keywords for searching by project name.
-     * 
-     * @return the index keywords for searching by project name
-     */
-    @Transient
-    @FullTextField(name = "searchProject")
-    @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.NO)
-    public String getKeywordsForSearchingByProjectName() {
-        return initializeKeywords().getSearchProject();
-    }
-
-    /**
      * When indexing, outputs the index keywords for searching for assignment to
      * batches.
      * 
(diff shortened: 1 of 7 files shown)


BartChris:
Works good in my first tests :+1: .
One thing i noticed is that "project_loose" gets shortened to "loose" after the search is executed:

image
image

solth:
> One thing i noticed is that "project_loose" gets shortened to "loose" after the search is executed:

Could this be caused by horizontical scrolling applied to long search terms?

BartChris:
> > One thing i noticed is that "project_loose" gets shortened to "loose" after the search is executed:
> 
> Could this be caused by horizontical scrolling applied to long search terms?

No, scrollbar is on the left.

image

Visually the problem is more that not the whole filter area is used to display active selections:

image

Functionality wise it still works, and "project_loose" is still transferred to the backend, so probably it is only visually cut off. (Maybe everything before "_" somehow gets lost during display?

henning-gerhardt:
@BartChris : this was the reason why I wanted the search index name prefix feature back. With this feature you can test this kind of change in a separate index without destroying the main index (or some other in future).

In general: changes to the index - in any way - are in most cases not back compatible so you must use different indexes.

henning-gerhardt:
> See also: https://docs.jboss.org/hibernate/search/6.2/reference/en-US/html_single/#mapper-orm-schema-management-strategy

I would not prefer that the application in a productive environment is re-creating the index schema on startup if there is some kind of mismatch. There must be reasons why this mismatch is happen and a person should may recreate the index and index schema. 

We had something similar in the past with the database. Some users let the application create and / or update the database schema and had some issues afterwards. It was a pain to figure out what happen there and how to fix it.

Creating schema by application may work if the application itself is designed in that way and the schema is not to complex itself or if schema is complex this is done in a right and stable way that creating and updating is working without any issue.


----

I tested this changes on a separate index and did not found an issue until now. Searching for different task states or project filtering works. Performance is not going down on my local system but maybe the amount of data is to less to see a performance issue.

henning-gerhardt (approved):
Changes looks correct even I can test them only with a few simple queries. I can not prove if they work in a correct way with more complex queries and other usage scenarios.

Issue #6590, opened 2025-07-08
Changing task state did not update sortHelperStatus


henning-gerhardt:
**Describe the bug**
I don't know yet if this is a regression of the Hibernate-Search implementation or was existing before.

If you change the task state through the task details page of the task list of a process the internal field `sortHelperStatus` got not updated even the task state itself is stored correct in the database.

This is important for closed processes if you want / need to reset the process workflow to a former state and use this way to change the task state. If you do this the process is still only "visible" in the process list if you still search for "closed processes".

**To Reproduce**
1. Search for an process in the database and note the value of `sortHelperStatus`
2. Search this process in the process list
3. Click on "Edit" button for this process
4. Switch to the "Task list" tab
5. Click on the "Task details" button
6. Chang the "Status" to an different one and click on "Save"
7. In the UI the new state is shown but in the database the value for `sortHelperStatus` is still the same

**Expected behavior**
After changing the state of a task though the task details the internal `sortHelperStatus` field should be updated to.

**Additional context**

If the process is not finished - means all tasks are done (green) and `sortHelperStatus` did not contain the value `100000000000` it is possible to change the task states on the Task list tab through the arrow buttons near the state symbol. If this way is used the `sortHelperStatus` value in the database is updated. But this way can not be used in first place for closed processes (all tasks green) as this arrows are only available if the last task is not done.

PR #6593, opened 2025-07-08
Sync the sortHelperStatus after change of task status

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6590


Kitodo/src/test/java/org/kitodo/production/forms/ProcessFormIT.java
```diff
@@ -109,4 +113,30 @@ public void testProcessesSelection() throws Exception {
                 .stream().map(Process::getId).sorted().collect(Collectors.toList());
         assertEquals(new ArrayList<>(Arrays.asList(1, 2, 5)), selectedIds);
     }
+
+    @Test
+    public void testSortHelperStatusIsUpdatedOnTaskSave() throws Exception {
```
henning-gerhardt, line 118: Personally I would use here a `DAOException` instead of the general `Exception` class as only `DAOException`s could be thrown and if other exceptions are thrown they may let the test be successful too.

henning-gerhardt (approved):
Changes looks good and are working. I would use an other exception (more strict) declaration for the new test method but this is may be not necessary.

BartChris:
I have not really checked it. Do we need the backport to 3.8?

solth:
> I have not really checked it. Do we need the backport to 3.8?

I just tested this. The same error seems to occur in 3.8, so a backport of this fix would be much appreciated! :)

PR #6606, opened 2025-07-15
[3.8] Fix adding physical divisions

oliver-stoehr:
Backport of #6583


thomaslow (approved):
Tested it. Works fine.

PR #6607, opened 2025-07-15
[3.8] Fix for retrieving processes with a given title

### LINKED ISSUE #6605: Search for available titles does not perform exact search
**Describe the bug**
When trying to save a process with a title that should be available, saving can fail when processes exist in the system with process titles that have the title of the new process as a substring, even if the _exact_ process title should be available. (for example when working with newspapers, whose process titles are constructed in a similar manner)

**To Reproduce**
1. Edit a newspaper process (that is: the top most process in the newspaper process hierarchy)
2. Change the title to some completely different, temporary value that does not exist anyway and save the process
3. Edit the same process again and try to reinstate the original title
4. Save the process and see the error message about the unavailable title

**Expected behavior**
The search for available process titles should be exact and not prevent saving processes with a title that only appears as a substring in other process titles

**Release**
~~3.8.3~~ (fixed by #6607), main

**Additional context**
~~Not tested in current `main` branch~~

**Edit:** just tested it, and the problem is not limited to the 3.8 release of Kitodo, but only occurs in the `main` branch

solth:
Fixes #6605

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -1787,8 +1787,7 @@ public class ProcessService extends ProjectSearchService<Process, ProcessDTO, Pr
      * @return amount as Long
      */
     public Long findNumberOfProcessesWithTitle(String title) throws DataException {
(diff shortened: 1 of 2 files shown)


BartChris (approved):
Tested and works!

solth:
Unlinked issue #6605 as it is only resolved by this pull request for Kitodo 3.8, but not for the `main` branch.

Issue #6602, opened 2025-07-14
Searching for properties does not work as expected in Hibernate Search, new metadata search needs readjustments


BartChris:
**Describe the bug**

The new Hibernate Search branch still allows searching for process properties using the ui:

So if you have a title like `Die Verlorenen`, you would expect a search like that to work:

`property:Haupttitel:verlorenen`

<img width="861" height="166" alt="Image" src="https://github.com/user-attachments/assets/71abb888-7991-4df2-bab5-ac852f6138bb" />

But the search result is empty, although there is a title like that.

The current code treats those searches for properties as something which is done in the search Index. The search index however does, in my tests, not index the process properties and i do not see any indication of that in the Indexing class - 

https://github.com/kitodo/kitodo-production/blob/main/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ProcessKeywords.java

What gets indexed instead is the metadata stored in the XML.

https://github.com/kitodo/kitodo-production/blob/32f4b41bb18bc16e041f83c87c63ad2be2e35fbb/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ProcessKeywords.java#L194

So what happens right now, when you search a property? 

There is special handling for properties in the query logic:

https://github.com/kitodo/kitodo-production/blob/32f4b41bb18bc16e041f83c87c63ad2be2e35fbb/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java#L399

https://github.com/kitodo/kitodo-production/blob/32f4b41bb18bc16e041f83c87c63ad2be2e35fbb/Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryPart.java#L83-L85

But this logic does only seem to convert the query:

`property:Haupttitel:verlorenen`

to the query `haupttitel:verlorenen`, formatted as `haupttitelqverlorenen`, which is the new style of making specific fields in the index searchable.

this query is then issued on the generic `search` field in the index:

In my case this generic `search` field looks like this in the index (massively shortened):

```
"search": "titledocmainqverlorenen  titledocmainqdie"
```

So we have the title indexed under the metadata key defined by the ruleset. But there is no property indexed there, if i am not mistaken. So this seems to explain for me why a search for `hauptitel:verlorenen` does not return anything, since there is no match for what gets constructed for property queries. 

-----
With the switch to Hibernate Search Kitodo introduced the possibility to search directly in the metadata  provided by the meta.xml. 

[...]

PR #6615, opened 2025-07-22
Fix parsing issues from issue 6602

matthias-ronge:
- Detect ruleset keys which contain attributes
- Not recognize any labels of options

See #6602


matthias-ronge:
Yes, now I realize where I was wrong. There are of course line breaks between the opening `<option>` and the closing `</option>`. In my mind, I was only thinking about the `<label>`, though I'm replacing `<option>`. :see_no_evil:

PR #6618, opened 2025-07-24
Find processes by their title, even if tokens are too short

### LINKED ISSUE #6553: Search for some process titles does not work
**Describe the bug**
The following process title can no longer be found with the process search

`process:retro_3117534-X_1940_1005`

The problem is, that the tokenizer kicks out the 'X', since it is only indexing tokens with a min length of 3. (see: https://github.com/kitodo/kitodo-production/blob/5a7c69c6bdc0f39e1c99f9d2ee20ea2fa5f0f254/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ProcessKeywords.java#L79 )

This is what gets stored in the index:

```
etro retr 194 311 3117 31175 3117534 534 ret 7534 retro tro 100 311753 1940 1005 940 005 117534 17534
```

When searching the full process title 
`process:retro_3117534-X_1940_1005`
this breaks.

**Expected behavior**
It should be possible to search for process titles like the above one. Maybe by also indexing the full process title.

**Release**
Kitodo Master

matthias-ronge:
Fixes #6553

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ProcessKeywords.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ProcessKeywords.java
@@ -68,6 +68,28 @@ public class ProcessKeywords {
     private final Set<String> batchKeywords;
     private final Set<String> taskKeywords;
 
+    /**
+     * Creates the individual subsets of keywords. These will later be indexed
+     * in different combinations into the various index fields. What is
+     * tokenized is the process title, the project title, the numbers and names
+     * of the batches containing the process, the process ID, the metadata, the
+     * titles of the process's tasks, and any processing comments. How it is
+     * tokenized: Only tokens with a minimum length of 3 are retained, except
+     * for the project title, for which it is 1. The metadata is tokenized both
+     * itself, and fielded into pseudowords to enable searches for specific
+     * metadata fields that are only known at runtime.
+     * 
+     * <p>
+     * In this implementation, most of the work actually happens in the
+     * constructor because it is placed as a singleton in a transient field of
+     * the respective process object. Different subsets of the keywords are
+     * retrieved via different getters whose call order is undefined. However,
+     * the creation of the keywords and the necessary loading and parsing of the
+     * ruleset and meta.xml files should only happen once.
+     * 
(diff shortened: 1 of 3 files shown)


solth:
@matthias-ronge Please rebase against `main` and add missing java doc!

matthias-ronge:
I'm surprised that Checkstyle demands for a Javadoc for a method that I haven't touched in this development.

BartChris:
I see the benefits of the indexing strategy and i suppose we are approaching a point, where we have it in a workable state. One disadvantage i see is that all that flexibility of tokenization make the common use cases (Doing a search on the full ttile) harder and i am not sure wethere there are still search requests which do not work or might render suprising results. 

I am not against the current approach and we should probably keep it if there are not major roadblocks left, but would like to see if we can somehow get feedback on what is `really` needed.

matthias-ronge:
it works if you type `process:retro_3117534-X_1940_1006` in the filter box. It searches for the process title on the `searchTitle` seach field, and it understands that the token “x” must be omitted because it was not indexed. If you only type `retro_3117534-X_1940_1006` in the filter box, the box renders a “process:” prefix, which is wrong. This is a graphics error in the filter box. The search searches on the `search` field for “retro 3117534 x 1940 1006”. Because for project title tokenization, single characters are allowed, it doesn’t throw out the “x”. If we set this to 3 characters for process title tokenization as well, then we can filter shorter strings, but I think the problem was that there can be institutions that have “Project Digitization A” and “Project Digitization B”. I do not remember in detail, but these questions need to be discussed, and with which consequences they go along.

Typing into the blue search box should do exactly the same as typing in the filter box. Do only consider what you typed in yourself, not what the filter box graphics enhancement with the blue boxes might add.

BartChris:
> Typing into the blue search box should do exactly the same as typing in the filter box. Do only consider what you typed in yourself, not what the filter box graphics enhancement with the blue boxes might add.

Right now it does not do exactly the same thing, but - as you have also indicated - this brings up more broader questions. I approve the PR but keep the remark that just typing in the process title in the filter box not always lead to finding the title, which is unfortunate. We need more input from the users at this point.

BartChris (approved):

PR #6623, opened 2025-08-04
Prevent excessive SQL queries when using action buttons in the process list

BartChris:
When just opening the KitodoScript modal window a lot of SQL queries are triggered because the form and all its data is reprocessed. 

image

This is visible when observing the loading spinner or by activating the SQL query log.

This is not necessary in my opinion. I therefor tried to prevent the server side reprocessing.

Fix only necessary on Master, as 3.8 is already fast.

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/processActionsColumn.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/processActionsColumn.xhtml
@@ -89,8 +89,7 @@
                            styleClass="action"
                            title="#{msgs.linkHomeDirectory}"
                            rendered="#{SecurityAccessController.hasAuthorityToEditProcessImages()}"
-                           process="@this"
-                           update="processesTable">
+                           process="@this">
                 <h:outputText><i class="fa fa-download"/></h:outputText>
                 <!-- TODO: add separate message in case it already exists and alert -->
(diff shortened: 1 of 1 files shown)


solth:
I wonder if just adding `process="@this"` to the individual `<p:menuItem>` elements would perhaps be enough to reduce the number of SQL queries and thus speed up the dialogs (not just import configuration and kitodo script, but renaming media as well - and probably some others, too)? 

I just tested this locally and it does indeed seem to do the trick, but the number of processes in my development system is probably not high enough to verify the speed improvement reliably.

BartChris:
I now added it all the other buttons in the Action Menu at the bottom of the page except PDF and excel export. The latter have an `ajax=false ` annotation so the speedup does probably not work here. Thise to functions do have to process a lot of data either way, so it is harder to judge wether you get a speedup or not.

Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/processActionsColumn.xhtml
```diff
@@ -116,26 +122,31 @@
                            styleClass="action"
                            title="#{msgs.docketPrint}"
                            ajax="false"
+                           immediate="true"
```
BartChris, line 125: This also seem to bring a speedup when `ajax=false`.

Issue #6624, opened 2025-08-05
On first DMS export of a child the parent process does not contain links to the child


BartChris:
**Describe the bug**
When exporting a child process for the first time i would aussume, that the automatically exported parent process does contain a link to the child. 
But on first export the child link is missing in the exported parent METS, when using the DMS export (async as automatic step). 
(In the cases i tested so far i had no already exported children)

I assume the that logic below is supposed to prevent the observed behaviour, but somehow it does not work as expected.

https://github.com/kitodo/kitodo-production/blob/a297f95d820004ec150ee73f2f03eab925b576ff/Kitodo/src/main/java/org/kitodo/export/ExportDms.java#L107-L110

This logic appears to enforce that the child is explicitly set as exported even before the export of it starts so that the parent includes it in its METS. But when using the DMS export it seems that the link to the child is only constructed when i export the child a second time. 

**To Reproduce**
1. Export as child process which was not yet exported
2. Parent export starts automatically
3. Check if the parent process has a link to the child in the METS
4. There is no link to the child.

**Expected behavior**
The parent process should contain a link to the child on the first export of the child.

**Release**
3.8.3, probably also in Master

BartChris:
This also happens when i have on child process already exported and then export the second process. Now the exported parent METS contains only the link to the first child, which was exported in a previous iteration:


```xml
 <mets:structMap TYPE="LOGICAL">
      <mets:div ID="uuid-99ec782b-8d45-416a-b56a-b985bb60b1fa"
                DMDID="uuid-4b4b28a4-057d-3422-8dc3-7f4690ff34f5"
                ADMID="AMD"
                TYPE="multivolume_work"
                LABEL="Homers Ilias">
         <mets:div ID="uuid-e2f05e2e-dac8-4174-81e4-bcc17ebf5aff"
                   TYPE="volume"
                   ORDER="1"
                   LABEL="2. Homers Ilias">
            <mets:mptr xlink:href="https://api-digital....."
                       LOCTYPE="URL"/>
         </mets:div>
      </mets:div>
```

The first export rendered no children at all:


```xml
 <mets:structMap TYPE="LOGICAL">
      <mets:div ID="uuid-99ec782b-8d45-416a-b56a-b985bb60b1fa"
                DMDID="uuid-4b4b28a4-057d-3422-8dc3-7f4690ff34f5"
                ADMID="AMD"
                TYPE="multivolume_work"
                LABEL="Homers Ilias">
      </mets:div>
```

[...]

PR #6625, opened 2025-08-05
[3.8] Fix missing link to child process in asynchronous export

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6624

As elaborated in the linked issue, in my tests the child process is not linked reliable in the exported parent process, when using the async DMS export (Manual and automatic) the first time. It is only linked, when it is exported a second time. (Checked on my dev system as well as on the server, both running with local db)

The problem appears to be, that the code does the following things (in async mode):

a) mark the to-be-exported process as exported in the database, so that it is included in the export of the parent process
b) export the child process in a different thread
c) export the parent process(es) in another thread

While those two (or more) exports are running in the background (in the case of an asynchronous) export 

d) the code continues and sets the state of the process to the result of the export operation. 

https://github.com/kitodo/kitodo-production/blob/fa8e2256ebefc05ea10c1a88411b96f1961eebe1/Kitodo/src/main/java/org/kitodo/export/ExportDms.java#L117-L120

This works in the synchronous case, but in the asynchronous case the export state is always immediately set to false (to be set to the right state later).

Only when the background thread is finished it is set to true.

Under specific timing conditions - it probably plays a role how long the export takes or how fast database writes are (local db vs remote db e.g.) - the database set the process to `not exported` and this value is read while generating the parent export. In consequence no link is written to the METS file.

I therefor in the PR's code call the logic from line 117-120 only in the synchronous case. The state of the exported processes should be recorded in the thread the export is happening.

There i added another condition

 https://github.com/kitodo/kitodo-production/blob/c2eec2c00040cb064d7c19c6a06646963978b246/Kitodo/src/main/java/org/kitodo/production/helper/tasks/ExportDmsTask.java#L94-L97

to ensure that the database export status of a process is only changed if this process has never been exported before. 

Edit: Will check and forward port to main, if this fix is accepted.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/export/ExportDms.java
+++ b/Kitodo/src/main/java/org/kitodo/export/ExportDms.java
@@ -62,7 +62,7 @@ public class ExportDms extends ExportMets {
     private final ProcessService processService = ServiceManager.getProcessService();
 
     private boolean exportWithImages = true;
-
+    private boolean optimisticExportFlagSet = false;
     private Task workFlowTask;
 
     public ExportDms() {
@@ -108,6 +108,7 @@ public class ExportDms extends ExportMets {
         if (wasNotAlreadyExported) {
             process.setExported(true);
             processService.save(process);
+            this.optimisticExportFlagSet = true;
         }
 
         boolean exportSuccessful = startExport(process, (URI) null);
@@ -132,7 +133,7 @@ public class ExportDms extends ExportMets {
     @Override
     public boolean startExport(Process process, URI unused) {
         if (ConfigCore.getBooleanParameterOrDefaultValue(ParameterCore.ASYNCHRONOUS_AUTOMATIC_EXPORT)) {
-            TaskManager.addTask(new ExportDmsTask(this, process));
+            TaskManager.addTask(new ExportDmsTask(this, process, this.optimisticExportFlagSet));
             Helper.setMessage(TaskSitter.isAutoRunningThreads() ? "DMSExportByThread" : "DMSExportThreadCreated",
                 process.getTitle());
             return false;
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/java/org/kitodo/production/helper/tasks/ExportDmsTask.java
```diff
@@ -45,10 +46,11 @@ public class ExportDmsTask extends EmptyTask {
      * @param process
      *            the process to export
      */
-    public ExportDmsTask(ExportDms exportDms, Process process) {
+    public ExportDmsTask(ExportDms exportDms, Process process, boolean optimisticExportFlagSet) {
```
henning-gerhardt, line 49: Missing JavaDoc for new method parameter.

henning-gerhardt (commented):
To reproduce the issue was really difficult as many parameters from configuration (`taskmanager.autoRunLimit`, `asynchronousAutomaticExport`), database (`sortHelperStatus`, `exported`) and the way how the process is exported (as task from task list, as part of a (automatic) workflow, different ways from process list) are involved. An other reason is that the issue is not always reproducible which made it even difficult to see if the changes are working or not.

So I'm not sure if the issue is really solved and no new issues are introduced which may be found later after many / thousands processes are already exported.

Code change are good except one point for missing JavaDoc of a new method parameter.

Kitodo/src/main/java/org/kitodo/export/ExportDms.java
```diff
@@ -345,6 +346,15 @@ public Task getWorkflowTask() {
...
 
+    /**
+     * Returns whether the optimistic export flag is set.
+     *
+     * @return true if the optimistic export flag is set; false otherwise
+     */
+    public boolean isOptimisticExportFlagSet() {
+        return optimisticExportFlagSet;
+    }
+
```
henning-gerhardt, line 357: Method can now be removed as not used. Access to class field is only inside the class by direct access.

PR #6628, opened 2025-08-07
Reduce SearchResultForm to a minimum

BartChris:
As The SearchResultForm does not really exist anymore i suppose we can remove the view safely and reduce the functionality of the class to the absolute minimum.


solth:
Removing the `searchResult` page sounds reasonable, since @matthias-ronge thankfully unified the search result page with the process list page. There is one reference left in `ProcessForm.java`, though, which should probably also be removed then, I think:

https://github.com/BartChris/kitodo-production/blob/3f162178ce02df7a45099c1f8bf5159ba9ac002d/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java#L1001-1002

solth (approved):

PR #6629, opened 2025-08-07
Show functional metadata in ruleset edit mask

solth:
This pull request adds a little QoL feature by adjusting the "edit ruleset" page to show all functional metadata of the currently selected ruleset file in individual lists. This way the user can quickly check which functional metadata is defined in specific ruleset without requiring server access to check the ruleset xml file itself in a text editor. 

Bildschirmfoto 2025-08-07 um 12 50 16

Lists of functional metadata types not currently defined in the ruleset are greyed out to make the overview easier to read, even with just a quick glance.

Note that this currently only supports _displaying_ the functional metadata of the ruleset, but not _changing/writing/adjusting_ it. That could be done in a future development but since the whole XML file is written in such a process any existing format would probably be lost in such an update, even if just a small change is made to the ruleset.

Additional note: developing this feature revealed some functional metadata types that currently are only used in tests, like `authorLastName`, `dataSource` (the corresponding information is currently saved as "ImportConfiguration" directly with the process object in the database) and `title`, which therefor are probably not required anymore. Those metadata could be removed from the system, but that should be discussed separately from the changes in this pull request.


BartChris:
> _Edit_: judging by your comments in other threads I suppose you didn't mean `divisions` (e.g. structure element types) but types of functional metadata when you wrote "on the same **element**", right? That is supported in this feature: if more than one metadata is - for example - configured as `recordIdentifier`, the list for `recordIdentifier` will become scrollable and contain _all_ functional metadata configured as `recordIdentifier`.

Yes, this is what i meant, sorry for the unclear question. I can confirm that this works. What also works is assigning two different functional identifiers to the same key like this:

```xml
<key id="TitleDocMain" use="displaySummary structureTreeTitle">
            <label>Main Title</label>
            <label lang="de">Haupttitel</label>
</key>
```

"Haupttitel" appears under `displaySummary` AND `structureTreeTitle`, so this seem to work well. :+1:

BartChris:
You added descriptions to every functional key, which is great. I think it is good that the application thereby also documents some features, but we have to make sure that this is coherent with what is in the wiki:

https://github.com/kitodo/kitodo-production/wiki/Regelsatz-3.x#schl%C3%BCssel-mit-besonderer-verwendung

In the wiki we have `mediaPartial` but i do not see it in the app. I will also point the @kitodo/kitodo-board to this, so they can crosscheck if the descriptions meet their expectations or if something needs to be refined:

https://github.com/solth/kitodo-production/blob/cb9da5fe34d8681526051e7ca3273e63f35d6de2/Kitodo/src/main/resources/messages/messages_de.properties#L509-519.

solth:
> The displayed labels are always in German, although my instance is configured to use English as the language.

Metadata language and UI language are separately configured in Kitodo. The first can be defined in "Main menu" -> "User data & Settings" -> Tab "Metadata editor" -> "Language for metadata" and controls the labels of metadata used in the metadata, during process creation in the import mask and here, on this new overview page.

solth:
> What would probabably be the best is to use identical texts in the wiki as in the application.

Yes, I guess that is a good idea. I mostly used the texts from the JavaDoc of the individual functional metadata and adjusted it where I thought necessary. Any specific suggestions to improve the translations are of course welcome!

BartChris:
is this maybe because i have a lot of keys assigned to `displaysummary`?

image

solth:
> is this maybe because i have a lot of keys assigned to `displaysummary`?
> image

This, too, looks like outdated CSS to me

solth:
> Can you maybe also check on Firefox, the problem only appears there for me, not in Chrome. I will further try to invalidate all the caches.

Actually I am always using Firefox, the screenshot above is also from that browser.

solth:
> Interesting, when i disable this in Firefox it looks way better.

Would you mind sending me the ruleset file used above via e-mail? I can check how it looks in my Firefox instance.

BartChris (approved):
Code changes look good and moving the methods to the `RulesetService` is more correct. 

I approve the PR. Outstanding issues are the small layout issue in (my) Firefox and the coherent and correct description of the keys between Wiki and Pull request:

Direct comparison:

https://github.com/kitodo/kitodo-production/wiki/Regelsatz-3.x#schl%C3%BCssel-mit-besonderer-verwendung

https://github.com/solth/kitodo-production/blob/cb9da5fe34d8681526051e7ca3273e63f35d6de2/Kitodo/src/main/resources/messages/messages_de.properties#L509-L519 .

BartChris:
@solth I think you german translations are good. I will just copy paste them to the Wiki so that you can merge here.

PR #6631, opened 2025-08-07
Optimize UI update in process actions

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6623


BartChris:
Interesting, this change seems to break tests in the Searching Selenium test. Those test were refactored in main with the introduction of Hibernate Search which effectively removed the SearchResultPage. I do not really now which of the changes in the PR make the 3.8 search tests fail.

solth:
@BartChris Backporting this improvement seems to be more effort than anticipated. I would suggest you close the pull request and we pass on having it available on 3.8, since it does not have high priority, as you initially suggested.

PR #6634, opened 2025-08-13
Update missing Spanish messages and verification Spanish messages

danilopenagos:
Missing Spanish messages were added and some machine translation Spanish messages were verified.


solth (approved):
@danilopenagos Thank you very much for providing the missing Spanish translations and checking those created with Google!

PR #6635, opened 2025-08-15
Update Kitodo.Production to version 3.9.0.RC1

solth:
This pull request updates the version of Kitodo.Production to 3.9.0.RC1 ("Release Candidate 1"). A new branch `3.9.x` will be created from this version which will eventually result in 3.9.0 proper (and probably patch releases 3.9.1, 3.9.2 etc. afterwards). 

This makes the way free for feature development for the next [major version 4.0.0](https://github.com/kitodo/kitodo-production/milestone/24) to be introduced into the `main` branch. All future bug fixes for features introduced in 3.9.0 - e.g. fixes for bugs related to HibernateSearch, the new long term preservation functionalitites and metadata import from CSV data - will have to be provided for both branches, `3.9.x` and `main`.

### DIFF
--- a/Kitodo-API/pom.xml
+++ b/Kitodo-API/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <artifactId>kitodo-production</artifactId>
         <groupId>org.kitodo</groupId>
-        <version>3.9.0-SNAPSHOT</version>
+        <version>3.9.0-RC1</version>
     </parent>
 
     <name>Kitodo - API</name>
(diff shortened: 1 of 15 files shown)


solth:
> An other approach could be: apply this pull request in the `3.9.x` branch and change the version in the `main` branch to `4.0.0-SNAPSHOT` as 3.9.x bug fixes and releases will be only be made in the `3.9.x` branch.

Yes, that would work, too, but I think it's mostly the same as what I plan now (_first_ change the version to 3.9.0.RC1, then create branch `3.9.x` from that code state). I am planning to change the version in the `main` branch to `4.0.0-SNAPSHOT` after creating the `3.9.x` branch anyway. 

Concerning the bugfixes, I do think we might still find bugs introduced in `3.9` that would not only have to be fixed in that branch, but be forward-ported to `main`/4.0.0 as well. 

Anyway, thanks for the quick review!

PR #6645, opened 2025-08-21
[3.9] Optimize task related buttons in process list

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6644


solth (approved):
Just tested the changes and can confirm that they do speed up the process of accepting tasks from the row expansion of the process list.

Issue #5813, opened 2023-10-19
Resetting process state results in Java out of memory error


henning-gerhardt:
**Describe the bug**
Resetting a lot of processes over a lot of process tasks is resulting in Java Out Of Memory Errors.

Output in Tomcat9 catalina.out
```
[2023-10-19 09:50:59] [info] Exception in thread "I/O dispatcher 1" java.lang.OutOfMemoryError: Java heap space
[2023-10-19 09:50:59] [info] Exception in thread "C3P0PooledConnectionPoolManager[identityToken->1n1gln1azjdzd8cq7lyzy|51c2b342]-AdminTaskTimer" java.lang.OutOfMemoryError: Java heap space
[2023-10-19 09:50:59] [info] #011at java.base/java.util.LinkedList.superClone(LinkedList.java:1012)
[2023-10-19 09:50:59] [info] #011at java.base/java.util.LinkedList.clone(LinkedList.java:1025)
[2023-10-19 09:50:59] [info] #011at com.mchange.v2.resourcepool.BasicResourcePool.cloneOfUnused(BasicResourcePool.java:1856)
[2023-10-19 09:50:59] [info] #011at com.mchange.v2.resourcepool.BasicResourcePool.cullExpired(BasicResourcePool.java:1642)
[2023-10-19 09:50:59] [info] #011at com.mchange.v2.resourcepool.BasicResourcePool.access$1900(BasicResourcePool.java:44)
[2023-10-19 09:50:59] [info] #011at com.mchange.v2.resourcepool.BasicResourcePool$CullTask.run(BasicResourcePool.java:2187)
[2023-10-19 09:50:59] [info] #011at java.base/java.util.TimerThread.mainLoop(Timer.java:556)
[2023-10-19 09:50:59] [info] #011at java.base/java.util.TimerThread.run(Timer.java:506)
[2023-10-19 09:50:59] [info] Exception in thread "pool-2-thread-1" java.lang.OutOfMemoryError: Java heap space
[2023-10-19 09:51:03] [crit] [55065ad6-9402-416c-9049-fc067c19c47d][172.16.4.43] FullAjaxExceptionHandler: An exception occurred during processing JSF ajax request. Error page '/pages/error.jsf' will be shown.
[2023-10-19 09:51:03] [crit] java.lang.OutOfMemoryError: Java heap space
[...]

PR #6647, opened 2025-08-21
Optimize Update of process state

BartChris:
Adresses  https://github.com/kitodo/kitodo-production/discussions/6177

As discussed in GitHub discussion https://github.com/kitodo/kitodo-production/discussions/6177, there is potential to optimize the update of the workflow state of processes. Currently, whenever a task within a child process is closed, we check whether all other child processes under the same parent are also closed. If they are, the parent process can be marked as complete as well.

However, this check seems necessary only when the task being closed is the final task in the process. If the process is not yet complete, there is no need to verify the status of other sibling processes. Additionally, if checking whether the other processes are completed is required, we can delegate this to the database to improve performance.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/BaseBeanService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/BaseBeanService.java
@@ -172,19 +172,6 @@ public abstract class BaseBeanService<T extends BaseBean, S extends BaseDAO<T>> 
     }
 
     /**
-     * Checks if there is at least one row in the database according to the given query.
-     *
-     * @param query
-     *            for database search
-     * @param parameters
-     *            for query
-     * @return true if at least one row exists according to the given query
-     */
-    public boolean has(String query, Map<String, Object> parameters) throws DAOException {
-        return dao.has(query, parameters);
-    }
-
-    /**
      * Returns the number of objects of the implementing type that the filter
      * matches.
      * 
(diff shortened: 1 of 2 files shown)


solth (approved):
Code looks good and I understand the reasoning behind the changes. However, it is difficult for me to verify in my development system whether these changes improve the performance when changing the state of many processes at once that was reported #6177. 

@aetherfaerber would you be able to test the changes in this pull request with your data to check whether it improves the situation for you?

BartChris:
Just a small test:

From discussion

"Advancing 1177 processes takes roughly 3 minutes (this is reproducible after several tests)"

on my system advancing 1864 (newspaper item) process takes around 30 seconds. Main is still working after way more than a minute. But it would be better if this is tested on the same system with the same data.

aetherfaerber:
Thanks for mentioning us here. We will look into it. @gerlingsm could you please apply this PR?

04.12.2025 11:47:37 Henning Gerhardt ***@***.***>:

BartChris:
@gerlingsm I backported to 3.9 so that you do not have to update Tomcat and Java.

https://github.com/kitodo/kitodo-production/pull/6795 

Backporting to 3.8 would be more involved, but i would do it,if you need it for testing.

aetherfaerber:
> > Setting down our manual tasks (see #5813 for workflow overview) like "Erfassen der Meta- und Strukturdaten" from "in process" to "open" takes a lot of time even with this changes. Even our other two tasks "Scannen" (manual task) and "Export / Import in das DMS" (automatic task) takes a long time from "in process" to "open". The other tasks can be set down fast even without this change. Maybe this is a new and separate issue but I would mention it.
> 
> I think the logic touched here is only triggered when you set the task status "up", not when setting it "down".

Interesting to know that the two have different logic or are completely distinct functions. Just as a side note: from memory I would say that advancing steps has performance issues for us, while setting down the status results in crashes more often.

BartChris:
>  Just as a side note: from memory I would say that advancing steps has performance issues for us, while setting down the status results in crashes more often.

I am also investigating the memory issues, but maybe we keep those two seperate for now.  (Fixing the memory issues probably has a positive effect on performance as well)

https://github.com/kitodo/kitodo-production/issues/5813#issuecomment-3613275301

solth:
@aetherfaerber & @gerlingsm did you find the opportunity to check whether the changes in this pull request do indeed solve the issue you observed?

PR #6649, opened 2025-08-25
Optimization of the process list

BartChris:
This should speed up the process list (especially paging through).


BartChris:
I was able to bring down the query count from around 2000 to **1260** by using the batch annotation for some of the properties.:

```java
OneToMany(mappedBy = "parent", cascade = CascadeType.PERSIST, fetch = FetchType.LAZY)
@BatchSize(size = 50) private List<Process> children;

@OneToMany(mappedBy = "process", cascade = CascadeType.ALL, orphanRemoval = true)
@BatchSize(size = 50)
@OrderBy("ordering")
private List<Task> tasks;

@LazyCollection(LazyCollectionOption.FALSE)
@OneToMany(mappedBy = "process", cascade = CascadeType.ALL, orphanRemoval = true)
@BatchSize(size = 50)
private List<Comment> comments;

@LazyCollection(LazyCollectionOption.FALSE)
@ManyToMany(mappedBy = "processes")
@BatchSize(size = 50)
private List<Batch> batches = new ArrayList<>();
``

**Review by henning-gerhardt — APPROVED** [review]
I tried your changes again and did look into the number of queries to the database. The amount of queries is with this changes decreasing about round 50%. So it is a improvement even I did not see it in a visible way.

**Comment by BartChris** [comment]
After setting up the batchsize in the process entitiy, swtiching from onee to another page takes around 350 _real_ queries (No autocommit and other transaction related stuff). On a page of 100 those three are executed ach 100 times...

```sql
select process?_.id as col_?_?_ from process process?_ where process?_.parent_id=? limit ? 

SELECT t.processingStatus as status, COUNT(*) as count FROM task t, (    WITH RECURSIVE process_children(id) as (        (SELECT id FROM process WHERE id = ?)        UNION ALL        (SELECT p?.id from process as p?, process_children as p? WHERE p?.id = p?.parent_id)    ) SELECT id FROM process_children) as p WHERE t.process_id = p.id GROUP BY t.processingStatus 

select tasks?_.process_id as process?_?_?_, tasks?_.id as id?_?_?_, tasks?_.id as id?_?_?_, tasks?_.batchStep as batchste?_?_?_, tasks?_.concurrent as concurre?_?_?_, tasks?_.correction as correcti?_?_?_, tasks?_.editType as edittype?_?_?_, tasks?_.homeDirectory as homedire?_?_?_, tasks?_.last as last?_?_?_, tasks?_.ordering as ordering?_?_?_, tasks?_.process_id as process?_?_?_, tasks?_.processingBegin as processi?_?_?_, tasks?_.processingEnd as process?_?_?_, tasks?
....
... truncated here
```

BartChris:
@solth I will keep the PR as is, so it can be merged. Will inspect the other optimizations (batching and query optimizations) later.  Maybe setting a batchsize already brings additional speedup on larger databases; apart from that those remaining queries outlined above can maybe be optimized.

BartChris:
As increasing the batch size reliable improves on the number of queries, i added the batchsize annotation. I have not discovered any problem with that so far. And i think it should not harm doing larger operations with more entities as Hibernate will just fetch multiple times with the specified chunk size. But maybe we should have an eye on that.

BartChris:
> What is most annoying right now on my system is the time it takes to select or unselect processes. When stepping through the execution flow it seems that Hibernate runs in LazyInitializationExceptions here and has to initialize a lot of data again. This leads to a query storm, when all we want is to select a line. Primefaces 12 improves the situation, but i do not know how to fix that in Primefaces 8 without a big refactor as the selection code also contains some manual stuff.

I was able to fix the slow speed of selections by eliminating an unnecessary double fetch of the process.

solth:
@BartChris thanks again for this optimization. Do you plan to backport these changes for 3.9.x?

BartChris:
> @BartChris thanks again for this optimization. Do you plan to backport these changes for 3.9.x?

Yes: https://github.com/kitodo/kitodo-production/pull/6652

PR #6651, opened 2025-08-27
Switch back to standard workbook generation

BartChris:
While implementing the export for new Excel versions i used the memory efficient variant of the workbook generation, which writes chunk files to disk and merges them afterwards. This mode is meant for huge Excel files and i thought it would enable us to export large process tables. 

https://github.com/kitodo/kitodo-production/pull/6174

I just noticed that this way of generating the Excel files has an effect on the generation of PDF files which is based on the same class. My PDF files now seem to only contain a chunk of the data. Therefor we should switch back to the less optimized but also default XSSFWorkbook which is hopefully enough for our purposes.


solth:
@BartChris how can I test the issue you described with only a chunk of the process list being exported to the PDF? Is there a fixed number of processes I have to have in my system to recreate the issue and verify that it is resolved with the changes in this pull request? Can you describe the steps I should take to check the system behaviour accordingly?

BartChris:
I just used the PDF export for the whole process list without any filter, so for 3000 processes on my system. It seems like only the last processes and not all are included in the PDF.

henning-gerhardt:
I can confirm this issue by 12.xxx open processes. Only the last 207 processes where listed in the created pdf file. I did not apply any filter or did not select any processes to create the pdf file. Creating PDF and Excel file works with the current used filter and there is no need to select the processes to get them into this files.

solth:
@henning-gerhardt thanks for your response. Would you mind reviewing this pull request (and the corresponding 3.9 backport), since you are able to reproduce the issue?

henning-gerhardt (approved):
Codes changes are good and are working.

I found an other issue but this is not related to this and for the new issue I will open a bug report.

Issue #6654, opened 2025-09-01
Creating PDF and Excel files ignores used client


henning-gerhardt:
**Describe the bug**
This is a regression with the introducing of Hibernate-Search. In 3.8.x the created PDF and Excel files containing only processes from the used client. With the integration of Hibernate-Search the files contains processes of all clients.

I found this while investigating and reviewing #6651.

**To Reproduce**
1. Have a system with more than one client
2. Login to one client
3. Go to process list view
4. Create a pdf or excel file through the actions menu on the left down side
5. Look into the pdf or excel file and found processes from an other client

**Release**
3.9.x, current master

**Additional context**
This behaviour did not occur in 3.8.x

PR #6655, opened 2025-09-01
Restrict Excel and PDF export to current client

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6654


henning-gerhardt (approved):
Change looks good and issue is fixed.

PR #6659, opened 2025-09-03
Delegate desired language to central `LocaleHelper`

BartChris:
I encountered the problem that localization does not work everywhere, e.g. here: https://github.com/kitodo/kitodo-production/pull/6646

It seems to me that there is a problem when the Faces context is not fully initialized. `FacesContext.getCurrentInstance().getViewRoot().getLocale()` does not seem to give reliable results.

 Looking around in the codebase i suppose we already have a mechanism to detect the locale in a more reliable way using the `LocaleHelper`. I am wondering if we can not reuse that in the central `Helper` class, but have not really tested it extensively.

This change fixes the translation problems in https://github.com/kitodo/kitodo-production/pull/6646

Edit: Tests are failing, so this seems to be a little bit more involved.


BartChris:
Ok, this change leads down a rabbit whole as it probably also brings up the question how the tests are interacting with the localization logic and the ressource bundles....

PR #6664, opened 2025-09-05
[3.8] Adapt error handling in script logging

BartChris:
Spec from: https://github.com/kitodo/kitodo-production/issues/6665

> Command module is only logging the output from standard output channel as INFO message. Output from the standard error output channel should be logged on ERROR level.

> Command Service class should output the standard output channel as normal message (Helper.setMessage) and maybe depending on exit state or nor the output from standard error channel as error message (Helper.setErrorMessage) but without in both cases to log this messages again as this is already done by the Command Module.

### DIFF
--- a/Kitodo-Command/src/main/java/org/kitodo/command/Command.java
+++ b/Kitodo-Command/src/main/java/org/kitodo/command/Command.java
@@ -17,8 +17,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Scanner;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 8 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -519,11 +519,6 @@ public boolean executeScript(Task task, String script, boolean automatic) throws
                 CommandService commandService = ServiceManager.getCommandService();
                 CommandResult commandResult = commandService.runCommand(script);
                 executedSuccessful = commandResult.isSuccessful();
-                if (executedSuccessful && !commandResult.getMessages().isEmpty()) {
-                    for (String message : commandResult.getMessages()) {
-                        Helper.setMessage(message);
-                    }
-                }
```
BartChris, line 526: CommandService already print to the UI, so we probably do not need this here.

henning-gerhardt:
I tried your changes and they look good: 
- standard out and standard error are displayed on manual script execution in the UI
- standard out and standard error are logged correct in the application log file

Kitodo-Command/src/main/java/org/kitodo/command/Command.java
```diff
@@ -49,32 +47,31 @@ public CommandResult runCommand(String command) {
...
-            errorMessages.add(e.getCause().toString());
-            errorMessages.add(e.getMessage());
-            commandResult = new CommandResult(command, false, errorMessages);
-            logger.error("Execution of Command {} failed!: {}", commandResult.getCommand(),
-                commandResult.getMessages());
-            return commandResult;
+            commandResult = new CommandResult(command, -1, Collections.emptyList(),
+                    Collections.singletonList("IOException: " + e.getMessage()));
+            logger.error("Execution of Command {} failed to start: {}", command, e.getMessage());
+
```
henning-gerhardt, line 74: Is this empty line needed?

Issue #6670, opened 2025-09-09
Search field throws an exception when a number is used


BartChris:
**Describe the bug**
When typing a number like `15` in the top left search bar an exception is thrown:

<img width="1839" height="495" alt="Image" src="https://github.com/user-attachments/assets/099d7c61-f232-4472-a529-770e7b187901" />

```
Exception message: org.hibernate.QueryException: could not resolve property: null of: org.kitodo.data.database.beans.Process [SELECT COUNT(*) FROM org.kitodo.data.database.beans.Process AS process WHERE process.project.client.id = :sessionClientId AND process.null = :userFilter1 AND (process.sortHelperStatus IS NULL OR process.sortHelperStatus != :completedState) AND process.project.id IN (:projectIDs)]
```

The problem seems to be, that searches in the top left search bar automatically get prefixed with `process:`.

https://github.com/kitodo/kitodo-production/blob/713df084ae086625e2773a9a40a9f5bbcbd1ae41/Kitodo/src/main/java/org/kitodo/production/services/data/FilterField.java#L82

That matches those queries to the `PROCESS_TITLE` field which holds a string and it seems like it breaks with integers.

The same thing happens when using the filter bar with `process:15`

<img width="1824" height="134" alt="Image" src="https://github.com/user-attachments/assets/a17afda4-fb7e-4bf6-a7f5-2c0cd18f7ccb" />

<img width="1836" height="414" alt="Image" src="https://github.com/user-attachments/assets/666b0611-7c07-4065-b823-a08633719001" />

It seems that here a wrong query gets constructed, which has the following part:

`process.null = :userFilter1`

https://github.com/kitodo/kitodo-production/blob/713df084ae086625e2773a9a40a9f5bbcbd1ae41/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java#L412-L414

**Expected behavior**
Searching for a number should not break the search and even better: find the process with that ID.

**Release**
Kitodo 3.9, Kitodo Main

Issue #6630, opened 2025-08-07
Advanced search throws exception for some terms


BartChris:
**Describe the bug**
When in Advance Search and using the ID field to search for "retro" i get an exception:

<img width="1263" height="243" alt="Image" src="https://github.com/user-attachments/assets/77a681a3-e562-4b5c-b9c1-23087edc2fb8" />

<img width="1741" height="521" alt="Image" src="https://github.com/user-attachments/assets/b09fa1d1-2a49-498b-8025-f3df957ad32c" />

**To Reproduce**
1. Go to 'Advanced Search'
2. Search the ID field for 'retro'
3. Get Stacktrace

**Release**
Kitodo Master

PR #6672, opened 2025-09-09
Harden query parsing: prevent numeric titles and non-numeric IDs from breaking search

BartChris:
fixes https://github.com/kitodo/kitodo-production/issues/6670
fixes https://github.com/kitodo/kitodo-production/issues/6630

Both bugs are a consequence of applying SQL queries which rely on a specific type (ID or string) and which break if wrong types are passed in.

Regarding https://github.com/kitodo/kitodo-production/issues/6670

As outlined in: https://github.com/kitodo/kitodo-production/issues/6670#issuecomment-3270725657

process titles are always strings and should be searched with string values (this is set formally set here: https://github.com/kitodo/kitodo-production/blob/713df084ae086625e2773a9a40a9f5bbcbd1ae41/Kitodo/src/main/java/org/kitodo/production/services/data/FilterField.java#L28). Even if the user searches for a number the query should not be Integer based. 

Regarding https://github.com/kitodo/kitodo-production/issues/6630

Queries on the `processID` field have at least to contain one number. It can even be strings like `12SEPERATOR13`, which are parsed as "12 13" and result in a query for multiple IDs. But there has to be at least one number. 

I tried to the exceptions without a big refactoring and as small targeted fix.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/FilterService.java
@@ -416,7 +416,7 @@ public class FilterService extends BaseBeanService<Filter, FilterDAO> {
 
         if (indexed && Objects.nonNull(filterField.getSearchField())) {
             return new IndexQueryPart(filterField, value, operand);
-        } else if (Objects.equals(filterField, FilterField.PROCESS_ID)) {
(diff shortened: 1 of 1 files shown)


oliver-stoehr (approved):
Works as expected now!

PR #6676, opened 2025-09-11
Filter ID to be an integer value

### LINKED ISSUE #6630: Advanced search throws exception for some terms

matthias-ronge:
Fixes #6630


matthias-ronge:
It's probably not necessary, but it would still be a useful GUI improvement, wouldn't it? You immediately notice that you can only enter numbers here.

Issue #6690, opened 2025-09-26
Creating big excel file takes a long time


henning-gerhardt:
**Describe the bug**

Creating an excel file for f.e. 390.000 processes takes a long time (more than 5 minutes) which results for the user in timeout messages from the browser itself as the response takes a to long time or in proxy timeouts if the application is behind an application proxy.

If you configure your browser and / or proxy for unlimited timeouts than you get the excel file downloaded with - so far as I can tell - all entries.

**To Reproduce**
1. Got to process list
2. Filter for f.e. an project with a lot of processes inside (maybe including already done processes)
3. Use the action to create an excel file
4. Wait a long time and get an timeout error message from browser

**Expected behavior**
If possible the creating of the excel list even for many processes should be improved so it is not necessary to increase timeout limits to values which are very high and the user did not want to wait such a long time.

**Release**
- 3.9.0-RC1, 3.9.x and maybe even in the main branch
- 3.8.x is may effected to but there is an other issue - see #6689 

**Additional context**
Using high timeout values in browser and / or application proxy settings may work around the issue but can not deployed everywhere as there other side effects which are not wanted.

BartChris:
I have not inspected in detail yet. I unforunately had to revert an aspect of the modern style Excel export in Kitodo 3.9: The more memory efficient variant does break the export:

https://github.com/kitodo/kitodo-production/pull/6651

So when generating thousands of lines we might use a lot of memory. But i suspect that memory consumption is not the core issue. I suppose the problem is the huge number of SQL queries which probably gets send to the database when generating the export. My assumption would be that 3.9.x is more efficient than 3.9.0-RC1 because it includes optimizations. See https://github.com/kitodo/kitodo-production/pull/6652

But even with those optimizations just scrolling through the process list does more than 300 queries. So this problem is only amplified when generating the Excel. We should probably optimize the SQL query to retrieve only what is needed for the Exel file and find an efficient way to stream that.

BartChris:
Speed is improved; Kitodo 4.0 will also have a fast CSV export. Closing for now. Issue can be reopened if there are still problems.

PR #6691, opened 2025-09-30
Reduce number of SQL queries during Excel export

BartChris:
Addresses https://github.com/kitodo/kitodo-production/issues/6690

One of the reasons the Excel export takes a long time right now is that it fetches data which is not actually needed for the exported file (e.g. comments). This results in a large number of (batched) queries to retrieve unnecessary data.

I made an initial attempt to fix this by introducing a DTO that holds only the required data, populated by a single SQL query. The query count is reduced massively.

I suppose there are ways to hydrate the DTOs directly without going through intermediary objects, which might improve query performance further. The more severe problem, however, is that we are holding a large number of objects in memory when generating Excel files with more than 100,000 entries.

Maybe @henning-gerhardt  can give this a try to see whether the change already improves loading times. There are certainly more efficient ways to generate these Excel files. (using SXSSFWorkbook and streaming the DB result in batches).

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/dto/ProcessExportDTO.java
+++ b/Kitodo/src/main/java/org/kitodo/production/dto/ProcessExportDTO.java
@@ -1,0 +1,115 @@
+/*
+ * (c) Kitodo. Key to digital objects e. V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo project.
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.production.dto;
+
+import java.io.Serializable;
+import java.util.Date;
+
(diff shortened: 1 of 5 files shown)


BartChris:
> @BartChris Is it possible to back port this first to 3.9.x branch instead of the main branch? Currently I can not switch and test changes on the main branch and this changes can ~not be applied to the 3.9.x branch.~ applied to 3.9.x with small manual adjustments.

sure https://github.com/kitodo/kitodo-production/pull/6693

BartChris:
> Has this change something to do with the pdf creation as excel and pdf creation share some code? I did not check this as we did not use the pdf creation.

The PDF generation is based upon the generated Excel workbook. My change fixes the underlying datafetching for the workbook, so it speeds up the whole process, no matter if it outputs Excel or PDF. What i think might be a good idea would be to add an additional CSV export. My assumption would be that you can stream way faster to CSV than to an Excel Workbook. So if somebody wants to process a real large batch, he can switch to CSV.

henning-gerhardt (approved):

PR #6693, opened 2025-10-01
[3.9] Reduce number of SQL queries during Excel export 

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6691


BartChris:
@henning-gerhardt I just added an experimental CSV export in the menu for performance testing.

Kitodo/src/main/java/org/kitodo/production/dto/ProcessExportDTO.java
```diff
@@ -0,0 +1,119 @@
...
+        this.status = status;
+    }
+
+    /**
+     * Returns the process identifier.
+     *
+     * @return process ID
+     */
+    public Integer getId() {
+        return id != null ? id : 0;
```
henning-gerhardt, line 53: Please use `Objects.nonNull()` methods for comparing nullable values.

Kitodo/src/main/java/org/kitodo/production/dto/ProcessExportDTO.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Returns the process title.
+     *
+     * @return title of the process
+     */
+    public String getTitle() {
+        return title != null ? title : "";
```
henning-gerhardt, line 62: see above.

Kitodo/src/main/java/org/kitodo/production/dto/ProcessExportDTO.java
```diff
@@ -0,0 +1,119 @@
...
+        return creationDate;
+    }
+
+    /**
+     * Returns the number of images.
+     *
+     * @return image count
+     */
+    public Integer getSortHelperImages() {
+        return sortHelperImages != null ? sortHelperImages : 0;
```
henning-gerhardt, line 80: see above.

Kitodo/src/main/java/org/kitodo/production/dto/ProcessExportDTO.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Returns the number of docstructs.
+     *
+     * @return docstruct count
+     */
+    public Integer getSortHelperDocstructs() {
+        return sortHelperDocstructs != null ? sortHelperDocstructs : 0;
```
henning-gerhardt, line 89: see above.

henning-gerhardt (changes requested):
Code looks good in general and did work for a lot of processes. Some small changes should be done.

Question: can you add an other test which is not using the default values from the `ProcessExportDTO` object?

BartChris:
> Code looks good in general and did work for a lot of processes. Some small changes should be done.
> 
> Question: can you add an other test which is not using the default values from the `ProcessExportDTO` object?

Yes, good idea, added that.

henning-gerhardt (approved):

Issue #6460, opened 2025-03-11
Add notification menu


oliver-stoehr:
## Description

The results from scripts started by the user or automatic tasks executed by the system are currently visible in the task manager.
1. Not every user might have access to the task manager but it still could be useful if the user could see the result of the script he started.
2. Some scripts need more time to execute, so it is not always possible to display the result directly after starting the execution to the user. 
3. After some time the finished/crashed tasks are removed from the task manager.
When the user visits the task manager e.g. after the weekend, the list is most likely already cleared.
(I believe this is done to destroy the underlying objects and let the garbage collection clear the memory. So increasing the time to keep those objects might have a performance impact and might not be wise.)
Currently it is possible to guess from the task status "open" of an automatic task, that it failed. But when the result is no longer displayed in the task manager, the log files must be viewed or the task must be restarted only to get the error message again.

A new **notification menu** could address these issues.
In the beginning of Kitodo.Production 3 there was already a notification icon in the top menu bar planned, but never used and eventually removed. This icon could be reused to display a new popup menu, to display task and script results (and maybe other, more general notifications in the future) more easily accessible for the user.

The menu should consist of two sections:
- Personal:
All results of scripts started by the user should be displayed here.
- Automatic:
All results of automatic tasks independent of the user should be displayed here. (The user should of course only see processes, he is allowed to see.

![Image](https://github.com/user-attachments/assets/6687ca0d-7331-4e18-89e6-e344febd9afe)
![Image](https://github.com/user-attachments/assets/f79f109e-2832-4675-817c-ebd9b081a76b)
I created two mockups showing how the two sections might look.

- Maybe a mechanism would be useful to clear the list of _personal_ notifications, e.g. a "clear" button (not shown in the mockups).
- For failed scripts/tasks a more detailed information might be useful, to understand why it failed (addressing point 3 from above).
A click on an entry in the notification menu might expand or open a dialog showing the stack trace. 
- It should be discussed how long the notifications should be persisted/displayed in both sections by default (Keeping the results from one year ago is probably not desirable).

To achieve this technically all results from the task manager could be persisted in a new database table to be able to keep the results without blocking garbage collection of the actual tasks/scripts. 

## Related Issues

#6449, #6370

## Estimated Costs and Complexity
**medium** to **high**

henning-gerhardt:
I opened a discussion #6461 where this issue is referred as I like this idea in general but there are some points which should be discussed first before an implementation is done.

[...]

PR #6698, opened 2025-10-06
Remove not needed IOException call

henning-gerhardt:
Forward port of #6696

### DIFF
--- a/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/CommandService.java
+++ b/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/CommandService.java
@@ -12,7 +12,6 @@
 package org.kitodo.filemanagement;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.List;
 
 import org.kitodo.api.command.CommandInterface;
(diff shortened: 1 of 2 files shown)


matthias-ronge (changes requested):
This pull request breaks the error handling on the call to `ProcessBuilder.start()` in [`Command` lines 44−74](/kitodo/kitodo-production/blob/35650a866eb99798d1cb5375e406bf993d55f13b/Kitodo-Command/src/main/java/org/kitodo/command/Command.java#L44-L74). From the documentation of [`processBuilder.start()`](https://docs.oracle.com/javase/8/docs/api/java/lang/ProcessBuilder.html#start--):

> Starting an operating system process is highly system-dependent. Among the many things that can go wrong are:
> • The operating system program file was not found.
> • Access to the program file was denied.
> • The working directory does not exist. 
> In such cases an exception will be thrown. The exact nature of the exception is system-dependent, but it will always be a subclass of `IOException`.

Because the `CommandInterface` does not support throwing `IOException`s, and modifying interfaces was not desirable at the time, this solution was implemented. A cleanup is certainly commendable, but it must be done in a way that preserves functionality.

Possible solutions that preserve functionality could be:

- wrapping the `IOException` in `Command` in an `UncheckedIOException` and unwrapping it in `CommandService`
- modifying the `CommandInterface` so that `IOException`s can be passed directly, and then removing the `catch`, wrap, and unwrap code.

_I also noticed the inconsistency, because related, although not part of this pull request, that `FileManagement.createSymLink()` returns boolean false on exceptions, which is not checked when it is called in `WebDav.downloadToHome()`. There, only the `IOException` is handled, but it is not passed up from `createSymLink()`._

matthias-ronge:
> For not reporting issues while execution of commands in FileManagement Module I'm unsure if there are an UI response is wanted or not and @matthias-ronge did even report this too in his review of this pull request

Intuitively, yes. Furthermore, the error message should make its way into the process comments. However, that would constitute a new feature (or two).

solth:
> Intuitively, yes. Furthermore, the error message should make its way into the process comments. However, that would constitute a new feature (or two).

This would probably be #6460

solth:
@henning-gerhardt can we close this pull request now or do you think it is still required?

henning-gerhardt:
I think we can close this pull request.

PR #6705, opened 2025-10-07
Add logic for enriching LABELs on export

BartChris:
This pull request adresses a set of long standing issue extensively discussed in https://github.com/kitodo/kitodo-production/discussions/6277

Kitodo currently does not automatically fill the METS attributes LABEL and ORDERLABEL, which are required by presentation systems.
Users therefore rely on various workarounds:

- Filling LABEL/ORDERLABEL during import via custom XSLT extracting values (e.g. from title) — ineffective for manually created processes.
- Filling them during export via XSLT — but this leaves parent processes without child labels, since the internal meta.xml remains unchanged.
- Running external scripts to post-process and patch meta.xml.

This PR introduces an optional export-time enrichment step that reuses the export XSLT output to write the derived LABEL and ORDERLABEL values back into the internal meta.xml.

Concept:

- The export XSLT remains the single source of truth for generating derived LABELs.
- During export, these LABELs are written back to the internal METS so that parent exports can include them automatically.
- Existing processes can be enriched simply by re-exporting — no manual metadata edits or scripts required.
- Once stored in meta.xml, the LABELs are available to all future exports, including parents.

There are some caveats here. Since the correct enrichment of the exported parent METS with child's labels relies on the existence of those labels in the internal XML of the files, it has to be ensured that the parent's export runs after the child's export. The only way to guarantee that right now is setting ` taskManager.autoRunLimit =1` in the `kitodo_config.properties ` file. There are other issues with task parallelism in Kitodo right now, which are discussed in https://github.com/kitodo/kitodo-production/discussions/6633. 

Draft PR for @andre-hohmann and @henning-gerhardt to test.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/export/ExportMets.java
+++ b/Kitodo/src/main/java/org/kitodo/export/ExportMets.java
@@ -22,16 +22,27 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.stream.StreamSource;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.kitodo.api.dataformat.LogicalDivision;
 import org.kitodo.api.dataformat.Workpiece;
 import org.kitodo.config.ConfigCore;
+import org.kitodo.config.enums.ParameterCore;
 import org.kitodo.data.database.beans.Process;
 import org.kitodo.data.database.beans.User;
 import org.kitodo.data.database.exceptions.DAOException;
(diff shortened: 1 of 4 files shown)


henning-gerhardt:
Only a small correction: not the system has to be shut down only the tomcat service must be restarted. 

Reason was that all available memory for tomcat service which was 5 GB was used and more memory needed to be allocated. I think this has something to do with this changes or this changes made this behaviour more visible. I tried exporting with this changes on my local development system and attached VisualVM to this process and I saw a far more memory usage with this change as without. Difference was that on my local system more memory could be used by tomcat service and on my system the allocated memory was earlier / faster deallocated than in our test system where we had this memory allocation over 20 minutes (on my system it was deallocating after 30 seconds).

andre-hohmann:
@BartChris : 
Thanks again for this PR! I think you can change the status `draft` to `open`. 

Could you assess, if it is possible to replace `LABEL` and `ORDERLABEL` in the `meta.xml` if it differs from the exported METS file?
Now, i think `LABEL` and `ORDERLABEL` are added only if they do not exist in the `meta.xml`. 

In case of corrections of title information, it must be corrected in `LABEL` and `ORDERLABEL` manually. In my opinion, this is not necessary and should be avoided. 

If this is complicated, the current state can be used - it is very helpful, too.

andre-hohmann:
No, the "memory-problem" did not occur in our tests anymore. 

We recognized, that the export of superordinate processes with many references to subordinate processes is very slow - more than 30 seconds for 
- https://digital.slub-dresden.de/oai/?verb=GetRecord&metadataPrefix=mets&identifier=oai:de:slub-dresden:db:id-408113715. 

But this occurs in 3.8.0, too. Thus, neither your changes nor the hibernate search are the reason. 
@henning-gerhardt and i think, that it was faster in 3.6 or 3.7, but we cannot verufy it.

Kitodo/src/main/java/org/kitodo/export/ExportMets.java
```diff
@@ -177,4 +196,79 @@ protected boolean writeMetsFile(Process process, URI metaFile, LegacyMetsModsDig
         Helper.setMessage(process.getTitle() + ": ", "exportFinished");
         return true;
     }
+
+    /**
+     * Extract LABEL/ORDERLABEL from export and update the internal meta.xml
```
henning-gerhardt: Suggest change "internal meta.xml" to "internal meta data"

henning-gerhardt (changes requested):
I suggest not to name `meta.xml` or `internal mets` as it more neutral to use `internal meta data` as the format and name of the persistence file of internal meta data could change. I'm open even for other name rules as long as they a more neutral as it could be confusing which file in which format get changed.

PR #6715, opened 2025-10-10
Fix issues to generate valid METS structures

### LINKED ISSUE #6714: Kitodo creates invalid METS files
**Describe the bug**
While implementing the XML validation described in #6367 I have observed that metadata files created with Kitodo.Production contain a few errors when validating them against the METS schema definition, which can be found at https://www.loc.gov/standards/mets/mets.xsd:

1. `<mets:mdWrap>` elements are missing the mandatory `MDTYPE` attribute
2. `<mets:fileSec>` elements are created without any `<mets:fileGrp>` elements, but at least one `fileGrp` is required, even if it is empty
3. `<mets:structLink/>` element is created without any child elements, but should be omitted if no `smLink` or `smLinkGrp` elements exist

**To Reproduce**
1. Validate the contents from the `meta.xml` file of a process created with Kitodo against the METS schema definition linked above, for example using the command line tool `xmllint` (see https://github.com/kitodo/kitodo-production/discussions/6682#discussioncomment-14494999) or an online validator like https://www.utilities-online.info/xsdvalidation

**Expected behavior**
Kitodo has to create valid METS files in order to facilitate validation against the METS schema as described in #6367

**Release**
current `main` branch and most likely all versions up to now

**Additional context**
This bug does not apply to processes created with "prestructured import" which uses custom XSLT files to create the METS structure in addition to the descriptive metadata.

solth:
Fixes #6714 

**Note**: this pull request only adds changes ensuring that _new_ processes will contain valid METS structures. A function to fix the structure of existing processes should be developed separately.

### DIFF
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/DivXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/DivXmlElementAccess.java
@@ -316,6 +316,8 @@ public class DivXmlElementAccess extends LogicalDivision {
         xmlData.getAny().add(new JAXBElement<>(KITODO_QNAME, KitodoType.class, kitodoType));
         MdWrap mdWrap = new MdWrap();
         mdWrap.setXmlData(xmlData);
+        mdWrap.setMDTYPE("OTHER");
+        mdWrap.setOTHERMDTYPE("KITODO");
         MdSecType dmdSec = new MdSecType();
         dmdSec.setMdWrap(mdWrap);
         return Optional.of(dmdSec);
--- a/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
+++ b/Kitodo-DataFormat/src/main/java/org/kitodo/dataformat/access/MetsXmlElementAccess.java
@@ -272,7 +272,12 @@ public class MetsXmlElementAccess implements MetsXmlElementAccessInterface {
         mets.setMetsHdr(generateMetsHdr());
 
         Map<URI, FileType> mediaFilesToIDFiles = new HashMap<>();
-        mets.setFileSec(generateFileSec(mediaFilesToIDFiles));
+        FileSec fileSec = generateFileSec(mediaFilesToIDFiles);
+
+        // Omit empty fileSecs as they are not valid in METS
+        if (!fileSec.getFileGrp().isEmpty()) {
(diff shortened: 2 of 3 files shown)


henning-gerhardt (approved):
Changes looks good and are working. Especially that even old processes could be repaired by saving them through meta data editor.

But there is a disadvantage which must be communicated with the release which contains the change:
If you do some corrections on export with your export XSLT like

```xslt
    <!-- Add attribute @MDTYPE="MODS" in mets:dmdSec/mets:mdWrap> -->
    <xsl:template match="mets:dmdSec/mets:mdWrap">
        <xsl:copy>
            <xsl:attribute name="MDTYPE">MODS</xsl:attribute>
            <xsl:apply-templates select="@* | node()"/>
        </xsl:copy>
    </xsl:template>
```

as previous `mets:mdWrap` did not contain any `MDTYPE` attribute and for METS/MODS a value like `MDTYPE="MODS"` is needed will won't work anymore without adjustment of the XSLT rule.

Without adjusting export result will look like that

```xml
<mets:mdWrap MDTYPE="OTHER" OTHERMDTYPE="KITODO">
```

The "new" rule should be like (without any warranty):

```xslt 
    <!-- Set attribute to @MDTYPE="MODS" in mets:dmdSec/mets:mdWrap> -->
<!-- DID NOT WORK - DO NOT USE IT
    <xsl:template match="mets:dmdSec/mets:mdWrap[@MDTYPE='OTHER' OTHERMDTYPE='KITODO']">
        <xsl:copy>
            <xsl:attribute name="MDTYPE">MODS</xsl:attribute>
            <xsl:apply-templates select="@* | node()"/>
        </xsl:copy>
    </xsl:template>
-->
```

Maybe the XSLT files in the source code and provided release package should / must be updated? As we did not use them I can not tell if there is a change needed or not.

henning-gerhardt:
In general I would support this back porting. But I will this first discuss with @andre-hohmann as we at SLUB must adjust our all export XSLT files and maybe there are some more issues which I did not discovered yet or I'm not aware of them.

henning-gerhardt:
@solth : I had a short discussion with Andre and he agreed to back port this changes to 3.9.x. The change at our export XSLT files must be done in any case so doing it sooner than later is better. Hopefully there will be a repair tool that is fixing this issue for many processes as it is difficult to support old and new meta data on export. 

My "fix" for the XSLT transformation from my former comment did not work and should not be used. I will change my comment to reflect this.

PR #6719, opened 2025-10-13
Scroll through list of IDs from Search index

BartChris:
One of the problems with the current Hibernate Search based setup is that it relies on the following flow:
- First use the provided filter fields to search the Search backend (e.g. Elasticsearch) for a specific term (`TitleDocMain:Rhein`). 
- Use the returned IDs to restrict the database query, which contains the actual results, to the ones coming from the search query

Because of the hard limit of 10.000 results (from Elasticsearch/Opensearch) in a Search response the ID list might not contain all relevant (or IDs of non active processes) so the returned result from the DB is wrong. This experimental PR uses for the Excel export result set scrolling to retrieve all IDs.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
@@ -11,6 +11,7 @@
 
 package org.kitodo.production.services.index;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
@@ -20,6 +21,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.hibernate.exception.DataException;
 import org.hibernate.search.engine.search.projection.SearchProjection;
+import org.hibernate.search.engine.search.query.SearchScroll;
+import org.hibernate.search.engine.search.query.SearchScrollResult;
 import org.hibernate.search.mapper.orm.Search;
 import org.hibernate.search.mapper.orm.massindexing.MassIndexer;
 import org.hibernate.search.mapper.orm.session.SearchSession;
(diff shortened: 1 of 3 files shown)


BartChris:
> I implemented something like this years ago in a different context, but then it was said that it unnecessarily complicated the code, and it was removed during the review. That's why I've omitted it this time. The code is naturally easier to read this way. If we really need it, we can of course reimplement it.

What i tried locally is sending multiple terms to Elasticsearch. this works as well:

```java
public void performIndexSearches() {
        List<Pair<String, String>> terms = new ArrayList<>();
        for (var iterator = indexQueries.entrySet().iterator(); iterator.hasNext();) {
            Entry<String, Pair<FilterField, String>> entry = iterator.next();
            String field = entry.getValue().getLeft().getSearchField();
            String token = entry.getValue().getRight();
            terms.add(Pair.of(field, token));
        }
        Collection<Integer> ids = indexingService.searchIds(Process.class, terms);
        Collection<Integer> finalIds = ids.isEmpty() ? NO_HIT : ids;

        for (var iterator = indexQueries.entrySet().iterator(); iterator.hasNext();) {
            Entry<String, Pair<FilterField, String>> entry = iterator.next();
            parameters.put(entry.getKey(), finalIds);
            iterator.remove();
        }
    }

public Collection<Integer> searchIds(Class<? extends BaseBean> beanClass, List<Pair<String, String>> terms) {
      SearchSession searchSession = Search.session(HibernateUtil.getSession());
      SearchProjection<Integer> idField = searchSession.scope(beanClass).projection().field("id", Integer.class)
              .toProjection();
      var query = searchSession.search(beanClass)
              .select(idField)
              .where(f -> {
                  var bool = f.bool();
                  for (var term : terms) {
                      bool.must(
                              f.match()
                                      .field(term.getLeft())
                                      .matching(term.getRight())
                      );
                  }
                  return bool;
              });
      List<Integer> ids = query.fetchAll().hits();

      logger.debug(
              "Searching {} IDs with terms {}: {} hits",
              beanClass.getSimpleName(),
              terms.stream()
                      .map(t -> t.getLeft() + "=\"" + t.getRight() + "\"")
                      .collect(Collectors.joining(", ")),
              ids.size()
      );
      return ids;
  }
```

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
```diff
@@ -124,23 +127,54 @@ public CompletionStage<?> startIndexing(Class<? extends BaseBean> type, MassInde
...
+     *            whether to use scrolling (for large result sets like exports)
      * @return ids of the found beans
      */
-    public Collection<Integer> searchIds(Class<? extends BaseBean> beanClass, String searchField, String value) {
+    public Collection<Integer> searchIds(Class<? extends BaseBean> beanClass,
+                                         String searchField,
+                                         String value,
+                                         boolean useScroll) {
+
         SearchSession searchSession = Search.session(HibernateUtil.getSession());
```
henning-gerhardt: While reading the code changes here: `HibernateUtil.getSession()` is returing an Hibernate-Session which implements `AutoCloseable` interface. Should here and the other places in this class this usage replaced by a try-with-ressources statement? I noticed that using the search beginning with 3.9.x consumes more resources than before and this could be a reason. Maybe this should be fixed in a separate pull request. I did not start a discussion why in a `IndexingService` labeled class search methods are defined and used instead doing this in a `SearchService` class like in 3.8.x.

matthias-ronge:
Let's discuss only this code review in this pull request. For the underlying problem, I've opened a new issue #6772

Issue #6721, opened 2025-10-14
Cannot search for a process title in the task list


BartChris:
**Describe the bug**
Right now searching for a process title like `AlbeTrio_PPN1725319225` in the task list does not return any results. Searching for the same title in the process list works.  As far as i see it works for no process title. 

**Screenshots**
<img width="1769" height="330" alt="Image" src="https://github.com/user-attachments/assets/f88ab77b-9b24-4a79-81f6-01392fcf3d12" />

<img width="1769" height="330" alt="Image" src="https://github.com/user-attachments/assets/1fc9b32a-93a0-42e3-a816-c9b8ca2b6730" />

**Context**

One possbile cause of it is probably that the central `restrictWithUserFilterString` method

https://github.com/kitodo/kitodo-production/blob/293a722ee5e0cb0b18d4072604b7b8e94ed520df/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java#L307-L317

calls the `filterService.parse(` method always with `indexed=false` if we are not operating on the `process` bean. And in the task list we are not operating on the `process`, but the `task` bean. This has the consequence that the logic tries to construct db queries for queries which should be an index query and therefor the whole query fails. 

(Setting this to true does produces other problems with the query string, so the whole query construction method has to be inspected in case of the task list) 

**Release**
Kitodo 3.9, main

solth:
@BartChris After testing your fix in #6722 for this bug I wanted to verify that it does indeed occur in the current state of the `3.9.x` branch, but alas, I was able to find the task of a process with the same filter that is used in the process list even _without_ this bugfix. Does this bug only occur for processes with specific title patterns? The process with which I was testing had the title `valid_mets_kitodo` (stems from the work on #6367 ) and the filter I used was `process:valid_mets_kitodo`. 

As I said, this did give me the correct process in the process list as well as the current task of that process in the task list. Does the process title have to contain numbers for the bug to be reproducable? Or perhaps I am missing something else?

PR #6722, opened 2025-10-14
[3.9] Fix Index queries in task list

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6721

Right now the central detection logic whether to use the index or the database is only activated for the `Process` class. https://github.com/kitodo/kitodo-production/blob/cae308d1f9ae4d4b2e581f570a0426be73c3a942/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java#L313-L317

by setting `filterService.parse(filterString, beanClass.isAssignableFrom(Process.class)`

Since only the ProcessService and the TaskService are calling into `restrictWithUserFilterString` and both lists need a reliable mixed query construction i just set it to `true` always. I have not yet discovered any problems.

When constructing the IN filters of the index query it is possible to pass in custom SQL fragments 

https://github.com/kitodo/kitodo-production/blob/cae308d1f9ae4d4b2e581f570a0426be73c3a942/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java#L332-L337

But in case of multiple filter conditions those custom fragments are not even used, instead an hardcoded ".id" is used. I made this dynamic.

https://github.com/kitodo/kitodo-production/blob/cae308d1f9ae4d4b2e581f570a0426be73c3a942/Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryPart.java#L109-L116

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java
@@ -29,6 +29,7 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.kitodo.data.database.beans.BaseBean;
 import org.kitodo.data.database.beans.Process;
 import org.kitodo.data.database.beans.Role;
+import org.kitodo.data.database.beans.Task;
 import org.kitodo.production.enums.ProcessState;
 import org.kitodo.production.services.ServiceManager;
 import org.kitodo.production.services.index.IndexingService;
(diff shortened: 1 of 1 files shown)


solth:
> Instead of always setting the second paramter to true we could maybe also do a check like that.
> 
> ```java
> boolean indexed = beanClass.isAssignableFrom(Process.class) || beanClass.isAssignableFrom(Task.class);
> ```

@BartChris Yes, this looks like a better solution. I just tested this and it does work as well! Just don't forget the additional import!

solth (approved):
Fix resolves the issue and code looks good!

BartChris:
Sure 
https://github.com/kitodo/kitodo-production/pull/6732

Issue #6713, opened 2025-10-09
Deleting or adding a role  association of a task leads to an exception


BartChris:
**Describe the bug**
Editing a task and deleting or adding an associated role leads to an exception. This happens all the way up from Kitodo 3.8 to Kitodo main.

<img width="1095" height="321" alt="Image" src="https://github.com/user-attachments/assets/f275fd2a-6b35-48df-ac4d-3901dd3c88f7" />

<img width="1823" height="787" alt="Image" src="https://github.com/user-attachments/assets/26c346e2-d34f-4faf-8d85-6c39de04edbc" />
 

**To Reproduce**
1. Go to task list  
2. Edit one task  
3. Do one of the following:  
    a. Delete an associated role  
    b. Add a new role  
4. Save  
5. See Error

**Release**
3.8, 3.9, Main

Stacktrace:
```
java.lang.IllegalArgumentException: id to load is required for loading
	at org.hibernate.event.spi.LoadEvent.<init>(LoadEvent.java:96)
	at org.hibernate.event.spi.LoadEvent.<init>(LoadEvent.java:64)
	at org.hibernate.internal.SessionImpl$IdentifierLoadAccessImpl.doLoad(SessionImpl.java:2845)
	at org.hibernate.internal.SessionImpl$IdentifierLoadAccessImpl.lambda$load$1(SessionImpl.java:2825)
	at org.hibernate.internal.SessionImpl$IdentifierLoadAccessImpl.perform(SessionImpl.java:2781)
	at org.hibernate.internal.SessionImpl$IdentifierLoadAccessImpl.load(SessionImpl.java:2825)
	at org.hibernate.internal.SessionImpl.get(SessionImpl.java:1019)
	at org.kitodo.data.database.persistence.BaseDAO.retrieveObject(BaseDAO.java:451)
	at org.kitodo.data.database.persistence.ProcessDAO.getById(ProcessDAO.java:23)
	at org.kitodo.data.database.persistence.ProcessDAO.getById(ProcessDAO.java:19)
	at org.kitodo.production.services.data.BaseBeanService.getById(BaseBeanService.java:217)
	at org.kitodo.production.forms.ProcessForm.saveTaskAndRedirect(ProcessForm.java:386)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #6734, opened 2025-10-22
Performance Improvements For Hibernate ORM 6

thomaslow:
Includes #6737

This pull request improves the default fetch strategy of Kitodo.Productions's persistence model in preparation of the Hibernate ORM 6 upgrade.

In Hibernate ORM 6, the algorithm to calculate the required database table joins between entities was changed from a "depth-first" to a "breadth-first" strategy (see [documenation](https://docs.hibernate.org/orm/6.0/migration-guide/#_fetch_circularity_determination)). This affects how table joins are calculated for circular fetch dependencies (e.g. process.getComments() - comment.getCorrectionTask() - task.getProcess()). As a result, many more database joins are generated than in Hibernate 5, reducing the overall performance of Kitodo.Production.

In order to improve this problem, I had a look at all "eager" fetch dependencies and created the following overview:

Kitodo Production

[Kitodo.Production.graphml.zip](https://github.com/user-attachments/files/23057123/Kitodo.Production.graphml.zip)

In this PR, the following circular eager fetch dependencies are broken up:

- `Process.comments` to `Comment.[current|correction]Task` - `Task.process` by marking `Comment.currentTask`, `Comment.correctionTask` and `Comment.process` as lazy

- `Process.importConfiguration` to `ImportConfiguration.defaultTemplateProcess` by marking `ImportConfiguration.defaultTemplateProcess` as lazy

- `ImportConfiguration.parentMappingFile` to `MappingFile.importConfigurations` by marking `ImportConfiguration.parentMappingFile` as lazy

- `Process.parent` to `Process.parent` by marking `Process.parent` as lazy

- `Project.generatorSource` to `Folder.project` by marking `Project.generatorSource` as lazy

I also noticed some weird design decisions, which do not make sense to me, e.g.

- ~~the connection between `Role` and `Task`~~
- ~~the connection between `Role` and `Client`~~
- DataEditorSetting contains columns for `userId` and `taskId`, which are not mapped to User and Task

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/MetadataST.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/MetadataST.java
@@ -267,6 +267,8 @@ public class MetadataST extends BaseTestSelenium {
     public void updateMediaReferencesTest() throws Exception {
         login("kowal");
         Pages.getProcessesPage().goTo().editMetadata(MockDatabase.MEDIA_REFERENCES_TEST_PROCESS_TITLE);
+        await().ignoreExceptions().pollDelay(300, TimeUnit.MILLISECONDS).atMost(5, TimeUnit.SECONDS)
+                .until(() -> Pages.getMetadataEditorPage().isFileReferencesUpdatedDialogVisible());
         assertTrue(Pages.getMetadataEditorPage()
                 .isFileReferencesUpdatedDialogVisible(), "Media references updated dialog not visible");
         Pages.getMetadataEditorPage().acknowledgeFileReferenceChanges();


henning-gerhardt:
Regarding to the weird design decisions and hopefully understand this correct:

- Table connection `Role` and `Task` this could be come from the 2.x migration. It may be possible to delete this connection but I'm unsure how this would "break" existing access solutions.
- Table connection `Role` and `Client`:  one role should be assignable to one client and should not be assignable to all clients as not every same called role has the same used authorities on every client
- DataEditorSettings: maybe here was some mistakes made. I think that @solth or @oliver-stoehr  can answer here what is may wrong or correct

thomaslow:
> * Table connection `Role` and `Client`:  one role should be assignable to one client and should not be assignable to all clients as not every same called role has the same used authorities on every client

Ah. Yes. That makes sense. Thank you.

thomaslow:
> I must change my mind: this connection between task and role must stay as we have no other way to show the right tasks to a person which they should have access to it. Without this role task assignment it would not possible that a scann operator can only see and use the processes where the scan task is open / in-use or how this should be made that the user is only seeing this tasks which he has "rights" / "access" to them?

Yes. I get the idea.

It still feels a bit weird to me that each and every task has (probably the same) associations to (probably the same) roles. There could have been another table assigning general task types to roles, e.g. "a Task of type Scanning should be shown to Users with these Roles". Of course, then you would loose the ability to specifically assign additional roles to a specific task, which is the scenario of #6713.

Anyway. Thanks for explaining this.

thomaslow:
@henning-gerhardt Thank you for testing this. This PR might cause additional SQL queries that could slow down the performance in some situations. However, I wouldn't like to optimize this PR for the current master branch, since this PR is supposed to improve performance for the Java 17 / Tomcat 10 update. If necessary, I am happy to try to optimize the performance a bit further, but only for the Java 17 / Tomcat 10 branch.

@solth Please don't merge this branch for a different (earlier) release than the Java 17 / Tomcat 10 branch.

PR #6745, opened 2025-11-05
Add logic for enriching LABELs on export

BartChris:
Forward port of https://github.com/kitodo/kitodo-production/pull/6705


henning-gerhardt (approved):
Code is identical to #6705 , looks good and should work. I did not do a functional review.

Issue #6751, opened 2025-11-17
Import did not check for amount of allowed meta data fields


henning-gerhardt:
**Describe the bug**

If you define in your ruleset that a meta data fields can occur only one time but on importing data from catalogue this meta data field is many more times available then you get no information that this meta data field is imported too many times. As later there is no validation / check that this meta data exists many more times than expected (see #6750) there could be more issues happen.

**Expected behavior**
On importing data from an catalogue or similar there should be an check if  the meta data fields could be imported as they may be restricted by the amount of occur.

**Release**
At least in 3.8.4 but even possible in older and newer versions.

Issue #6750, opened 2025-11-17
No validation error on to many meta data fields


henning-gerhardt:
**Describe the bug**
Have a meta data field with `maxOccurs="1"` defined in your ruleset. Open a process where this meta data field used more times than defined in the ruleset. F.e. defined as 1 in ruleset but 5 times in meta data.

If you click on the validation button in the meta data editor no validation error message appear - everything is fine. There is even no yellow mark sign (or something different) on this meta data fields if you look on them.

Depending on how good or bad your export XSLT is written the export is failing or is successful but maybe with a not expected output.

**Expected behavior**
If there is an mismatch between maximum allowed amount of meta data fields and there real usage this should be shown on validation

**Release**
at least in 3.8.4 but maybe even in older and newer versions

PR #6753, opened 2025-11-17
Fix maxOccurs validation by grouping metadata values under a single view per ID

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6750
Fixes https://github.com/kitodo/kitodo-production/issues/6751


henning-gerhardt (changes requested):
Thank you @BartChris for this fix. It now shows that a multiple occurring meta data field which is allowed only once is too many in place but the validation error message in meta data editor is wrong

```
Volume ".....": Projekt-ID occurs 5 times, but has to occur at most 0 times.
```

This message even appears on creating a new process where a meta data field occurs more times than allowed.

Having tests which cover the import process and even the validation of existing meta data would be really nice.

BartChris:
> Having tests which cover the import process and even the validation of existing meta data would be really nice.

I will have a look how the test coverage is so far and if i can cover my changes as well.

henning-gerhardt (approved):
Works now for me - so far as I can test it and the validation error message looks correct for our usage.

PR #6762, opened 2025-11-20
Improve Selenium Test Reliability via Custom Locating Handler

thomaslow:
This pull request aims to improve the reliability of selenium tests.

For unknown reasons, the frequency of random selenium test failures seems to be a lot higher after the Java/Tomcat update, see #6692. Since these problems are random and often cannot be reproduced locally, they are very difficult to debug.

One problem is that Selenium `WebElement`s annotated with `@FindBy` randomly cause `StaleElementReferenceException`s, e.g., when clicking the "save" button in a form. The problem seems to be that a `WebElement` annotated with `@FindBy` is searched for in the DOM as soon as any of its methods is invoked (e.g. clicking a button). However, this process is not re-tried in case there is any error (e.g. a `StaleElementReferenceException`). Since the browser is running in a separate process, there is no guarantee that a DOM element is still available after it was found, even if it was found immediately before invoking a method. This problem seems to be especially frequent when entering form data, because each key-press triggers an ajax-call for validation and updating of the form's save button, which reloads all related DOM elements and makes the corresponding `WebElement`s become stale.

This pull request modifies this basic Selenium logic such that method calls on `WebElement`s annotated with `@FindBy` are tried multiple times until there is no more exception or a timeout of 5 seconds is reached.

This modification seems to improve the reliability of selenium tests at the expense of a performance degradation. With this PR, the CI build pipeline runs about 2min longer.

Another downside is that potential important Exceptions are hidden behind a timeout exception, which makes debugging even more difficult. On the other hand, each method invokation on a `WebElement` annotated with `@FindBy` can now be logged with the `trace` logging level.

Related Selenium source code:
- [LocatingElementHandler.java](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/support/pagefactory/internal/LocatingElementHandler.java)
- [DefaultFieldDecorator](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/support/pagefactory/DefaultFieldDecorator.java)

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/RepeatingLocatingElementHandler.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/RepeatingLocatingElementHandler.java
@@ -26,7 +26,7 @@ import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.pagefactory.ElementLocator;
 
 /**
- * Custom implementation of the default Selenium LocationngElementHandler.
+ * Custom implementation of the default Selenium LocatingElementHandler.
  * 
  * <p>The default implementation tries to invoke a WebElement method just once. However, there can be 
  * instances where a WebElement is already stale before the method can be invoked. In contrast, this
@@ -50,13 +50,13 @@ public class RepeatingLocatingElementHandler  implements InvocationHandler {
     @Override
     public Object invoke(Object object, Method method, Object[] objects) {
         Callable<Object> findAndInvoke = () -> {
-            logger.trace("invoke method " + method.getName() + " on WebElement with locator " + locator.toString());
+            logger.trace("invoke method {} on WebElement with locator {}", method.getName(), locator.toString());
             WebElement element;
             try {
                 element = locator.findElement();
             } catch (NoSuchElementException e) {
                 if ("toString".equals(method.getName())) {
-                    return "Proxy element for: " + locator.toString();
+                    return "Proxy element for: " + locator;
                 }
                 throw e;
             }


solth:
@thomaslow thanks a lot for these changes. I really hope they help to improve the reliability of the Selenium tests. Those `StaleElementReferenceException`s have haunted as for as long as I can remember. 

I don't think their frequency has been increased by the Java update, though, since that pull request has not been merged, yet, and similar exceptions have occured in multiple other pull requests that are based on the current `main` and `3.9.x` branches, as far as I can tell. 

Edit: The tests that often fails seems to be `LtpValidationConfigurationST.canAddNewLtpValidationConfigurationTest:138`

thomaslow:
> Edit: The tests that often fails seems to be LtpValidationConfigurationST.canAddNewLtpValidationConfigurationTest:138

@solth Yes, the tests that often fail seem to follow the pattern:

1. Enter data into the form via `sendKeys`
2. Click on the save button

There is either a `StaleElementReferenceException` or a general `ConditionTimeoutException` from https://github.com/kitodo/kitodo-production/blob/c095513ba9dae5617292b38da01a0e2bdf11613a/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/Page.java#L199-L200

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/RepeatingLocatingElementHandler.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.pagefactory.ElementLocator;
+
+/**
+ * Custom implementation of the default Selenium LocationngElementHandler.
```
solth:
```suggestion
 * Custom implementation of the default Selenium LocatingElementHandler.
```

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/RepeatingLocatingElementHandler.java
```diff
@@ -0,0 +1,92 @@
...
+        this.locator = locator;
+    }
+
+    /**
+     * Invokes a method on a WebElement.
+     */
+    @Override
+    public Object invoke(Object object, Method method, Object[] objects) {
+        Callable<Object> findAndInvoke = () -> {
+            logger.trace("invoke method " + method.getName() + " on WebElement with locator " + locator.toString());
```
solth:
```suggestion
            logger.trace("invoke method {} on WebElement with locator {}", method.getName(), locator.toString());
```
I think parameterized log messages are encouraged when logging non-constant strings.

Issue #6709, opened 2025-10-08
Different filter behaviour with and without process:


henning-gerhardt:
**Describe the bug**
We found a strange behaviour if we use the filter for processes with a more or less complex process title like `Heutwia_898482011-1794081501_01-s`.

When we enter only `Heutwia_898482011-1794081501_01-s` in the filter input field on the process page we got no hit. But if we enter `process:Heutwia_898482011-1794081501_01-s` in the filter input field we got the process.

But in both cases the filter looks after the filtering is executed like that:

<img width="514" height="92" alt="Image" src="https://github.com/user-attachments/assets/68462e12-f352-45fd-88dd-2bdf7b1b64ba" />

We thought / assumed that a complete process title search is executed in the case no `<field>:` f.e. `process:`, `id:`, ... is entered.

**Expected behavior**
Filter behaviour should be identical or should be displayed correct.

**Release**
3.9.0-RC1, 3.9.x branch and maybe main branch

BartChris:
See for some implementation details: https://github.com/kitodo/kitodo-production/pull/6618#issuecomment-3113657079

Issue #6743, opened 2025-11-03
Kitodo Search improvement (post Hibernate Search)


BartChris:
**Is your feature request related to a problem? Please describe.**
The current search in Kitodo Production 3.9 and later has some issues which should be fixed.

1. Searching in the top left search bar should do a search over all field, right now only the process title is searched

<img width="1778" height="248" alt="Image" src="https://github.com/user-attachments/assets/824ed1e8-6d25-42a5-b882-da68511e071f" />

2. When doing a free text search over all fields using the filter field, a prefix "process" is injected which indicates a search over the process title only. Instead "all" or "search" should be displayed.
 
<img width="1789" height="204" alt="Image" src="https://github.com/user-attachments/assets/c2593e53-6a70-4101-bff5-d513933d809b" />

<img width="1789" height="204" alt="Image" src="https://github.com/user-attachments/assets/1981e85c-271a-417a-b132-788819008ed6" />

3. Freetext search, especially for the process title sometimes renders less results than doing a search for a process title. At least for process titles search the free search should at least produces the same amount of results as the specific title search.

See: https://github.com/kitodo/kitodo-production/issues/6709

4. The process title search should for clarity purposes differentiate between an "exact" process title search and a "loose" search, to enable to find only titles with the exact name and not all processes starting with the same catalog identifier.
E.g.: `retro_555444` should bring `retro_555444` and not also `retro_555444_1` or `retro_555444_2`.

andre-hohmann:
The following discussion should be taken into account before extensive adjustments are conducted.
- #6772 

Especially:
> wäre meine Ideallösung eine durchgreifende: Der Verzicht auf Elasticsearch und die vollständige Ersetzung auch der Properties durch eine datenbankgestützte Suche in Volltextfeldern.

https://github.com/kitodo/kitodo-production/issues/6772#issuecomment-3580889906

PR #6764, opened 2025-11-24
Change tokenization for more coherent search

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6709

This addresses point 3) of https://github.com/kitodo/kitodo-production/issues/6743

This is a very small change which requires a longer explanation. As discussed extensively in the linked issue, the Kitodo search for process titles behaves different when searching a title in a search over `all` fields vs. a search in the `process title` field in the index.

The reason for that is, that Kitodo tokenizes the input string different at search time depending on which field is searched. In the `process title search` the Kitodo search effectively drops all tokens with a length of less then three.

So `Heutwia_898482011-1794081501_01-s` is effectively searched as `Heutwia_898482011-1794081501`. This works, because at index time the same happens. 

The behaviour when searching over all fields however is different. When searching over all fields Kitodo drops no token and searches for `Heutwia_898482011-1794081501_01-s`. This is however not what was indexed (only `Heutwia_898482011-1794081501`). As result the user gets no hits when searching this process title over all fields, which is super confusing.

The reason for this implementation was that when searching `projects` over all fields, it was considered necessary that "project A" can be differentiated from "project B". Therefor the tokenization for the project terms in the `all` field determines the search time tokenization of all other fields which leads to the confusing behavior. (See: https://github.com/kitodo/kitodo-production/pull/6618#issuecomment-3113657079)

As we already have a quite sophisticated `project` field search the @kitodo/kitodo-community-board decided after consulatation with @matthias-ronge, that this special tokenization behaviour is not necessary. We opt for using the same **default** tokenization rules for all fields indexed in the global search field and to drop the special constant.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ProcessKeywords.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/beans/ProcessKeywords.java
@@ -92,25 +92,25 @@ public class ProcessKeywords {
      */
     public ProcessKeywords(Process process) {
         // keywords for title search + default search
-        this.titleKeywords = filterMinLength(initTitleKeywords(process.getTitle()), LENGTH_MIN_DEFAULT);
+        this.titleKeywords = filterMinLength(initTitleKeywords(process.getTitle()));
 
         // keywords for project search in default search
         String projectTitle = Objects.nonNull(process.getProject()) ? process.getProject().getTitle() : "";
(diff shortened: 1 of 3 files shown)


matthias-ronge:
This is a business decision, not a technical one. The only consideration would be to omit the parameter in `FilterField` entirely if it is set to `LENGTH_MIN_DEFAULT` for **all** search fields.

BartChris:
@matthias-ronge Thanks a lot for the suggestions i implemented the changes, can you have another look?

PR #6768, opened 2025-11-25
IndexingService: Use try-with-resource statement

henning-gerhardt:
Use try-with-resource statement for auto closing the used Hibernate and Search session after they are not used anymore. Found by reviewing #6719 but this change is not related this pull request.


BartChris (approved):
Looks good to me.

PR #6683, opened 2025-09-25
Pagination editor improvements

oliver-stoehr:
This pull request adds several improvements for the pagination editor (#6455):

### Remove page selection:
Since multiselect is now available in the structure tree, a separate page selection in the pagination panel is no longer required and thus removed.

### Realign pagination editor fields:
The input fields and buttons are now rendered in a more responsive layout, that displays one or two columns, depending on the panel width.
Bildschirmfoto 2025-09-25 um 10 51 07
Narrow column layout

Bildschirmfoto 2025-09-25 um 10 50 45
Wider column layout

### Dynamically render input fields:
Not all input fields are required for all types of pagination. The input fields are now rendered conditionally, based on the selected type of pagination.

### Add alphabetic pagination type:
A new pagination type is now available using alphabetic characters (a, b, c, …, z, aa, ab, …). This pagination type can also be used in the advanced pagination by wrapping the start value in acute accents: `´a´`
(Resolves #6279)

### Add sheet pagination starting with Verso pages:
To enable recto-verso sheet pagination starting on the verso page a new Counting option is available: `VersoRecto Pagination`
(Resolves #6270)
Bildschirmfoto 2025-09-25 um 11 14 47

### Add option to select pages of a specific structure element:
The new option `Select assigned media` is added to the context menu of the logical structure tree. This selects all pages assigned to the currently selected logical structure element.
(Resolves #6458)
Bildschirmfoto 2025-09-25 um 10 53 01

### Add link to documentation for advanced pagination:
When the advanced pagination is active a new "help" icon is displayed next to the advanced pagination controls. This icon is linked to the documentation in the GitHub wiki.
Bildschirmfoto 2025-09-25 um 10 51 32

### Rename "Cancel" button:
The "Cancel" button is renamed to "Reset" to be more specific what this button actually does (resetting the values of pagination editor inputs).

### Add user settings:
Two new options are available in the user settings' metadata editor tab.
`Default pagination type` allows to set a type of pagination that will be preselected in the metadata editor. (Resolves #6045)
`Show logical page number below thumbnails` renders, when active, the logical page number (= order label) below each thumbnail in the structured view and detail view.
Bildschirmfoto 2025-09-25 um 10 56 23
New options in the user settings

Bildschirmfoto 2025-09-25 um 10 54 21
Logical page number under each thumbnail, when option is activated

Issue #6456, opened 2025-03-07
Fix scopes of CDI managed named beans


solth:
## Description

Most named CDI backing beans in Kitodo.Production - like for example `ProcessForm` or `UserForm` -  currently have the `@SessionScope`, which stems from the old version of the application that used JSF 1 and is not the correct scope for backing beans in modern JSF applications. The [correct scope for these beans](https://stackoverflow.com/a/7031941/5467214) would be `@ViewScoped`, but that wasn't available in JSF 1 and was only introduced in JSF 2 to which the application was updated during the development of Kitodo.Production 3.  A `ViewScoped` bean only lives as long as a particular view/page is loaded in a browser window of a user, and is destroyed as soon as the page unloads, thus freeing resources that currently clutter the session of a user. Just a [few](https://github.com/kitodo/kitodo-production/pull/5467) of the beans in Kitodo have been switch to this new scope though far, however. Additionally, Kitodo used the same backing bean for multiple pages/view - like `ProcessForm` for process **list** and process **edit** pages - which is conceptionally wrong.

Thus, the goal of this issue is to split existing `SessionScoped` beans that are used for multiple pages into multiple, `ViewScoped` beans, one for each individual page (e.g. `ProcessForm` -> `ProcessListView` + `ProcessEditView`, `UserForm` -> `UserListView` + `UserEditView`, `TemplateForm` -> `TemplateListView` + `TemplateEditView` etc.)

## Related Issues

The suggested improvements in this issue are a subset of the extensive list of proposed changes in #6418. I highly support nearly all points that @matthias-ronge makes in that issue. Since funding the whole list of changes at once seems not feasible, though, I would suggest to alternatively start with what I see as the highest priority by fixing the scope of the backing beans as described above, and implement further suggested changed in successive, later projects. If possible/feasible, individual improvements described #6418 can also be included, but that should be optional at this point.

## Expected Benefits of this Development

The correct scope should for backing beans should improve performance and memory usage, as unnecessary data is not kept in session scope but will be discarded once a particular page is unloaded. It will also prevent various bugs that might stem from data being kept in session unintentionally even when the user leaves a particular page.

## Estimated Costs and Complexity

* **high** ~ 10 - 15 working days

solth:
17 votes

solth:
All forms mentioned above have been refactored and put into the correct scope, thus this issue is completed. Thanks for the contribution, @thomaslow !

PR #6770, opened 2025-11-25
Refactor Bean Scopes: User, Role, Authority, Client, LdapGroup, LdapServer forms

### LINKED ISSUE #6826: Newly saved filters are sometimes missing in the list of saved filters
**Describe the bug**
When saving a new filter, it is saved to the database, but not always shown in the list of saved filters. The problem is that the saved filters are loaded from the "authenticated user object", which isn't updated when saving a new filter.

As of now, this problem only appears when opening the filter menu on a different form, because both the `UserForm` and `ProcessForm` are `@SessionScoped` and load a new user instance in case a new filter is saved. However, the respective other form is unaware of the new filter.

This is more problematic after switching to `@ViewScoped`, because then filters are always loaded from the "authenticated user object", meaning only outdated filters are loaded.

**To Reproduce**
1. Go to process list
2. Add a filter
3. See that it was added correctly
4. Go to user list
5. See that the new filter is not included in the list without re-logging or adding another filter (with updates the list)

See demonstration:

https://github.com/user-attachments/assets/509988c0-a583-445f-ad59-6af084d1ffb6

**Release**
Current main branch

 - OS: Fedora
 - Browser: Firefox

thomaslow:
This PR refactors all forms on the "users" page from `@SessionScoped` to `@ViewScoped`. In addition, a few minor cleanup improvements are made.

Related to: #6456

Updated forms:
- `UserForm`
- `RoleForm`
- `AuthorityForm`
- `ClientForm`
- `LdapGroupForm`
- `LdapServerForm`

Changes:
- Splitting form classes into list view and edit view
- Splitting user edit form into multiple classes, one for each tab view
- Column selection dialog only shows columns of the currently active tab
  - `tab` - the currently active tab id
  - `referrerListOptions` - the list options `sortField`, `sortOrder`, `firstRow` and `filter` encoded in a single parameter such they can be restored when navigating back to the list after editing a single entity
- Various cleanups and refactoring, e.g. 
  - moving database retrieval calls from getter to bean init methods
  - removing unused methods
  - adding Javadoc
- Fixes #6826

Due to the replacement of `UserForm.java`, `RoleForm.java`, ..., merging with the following PR needs to be done more carefully such that any improvements are not lost:

- #6803 
- #6802 
- #6683 
- #6851

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/user/UserListView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/user/UserListView.java
@@ -87,7 +87,7 @@ public class UserListView extends BaseListView {
      *         are "INWORK" and belong to process, not template
      */
     public static List<Task> getTasksInProgress(User user) {
-        return ServiceManager.getUserService().getTasksInProgress(user);
+        return ServiceManager.getTaskService().getTasksInProgress(user);
     }
 
     /**
--- a/Kitodo/src/main/java/org/kitodo/production/forms/CurrentTaskForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/CurrentTaskForm.java
@@ -834,18 +834,6 @@ public class CurrentTaskForm extends ValidatableForm {
     }
 
     /**
-     * Retrieve and return the list of tasks that are assigned to the user that are
-     * currently in progress.
-     *
-     * @return list of tasks that are currently assigned to the user that are
-     *         currently in progress.
-     */
-    public List<Task> getTasksInProgress() {
-        Stopwatch stopwatch = new Stopwatch(this, "getTasksInProgress");
-        return stopwatch.stop(ServiceManager.getUserService().getTasksInProgress(this.user));
-    }
-
-    /**
      * Get taskListPath.
      *
      * @return value of taskListPath
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
@@ -306,22 +306,35 @@ public class TaskService extends BaseBeanService<Task, TaskDAO> {
      */
     public void replaceProcessingUser(Task task, User user) {
         User currentProcessingUser = task.getProcessingUser();
-
         if (Objects.isNull(user) && Objects.isNull(currentProcessingUser)) {
             logger.info("do nothing - there is neither a new nor an old user");
         } else if (Objects.isNull(user)) {
-            currentProcessingUser.getProcessingTasks().remove(task);
             task.setProcessingUser(null);
         } else if (Objects.isNull(currentProcessingUser)) {
-            user.getProcessingTasks().add(task);
             task.setProcessingUser(user);
         } else if (Objects.equals(currentProcessingUser.getId(), user.getId())) {
             logger.info("do nothing - both are the same");
         } else {
-            currentProcessingUser.getProcessingTasks().remove(task);
-            user.getProcessingTasks().add(task);
             task.setProcessingUser(user);
         }
+    }
+
+    /**
+     * Retrieve and return all tasks assigned to the given user
+     * that are currently in progress and linked to a process.
+     *
+     * @param user the processing user
+     * @return list of tasks in progress for the given user
+     */
+    public List<Task> getTasksInProgress(User user) {
+        String hql = "FROM Task t WHERE t.processingUser = :user "
+                + "AND t.processingStatus = :status "
+                + "AND t.process IS NOT NULL";
+        Map<String, Object> params = Map.of(
+                "user", user,
+                "status", TaskStatus.INWORK
+        );
+        return dao.getByQuery(hql, params);
     }
 
     /**
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java
@@ -461,20 +461,6 @@ public class UserService extends BaseBeanService<User, UserDAO> implements UserD
     }
 
     /**
-     * Retrieve and return the list of tasks that are assigned to the user and
-     * that are "INWORK" and belong to process, not template.
-     *
-     * @return list of tasks that are currently assigned to the user and that
-     *         are "INWORK" and belong to process, not template
-     */
-    public List<Task> getTasksInProgress(User user) {
-        return user.getProcessingTasks().stream()
-                .filter(
-                    task -> task.getProcessingStatus().equals(TaskStatus.INWORK) && Objects.nonNull(task.getProcess()))
-                .collect(Collectors.toList());
-    }
-
-    /**
      * Changes the password for given User object.
      *
      * @param user
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/TaskServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/TaskServiceIT.java
@@ -17,12 +17,15 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.kitodo.MockDatabase;
 import org.kitodo.data.database.beans.Task;
+import org.kitodo.data.database.beans.User;
 import org.kitodo.data.database.enums.TaskStatus;
 import org.kitodo.data.database.exceptions.DAOException;
 import org.kitodo.production.helper.Helper;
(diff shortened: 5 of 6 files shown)


thomaslow:
@solth After refactoring the user list view from `@SessionScoped` to `@ViewScoped` I noticed that various state information is missing because of that, in particular:

- the current active tab
- the current page in the list
- sorting criteria for a list
- any search filter

I moved these state information into the URL as query parameters, which seems to work fine. I even keep these state information available by forwarding them to the respective edit views and reloading them after navigating back to the source list view (via another query parameter `referrerListOptions` similar to `keepPagination`).

In general, I think this approach works fine and now would allow to, e.g., create bookmarks with existing sorting and filter criteria.

However, I see a problem for the task and process list, and their respective filter options. When navigating to the task or process list, the current page and sorting options are reset to their defaults, but the filter option is loaded from the session scope. You can navigate anywhere in Kitodo.Production and still have the same filter option active when returning to the task or process list. 

I am not entirely sure, but I think this is most certainly desired behavior.

Anyway, this long-term information about the current active filter in the task and process list cannot be modeled as query parameters (alone). We could try cookies? What do you think?

BartChris:
> Anyway, this long-term information about the current active filter in the task and process list cannot be modeled as query parameters (alone). We could try cookies? What do you think?

Would it be an option to store only the filter-related information in a small, dedicated @SessionScoped bean and inject it into the @ViewScoped process and task list views?

thomaslow:
> Would it be an option to store only the filter-related information in a small, dedicated @SessionScoped bean and inject it into the @ViewScoped process and task list views?

Yes, of course this would be possible. Other options would be, storing it:

- in the database in a field `lastProcessListFilter` and `lastTaskListFilter`
- in a @SessionScoped bean
- in a Cookie with some lifetime (session, one day, one week, forever), or the browser's LocalStorage (without a lifetime)

In any case I would suggest storing all relevant list view options (including the filter) in URL query parameters as well (overwriting any long-term preferences), such that you can easily save various filters via browser bookmarks.

Kitodo/src/main/java/org/kitodo/production/forms/RoleListView.java
```diff
@@ -0,0 +1,162 @@
...
+        return this.stayOnCurrentPage;
+    }
+
+    /**
+     * Remove a role from the database.
+     * 
+     * @param role the role to be removed
+     * @return true if role was removed
+     */
+    public static boolean deleteRole(Role role) {
```
solth: Maybe we can move this method to the `RoleService`, as it contains quite a bit of business logic and isn't directly called from the template? If it stays in this class, it could probably be made private, since it is only called from within this class itself, isn't it?

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewClientsTab.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+
+    /**
+     * Save clients tab.
+     *
+     * @return true if user clients data was saved
+     */
+    public boolean save() {
+        return true;
```
solth: I am not quite sure why we need this method. It only ever returns `true`. Have you added this because some tabs have actual conditions that are checked before the user can be saved?

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
```diff
@@ -0,0 +1,278 @@
...
+     * Validate the entered password.
+     * 
+     * @return the set of validation violations
+     */
+    private Set<ConstraintViolation<KitodoPassword>> getPasswordViolations() {
+        if (isLdapServerReadOnly()) {
+            return Collections.emptySet();
+        }
+        KitodoPassword validPassword = new KitodoPassword(passwordToEncrypt);
+        ValidatorFactory factory = Validation.buildDefaultValidatorFactory();
```
solth: My IDE tells me that `ValidatorFactor` is `AutoClosable` and should be used with `try-with-resources`. Would something like 
```
        try (ValidatorFactory factory = Validation.buildDefaultValidatorFactory()) {
            Validator validator = factory.getValidator();
            return validator.validate(validPassword);
        }
```
work?

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
```diff
@@ -0,0 +1,278 @@
...
+    /**
+     * Check whether the old password matches the password stored in the database.
+     * 
+     * @return true if the old password does not match the password stored in the database
+     */
+    private boolean isOldPasswordInvalid() {
+        if (!ServiceManager.getSecurityAccessService().hasAuthorityToEditUser()) {
+            return !Objects.equals(this.oldPassword, passwordEncoder.decrypt(this.userObject.getPassword()));
+        }
+        return false;
```
solth: Despite being just a few lines, I find this whole method confusing because of the multiple negations (e.g. "invalid" instead of "valid", `if(!ServicManager...)`, `return !Objects.equals` etc.). I know you just copied this method from the `UserForm` and didn't author it, but perhaps now would be a good opportunity to simplify the code?

Kitodo/src/main/webapp/WEB-INF/templates/includes/users/userList.xhtml
```diff
@@ -30,15 +30,15 @@
...
+                #{UserListView.formatString('confirmDeleteUserWithTasks', UserListView.confirmResetTasksDialogUser.fullName)}
             </div>
         </h:panelGroup>
         <h:panelGroup>
             <h:form>
                 <p:commandButton
                         id="yesButton"
                         value="#{msgs.yes}"
-                        action="#{UserForm.resetTasksAndDeleteUser}"
+                        action="#{UserListView.resetTasksAndDeleteUser(confirmResetTasksDialogUser)}"
```
solth:
```suggestion
                        action="#{UserListView.resetTasksAndDeleteUser(UserListView.confirmResetTasksDialogUser)}"
```
I think this needs the reference the bean that contains user object, doesn't it? Without it I get a `NullPointerException` when trying to delete a user with assigned tasks:
Bildschirmfoto 2026-01-09 um 14 39 41

thomaslow:
@solth Thank you for your feedback. I'll update this branch next week.

Kitodo/src/main/java/org/kitodo/production/forms/ClientListView.java
```diff
@@ -0,0 +1,107 @@
...
+import org.primefaces.model.SortMeta;
+import org.primefaces.model.SortOrder;
+
+@Named("ClientListView")
+@ViewScoped
+public class ClientListView extends BaseListView {
+
+    private static final Logger logger = LogManager.getLogger(ClientListView.class);
+
+    public static final String VIEW_PATH = MessageFormat.format(REDIRECT_PATH, "users") + "&tabIndex=2";
```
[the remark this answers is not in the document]
thomaslow: done

Kitodo/src/main/java/org/kitodo/production/forms/LdapServerListView.java
```diff
@@ -0,0 +1,105 @@
...
+import org.primefaces.model.SortMeta;
+import org.primefaces.model.SortOrder;
+
+@Named("LdapServerListView")
+@ViewScoped
+public class LdapServerListView extends BaseListView {
+
+    public static final String VIEW_PATH = MessageFormat.format(REDIRECT_PATH, "users") + "&tabIndex=5";
+
+    private static final Logger logger = LogManager.getLogger(LdapServerEditView.class);
```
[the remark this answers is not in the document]
thomaslow: done

Kitodo/src/main/java/org/kitodo/production/forms/RoleListView.java
thomaslow: done

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/BaseTabView.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * The list of allowed tab ids (in order) for a specific tab view (to sanitize URL parameter value) 
+     * and determine the correct active tab index. 
+     * 
+     * <p>This method needs to be overwritten by each tab view. It should only return tabs that are actually 
+     * shown to the user (considering authorities), such that the correct index can be calculated.
+     * 
+     * @return the set of allowed tab ids
```
solth:
```suggestion
     * @return the list of allowed tab ids
```
Maybe the JavaDoc should be slightly adjusted to reflect that this method returns a list and not just a set.

Kitodo/src/main/java/org/kitodo/production/forms/user/UserListView.java
```diff
@@ -0,0 +1,276 @@
...
+    /**
+     * Set filter based on the URL query parameter "filter", which can be any string.
+     * 
+     * @param encodedFilter the filter as URL query parameter to be set as new filter
+     */
+    public void setFilterFromTemplate(String encodedFilter) {
+        logger.error("UserListView.setFilterFromTemplate called with: " + encodedFilter);
+        if (Objects.nonNull(encodedFilter) && !encodedFilter.isEmpty()) {
+            String decodedFilter = encodedFilter.replace("%26", "&");
+            logger.error("UserListView.setFilterFromTemplate with decoded filter: " + decodedFilter);
```
solth:
```suggestion
            logger.error("UserListView.setFilterFromTemplate with decoded filter: {}", decodedFilter);
```

solth (commented):
The error I countered when deleting a user with assigned tasks does not occur anymore. Everything looks good and works now. Apart from the one code conflict that needs to be resolved I only found some very minor nitpicks you might want to take another look at.

thomaslow:
> @thomaslow one thing I noticed: all tabs on the user list and user edit pages are individual, named, view scoped beans. (for example `UserEditViewClientsTab.java`, `RoleListView.java` etc.).
> 
> Does it bring any advantages to have these classes as managed beans instead of "POJOs", e.g. plain old java objects? Since they are all part of `UserListView.java` and `UserEditView.java` respectively, which themselves are already `ViewScoped` beans, would it make the tabs automatically view scoped as well if they were simple java classes added as properties to those two ViewScoped beans as properties?

Good question. I don't think there are any drastic advantages or differences between them. If I remember correctly, the Metadata Editor uses the "POJO"-way to structure various panels into separate classes. Maybe one "advantage" of Managed Beans would be, that you can `@Inject` them from different classes and still only have one instance per JSF lifecycle. Otherwise, you would have to pass around the POJO instance manually. I remember having this "inconvenience" in the Metadata Editor code.

My previous experience with JavaEE and JSF is limited to a few small research projects in 2010-2015. Based on that, I don't really feel comfortable saying one way is better than another. In my experience, I think it is more important to consistently split code based on the principle of "separation of concerns" rather than the exact method of organizing the code.

PR #6778, opened 2025-11-27
Fix hasProcesses lookup and improve ProjectForm performance

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6777
Fixes https://github.com/kitodo/kitodo-production/issues/6776

PR #6779, opened 2025-11-27
[3.9] Fix hasProcesses lookup and improve ProjectForm performance

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6778

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProjectForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProjectForm.java
@@ -481,12 +481,18 @@ public class ProjectForm extends BaseForm {
      *            Project object
      */
     public void setProject(Project project) {
+        // has to be called if a page back move was done
         cancel();
         this.project = project;
         try {
             hasProcesses = ServiceManager.getProjectService().hasProcesses(project.getId());
         } catch (DAOException e) {
-            Helper.setErrorMessage("Error checking for assigned processes", logger, e);
+            Helper.setErrorMessage(
+                    ERROR_DATABASE_READING,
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/ProjectService.java
```diff
@@ -301,6 +301,21 @@ public String getProjectTitles(List<Project> projects) throws DAOException {
...
 
+    /**
+     * Checks whether the given project has any processes assigned to it.
+     *
+     * @param projectId
+     *          the ID of the project to check
+     * @return return true if at least one process belongs to the project,
+     * false otherwise
+     */
+    public boolean hasProcesses(int projectId) throws DAOException{
```
henning-gerhardt: - Missing white space between `DAOException` and `{` character.
- Having tests for this new method should be made as long an test for zero and an other test for at least one process assigned to a project. A test for the throwing of the `DAOException` would be difficult and can be omitted but if you found a simple way you can add this test too.

Kitodo/src/main/java/org/kitodo/production/forms/ProjectForm.java
```diff
@@ -475,16 +475,20 @@ public Project getProject() {
...
      */
     public void setProject(Project project) {
-        // has to be called if a page back move was done
         cancel();
         this.project = project;
-        hasProcesses = !project.getProcesses().isEmpty();
+        try {
+            hasProcesses = ServiceManager.getProjectService().hasProcesses(project.getId());
+        } catch (DAOException e) {
+            Helper.setErrorMessage("Error checking for assigned processes", logger, e);
```
henning-gerhardt: Sorry forget to add this one: the error message did not get translated right? As this is show to the user a translation would be good or?

PR #6782, opened 2025-11-27
Use multiple field/value terms in ES query

BartChris:
This PR addresses point 8) of https://github.com/kitodo/kitodo-production/issues/6743. The problem with the current implementation is, that when you have a search like` projectname:MDP_LIK `to retrieve all processes which have a metadata value of "MDP_LIK" in the project field, Kitodo first tokenizes the term to "MDP" and "LIK" and then issues **two** requests for Elasticsearch. 
Both requests return independent sets of IDs. And when both sets do not intersect we might get less hits then we actually have.

For each token constructed at query time we inject the list of IDs (retrieved from the Index) into SQL, so at the end we have sth like `WHERE id IN (x,y,z) and ID in (a,b,c)`. The individual ID lists might be huge because they contain all results which just contain the fragment "MDP".

My change does ensure that only one query is sent to the search index and only one list of IDs is returned, so we do not get multiple contradicting ID lists. It also in many cases ensures that the ID lists are way shorter because we already intersect at the Elasticsearch level and not just pass huge ID lists to the database which are not relevant.

This also frees us from having to construct complex SQL restrictions for multiple returned ID list. Since we only have one list we can just append a simple `WHERE id in (<list od ids>)`.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryPart.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryPart.java
@@ -13,9 +13,7 @@ package org.kitodo.production.services.data;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.List;
-import java.util.Map;
 
 import org.apache.commons.lang3.tuple.Pair;
 import org.kitodo.data.database.beans.ProcessKeywords;
@@ -25,8 +23,6 @@ import org.kitodo.data.database.beans.ProcessKeywords;
  * search index.
  */
 class IndexQueryPart implements UserSpecifiedFilter {
-
-    private static final String UNIQUE_PARAMETER_EXTENSION = "query";
 
     private static final char VALUE_SEPARATOR = 'q';
     private final List<String> lookfor = new ArrayList<>();
@@ -88,33 +84,14 @@ class IndexQueryPart implements UserSpecifiedFilter {
     }
 
     /**
-     * Inserts the search parameters into the database query logic.
-     * 
-     * @param varName
-     *            variable name of the HQL search
-     * @param parameterName
-     *            name of the search parameter for the results
-     * @param idField
-     *            field name of the process ID
+     * Adds the prepared index search terms for this filter to the list of index queries.
+     *
      * @param indexQueries
      *            puts the prepared tokens for the search queries here
-     * @param restrictions
-     *            puts the HQL restrictions here
      */
-    void putQueryParameters(String varName, String parameterName, String idField,
-            Map<String, Pair<FilterField, String>> indexQueries,
-            Collection<String> restrictions) {
-        if (lookfor.size() == 1) {
-            restrictions.add(varName + "." + idField + (operand ? " IN (:" : " NOT IN (:") + parameterName + ')');
-            indexQueries.put(parameterName, Pair.of(filterField, lookfor.getFirst()));
-        } else if (lookfor.size() >= 1) {
-            int queryCount = 0;
-            for (String lookingFor : lookfor) {
-                queryCount++;
-                String uniqueParameterName = parameterName + UNIQUE_PARAMETER_EXTENSION + queryCount;
-                restrictions.add(varName + "." + idField + (operand ? " IN (:" : " NOT IN (:") + uniqueParameterName + ')');
-                indexQueries.put(uniqueParameterName, Pair.of(filterField, lookingFor));
-            }
+    void putQueryParameters(List<Pair<FilterField, String>> indexQueries) {
+        for (String lookingFor : lookfor) {
+            indexQueries.add(Pair.of(filterField, lookingFor));
         }
     }
 
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryPart.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryPart.java
@@ -15,7 +15,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
-import org.apache.commons.lang3.tuple.Pair;
 import org.kitodo.data.database.beans.ProcessKeywords;
 
 /**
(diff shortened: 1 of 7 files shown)


Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
```diff
@@ -123,24 +125,43 @@
...
+        List<Integer> ids = query.fetchAll().hits();
+
+        String termSummary = terms.stream()
+                .distinct()
+                .map(t -> t.getLeft() + "=\"" + t.getRight() + "\"")
+                .collect(Collectors.joining(", "));
+        logger.debug(
+                "Searching {} IDs with terms {}: {} hits",
+                beanClass.getSimpleName(),
+                termSummary,
```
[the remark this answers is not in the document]
BartChris: It seems like we are not allowed to log the terms somebody searched for in the debug log. I will show only the search Keys for now.

Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
```diff
@@ -126,25 +127,45 @@ public CompletionStage<?> startIndexing(Class<? extends BaseBean> type, MassInde
...
             SearchProjection<Integer> idField = searchSession.scope(beanClass).projection().field("id", Integer.class)
                     .toProjection();
-            List<Integer> ids = searchSession.search(beanClass).select(idField).where(function -> function.match().field(
-                    searchField).matching(value)).fetchAll().hits();
-            logger.debug("Searching {} IDs in field \"{}\" for \"{}\": {} hits", beanClass.getSimpleName(), searchField,
-                    value, ids.size());
+            var query = searchSession.search(beanClass)
+                    .select(idField)
+                    .where(f -> {
+                        var bool = f.bool();
```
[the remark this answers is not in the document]
BartChris: You have a point, i will try to clarify. This structure is idiomatic Hibernate Search DSL:
https://docs.hibernate.org/search/6.2/reference/en-US/html_single/#query-predicate

matthias-ronge (commented):
There is only one question for me to clarify, if you can say that this isn’t a problem, then I approve the PR.

Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryPart.java
```diff
@@ -88,33 +84,14 @@ public FilterField getFilterField() {
...
-            Collection<String> restrictions) {
-        if (lookfor.size() == 1) {
-            restrictions.add(varName + "." + idField + (operand ? " IN (:" : " NOT IN (:") + parameterName + ')');
-            indexQueries.put(parameterName, Pair.of(filterField, lookfor.getFirst()));
-        } else if (lookfor.size() >= 1) {
-            int queryCount = 0;
-            for (String lookingFor : lookfor) {
-                queryCount++;
-                String uniqueParameterName = parameterName + UNIQUE_PARAMETER_EXTENSION + queryCount;
-                restrictions.add(varName + "." + idField + (operand ? " IN (:" : " NOT IN (:") + uniqueParameterName + ')');
```
matthias-ronge, line 115: Handling of NOT search (`operand`) was removed and has no replacement.

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -297,7 +297,7 @@ private BeanQuery createProcessQuery(Map<?, String> filters, boolean showClosedP
         Collection<Integer> projectIDs = ServiceManager.getUserService().getCurrentUser().getProjects().stream().filter(
             project -> showInactiveProjects || project.isActive()).map(Project::getId).collect(Collectors.toList());
         query.restrictToProjects(projectIDs);
-        query.performIndexSearches();
+        query.applyIndexRestriction("id");
```
matthias-ronge: Please use a static final string for all occurrences of `"id"`.

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -177,7 +177,7 @@ public Long countResults(Map<?, String> filters, boolean onlyOwnTasks, boolean h
             boolean showAutomaticTasks, List<TaskStatus> taskStatus) throws DAOException {
 
         BeanQuery query = formBeanQuery(filters, onlyOwnTasks, hideCorrectionTasks, showAutomaticTasks, taskStatus);
-        query.performIndexSearches();
+        query.applyIndexRestriction("process.id");
```
matthias-ronge: Please use a static final string for both occurrences of `"process.id"`.

matthias-ronge (changes requested):
The NOT search (value of `operand`) is lost.

solth:
@BartChris please rebase against `main` to resolve conflicts!

Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -142,6 +142,7 @@
 public class ProcessService extends BaseBeanService<Process, ProcessDAO> {
 
     private static final Map<String, String> SORT_FIELD_MAPPING;
+    private static final String FIELD_ID = "id";
```
solth, line 145: In the future such generic constants - whose values are used in many other classes as well - could be moved to `Kitodo-API/src/main/java/org/kitodo/constants/StringConstants.java`

BartChris:
> The NOT search (value of `operand`) is lost.

Very good point, will have to look deeper here. A negation index query is right now not documented. https://github.com/kitodo/kitodo-production/wiki/Suchen-und-filtern

BartChris:
I added back the option to prefix metadata searches with "-", e.g. `-TitleDocMain:Urkunde` to allow for "Does not contain"-searches. To pass the information (field, value, operand) around i introduced a Java `record`.
My hope is that we can reduce a lot of the complexity invoved in passing information between the DB search and Index search when switching to a DB only search.

matthias-ronge:
> A negation index query is right now not documented. https://github.com/kitodo/kitodo-production/wiki/Suchen-und-filtern

It is only mentioned here: https://github.com/kitodo/kitodo-production/wiki/Suche-und-filtern#ausschlie%C3%9Fende-suche

Kitodo/src/main/java/org/kitodo/production/services/data/IndexQueryTerm.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.production.services.data;
+
+public record IndexQueryTerm(
```
matthias-ronge, line 14: Please add javadoc.

```suggestion
/**
 * A token of an index search.
 * 
 * @param field
 *            search field selected by the user
 * @param token
 *            index query term
 * @param operand
 *            whether the search must match
 */
public record IndexQueryTerm(
```

matthias-ronge (approved):
This is OK now. You may want to add Javadoc to the `record`.

solth:
@BartChris would you mind rebasing your branch against `main` one more time?

PR #6785, opened 2025-11-27
[3.9] Improve Selenium Test Reliability via Custom Locating Handler

thomaslow:
Backport of #6762


solth (commented):
Same as the 3.8 backport: Code looks good and works, just one very small remark regarding a wrong article.

Thanks for these backports, @thomaslow !

PR #6796, opened 2025-12-05
Upgrade to Java 21

thomaslow:
This PR upgrades Kitodo.Production to Java 21.

Distro support would be:

|  | Distro | EOL | Java | Tomcat |
| ---------- | --------- | -----  | -----  | ----------- |
| ❌ | ~~Ubuntu 22.04~~ | ~~April 2027~~ | ~~11~~ | ~~9~~ |
| ❌ | ~~RHEL 9 / Rocky Linux 9~~ | ~~May 2027~~ | ~~17~~ | ~~9~~ |
| ✅ | Ubuntu 24.04 | April 2029 | 21 | 10 | 
| ❌ | ~~Debian 12~~ | ~~June 2026~~ | ~~17~~ | ~~10~~ |
| ✅ | Debian 13 | ~Mid 2028 | 21 | 10 (~~11~~) |
| ✅ | RHEL 10 / Rocky Linux 10 | May 2030 | 21 | 10 |


solth:
I feel a little bad asking this, since in https://github.com/kitodo/kitodo-production/pull/6692#issuecomment-3611960993 I supported @henning-gerhardt s idea to to upgrade even further to Java 21, but if the only difference in upgrading to Java 21 instead of Java 17 is _loosing_ support for one Linux distribution (e.g. Debian 12, when I compare the both tables of supported distributions that @thomaslow provided), I wonder whether this is a good idea after all? Do we _gain_ support for a different release instead? Or should we always upgrade the supported versions to the newest version available to make Kitodo.Production as future-proof as possible, even if it comes with loosing support for some specific current systems?

I am not really against this upgrade, but genuinely interested in everyones opinion (@henning-gerhardt , @BartChris , @thomaslow , @matthias-ronge etc.).

PR #6797, opened 2025-12-05
Fix Spring warning about `commons-logging` conflict

thomaslow:
After updating Spring in #6692 , there is a warning message printed during the startup (see [original comment](https://github.com/kitodo/kitodo-production/pull/6692#pullrequestreview-3486268165) by @henning-gerhardt ):

> Standard Commons Logging discovery in action with spring-jcl: please remove commons-logging.jar from classpath in order to avoid potential conflicts

Spring provides its own implementation of the `commons-logging` classes and wants that package to be removed from the classpath to avoid conflicts. This PR excludes `commons-logging` in pom dependences such that the warning disappears.

All logging seems to work the same as before.


henning-gerhardt (approved):
The message from `spring-jcl` did not appear anymore. So it should be fixed for now. As I see even normal log messages like before I think there should no regression.

This works only until a new or an existing dependency get updated and introduce the `common-logging` artifact as new dependency which will raise the message from `spring-jcl` again.

PR #6802, opened 2025-12-08
Remove reliance on bidirectional association in task-user relation

BartChris:
Adresses https://github.com/kitodo/kitodo-production/issues/5813 (But no full fix of the described memory usage since this probably requires more changes)

As outlined in https://github.com/kitodo/kitodo-production/issues/5813#issuecomment-3617287774 bidrectional access to the non owning side of a relation can be problematic if the collections are too large. I would prefer if we remove the collection property. 

```java
@OneToMany(mappedBy = "processingUser", cascade = CascadeType.PERSIST)
private List<Task> processingTasks;
```

to not encourage access to and hydration of potential large collections. But we use those accessors in tests so i only replaced the application-level usage of it with an HQL query. Since the `processingTasks` accessor is no longer used at runtime, we do not need to maintain manual synchronization between both sides of the association.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/CurrentTaskForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/CurrentTaskForm.java
@@ -828,18 +828,6 @@ public class CurrentTaskForm extends BaseForm {
     }
 
     /**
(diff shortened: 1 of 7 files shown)


BartChris:
Should be rebased on https://github.com/kitodo/kitodo-production/pull/6803

PR #6803, opened 2025-12-08
Optimize user list

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/4378

This Pull Request tries to adress the slow loading user list by using optimized SQL queries to retrieve the information we need for the user list. 
Some aspects of the PR have already been discussed here. https://github.com/kitodo/kitodo-production/pull/6802

One aspect is whether the SQL querie for loading tasks "in processing" is too restrictive as it only looks for tasks where the user is actually the processing user.

https://github.com/BartChris/kitodo-production/blob/96c5e8e32408acd42e414309e2f9810e45d13a59/Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java#L660-L665

However, i think this was the case before with the old implementation:

https://github.com/kitodo/kitodo-production/blob/96c5e8e32408acd42e414309e2f9810e45d13a59/Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java#L470-L475 

which only retrieved tasks where the user is set as `processing user `, if i am not mistaken. I believe this behavior is correct, as this check is only relevant in the user list when determining whether a user has tasks in progress, for example when deciding if the user can be deleted.

The other question was why we call Primefaces from the `LazyUsermodel` (see: https://github.com/kitodo/kitodo-production/issues/4378#issuecomment-3632531737). We however also do that in the `LazyBeanModel` and the `LazyProcessModel`

https://github.com/kitodo/kitodo-production/blob/53eec68a66e89d7c944f0172c15b1d1820369a01/Kitodo/src/main/java/org/kitodo/production/model/LazyProcessModel.java#L146

https://github.com/kitodo/kitodo-production/blob/53eec68a66e89d7c944f0172c15b1d1820369a01/Kitodo/src/main/java/org/kitodo/production/model/LazyBeanModel.java#L131

I basically adapted the implementation from there, so this maybe deserves a seperate discussion.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
@@ -300,11 +300,11 @@ public class UserForm extends BaseForm {
      * user from a connected LDAP service.
      */
     public void checkAndDelete() {
-        if (!hasTasksInProgress(userObject)) {
-            deleteUser(userObject);
-        } else {
+        if (hasTasksInProgress(userObject)) {
             PrimeFaces.current().ajax().update("usersTabView:confirmResetTasksDialog");
             PrimeFaces.current().executeScript("PF('confirmResetTasksDialog').show();");
+        } else {
+            deleteUser(userObject);
         }
     }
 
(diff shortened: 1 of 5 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/UserForm.java
```diff
@@ -300,14 +300,15 @@ public void resetTasksToOpen(User user) {
      * user from a connected LDAP service.
      */
     public void checkAndDelete() {
-        if (getTasksInProgress(userObject).isEmpty()) {
+        if (!hasTasksInProgress(userObject)) {
```
solth: Perhaps this `if...else` can be reverted to prevent the `!`? Something like
```
if (hasTasksInProgress(userObject)) {
    PrimeFaces.current().ajax().update("usersTabView:confirmResetTasksDialog");
    PrimeFaces.current().executeScript("PF('confirmResetTasksDialog').show();");
} else {
    deleteUser(userObject);
}
```

solth (changes requested):
@BartChris thank you very much, this does seem to improve the performance of the user list siginicantly. As far as I could test the caches did not create any erroneous lists. I just have a few suggestions concerning the code.

solth:
@thomaslow it would be great if you could take a look at this pull request, too, since it does touch on several aspects you refactored in #6770

solth (approved):

Issue #6805, opened 2025-12-09
Save button is not activated after change is made in edit mode


henning-gerhardt:
**Describe the bug**
It is not possible anymore to save changes in different "dialog" forms as the save button is not activated after a change is made. This happens so far at

- changing roles
- changing import configurations

Maybe there are more dialogs which are broken

**To Reproduce**
1. Go to Projects -> Roles
2. Try to edit an existing role
3. 'Save' button stays gray and is not clickable

**Expected behavior**
After a change is made the Save button should be clickable and the change itself should be stored.

**Release**
current main Branch (2025-12-09)

solth:
> Do we have more roles on some installations?

No, I think the problem lies in the database of my system, were I do have a few more authorities/permissions than in the current `main` branch,  for branches currently in development.

solth:
> As [@thomaslow](https://github.com/thomaslow) mentioned the amount of roles: I looked into my local system and there I had already 29 roles on two clients and in the cross table `role_x_authority` 389 entries and I know that we have a few more roles in our productive environment already as we stay or use many roles with small assigned amount of authorities than few roles with a lot of authorities assigned.

I think this might be a misunderstanding. Not the number of roles is relevant on the "roleEdit" page, but instead the number of permissions (or "authorities") that can be assigned to a role. And since I have some additional, experimental authorities in my database (see above), the limit of 150 parts was slightly to low in my case. 

But since these permissions will probably be added in the near future into the main branch, I think it would be wise to make for some "headroom" in our configuration (since we cannot change the Tomcat `server.xml` settings in our pull requests, it would be akward to always change the installation guide when a new permission is added and to inform all users that they have to change their server settings every time)

thomaslow:
> I think this might be a misunderstanding. Not the number of roles is relevant on the "roleEdit" page, but instead the number of permissions (or "authorities") that can be assigned to a role.

Yes, you are correct. My mistake, sorry.

henning-gerhardt:
Okay. I had forgot to add `maxPartCount` parameter to the Tomcat `server.xml`. I added the recommended value of `150` and there is no change on editing a role. After I changed this parameter to a value of `200` changing roles did work - at least the save button switched from gray to blue.

[...]

PR #6810, opened 2025-12-10
Remove multipart form encoding from all edit views except mass import

thomaslow:
See #6805

### DIFF
--- a/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/CommandService.java
+++ b/Kitodo-FileManagement/src/main/java/org/kitodo/filemanagement/CommandService.java
@@ -39,7 +39,7 @@ class CommandService {
         CommandResult commandResult = command.runCommand(script);
         List<String> commandResultMessages = commandResult.getMessages();
         if (!commandResultMessages.isEmpty() && commandResultMessages.get(0).contains("IOException")) {
-            throw new IOException(commandResultMessages.get(1));
+            throw new IOException(commandResultMessages.get(0));
         }
         return commandResult;
     }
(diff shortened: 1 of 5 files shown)


thomaslow:
> @thomaslow : is the `maxPartCount` in Tomcat `server.xml` file still needed?

Yes `maxPartCount` is still needed, but it will only make a difference on pages or dialogs with file uploads that still need to use the `multipart` encoding. 

The role edit page should be unaffected from any `maxPartCount` setting now and work reliably in any case.

I haven't tested it, but with this PR it might be possible to reduce `maxPartCount` to 50 or even 20. The mass import form submission currently uses 9 parts.

Anyway, in case we would like to extend these forms with additional inputs, it might be wise to still plan for some headroom and still ask people to set `maxPartCount` to 50 or so.

I am not sure how high `maxPartCount` can be set before it gets to "dangerous" levels with respect to the [CVE](https://lists.apache.org/thread/nzkqsok8t42qofgqfmck536mtyzygp18).

Also, please keep in mind that this also applies to Tomcat 9 and existing deployments in case they update to Tomcat v9.0.106 or later. So, this change should probably be backported as well.

henning-gerhardt:
Thanks for your answer @thomaslow . 

I looked into the default `server.xml` file for Tomcat 9 on Debian 11 (9.0.107-0+deb11u1 but I did not look yet if this CVE is fixed in Debian 11) but there is for `3.8.x` no change on `maxPartCount` for our usage needed and I can edit roles without this change here. Even for `3.9.0-RC1` (and maybe even `3.9.0`) there is no need for a defined `maxPartCount`on Debian 11. to edit roles.

PR #6815, opened 2025-12-11
Add XML schema validation for metadata, rulesets and mapping files

### LINKED ISSUE #5877: Validation of search responses (SRU/OAI at least)

Currently (January 2024) is assumpted that a search response is every time correct in the format of the used search type (OAI or SRU). If this is not the case an error message appear that the search is not returning a positive result. But this is misleading as the response format is complete wrong (f.e. instead of a XML response a HTTP resonse was send) or not in the expected format (f.e. using a SRU 1.1 search request and got a SRU 2.0 response back).

After s search specific request was send and a positive answer is returned (f.e. HTTP status code 200) the response should be validated against the used search type (f.e. SRU 2.0). If this request is validated correct all needed additional tasks like extracting number of hits, extracting embedded meta data and more should be startet. If the validation is negative a message should be appear for the user (and maybe in the log) that search was failing because of an incorrect response data format. If a response validation is executed or not should be configurable in the import configuration as it is possible that a used search catalogue can not response to 100% in response format.

For SRU 2.0 and OAI are schema xsd files for validating (the search response) are available. For other versions of SRU they may can be found or adjusted from the SRU 2.0 response format.

For custom import configurations there should maybe a possibility to store a custom xsd file which can be used too. So maybe storing used xsd files for response validation should be general available like the to be used mapping files on import?

### LINKED ISSUE #6367: Comprehensive validation of XML files
## Description

Kitodo.Production works with many different XML files: rulesets, metadata mappings and configuration files are created and changed by the institutions working with Kitodo. Additionally there are external files like the imported metadata files in MODS, MARCXML or EAD format. 

This makes the application somewhat vulnerable to mistakes made outside of Kitodo itself, because all of these files can contain malformed XML, may not adhere to the corresponding schema definitions or be not readable at all. Since the functionalities of Kitodo rely heavily on these XML files being correct, it would be very beneficial to introduce a set of validations that have to be performed on _all_ of them whenever they are processed:

- check if file exists -> otherwise: show `File XY does not exist` error message
- check if file is readable -> otherwise: show `File XY is not readable` error message 
- check if file contains well-formed XML -> otherwise: show `File does not contain well-formed XML` error message
- check if contained file is valid against corresponding XML schema definition (for example Kitodo ruleset schema): otherwise: show `File does not contain valid Kitodo ruleset XML` error message
  - in case of XML responses from search interfaces like SRU or OAI, check, if response contains valid SRU/OAI container XML -> otherwise: show `Response does not contain valid SRU XML`

Where applicable, specific detected errors should also be listed (e.g. "opening XML tag is not closed" for XML checks or "Field XYZ is not allowed in MODS" for schema validation errors), ideally in the error message in the frontend, but at least in the Kitodo log files.

Some of these checks are already performed at some points in the system (for example, when opening a process in the metadata editor, an error message is shown when the file could not be found), but many other XML files are not validated before usage (for example XML files imported from external sources during process creation is not validated against the schema definition of the metadata format configured in the corresponding `ImportConfiguration`, yet, and thus processing this imported data with - unvalidated - XSL mapping files sometimes fails because they do not actually contain the expected valid MODS or MACRXML; similarly ruleset files are not validated against `ruleset.xsd` before saving them or applying them during process creation or in the metadata editor - see discussion #6168).

## Related Issue

#5877

Many problems we encountered over the years in Kitodo stem from external or internal XML files not containing the expected, valid contents, so introducing more checks and adding corresponding feedback to the GUI will harden the application, prevent malformed or invalid XML files and therefor also support and simplify problem investigation and solving.

* **medium** ~ around 5 - 7 working days

solth:
This pull request adds schema based validation for XML content as described in #6367. 

Since Kitodo.Production makes heavy use of XML files in all parts of the application and depends on those files to conform to specific patterns in order for the programm to run correctly, the aim of this pull request is to harden the application against errors caused by schema invalid XML files. It should also help users and developers finding and fixing XML content related errors.

The validation is called whenever a metadata, ruleset or mapping file is loaded: 
- (internal) metadata files are validated against schemata `mets.xsd` and `kitodo.xsd`
- ruleset files are validated against the ruleset schema `ruleset.xsd`
- mapping files are validated against the XSLT schema definition file `xslt20.xsd`

If a metadata file is not valid against the corresponding schema definition, loading the file in the metadata editor fails and a popup dialog is displayed to inform the user about the specific validation errors:

Bildschirmfoto 2025-12-11 um 15 00 29

Additionally, saving a "Mapping file" or "Ruleset" with a schema invalid XML file will also be prevented:

Bildschirmfoto 2025-12-11 um 15 57 25

The user - or rather the admin - is then responsible for fixing the XML structure of the corresponding file before processing can continue. The specific validation errors listed in the popup dialog can be used to fix the invalid files.

In case of metadata files, Kitodo.Production used to save internal `meta.xml` with a slightly invalid METS structure. For details see #6714. That issue was resolved by pull request #6715, which fixed the internal logic so that all _future_ metadata files will have a valid METS structure. 

However, since fixing the METS structure of potentially millions of _existing_ processes manually is not feasible, this pull request also adds a new Kitodo-Script called `resaveMetadataFile`, which can be applied to multiple selected processes in the process list. It resaves the `meta.xml` file with the fixed program logic mentioned above, thus fixing the METS structure of existing legacy process and therefore allowing to open them in the metadata editor again.

Validation is also performed on XML data imported via search interfaces from external sources. Here, the schema definition used depends on the metadata format configured in the import configuration for the external source:
Bildschirmfoto 2025-12-11 um 15 07 11

The following external metadata formats are validated:
- `MODS` data is validated against the current mods schema definition `mods-3-8.xsd` 
- `MARC` data is validated against `MARC21slim.xsd` 
- `PICA` data is validated against `pica-xml-v1-0.xsd`
- `EAD` data is validated against `ead.xsd`

These schema definition files have been retrieved from the website of the Library Of Congress and added to the Kitodo.Production repository.

In contrast to the validation of _internal_ files, errors during the validation of _external_ XML content do not prevent the user from working with it, since normally a user does not have any way to fix the source of those XML documents retrieved from external interfaces. 

In these cases, the popup dialog listing the validation errors offers an additional "Continue without validation" button:
Bildschirmfoto 2025-12-11 um 15 33 48

XML Validation of data records retrieved from external sources can now also be deactived completely in the corresponding import configuration:
Bildschirmfoto 2025-12-11 um 15 47 01

When a data record is retrieved via a certain kind of search interface (specifically SRU and OAI), the descriptive metadata (for example MODS) will be embedded in a container XML format, resulting in a combined XML structure incorporating elements from multiple schema definitions. Since this has to be taken into account when performing XML validation of those combined records, schema definitions for SRU (`srw-types.xsd`) and OAI (`OAI-PMH.xsd`) have been added as well. (see #5877 for details)

In addition to the metadata standards listed above, the XML structure of internal documents is also validated against the Kitodo schema definition `kitodo.xsd` (see section above about loading internal metadata files). Because an issue in the current Kitodo schema definition prevents the original `kiodo.xsd` file from being used for schema validation (see discussion #6682 for details), a minimally adjusted copy of this schema definition has been added  to the repository for schema validation.

**Note**: Please make sure that your import XSLT mapping files create only "Kitodo" elements, not - potentially invalid or incomplete - METS structures (which are normally created by the application, unless you are using "Prestructured import", in which case the import mapping files are also responsible for creating the METS container of importerd records). This is because the validation of imported data will detect and mark incomplete or spurious METS elements around the real "kitodo:kitodo" metadata elements possibly created by existing import mapping files.

The functionality for the validation has been implemented in the existing module `Kitodo-Validation`. This module already contained an empty class `FileStructureValidation.java` which now contains the basic methods for the XML validation.

Fixes #6367 
Fixes #5877

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/dataformat/MetsService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/dataformat/MetsService.java
@@ -148,7 +148,7 @@ public class MetsService {
                 try (InputStream xmlValidationStream = new ByteArrayInputStream(outputStream.toByteArray())) {
                     XMLUtils.checkIfXmlIsWellFormed(new String(xmlValidationStream.readAllBytes(), StandardCharsets.UTF_8));
                 } catch (SAXException e) {
-                    logger.error("Malformed XML: {}", e.getMessage());
+                    logger.error("Error loading workpiece. Metadata file '{}' contains malformed XML: {}", uri, e.getMessage());
                     throw e;
                 }
                 if (validateAgainstSchema) {


Kitodo-Validation/src/test/resources/xmlschemadefinitions/mods-3-4.xsd
henning-gerhardt: File has Windows line endings. File should be converted to Unix file endings as all other files too.

Kitodo-API/src/main/java/org/kitodo/api/externaldatamanagement/XmlResponseHandler.java
```diff
@@ -125,7 +122,7 @@ private static LinkedList<SingleHit> extractHits(Document document, DataImport d
         return hits;
     }
 
-    static int extractNumberOfRecords(String content, SearchInterfaceType type) {
+    public static int extractNumberOfRecords(String content, SearchInterfaceType type) {
```
henning-gerhardt: Missing JavaDoc

Kitodo-API/src/main/java/org/kitodo/api/validation/filestructure/FileStructureValidationInterface.java
```diff
@@ -11,23 +11,48 @@
...
+    /**
+     * Validates an XML file at a given uri against a xsd at a given location.
+     *
+     * @param xmlString
+     *            The string with XML content to validate.
+     * @param xsdFileUri
+     *            The location of the schema to validate against.
+     * @return A ValidationResult, with result boolean and resultMessages.
+     */
+    ValidationResult validate(String xmlString, URI xsdFileUri) throws SAXException, IOException;
```
henning-gerhardt: Incomplete JavaDoc for thrown exceptions.

Kitodo-Validation/src/main/java/org/kitodo/validation/filestructure/FileStructureValidationErrorHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class FileStructureValidationErrorHandler implements ErrorHandler {
+
+    private final List<SAXParseException> validationErrors;
+
+    public FileStructureValidationErrorHandler() {
+        this.validationErrors = new ArrayList<>();
+    }
+
+    public List<SAXParseException> getValidationErrors() {
```
henning-gerhardt: Mussing JavaDoc.

Kitodo/src/main/java/org/kitodo/production/helper/XMLUtils.java
```diff
@@ -289,4 +292,26 @@ public static int getNumberOfEADElements(String xmlString, String eadLevel) thro
...
+    public static void checkIfXmlIsWellFormed(String xmlContent) throws IOException, SAXException {
+        SAXParser saxParser;
+        try {
+            SAXParserFactory saxParserFactory = SAXParserFactory.newInstance();
+            saxParserFactory.setValidating(false);
+            saxParserFactory.setNamespaceAware(true);
+
+            saxParser = saxParserFactory.newSAXParser();
+        } catch (ParserConfigurationException | SAXException e) {
+            throw new RuntimeException(e);
```
henning-gerhardt: Is changing this two exception classes into a `RuntimeException` intended?

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -554,9 +556,11 @@ public void executeScript(Task task, boolean automatic) throws DAOException {
      *             if the task cannot be saved
      * @throws IOException
      *             if the task cannot be closed
+     * @throws SAXException
+     *             if the task cannot be closed
      */
     private void finishOrReturnAutomaticTask(Task task, boolean automatic, boolean successful)
-            throws DAOException, IOException {
+            throws DAOException, IOException, SAXException, FileStructureValidationException {
```
henning-gerhardt: Incomplete JavaDoc

Kitodo-API/src/main/java/org/kitodo/api/validation/filestructure/FileStructureValidationInterface.java
solth: done

Kitodo-API/src/main/java/org/kitodo/exceptions/FileStructureValidationException.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * This exception is thrown when an error occurs during the validation process.
+ */
+public class FileStructureValidationException extends Exception {
+
+    private final ValidationResult validationResult;
+    boolean externalDataValidation = false;
+
+    public FileStructureValidationException(String message, ValidationResult validationResult) {
```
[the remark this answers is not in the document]
solth: done

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
solth: done

Kitodo/src/main/resources/kitodo_config.properties
```diff
@@ -44,6 +44,8 @@ directory.rulesets=/usr/local/kitodo/rulesets/
 # The path must be terminated by a directory separator ("/").
 directory.xslt=/usr/local/kitodo/xslt/
 
+directory.xsd=/usr/local/kitodo/schemata
```
henning-gerhardt: Maybe add a small description like the entries above.

Kitodo/src/main/java/org/kitodo/production/services/validation/FileStructureValidationService.java
```diff
@@ -11,14 +11,234 @@
...
+        schemata.add(KITODO_XSD);
+        // only validate against METS schema as well if the record was created using "prestructured import"
+        if (validateMets) {
+            schemata.add(METS_XSD);
+        }
+        ValidationResult validationResult = validateXmlFile(xmlString, schemata);
+        if (Objects.nonNull(validationResult) && !validationResult.getResultMessages().isEmpty()) {
+            logger.info("Validation errors for internal record: {}", validationResult.getResultMessages());
+            if (StringUtils.isBlank(mappingFiles)) {
+                throw new FileStructureValidationException(Helper.getTranslation(
```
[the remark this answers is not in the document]
solth: I am not sure about this - most log messages are currently not translated, but in English, so this keeps in line with our current approach.

henning-gerhardt (changes requested):
First round of technical and behaviour review.

#### Errors:
E1) Running the introduced KitodoScript "resaveMetadataFile" did not create a backup file before the internal meta data file is written. Any change should be saved in a backup so it is possible to revert if there are other issues created.

E2) If you try to open the meta data editor for an invalid process through the task list you got the error displayed but if you click on "cancel" button you get the following up error:

Screenshot_2026-01-06_10-17-50

Selecting text in this case is not possible.

E3) If you try to finish a meta data editor task on an non valid meta data and have the "Validation on exit" option in this task active you only get the error message "Error saving task" without an future information. This error information get logged into the application log but the user did not get any information.

#### Behaviour:

B1) Checking schema validation on import works and can be deactivated. Currently I must do this as we still use the SRU from BSZ which returns not schema valid SRU responses.

B2) Wrong created internal xml through the import XSLT is reported and after the issue is fixed importing data is possible. We added a `mets:mdWrap/mets:xmlData` around every created internal Kitodo data which is not necessary.

B3) Correcting simple issue in the meta data can be corrected with the KitodoScript call "resaveMetadataFile".


#### General open to dos (mostly for me and only for information): 
G1) I must check the behaviour of non-valid meta data if the Kitodo part is invalid how validation works in this case.

G2) I must check how the application reacts on repairs which are not successful through the KitodoScript call if there are more errors or more complex errors - maybe even hard to create such a scenario.

#### Feature:
F1) Through the limit of maximal 10.000 hits per search it is difficult to repair hundred thousands of processes. Maybe there must be add something similar like to update the internal meta information based on projects over the internal task manager. This is not the scope of this pull request and can be added later.

F2) Maybe it is possible to add some kind of report (project based) which reports how many and which processes have not valid internal meta data.

solth:
@henning-gerhardt thanks a lot for your review. I tried to fix all issues you reported. 

> E1) Running the introduced KitodoScript "resaveMetadataFile" did not create a backup file before the internal meta data file is written. Any change should be saved in a backup so it is possible to revert if there are other issues created.

I added a call to `createBackupFile` so that any changes can now be reverted, if necessary.

> E2) If you try to open the meta data editor for an invalid process through the task list you got the error displayed but if you click on "cancel" button you get the following up error:
> Screenshot_2026-01-06_10-17-50
> 
> Selecting text in this case is not possible.

I fixed the invalid parameter, so that canceling the "Validation errors" dialog does not lead to an unhandled exception anymore.

> E3) If you try to finish a meta data editor task on an non valid meta data and have the "Validation on exit" option in this task active you only get the error message "Error saving task" without an future information. This error information get logged into the application log but the user did not get any information.

I refactored the `CurrentTaskForm` to extend the `ValidatableForm` as well, so that validation errors on are now displayed when they occur while trying to close a task with `typeCloseVerify=true`.

henning-gerhardt:
@solth: your last change did fix the mentioned error cases which I found.

I did even check the "General open to dos" and the application behaviour was good.

But I found a new error: if the meta data file is XML invalid (f.e. missing `>` at end of an xml tag like `</kitodo:metadata` instead of correct `</kitodo:metadata>`) than this error is displayed in the UI - which is good - but there is no log entry and you get re-directed to a former page. I switched from the desktop page  in the process list page, searched for an id which has this invalid XML, tried to open this process in the meta data editor, got the error message and after clicking on the only "Ok" button I got re-directed to the desktop page which I did not expect.

solth:
@henning-gerhardt thanks for the hint. The reason for malformed XML not being logged was that this was done by a different function. I added the log message and ensured that the dialog showing the error message leads back to the referring view from which the user originally came.

solth:
> Can the used URI / path to the file be logged too, so it is easier to find the right data file which contain this invalid xml?

Sure, good idea. I adjusted the error message to occur the (relative) path to the metadata file. Does that work for you?

BartChris:
> B2) Wrong created internal xml through the import XSLT is reported and after the issue is fixed importing data is possible. We added a mets:mdWrap/mets:xmlData around every created internal Kitodo data which is not necessary.


Do we have this documented already somewhere?

I also stumbled over problems when doing a metadata reimport in the metadata editor, You have less info if something goes wrong there, i will probably create an issue.

image

PR #6824, opened 2025-12-16
Open linked processes in metadata editor

solth:
This pull request adds a new option to the structure tree context menu in the metadata editor to directly open linked parent or child processes in the metadata editor, either in the current browser tab/window, thus replacing the currently opened process in the editor, or in a new tab/window. 

Bildschirmfoto 2025-12-16 um 14 32 05

Before opening a linked process (either via this menu option or directly via URL) the system checks whether the process is currently "locked", e.g. opened by another user, and in that case shows a corresponding message:

Bildschirmfoto 2025-12-16 um 14 36 25

This is a convenience function meant to make it easier to manage metadata of process hierarchies.


thomaslow (changes requested):
I successfully tested the new feature:
- the metadata editor shows a warning dialog if it is currently edited by another user
- the metadata editor shows a warning dialog if the user opens a process for a project that he is not assigned to
- linked processes can be navigated to via the two new menu options

I am not familiar with the "lock" feature. Is it "normal" that you can cancel the dialog and still edit and save the metadata even though it is currently edited by another user?

An improvement could be to proactively grey-out both menu options (similar to the metadata editor icon in the process list) such that users cannot navigate to the locked metadata editor in the first place.

A minor observation: While testing, if the "user - project" assignment is removed while the user is logged in, the warning message does not appear. It requires a re-login of the user for the warning message to appear. 

An important problem in my opinion is, if any multiple nodes are selected in the logical structure tree, the context menu options to open metadata still shows up (incorrectly). Clicking it opens the metadata editor without an "id" query parameter, causing an error.

Looking over the code, the code looks ok. It would be nice to add minimal Javadoc to private methods as well. Otherwise, they look somewhat "naked" in my eye.

PR #6831, opened 2025-12-19
Optimize Process list by batching

BartChris:
This Pull Request tries to further optimize the process list in Kitodo.Production. The changes address bottlenecks which were identified earlier while preserving existing behavior  (See https://github.com/kitodo/kitodo-production/pull/6649#issuecomment-3224784567)

The linked issue identified the following bottlenecks which all stem from executing SQL logic for each process in the list. (100 times on a list with max size)

| Query (simplified) | Executions | Avg (ms) | Max (ms) | Total (s) | % of DB time |
|---------------------|------------|----------|----------|-----------|--------------|
| `tasks0_ ... WHERE process_id=?` | 100 | 0.45 | 0.95 | 0.045 | 42% |
| `t.processingStatus ... WITH RECURSIVE process_children` | 100 | 0.27 | 0.49 | 0.027 | 25% |
| `process0_.id ... parent_id=?` | 100 | 0.19 | 0.36 | 0.019 | 18% |
| `comments0_ ... JOIN user` | 3 | 0.59 | 0.81 | 0.0018 | 1.7% |
| `batches0_.process_id IN (...)` | 3 | 0.19 | 0.27 | 0.0006 | 0.5% |
| Other queries | ~10 | — | — | ~0.015 | ~13% |

The queries identified there can all be made more efficient by executing them only once for all processes, caching the result and reusing the cached result for the view.

The first optimization extends an idea introduced in https://github.com/kitodo/kitodo-production/pull/5360 (see esp. https://github.com/kitodo/kitodo-production/pull/5360#issuecomment-1270071800). In order to recursively calculate the progress for all processes in the list (including parents) we rely on native SQL queries which are now supported by current versions of MySQL and MariaDB. The changes here go one step further and recursively calculate the progress for all processes in the list at once. 

I think this can even be tested on the H2 database, so i also tried to write a test for that. https://www.h2database.com/html/commands.html#with:

> Can be used to create a recursive or non-recursive query (common table expression). For recursive queries the first select has to be a UNION. One or more common table entries can be referred to by name. Column name declarations are optional - the column names will be inferred from the named select queries. 
Example:

```sql

WITH RECURSIVE cte(n) AS (
        SELECT 1
    UNION ALL
        SELECT n + 1
        FROM cte
        WHERE n < 100
)
SELECT sum(n) FROM cte;

Example 2:
WITH cte1 AS (
        SELECT 1 AS FIRST_COLUMN
), cte2 AS (
        SELECT FIRST_COLUMN+1 AS FIRST_COLUMN FROM cte1
)
SELECT sum(FIRST_COLUMN) FROM cte2;
```

The second optimization is directed at the calculation of the task title of open/in work tasks of a process, which is used in a tooltip in the list. We can use default HQL to retrieve the information for all processes at once and cache it for reuse in the view. The same is true for identifying all processes with children, which can also be done in one batch query.

The same general pattern has also been applied in another PR to optimize the user list (https://github.com/kitodo/kitodo-production/pull/6803): Calculate the values for all processes in the derived `LazyBeanModel` for this view and store them in a HashMap which serves as a cache, which is accessed by the view.  

To asses whether this actually improves on performance maybe @solth or @henning-gerhardt can give it a try.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
@@ -1131,7 +1131,7 @@ public class ProcessForm extends TemplateBaseForm {
      * @return amount of processes
      */
     public Integer getAmount() {
-        return lazyBeanModel == null ? 0 : lazyBeanModel.getRowCount();
+        return Objects.isNull(lazyBeanModel) ? 0 : lazyBeanModel.getRowCount();
     }
 
     /**
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
@@ -31,7 +31,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.SystemUtils;
@@ -48,8 +47,8 @@ import org.kitodo.api.dataformat.Workpiece;
 import org.kitodo.api.dataformat.mets.LinkedMetsResource;
 import org.kitodo.config.ConfigCore;
 import org.kitodo.config.enums.ParameterCore;
-import org.kitodo.data.database.beans.Role;
 import org.kitodo.data.database.beans.Process;
+import org.kitodo.data.database.beans.Role;
 import org.kitodo.data.database.beans.Task;
 import org.kitodo.data.database.beans.User;
 import org.kitodo.data.database.converter.ProcessConverter;
@@ -787,7 +786,7 @@ public class ProcessServiceIT {
     }
 
     private boolean hasVisibleTaskForUser(Process process, Set<Integer> userRoles) {
-        if (process.getTasks() == null) {
+        if (Objects.isNull(process.getTasks())) {
             return false;
         }
         for (Task task : process.getTasks()) {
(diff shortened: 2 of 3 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/ProcessForm.java
```diff
@@ -1145,14 +1130,8 @@ public List<Task> getCurrentTasksForUser(Process process) {
...
-    public String getAmount() throws DAOException {
-        Stopwatch stopwatch = new Stopwatch(this, "getAmount");
-        HashMap<String, String> filterMap = new HashMap<>();
-        if (!StringUtils.isBlank(this.filter)) {
-            filterMap.put(FilterService.FILTER_STRING, this.filter);
-        }
-        return stopwatch.stop(ServiceManager.getProcessService().countResults(filterMap,
-                isShowClosedProcesses(), isShowInactiveProjects()).toString());
+    public Integer getAmount() {
+        return lazyBeanModel == null ? 0 : lazyBeanModel.getRowCount();
```
henning-gerhardt: Instead of `lazyBeanModel == null` is use `Objects.isNull(lazyBeanModel)`.

Kitodo/src/test/java/org/kitodo/production/services/data/ProcessServiceIT.java
```diff
@@ -732,4 +738,70 @@ public void shouldReturnCorrectProcessExportDTOWithCorrectStats() throws Excepti
...
+            return false;
+        }
+        for (Task task : process.getTasks()) {
+            if (!(TaskStatus.OPEN.equals(task.getProcessingStatus())
+                    || TaskStatus.INWORK.equals(task.getProcessingStatus()))) {
+                continue;
+            }
+            for (Integer roleId : task.getRoleIds()) {
+                if (userRoles.contains(roleId)) {
+                        return true;
```
henning-gerhardt: Wrong indent level of return statement.

PR #6834, opened 2025-12-23
[3.9] Optimize Process list by batching

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6831


BartChris:
@solth I can do it. For large databases one of the most important optimizations seems to be having an index on the process table (sorthelperStatus). This requires a database migration so we probably would not include in a 3.9 release.

PR #6847, opened 2026-01-07
Refactor Bean Scopes: Project, Template, Workflow, Docket, Ruleset forms

thomaslow:
This PR refactors all forms on the "projects" page from `@SessionScoped` to `@ViewScoped`. In addition, a few minor cleanup improvements are made.

Related to issues: 
- #6456

Updated forms:
- `ProjectForm`
- `TemplateForm`
- `WorkflowForm`
- `DocketForm`
- `RulesetForm`

Changes:
- Splitting form classes into list view and edit view
- Column selection dialog only shows columns of the currently active tab
  - `tab` - the currently active tab
  - `referrerListOptions` - the list options `sortField`, `sortOrder`, `firstRow` and `filter` encoded in a single parameter such they can be restored when navigating back to the list after editing a single entity

Due to the replacement of `ProjectForm.java`, `TemplateForm.java`, ..., merging with the following PR needs to be done more carefully such that any improvements are not lost:

- #6815
- #6778
- #6408

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/DocketService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/DocketService.java
@@ -12,7 +12,6 @@
 package org.kitodo.production.services.data;
 
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


henning-gerhardt:
Sorry for late asking but why all the `VIEW_PATH` constant are declared as public as they used only in the class itself so private declaration would be better?

Edit: Ah, they are used in corresponding classes as return path information but why the are not declared as normal class fields / members and are accessed by (maybe static) get access method?

thomaslow:
@henning-gerhardt I guess writing it as a getter could be beneficial, since you then could also access view paths from JSF templates, which currently is not done. Instead, view paths are repeated in JSF templates as simple text.

Please keep in mind that I didn't change this code other than rephrasing it a bit. I simply unified the naming of these variables or constants to `VIEW_PATH`, which is - in my opinion - an improvement in readability with respect to the previous code.

There are many things that can be improved further in these view classes. I can't improve everything as part of this PR, which solely focuses on refactoring them from @SessionScoped to @ViewScoped.

Issue #6855, opened 2026-01-19
Schema files with local and external schema references


henning-gerhardt:
**Describe the bug**
With #6815 there are schema files introduced and used on importing data and internal meta data file access. But some of them use references to local schema files and some of them are using references to external URL schema files.

F.e. `ead.xsd`, `mods.xsd` and `xslt20.xsd` are using references to other xsd over an URI path which results that on validating with this schema's the additional external schema files get downloaded. This may be non issue if you only validate a few times but if you do f.e. an mass import than there will be a load of validations and accesses to this external schema files.

**Expected behavior**
The external referenced schema files should be added to the repository and the reference should be changed to this local files.

**Additional context**
`mods.xsd` and `xslt20.xsd` are referencing to similar named xsd file `xml.xsd` but they have different URI's and are not similar and should be stored and references in different file names.

BartChris:
> I. Maybe there is in future a solution for `xslt20.xsd`.

The complexity of having everything local is massively increased by having to bring in the whole `XMLSchema.xsd` which brings a bunch of other dependencies. The question is whether we actually have XSL (mapping) files which need the in-line schema using xsl:import-schema. 

The comment in  `xslt20.xsd` says:
> An XSLT stylesheet may contain an in-line schema within an xsl:import-schema element,
so the Schema for schemas needs to be imported

We could restrict the schema down and conventional XSL validation still seems to work.

Experimental branch:

https://github.com/kitodo/kitodo-production/compare/main...BartChris:kitodo-production:xsl_schema?expand=1

https://github.com/BartChris/kitodo-production/tree/xsl_schema

Downside is that we lose support for stuff like the following, but i am not sure whether this is something which might appear in normal xsl files (i never used it):

```xml
<xsl:import-schema>
  <xs:schema targetNamespace="http://example.org/test">
    <xs:element name="dummy" type="xs:string"/>
  </xs:schema>
</xsl:import-schema>
```

henning-gerhardt:
I don't know if we can drop the support of `xsl:import-schema` support inside the `xslt20.xsd` as I don't know how the xslt files are created or which content they have.

BartChris:
Yes, understandable. I hope the fact that those external URLs seemed blocked for me two weeks ago was just a rare event. Because at that point i was not able to import anything into Kitodo.

henning-gerhardt:
This blocking is a security mechanism by `loc.gov` as I have rarely this blocking on my local development system too.

[...]

PR #6858, opened 2026-01-20
Use local references xsd schema files

henning-gerhardt:
Partial fix for #6855

Replace URI schema references with local schema files for `ead.xsd` and `mods-3-8.xsd`.

There are still URI schema references in `xslt20.xsd`as replacing them is difficult.


solth (approved):
Tested, works well, thanks @henning-gerhardt !

PR #6860, opened 2026-01-22
Move linking child processes to a separate dialog in metadata editor

solth:
I have spoken to quite a few Kitodo users over the years who were not aware of the option to link subordinate processes via the "Neues Strukturelement" dialog in the metadata editor. 

That dialog can be opened using the context menu option "Element hinzufügen" of the structure tree in the metadata editor of Kitodo:

Bildschirmfoto 2026-01-22 um 20 30 35

The reason for the option to link processes being so little known among users is that the corresponding functionality is well hidden behind a inconspicuous little checkbox in the dialog that many users don't notice:

Bildschirmfoto 2026-01-22 um 20 22 06

Activating this checkbox completely replaces the content of the dialog with a search mask to link a subordinate process instead of adding a new structural element:

Bildschirmfoto 2026-01-22 um 20 23 33

Apart from effectively hiding this option, the whole dialog with conditionally rendered components and buttons makes for very awkward template code.

To improve this situation I separated the two functionalities into separate dialogs that can now be opened directly via individual context menu options.

The first option, "Element hinzufügen":

Bildschirmfoto 2026-01-22 um 20 36 14

leads to the dedicated dialog for adding structural elements:

Bildschirmfoto 2026-01-22 um 20 36 03

and the second option, "Untergeordneten Vorgang verknüpfen":

Bildschirmfoto 2026-01-22 um 20 26 44

leads to the dedicated dialog for linking processes:

Bildschirmfoto 2026-01-22 um 20 36 23

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/dataeditor/DataEditorService.java
@@ -636,7 +636,7 @@ public class DataEditorService {
                             dataEditor.getAcquisitionStage(), locales);
                     Map<String, String> allowedSubstructures = structureElement.getAllowedSubstructuralElements();
                     allowedSubstructures.keySet().retainAll(documentTypes.keySet());
-                    return (!(structureTreeNode.isLinked() || allowedSubstructures.isEmpty()));
+                    return !structureTreeNode.isLinked() && !allowedSubstructures.isEmpty();
                 }
             }
         }


BartChris:
I tracked your changes and so far i cannot see any difference in behaviour, all the refactorings are good and make the code better.

One functional thing: Right now it seems, that it is not checked whether the selected doctype can actually have child processes. The button is always displayed although e.g. "volumes" in my test system cannot  have child processes. Should we also check whether the selected doctyp can have children?

image

solth:
@BartChris I added a check for document types that are allowed as child structures for the currently selected tree node. I no such document type is allowed as child structure, the context menu option to link subordinate processes is not displayed.

Due to the complexity of this check I moved it from `DataEditorForm` to `DataEditorService`. I am unsure if the condition checks for the _other_ context menu items (like `isAddingElementPossible`) should be moved to the service class as well - for uniformity - because they are less complex. Thus, creating corresponding methods in the service class does not seem mandatory to me, in this case, but if you disagree I can adjust that as well.

BartChris (approved):
I am fine with your changes and only added one suggestion for simplification; i had hoped that we could do this useful check for substructures with less code, or could reuse existing methods from ProcessService (`canCreateChildProcess()`) but underestimated the complexity.

PR #6869, opened 2026-01-28
Paginate every second page

### LINKED ISSUE #6457: Sheetpagination with empty page
In Kitodo you can select "Blattzählung" as pagination mode, which leads to the the following numbering of pages with Arabic numbers.
`1 1 2 2 3 3 4 4`

What we would like to have is a "Blattzählung mit Leerseite" which would generate the following pagination
`1 UNCOUNTED 2 UNCOUNTED 3 UNCOUNTED 4 UNCOUNTED`

The selection menu should be extended with the option "Blattzählung mit Leerseite" or "Sheet Counting with empty page"

Image

Image

oliver-stoehr:
This pull request adds a new option to the pagination mode menu.
"Sheet counting with empty page"/"Blattzählung mit Leerseite" allows a pagination with a counter increased for every second page while alternating between the counter and ~"UNCOUNTED"~ "uncounted".
This produces a pagination like this:
~`1, UNCOUNTED, 2, UNCOUNTED, 3, UNCOUNTED, …`~
`1, uncounted, 2, uncounted, 3, uncounted, …`

This mode is also compatible with:
roman (`I, uncounted, II, uncounted, …`) and
alphabetic (`a, uncounted, b, uncounted, …`) pagination.

Technical base for this new mode is a String like this:
`¿uncounted¡1½` 
It can also be used in the advanced pagination (The `1` can be replaced with a different start value, also using a start value for roman or alphabetic pagination, like `I` or `´a´`).

To be able to output the counter conditionally on every second page, a code adjustment was necessary.
**This adjustment has one constraint for advanced pagination:**
The characters following `¿` and `¡` are no longer interpreted as free text (implicitly!).
For example, using a string like `¿v` to output `v` for every second page now requires (explicit) escaping to prevent an interpretation as roman numeral: `` ¿`v` ``.
(That means configurations like recto verso pagination change, e.g. from `1°¡r¿v½` to ``1°¡r¿`v`½``.)

_This only affects the advanced pagination. Additional escaping is only necessary for characters that should be free text, but could be interpreted as roman numerals or decimals (`I, i, V, v, X, x, 1, 2, 3, …`)._
<details>
<summary><strong>More details on why this is necessary</strong></summary>

Generally, it is necessary to escape free text with back ticks (`` ` ``) to prevent unwanted evaluation.
This was not necessary for text following `¿` or `¡` (e.g. in `1°¡r¿v½` for recto verso pagination = `1r, 1v, 2r, 2v, …`), as text following `¿` and `¡` was implicitly interpreted as free text.

To output the counter on every second page, a String like `¡1` would be necessary, but  did not work in this combination, as `¡` prevented the interpretation of `1` as counter).

This implicit interpretation as free text is removed in this PR, as it prevented any conditional output of a counter.
This change is only relevant to users of the advanced pagination, as all other counting modes like "Recto Verso Pagination" have already been adjusted accordingly in this PR. 

</details>

Resolves #6457.

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/production/helper/metadata/pagination/PaginatorTypeTest.java
+++ b/Kitodo/src/test/java/org/kitodo/production/helper/metadata/pagination/PaginatorTypeTest.java
@@ -274,7 +274,7 @@ public class PaginatorTypeTest {
     }
 
     @Test
-    public void testFreetextFormatPagesFictitious) {
+    public void testFreetextFormatPagesFictitious() {
         assertEquals("[`Hello world!`]",
             PaginatorType.FREETEXT.format(PaginatorMode.PAGES, HELLO_WORLD_STRING, true, UNUSED_STRING));
     }


oliver-stoehr:
@andre-hohmann would you mind taking a look at my description of this change?
The change has one constraint for the advanced pagination as described above. I would like to get your opinion on whether this constraint might be critical for users of the advanced pagination.

solth:
@kitodo/kitodo-community-board please take a look at the changes in this pull request!

andre-hohmann:
In the Community Board is also little experience with the specific functionality of the extended pagination. Therefore, we cannot offer a definitive assessment. 

In our view, the change is possible despite the constraint of escaping, but the necessary escaping process should be described in the [documentation](https://github.com/kitodo/kitodo-production/wiki/Paginierung).

PR #6877, opened 2026-02-05
Fix some deprecated methods

henning-gerhardt:
There are a lot of method calls which a marked as deprecated. This pull request fix some of this deprecated method calls.


solth:
@BartChris would you mind taking a look at this pull request?

PR #6880, opened 2026-02-05
Sequenced Collections

solth:
This pull request replaces usages of `get(0)` to access the first element of a list with the "Sequenced Collection" method `getFirst()` introduced in Java 21. This method - and its counterpart `getLast()` - aims to provide a unified way for all collection types in Java to access the first and last element and is recommended to be used from Java 21 onwards (see https://www.baeldung.com/java-21-sequenced-collections for details)


solth:
Alerts by automatic GitHub security scanning about other deprecated methods in the changed lines like `getByQuery` and `getLogicalDocStruct` have to be handled separately.

PR #6882, opened 2026-02-09
Replace deprecated URL constructor

solth:
Replacing `URL.URL` constructor, which has been deprecated in Java 21, with
- `Paths.get(String).toUrl()` for local files
- `new URI(String).toURL()` for online resources

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
@@ -13,6 +13,7 @@ package org.kitodo.api.schemaconverter;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -46,7 +47,7 @@ public enum MetadataFormatConversion {
     MetadataFormatConversion(String filename, String source, MetadataFormat targetFormat) {
         this.fileName = filename;
         try {
-            this.source = new URL(source);
+            this.source = Paths.get("xslt/" + source).toUri().toURL();
         } catch (MalformedURLException e) {
             this.source = null;
         }
--- a/Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
+++ b/Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
@@ -12,8 +12,9 @@
 package org.kitodo.api.schemaconverter;
 
 import java.net.MalformedURLException;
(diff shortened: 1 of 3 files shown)


Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
```diff
@@ -46,7 +47,7 @@ public enum MetadataFormatConversion {
     MetadataFormatConversion(String filename, String source, MetadataFormat targetFormat) {
         this.fileName = filename;
         try {
-            this.source = new URL(source);
+            this.source = Paths.get("xslt/" + source).toUri().toURL();
```
henning-gerhardt: Why did you add the `xslt/` as hard coded path prefix?

henning-gerhardt (commented):
Changes looks good in general except the `xslt/` hard coded path information.

henning-gerhardt (approved):

Kitodo-API/src/main/java/org/kitodo/api/schemaconverter/MetadataFormatConversion.java
```diff
@@ -46,8 +48,8 @@ public enum MetadataFormatConversion {
     MetadataFormatConversion(String filename, String source, MetadataFormat targetFormat) {
         this.fileName = filename;
         try {
-            this.source = new URL(source);
-        } catch (MalformedURLException e) {
+            this.source = new URI(source).toURL();
+        } catch (NullPointerException | MalformedURLException | URISyntaxException e) {
```
stweil, line 52: @solth, this change is the reason for two new Coverity warnings:

503913 Explicit null dereferenced

A null pointer exception will occur.

In org.​kitodo.​api.​schemaconverter.​MetadataFormatConversion.​MetadataFormatConversion(): Dereference of an explicit null value ([CWE-476](http://cwe.mitre.org/data/definitions/476.html))


503915 DCN: Don't Catch NullPointer Exception

In org.​kitodo.​api.​schemaconverter.​MetadataFormatConversion.​MetadataFormatConversion(java.​lang.​String, int, java.​lang.​String, java.​lang.​String, org.​kitodo.​api.​schemaconverter.​MetadataFormat): According to SEI CERT rule ERR08-J NullPointerException should not be caught. ([From SpotBugs description](https://scan4.scan.coverity.com/doc/en/spotbugs/sb_checker_ref.html#FB.DCN_NULLPOINTER_EXCEPTION))

stweil: I think the correct solution should either avoid the calls which pass `null` or add explicit handling of a `null` parameter instead of catching the exception which it causes. I suggest to add explicit handling of `null`. Do you want to fix it yourself? Or should I provide a pull request?

PR #6886, opened 2026-02-13
Optimize desktop view

BartChris:
This Pull Requests optimizes the Desktop View by resusing the methods developed for https://github.com/kitodo/kitodo-production/pull/6831.

Both process list as well as desktop view show processes and therefor need the same data to be provided; especially for calculating the progress.

I introduced a new `ProcessProgressHelper` which provides calculation of progress process for both the ProcessList as well as the Desktop View in order to not having to duplicate methods.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
@@ -22,6 +22,7 @@ import java.util.Set;
 
 import javax.json.JsonException;
 
+import jakarta.annotation.PostConstruct;
 import jakarta.faces.view.ViewScoped;
 import jakarta.inject.Named;
 
@@ -66,6 +67,26 @@ public class DesktopForm extends BaseForm {
         super();
     }
 
+    @PostConstruct
+    public void init() {
+        loadProcesses();
+    }
+
+    /**
+     * Loads processes and initializes related caches.
+     */
+    private void loadProcesses() {
+        try {
+            if (ServiceManager.getSecurityAccessService().hasAuthorityToViewProcessList()) {
+                processList = ServiceManager.getProcessService().loadData(0, 10, SORT_ID, SortOrder.DESCENDING, null);
+                preloadProcessCaches();
+            }
+        } catch (DAOException | JsonException e) {
+            Helper.setErrorMessage(ERROR_LOADING_MANY,
+                    new Object[]{ObjectType.PROCESS.getTranslationPlural()}, logger, e);
+        }
+    }
+
     /**
      * Get values of ObjectType enum.
      *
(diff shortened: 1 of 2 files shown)


Kitodo/src/main/webapp/WEB-INF/templates/includes/processes/progressColumn.xhtml
```diff
@@ -16,29 +16,30 @@
         xmlns:h="http://xmlns.jcp.org/jsf/html"
         xmlns:ui="http://xmlns.jcp.org/jsf/facelets">
     <!--@elvariable id="process" type="org.kitodo.production.dto.ProcessDTO"-->
+    <!--@elvariable id="progressBean" type="org.kitodo.production.forms.ProcessForm"-->
     <h:panelGroup
             class="process-progress"
-            title="#{ProcessForm.getCurrentTaskTitles(process)}"
-            rendered="#{ProcessForm.hasAnyTasks(process) or ProcessForm.hasChildren(process)}"
+            title="#{progressBean.getCurrentTaskTitles(process)}"
```
BartChris: The widget is used by both the `ProcessForm` as well as the `DesktopView`, so i provide the bean context dependent.

Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
```diff
@@ -47,6 +54,10 @@ public class DesktopForm extends BaseForm {
     private List<Task> taskList = new ArrayList<>();
     private List<Process> processList = new ArrayList<>();
     private List<Project> projectList = new ArrayList<>();
+    private Map<Integer, EnumMap<TaskStatus, Integer>> taskStatusCache = new HashMap<>();
+    private Map<Integer, Map<TaskStatus, List<String>>> taskTitleCache = new HashMap<>();
+    private Set<Integer> processesWithChildren = new HashSet<>();
+    private final ProcessProgressHelper progressService = new ProcessProgressHelper();
```
thomaslow: Personally, I would prefer if the variable is not called `progressService`, which can be confusing because Kitodo has many "Service"-classes like "ProcessService", "TaskService", etc. that correspond to beans, and "Progress" does not exist as a bean. Maybe `progressHelper` would be better here?

Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
```diff
@@ -100,6 +111,7 @@ public List<Process> getProcesses() {
         try {
             if (ServiceManager.getSecurityAccessService().hasAuthorityToViewProcessList() && processList.isEmpty()) {
                 processList = ServiceManager.getProcessService().loadData(0, 10, SORT_ID, SortOrder.DESCENDING, null);
+                preloadProcessCaches();
```
thomaslow: I am not sure how it works here exactly, but `getProcesses` should ideally be a simple getter method that should not involve triggering database calls. There is no guarantee from JSF that a getter is only called once when rendering a page, even if it is only used once in a JSF template.

I realize that the current implementation does not consider this either, but it might be a good idea to not add additional "initialization" code (that is supposed to run only once per page view) to the getter.

PR #6887, opened 2026-02-16
Replace Java assert keyword usage with Junit assert method calling

henning-gerhardt:
Replace the Java `assert` keyword usage with calling Junit `assertXXX(...)` methods. May conflict with some changes in #6880 . IntelliJ IDEA notices some more improvements / simplifier code usage but this would I change in a separate change.


thomaslow (approved):
Thank you. I checked that there were no accidental copy/paste mistakes. All tests pass. Looks good!

PR #6889, opened 2026-02-19
Update Junit and Mockito to latest 5.x version

henning-gerhardt:
Updating Junit and Mockito to latest 5.x versions as Mockito currently only support Junit version 5.x.

Changes in the executing of sunfire and failsave Maven plugins are needed as the Mockito integration is recommend now as an additional Java agent. If not you get the message

```
Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
```

I used https://rieckpil.de/how-to-configure-mockito-agent-for-java-21-without-warning/ for the way of integrating Mockito as a Java agent (even recommended by Mockito itself but a little bit less documented how to do this)

### DIFF
--- a/Kitodo-API/pom.xml
+++ b/Kitodo-API/pom.xml
@@ -80,7 +80,7 @@
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-engine</artifactId>
+            <artifactId>junit-jupiter</artifactId>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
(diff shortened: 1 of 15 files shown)


thomaslow (approved):
Hmm. You are right. I checked again. Warning messages about bytebuddy disappear with your PR. I must have mixed up log files of the current main and your PR, sorry.

Then I'll withdraw my suggestions. Have a nice weekend!

PR #6898, opened 2026-02-26
[3.8] Fix reimport for groups with empty fields

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6495


BartChris:
I refactored the logic for Java 11, because it got rather ugly.

oliver-stoehr (approved):

PR #6905, opened 2026-03-02
Add CSV-Export

BartChris:
This pull requests adds a CSV export to Kitodo production to allow for faster exports of the (filtered) process list.

While adding the CSV export i encountered that the current implementation could be improved. Right now the PDF export relies on the generation of an Excel workbook, which is unnecessary. I also tried to seperate data retrieval from export generation. The `SearchResultGeneration` class now receives the data and is responsible for generating the output without having to call into `ProcessService` again.


Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -1207,19 +1178,28 @@ public void generateResultAsPdf(String filter, boolean showClosedProcesses, bool
...
-            ExternalContext response = prepareHeaderInformation(facesContext, "search.xlsx");
+            List<ProcessExportDTO> results =
+                    getProcessesForExport(
+                            filter,
+                            showClosedProcesses,
+                            showInactiveProjects,
+                            ServiceManager.getUserService().getSessionClientId()
+                    );
+            ExternalContext response = prepareHeaderInformation(facesContext, format.getFilename());
+            facesContext.responseComplete();
```
BartChris: Calling this later leads to an exception when exporting CSV and Excel in my case:

```
11-Mar-2026 18:47:46.175 SCHWERWIEGEND [http-nio-8080-exec-26] org.omnifaces.filter.FacesExceptionFilter.logException [45d36942-9d84-4dc2-a2ad-377951f0b9b5][127.0.0.1] FacesExceptionFilter: An exception occurred during processing servlet request. Error page '/pages/error.jsf' will be shown. java.lang.IllegalStateException: Cannot change buffer size after data has been written at org.apache.catalina.connector.ResponseFacade.checkCommitted(ResponseFacade.java:515) at org.apache.catalina.connector.ResponseFacade.setBufferSize(ResponseFacade.java:224) at jakarta.servlet.ServletResponseWrapper.setBufferSize(ServletResponseWrapper.java:152) at
```

PR #6906, opened 2026-03-03
Refactor Bean Scopes: Task List

thomaslow:
This PR refactors the CurrentTaskForm from `@SessionScoped` to `@ViewScoped` and splits it up into various smaller classes. In addition, a few minor cleanup improvements are made.

Related to issues: 
- #6456

Includes PR:
- #6903 

Updated forms:
- `CurrentTaskForm`

Changes:
- Splitting form class into list view and edit view
  - `taskStatus` - which tasks to show based on status
  - `taskFilter` - additional task-specific filter options
  - `referrerListOptions` - the list options encoded in a single parameter such they can be restored when navigating back to the list after editing a single entity
- When navigating to unrelated pages, the filter that was last used is stored in a small @SessionScoped bean `TaskListSessionState` and recovered if no other filter was provided as URL parameter


Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+    private static final String VIEW_PATH = MessageFormat.format(REDIRECT_PATH, "tasks");
     
+    private final WorkflowControllerService workflowControllerService = new WorkflowControllerService();
+
+    private List<Task> selectedTasks = new ArrayList<>();
+    private final WebDav myDav = new WebDav();
+    
+    private List<String> taskFilters;
+    private List<String> selectedTaskFilters;
+    private FilterMenu filterMenu = new FilterMenu(this);
```
solth:
```suggestion
    private final FilterMenu filterMenu = new FilterMenu(this);
```

Kitodo/src/main/java/org/kitodo/production/forms/BaseEditView.java
```diff
@@ -20,6 +27,8 @@
  */
 public class BaseEditView extends BaseForm {
 
+    private static final Logger logger = LogManager.getLogger(BaseEditView.class);
```
solth: Did you use this logger during development and then removed it's usages again? If it's not required anymore it should probably be remove as well (alongside the corresponding imports)

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskBatchEditView.java
```diff
@@ -0,0 +1,100 @@
...
+import org.kitodo.production.enums.ObjectType;
+import org.kitodo.production.forms.BaseForm;
+import org.kitodo.production.helper.Helper;
+import org.kitodo.production.helper.batch.BatchTaskHelper;
+import org.kitodo.production.services.ServiceManager;
+import org.kitodo.utils.Stopwatch;
+
+public class TaskBatchEditView extends BaseForm {
+
+    private static final Logger logger = LogManager.getLogger(TaskEditView.class);
```
solth:
```suggestion
    private static final Logger logger = LogManager.getLogger(TaskBatchEditView.class);
```

thomaslow:
@henning-gerhardt This PR also updates the ActiveMQ code in "FinalizeStepProcessor", because it depended on the refactored UI bean "CurrentTaskForm". There does not seem to be any automated tests for this processor. I currently do not have an ActiveMQ setup. Would this be easy to test for you?

henning-gerhardt:
@thomaslow : I can look into this.

henning-gerhardt:
@thomaslow : I tried to close a task through ActiveMQ and it was successful. As this close message has no `properties` nor `message` field I can not say that changes in the handling of this two fields is still working or not.

thomaslow:
@henning-gerhardt Thank you for testing the ActiveMQ command!

Kitodo/src/main/webapp/pages/taskBatchEdit.xhtml
```diff
@@ -14,7 +14,13 @@
 <ui:composition
         template="/WEB-INF/templates/baseListView.xhtml"
         xmlns:p="http://primefaces.org/ui"
-        xmlns:ui="http://xmlns.jcp.org/jsf/facelets">
+        xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
+        xmlns:f="http://xmlns.jcp.org/jsf/core">
+
+    <f:metadata>
+        <f:viewParam name="id"/>
+        <f:viewAction action="#{TaskBatchEditView.loadFromTemplate(id)}" />
```
solth, line 22: `TaskBatchEditView` is not declared as a named CDI bean and thus cannot be called as backing bean in this template. So you either need to add annotations for scope - probably `@ViewScoped` - and optionally `@Named` to `TaskBatchEditView` **or** move the `loadFromTemplate` method to another class which is already used as a bean and call it from there.

Independent of this problem I think the whole `TaskBatch` topic has to be discussed. The way I see it the corresponding fuctionality has been replaced by list functions and addtional Kitodo scripts in Kitodo 3, so this template and the corresponding class/bean can be removed, but I am not totally sure about this. Perhaps @andre-hohmann, @henning-gerhardt and/or @matthias-ronge can comment on this?

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskBatchEditView.java
solth, line 1: Since this class isn't declared as a `ManagedBean` or `ViewScoped`, it's methods cannot be called from the corresponding templates, hence all its methods are marked as "unused" by my IDE. 

 _If_ we still need this class, the missing annotations should be added. Otherwise the class and the corresponding template could probably be removed.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+    }
+
+    /**
+     * Encodes the filter describing which tasks of a certain task status is supposed to be shown in the task list.
+     * 
+     * @param taskStatus the list of task status
+     * @return encoded query parameter, e.g. `OPEN+INWORK`
+     */
+    private String encodeTaskStatusAsQueryParameter(List<TaskStatus> taskStatus) {
+        return taskStatus.stream().map(s -> s.toString()).collect(Collectors.joining("+"));
```
solth:
```suggestion
        return taskStatus.stream().map(Enum::toString).collect(Collectors.joining("+"));
```

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+    }
+
+    /**
+     * Encodes additional task-specific filter options as query parameter.
+     * 
+     * @param taskFilter the list of task-specific filter options
+     * @return encoded query parameter, e.g. `automaticTasks+correctionTasks`
+     */
+    private String encodeTaskFilterAsQueryParameter(List<String> taskFilter) {
+        return taskFilter.stream().collect(Collectors.joining("+"));
```
solth:
```suggestion
        return String.join("+", taskFilter);
```

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+     * Parses the query parameter "taskFilter" to a list of filter options. Ignores any unrelated values.
+     * 
+     * @param encodedTaskFilter the taskFilter query parameter, e.g. `automaticTasks+correctionTasks`
+     * @return the parsed list of filter options
+     */
+    private List<String> parseTaskFilterFromQueryParameter(String encodedTaskFilter) {
+        Set<String> allowed = Set.of(AUTOMATIC_TASKS_FILTER, CORRECTION_TASKS_FILTER, OTHER_USERS_TASKS_FILTER);
+        return Stream.of(encodedTaskFilter.split("\\+"))
+            .map(String::trim)
+            .filter((s) -> allowed.contains(s))
```
solth:
```suggestion
            .filter(allowed::contains)
```

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskWorkView.java
```diff
@@ -0,0 +1,364 @@
...
+            return stopwatch.stop(this.stayOnCurrentPage);
+        }
+        return stopwatch.stop(getReferrerViewPath());
+    }
+
+    /**
+     * Remember referring view such that user can be forwarded correctly once task is closed.
+     */
+    public void setReferrerFromTemplate(String referrer) {
+        if (Objects.nonNull(referrer) && referrer.equals("desktop")) {
```
solth:
```suggestion
        if ("desktop".equals(referrer)) {
```
Could this be simplified?

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
```diff
@@ -78,20 +78,20 @@ public class TaskService extends BaseBeanService<Task, TaskDAO> {
     static {
         SORT_FIELD_MAPPING = new HashMap<>();
         SORT_FIELD_MAPPING.put("title", "title");
-        SORT_FIELD_MAPPING.put("title.keyword", "title");
-        SORT_FIELD_MAPPING.put("processForTask.id", "process.id");
-        SORT_FIELD_MAPPING.put("processForTask.title.keyword", "process.title");
+        SORT_FIELD_MAPPING.put("title", "title");
```
solth: Is this addition required? The previous 80 line already contains the same key/value pair ("title", "title").

Kitodo/src/main/webapp/WEB-INF/templates/includes/taskBatchEdit/taskEditActions.xhtml
```diff
@@ -18,9 +18,9 @@
         xmlns:p="http://primefaces.org/ui"
         xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
 >
-    <h:panelGroup rendered="#{CurrentTaskForm.batchHelper.currentStep.processingUser.id == LoginForm.loggedUser.id}">
+    <h:panelGroup rendered="#{TaskBatchEditView.batchHelper.currentStep.processingUser.id == LoginForm.loggedUser.id}">
```
solth, line 21: I think this (and the following, changed lines in this file) cannot be resolved, because `TaskBatchEditView` is not annotated as a `Named` and scoped bean. I think this just hasn't been noticed because this template (`taskEditActions.xhtml`) is not really reachable in the application anymore, is it? (see my comment(s) above)

Kitodo/src/main/webapp/WEB-INF/templates/includes/taskBatchEdit/taskEditDetails.xhtml
```diff
@@ -30,8 +30,8 @@
                         <tr>
                             <td class="eingabeBoxen_row2">
                                 <h:outputText value="#{msgs.showDataForProcess}: "/>
-                                <h:selectOneMenu value="#{CurrentTaskForm.batchHelper.processName}">
-                                    <f:selectItems var="process" value="#{CurrentTaskForm.batchHelper.processNameList}"
+                                <h:selectOneMenu value="#{TaskBatchEditView.batchHelper.processName}">
```
solth, line 33: See above, the same applies to this file.

Kitodo/src/test/java/org/kitodo/production/forms/TaskListViewIT.java
```diff
@@ -26,12 +26,13 @@
...
 import org.kitodo.data.database.exceptions.DAOException;
+import org.kitodo.production.forms.task.TaskListView;
 import org.kitodo.production.services.ServiceManager;
 import org.kitodo.production.services.data.TaskService;
 
-public class CurrenTaskFormIT {
+public class TaskListViewIT {
 
-    private CurrentTaskForm currentTaskForm = new CurrentTaskForm();
+    private TaskListView taskListView = new TaskListView();
```
solth:
```suggestion
    private final TaskListView taskListView = new TaskListView();
```

Kitodo/src/main/webapp/pages/taskBatchEdit.xhtml
henning-gerhardt: At SLUB we had never used the batch task feature. The only "batch" feature which we used was in 2.x the batches of processes to export newspaper like processes at once.

BartChris: The `TasKBatch`-feature sounds interesting to me. I have never used it, but i think that what many people do in the wild is setting the status of many processes "up" to implicitely bulk-trigger automatic tasks. This has many downsides. A mechanism which would allow to batch trigger task and afterwards close the step of each task would be a better pattern in my opinion.
I do not know, what the `BatchTasks` provide in detail but i have the impression there is a gap in Kitodo functionality wise. This needs a more in depth study. I am not really sure whether it helps or not to keep the broken feature, but i would maybe keep it as a reminder, if it does not do any harm.

BartChris: Related: https://github.com/kitodo/kitodo-production/discussions/6177 (and more general but related to providin more controlled task execution: https://github.com/kitodo/kitodo-production/issues/6460)

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/BaseEditView.java
thomaslow: Removed

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskBatchEditView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskBatchEditView.java
thomaslow: I have added a @ViewScoped and @Named annotations.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskWorkView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java
thomaslow: Removed duplicate

Kitodo/src/main/webapp/WEB-INF/templates/includes/taskBatchEdit/taskEditActions.xhtml
thomaslow: Unfortuantely, my IDE doesn't show these problems, and I can't test this because it is broken in main. With the new @Named property, hopefully the warning message is gone.

Kitodo/src/test/java/org/kitodo/production/forms/TaskListViewIT.java
thomaslow: Done

thomaslow:
@solth Thank you for your feedback. I implemented your proposed changes, which all make sense to me.

Kitodo/src/main/webapp/pages/taskBatchEdit.xhtml
solth: @thomaslow sorry, I totally overlooked the issue #6917 you opened that already discusses this topic.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskBatchEditView.java
```diff
@@ -0,0 +1,105 @@
...
+     * 
+     * @param id the id of the batch task
+     */
+    public void loadFromTemplate(int id) {
+        try {
+            if (id != 0) {
+                Task batchTask = ServiceManager.getTaskService().getById(id);
+                List<Batch> batches = batchTask.getProcess().getBatches();
+                if (batches.size() == 1) {
+                    Integer batchId = batches.get(0).getId();
```
solth:
```suggestion
                    Integer batchId = batches.getFirst().getId();
```

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+            if (Objects.isNull(task.getProcessingBegin())) {
+                task.setProcessingBegin(new Date());
+            }
+
+            if (task.isTypeImagesRead() || task.isTypeImagesWrite()) {
+                try {
+                    URI imagesOrigDirectory = ServiceManager.getProcessService().getImagesOriginDirectory(false,
+                        task.getProcess());
+                    if (!ServiceManager.getFileService().fileExist(imagesOrigDirectory)) {
+                        Helper.setErrorMessage("errorDirectoryNotFound", new Object[] {imagesOrigDirectory });
```
solth:
```suggestion
                        Helper.setErrorMessage("errorDirectoryNotFound", imagesOrigDirectory);
```

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+            return stopwatch.stop(this.stayOnCurrentPage);
+        }
+    }
+
+    /**
+     * Get list of selected Tasks.
+     *
+     * @return List of selected Tasks
+     */
+    public List<Task> getSelectedTasks() {
```
solth: This method seems to be unused. Is it sill required?

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+        return stopwatch.stop(this.selectedTasks);
+    }
+
+    /**
+     * Set selected tasks: Set tasks in old list to false and set new list to true.
+     *
+     * @param selectedTasks
+     *            provided by data table
+     */
+    public void setSelectedTasks(List<Task> selectedTasks) {
```
solth: This method seems to be unused. Is it sill required?

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+        ((LazyTaskModel)this.lazyBeanModel).setShowAutomaticTasks(showAutomaticTasks);
+        stopwatch.stop();
+    }
+
+    /**
+     * Check if it should hide correction tasks.
+     *
+     * @return boolean
+     */
+    public boolean isHideCorrectionTasks() {
```
solth: This method seems to be unused. Is it sill required?

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -11,14 +11,711 @@
...
+        ((LazyTaskModel)this.lazyBeanModel).setHideCorrectionTasks(hideCorrectionTasks);
+        stopwatch.stop();
+    }
+
+    /**
+     * Get taskListPath.
+     *
+     * @return value of taskListPath
+     */
+    public String getTaskListPath() {
```
solth: This method seems to be unused. Is it sill required?

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskBatchEditView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Done

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Seems to be unused for many years now. Is removed now.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Same.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Replaced by iterating over task status constraints list a few months back. Is removed now.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Not needed any more. Is removed now.

solth:
@thomaslow if I am not mistaken, this should now cover all bean refactorings for which the scope was to be corrected, or do you have any further pull requests in the pipeline for this topic? Otherwise I think we could probably close #6456 now. (the only remaining `@SessionScoped` bean that _could_ possibly be changed would be the `CommentForm`, I guess? But I think that form doesn't work exactly the same as the other forms, so I am not completely sure right now; what do you think?)

thomaslow:
@solth Yes, although not strictly mentioned in #6456, I think the CommentForm is a candidate, too. I can't think of a good reason why it would be session scoped. 

Let's see whether I have to remove/redesign the query parameter navigation again. Depending on that, I'll look at the CommentForm. However, the CommentForm is a minor thing and should not influence the 4.0 release.

Issue #6910, opened 2026-03-06
Pagination crashes editor (K.P 3.9.1)


anspallek:
**Describe the bug**
If I choose a image in the pagination field as the first step, the editor crashes.
If I start assigning the the structure data as the first step and assign the pagination as the second step, the editor works properly and the data is saved.

**To Reproduce**
1. Go to Pagination in the editor
2. Click on a image 
3. Editor crashes
4. See error message 

**Expected behavior**
Editor doesn't crash, if I choose a image as the first step

**Screenshots**

<img width="243" height="168" alt="Image" src="https://github.com/user-attachments/assets/ecfc2c09-144e-4a76-a4fe-3227d61afeba" />

<img width="1213" height="759" alt="Image" src="https://github.com/user-attachments/assets/da6050ab-a67b-4927-aa61-f949db17a9d0" />

**Release**
K.P 3.9.1 

**Desktop (please complete the following information):**
 - MacOS and Chrome 145
 - MacOS and  Firefox 148
 - Windows and Firefox

<details><summary>Bug report</summary>

<p>
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.base/java.util.LinkedList.checkElementIndex(LinkedList.java:559)
	at java.base/java.util.LinkedList.get(LinkedList.java:480)
	at org.kitodo.production.forms.dataeditor.PaginationPanel.lambda$setPaginationSelectionSelectedItems$0(PaginationPanel.java:120)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
[...]

PR #6912, opened 2026-03-06
 [3.9] Prevent crash on pagination of unstructured doc

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6910


solth (approved):
Works and code looks good. Since the current `main` branch does not have a page selection in the pagination panel anymore, a corresponding port is not required.

PR #6930, opened 2026-03-16
Fix media cannot be deleted in metadata editor

### LINKED ISSUE #6684: Media cannot be removed via context menu in metadata editor
**Describe the bug**
Deleting media via the context menu option of the structure tree and gallery of the metadata editor does not remove the media anymore.

**To Reproduce**
1. Open a process with media in the metadata editor
2. Open the context menu for a node in the structure tree representing a media element (e.g. a page or track)
3. Click the option "Delete media" in the context menu
4. See that media is not removed

**Expected behavior**
Deleting media via the context menu of the structure tree of gallery in the metadata editor should work

**Screenshots**

https://github.com/user-attachments/assets/52c4109f-b0fc-4133-98b1-988457477c3d

**Release**
3.9.0.RC1, current `main` branch

**Additional context**
~~Deleting media via the context menu does work in Kitodo 3.8.x and presumably has been broken by the integration of `hibernate-search`.~~

thomaslow:
Fixes #6684

For some reason, the delete code for the context menu option "delete media" only iterates over "selectedPhysicalNodes", meaning selected tree nodes of the physical structure tree. Because of that, selected media in the combined structure tree is not deleted.

I also added a selenium test for deleting media in the combined structure tree.


BartChris:
Can we backport this to 3.9? And maybe even 3.8?

solth:
> Can we backport this to 3.9? And maybe even 3.8?

3.8 is not affected, I think (at least I couldn't reproduce the error in an installation where 3.8.5 is installed and #6684 does only lists 3.9.x and main)

BartChris:
> > Can we backport this to 3.9? And maybe even 3.8?
> 
> 3.8 is not affected, I think (at least I couldn't reproduce the error in an installation where 3.8.5 is installed and #6684 does only lists 3.9.x and main)

Would be good. I just noted it because the changes which seem to have introduced it were backported.

https://github.com/kitodo/kitodo-production/pull/6440

thomaslow:
I added a backport PR for 3.9, see #6961.

> Yes, I saw that, too, which I find confusing.

Multi drag & drop was backported to 3.8 upon @BartChris's request, see [comment](https://github.com/kitodo/kitodo-production/pull/6340#issuecomment-2669137560).

BartChris:
I am totally fine to not backport to 3.8, especially since you cannot reproduce it. (I have not checked 3.8) Having this fix in 3.9 and 4.0 is enough.

Issue #6024, opened 2024-04-05
metadata "processTitle" creates an an additional <mets:dmdSec> for some document types


andre-hohmann:
**Describe the bug**
If a new process is created, the metadata element `processTitle` is written in an own `<mets:dmdSec>` in the internal METS file. 
This element is connected with the first `<mets:div>`-element in `<mets:structMap TYPE="PHYSICAL">` - instead of `<mets:structMap TYPE="LOGICAL">`. 
This leads to some empty metadata elements in the exported METS-file. Probably, this depends on the XSLT configuration. 

This behavior can be observed for processes of the type _manuscript_, _monographs_, _volumes_, ... For processes of the type _newspaper year_, _newspaper issue_, _ephemera year_ and _ephemera issue_, the metadata `processTitle` is written in the existing `<mets:dmdSec>`, which is connected with `<mets:structMap TYPE="LOGICAL">`. 

Probably, this is caused by: 
- #4459

See also: 
- #6025

**To Reproduce**
1. Create a process (monograph, volume) in Kitodo.Production 3.6.0 or a newer version. 
2. Open the internal METS file
3. See the two `<mets:dmdSec>`-elements

**Expected behavior**
The metadata `processTitle` should always be written in the existing `<mets:dmdSec>`, which is connected with `<mets:structMap TYPE="LOGICAL">`. According to the behavior of the processes of the type newspaper, ephemera (tile, year, issue). 

Probably, this must be analyzed more deeply. It seems there are more consequences: 
- https://github.com/kitodo/kitodo-production/issues/6024#issuecomment-2039310575

**Release**
3.6.2 - probably since 3.6.0

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome

**Additional context**

<details>
<summary>Internal METS file: Manuscript</summary>

``` xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<mets:mets xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:kitodo="http://meta.kitodo.org/v1/" xmlns:mets="http://www.loc.gov/METS/">
    <mets:metsHdr CREATEDATE="2024-04-05T09:09:49.918+02:00" LASTMODDATE="2024-04-05T09:09:58.421+02:00">
        <mets:metsDocumentID>761642</mets:metsDocumentID>
    </mets:metsHdr>
[...]

Issue #4362, opened 2021-04-28
Newspaper Calendar Function - Exported METS does not include "type" for "month" and "day"


stefanCCS:
Using Version 3.2.1-build 24.03.21
I am using calendar newspaper function to create processes.
To do so I have in my rule set:
```
       <division id="newspaper" processTitle="MODSrecordIdentifier" use="createChildrenWithCalendar" withWorkflow="false"> 
           <label>newspaper</label>
            <label lang="de">Zeitung</label>
            <subdivisionByDate>
                <division dates="ORDERLABEL" id="year" scheme="yyyy" processTitle="+'-'+#YEAR" withWorkflow="false"/>
                <division dates="ORDERLABEL" id="month" scheme="yyyy-MM"/>
                <division dates="ORDERLABEL" id="day" scheme="yyyy-MM-dd"/>
            </subdivisionByDate>
        </division>
        
        <division id="issue"> <!-- DFG-Viewer compatible -->
            <label>Newspaper (one Issue)</label>
            <label lang="de">Zeitung (eine Ausgabe)</label>
        </division>


        <restriction division="newspaper" unspecified="forbidden">
            <permit division="year"/>
            <permit key="MODSrecordIdentifier" maxOccurs="1" minOccurs="1"/>
            <permit key="MODStitle" minOccurs="1"/>
        </restriction>
        <restriction division="year" unspecified="forbidden">
            <permit division="month"/>
            <permit key="ORDERLABEL" minOccurs="1"/>
        </restriction>
        <restriction division="month" unspecified="forbidden">
            <permit division="day"/>
            <permit key="ORDERLABEL" minOccurs="1"/>
        </restriction>
        <restriction division="day" unspecified="forbidden">
            <permit division="issue"/>
            <permit key="ORDERLABEL" minOccurs="1"/>
        </restriction>
        <restriction division="issue" unspecified="forbidden">
            <permit key="LABEL" />            
            <permit key="MODStitle" minOccurs="1"/>

```

The process generation function runs fine - it creates hierarchical processes as follows:
- 1 process for the newspaper title
- 1 process per year (below title)
- 1 process per issue (below according year)
 
The DMS Export of an issue creates exported METS for this issue, its according year and the title - so far, so good.
The exported METS for the title has mptr references to the "year" METS files - also good.
The exported "year" METS has full logical hierarchy (title, year,month,day, issues), with the issues having mptr references to the "issue" METS - also good.

[...]

PR #6931, opened 2026-03-17
Skip metadata assignment for untyped divisions

BartChris:
This Pull request addresses the issue that Kitodo right now often injects unwanted metadata into the METS file. The issue has been described in different places, e.g. 

https://github.com/kitodo/kitodo-production/issues/4362#issuecomment-4073939262
or
https://github.com/kitodo/kitodo-production/issues/6024#issuecomment-2061639760

The problem is, that right now it is not mandatory to define special rules in the ruleset which prevent the uncontrolled insertion of e.g. the `processTitle`:

```xml
<restriction division="" unspecified="forbidden">
    <permit division="page"/>
    <permit division="track"/>
    <permit division="other"/>
</restriction>

<restriction division="page" unspecified="forbidden">
     <permit key="ORDERLABEL" maxOccurs="1"/>
</restriction>
```

If the institution does not have those rules in place, `page` elements or `unspecified` (untyped) elements which are created for newspaper issues might get unwanted metadata injections.
My fix therefor does two things:
- If the type of a logical section is undefined, do not inject metadata there. (And in consequence inject unwanted DMDSecs)
- Do not inject unwanted metadata into pysical pages

If both things are wanted it has to be implemented in a safer way. We cannot guarantee that all institutions know about those settings and have them in place. The behavior should therefor by default prevent unwanted insertions.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -45,6 +45,7 @@ import org.kitodo.api.dataeditor.rulesetmanagement.RulesetManagementInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.SimpleMetadataViewInterface;
 import org.kitodo.api.dataeditor.rulesetmanagement.StructuralElementViewInterface;
 import org.kitodo.api.dataformat.Division;
+import org.kitodo.api.dataformat.LogicalDivision;
 import org.kitodo.exceptions.InvalidMetadataValueException;
 import org.kitodo.exceptions.NoSuchMetadataFieldException;
 import org.kitodo.production.services.dataeditor.DataEditorService;
@@ -672,6 +673,10 @@ public class ProcessFieldedMetadata extends ProcessDetail implements Serializabl
     @Override
     public void preserve() throws InvalidMetadataValueException, NoSuchMetadataFieldException {
         try {
+            if (isDivisionUntyped()) {
+                logger.warn("Skipping metadata preservation for untyped division.");
+                return;
+            }
             if (Objects.nonNull(division)) {
                 division.getContentIds().clear();
                 division.setOrderlabel(null);
@@ -715,6 +720,18 @@ public class ProcessFieldedMetadata extends ProcessDetail implements Serializabl
             container.metadata.add(metadataGroup);
             copy = false;
         }
+    }
+
+    private boolean isDivisionUntyped() {
+        if (division instanceof LogicalDivision logicalDivision) {
+            if (Objects.isNull(logicalDivision.getType()) || logicalDivision.getType().isBlank()) {
+                logicalDivision.getContentIds().clear();
+                logicalDivision.setOrderlabel(null);
+                logicalDivision.setLabel(null);
+                return true;
+            }
+        }
+        return false;
     }
 
     private void updateDivisionFromProcessDetail(String key, ProcessSimpleMetadata processDetail) throws InvalidMetadataValueException {
--- a/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
@@ -673,9 +673,8 @@ public class ProcessFieldedMetadata extends ProcessDetail implements Serializabl
     @Override
     public void preserve() throws InvalidMetadataValueException, NoSuchMetadataFieldException {
         try {
-            if (isDivisionUntyped()) {
-                return;
-            }
+            boolean untyped = isDivisionUntyped();
+
             if (Objects.nonNull(division)) {
                 division.getContentIds().clear();
                 division.setOrderlabel(null);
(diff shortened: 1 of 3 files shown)


BartChris:
> This Pull request addresses the issue that Kitodo right now often injects unwanted metadata into the METS file. The issue has been described in different places, e.g.

It might be that this only fixes parts of the issue. As outlined by @michaelkubina in https://github.com/kitodo/kitodo-production/issues/4362#issuecomment-4073604772 Kitodo also seems to inject default values (presets) into the untyped metadata divisions. It would have to be traced where this happens exactly (process creation? Save in Metadata editor?).

Kitodo/src/main/java/org/kitodo/production/forms/createprocess/ProcessFieldedMetadata.java
```diff
@@ -672,6 +673,10 @@ void markLeadingMetadataFields(List<Map<MetadataEntry, Boolean>> leadingMetadata
     @Override
     public void preserve() throws InvalidMetadataValueException, NoSuchMetadataFieldException {
         try {
+            if (isDivisionUntyped()) {
+                logger.warn("Skipping metadata preservation for untyped division.");
```
henning-gerhardt: How helpful is this warning message? There are no context information nor an administrator or any user with access to the log files can do here anything or inform anyone as context information are missing. Maybe the user in the UI should be informed about this case.

BartChris: You are right; i am trying to adress a behaviour which should be fixed by the application, but logging it seems overkill. I think "preserve" is called way to often so that the user will also be annoyed by those messages.

BartChris:
After having discussed the problems with @michaelkubina the unwanted enrichments happen, when an instiution has defined metadata keys as "always showing" and with a default preset, e.g:

```xml
 <key id="docType">
            <label>Document Type</label>
            <label lang="de">Dokumenttyp</label>
            <option value="monograph">
                <label>Monograph</label>
                <label lang="de">Monographie</label>
            </option>
            <option value="multivolume_work">
                <label>Multivolume Work</label>
                <label lang="de">Mehrbändiges Werk</label>
            </option>
            <preset>monograph</preset>
        </key>
```
and:

```xml
<setting key="docType" editable="true" alwaysShowing="true"/>
```

When you have those keys defined and select in the metadataeditor view of the issue one of the divisions without a type, those default values are injected in the UI:

image

Those UI value are then preserved in the meta.xml upon save. To quote @michaelkubina:

> Flow wenn man einen Vorgang öffnet:
> "Ausgabe" ist angewählt -> Speichern + Schließen -> alles Okay
> "Ausgabe" ist angewählt -> man klickt auf die erste "Aufteilung ohne Typ" und geht zurück auf Ausgabe -> Speichern + Schließen -> DMD-Sec für die Aufteilung ohne Typ wurde angelegt
> "Ausgabe" ist angewählt -> man klickt auf die zweite "Aufteilung ohne Type" und geht zurück auf Ausgabe -> Speichern + Schließen -> DMD-Sec für die zweite Aufteilung ohne Typ wurde angelegt.

When this setting is defined in the ruleset:

```xml
<restriction division="" unspecified="forbidden">
    <permit division="page"/>, 
    <permit division="track"/>
    <permit division="other"/>
</restriction>
```
this does not happen, because the fields in questions are not actually rendered in the UI. My fix does not prevent them from appearing in the UI, but prevents that those values are serialized in the meta.xml as untyped divisions are skipped when calling `preserve`. (Which maps the UI data to the Kitodo data model, which then gets serialized into the XML).

This PR therefor introduces a safety net for institutions which do not have the necessary ruleset rules defined. 
Another step might be to block the metadata display in the UI for those untyped divisions so that they are not part of the serialized data.
The long term solution would probably be to get rid of untyped divisions in Kitodo.

BartChris:
My commit https://github.com/kitodo/kitodo-production/pull/6931/commits/1f0d82a1a08fda1114f9e64ec82ad7441d8e5bbe goes one step further. If the division is untyped we should not inject metadata in the UI layer. Untyped divisions do not show any data now, which might get serialized into the meta.xml:

image

BartChris:
I am not exactly sure if i introduce a behaviour change here. It might be that people actually assign metadata to the untyped divisions. This is of course only possible if the metadata is actually rendered in the UI and preserved on Save. My fix basically enforces that the untyped divisions do not get any metadata assigned. 

Maybe @andre-hohmann can comment here.

BartChris:
> My fix basically enforces that the untyped divisions do not get any metadata assigned.

This does of course not exclude the option to enrich metadata here on export or via XSL. See: https://github.com/kitodo/kitodo-production/issues/4362#issuecomment-843056583

andre-hohmann:
I hope I understood correctly that "_untyped divisions do not get any metadata assigned_" refers to both:
1. **Manual assignment** via the metadata editor, and
2.  **Automatic assignment** during the creation process using the calendar.

Regarding 1 (Manual):
At SLUB, we correct issues, which are assigned to wrong dates. In these cases, the `ORDERLABEL` of the second untyped division is corrected, too. 

Regarding 2 (Automatic):
This is required to generate `ORDERLABEL` for the _day_, in the `<mets:structMap TYPE="LOGICAL">` for the issue. 
This is then used to derive the `ORDERLABEL` for the _month_ by XSLT as suggested by @stefanCCS : 
- https://github.com/kitodo/kitodo-production/issues/4362#issuecomment-847809209


Wouldn't this be solved, if the container levels would be eliminated?: https://github.com/kitodo/kitodo-production/issues/4362#issuecomment-4243358729 
As written in the comment, it would be necessary to generate the values for the month and day level in the issue from the information in the year process.

BartChris:
> Wouldn't this be solved, if the container levels would be eliminated?: [#4362 (comment)](https://github.com/kitodo/kitodo-production/issues/4362#issuecomment-4243358729)

I agree, but i am not really sure what this means exactly and what the consequences would be. But thanks a lot for your comment. Given what you say my changes are probably to radical as i think it would break your existing workflows. On the other hand the current behaviour of metadata injections can have really destructive consequences. I have to think more about this.

PR #6937, opened 2026-03-20
Avoid possible confusion between local variable and field with the same name

solth:
Rename some local variables to avoid possible confusion between local variables and class fields with the same name. 

Detected by CodeQL.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Block.java
+++ b/Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Block.java
@@ -248,14 +248,14 @@ public class Block {
         }
         ArrayList<IndividualIssue> result = new ArrayList<>(this.issues.size());
 
-        List<Issue> individualIssues = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Block.java
```diff
@@ -246,16 +246,16 @@ public List<IndividualIssue> getIndividualIssues(LocalDate date) {
         if (!isMatch(date)) {
             return Collections.emptyList();
         }
-        ArrayList<IndividualIssue> result = new ArrayList<>(issues.size());
+        ArrayList<IndividualIssue> result = new ArrayList<>(this.issues.size());
 
-        List<Issue> issues = new ArrayList<>();
+        List<Issue> individualIssues = new ArrayList<>();
```
matthias-ronge: I am not happy with this one. `List<Issue>` should be something like `issueList` or `appearedIssues`, because it contains `Issue` instances, not `IndividualIssue` instances. Those are created below in line 259

Issue #6940, opened 2026-03-20
Updating an existing project folder leads to exception and the folders cannot be deleted


BartChris:
**Describe the bug**

### 1. Failing update of folders

If you have more than one project folder configured and try to update any folder beside the first one, an exception is thrown.

<img width="1758" height="264" alt="Image" src="https://github.com/user-attachments/assets/68192c03-7ccb-4ac6-a28b-266c428dc41e" />

<img width="717" height="791" alt="Image" src="https://github.com/user-attachments/assets/0c1a8631-6785-4223-9e7b-c05bdd342d09" />

A new page is showing the following exception:

```
"Oops!
A problem occurred during processing the ajax request. Subsequently, another problem occurred during processing the error page which should inform you about that problem.
If you are the responsible web developer, it's time to read the server logs about the bug in the error page itself."
```

I suspect that this is related to the comparison code here:

https://github.com/kitodo/kitodo-production/blob/7631816ee2d20d93e13b6243fe8f3f6bf365cd5e/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java#L289-L298

which relies on comparing folders using the modified `equals` implementation in Folder.java. 

It seems as is if the projects folders (at least a part of them) are not fully initialized when this comparison happens. The first folder seems to be initialized and i can update it. 

<img width="591" height="106" alt="Image" src="https://github.com/user-attachments/assets/5ee10a17-f67b-4a5b-8a9a-a6a29860f9f9" />

When trying to update the unitialized folders, the equals-comparison fails and Kitodo assumes that there is a new folder created which is not the case. I do not really understand how this leads to the cited error page. But since i do not get the error page for the initialized folder, i suspect there is a connection.

The code is hard to trace unfortunately as it builds upon a variable named `myFolder` for which it is not quite clear to me if there is a difference between the use case of adding a folder vs editing a folder.

**To Reproduce**
1. Go to a project with more than one folder configured
2. Edit one of the folders
3. Click Save/Apply

### 2. Delete fails for folders

It is also impossible to delete the uniitialized folders and i suppose the reason is similar. 

**To Reproduce**
1. Go to a project with more than one folder configured
2. Try to delete one of the folders
3. Folder does not disappear 
4. Saving the form does not delete the folder

**Release**

Kitodo Main

**Additional context**

This reminds me of a problem fixed for 3.9 and i suppose the switch to `@ViewScoped` uncovered some depper flaws in the folder logic on the project page.
[...]

Issue #6337, opened 2024-12-02
Incoherent behaviour in project folder settings


BartChris:
**Describe the bug**
Adding or deleting folders in the project settings works different. When i add a folder, it is not necessary to save the new setting. It seems to be saved automatically. When i delete a folder, i have to explicitely save the settings in order for the settings to take effect.

Screen in question is this.
![image](https://github.com/user-attachments/assets/18515458-5cbd-4c29-9381-d0710bdfe4b7)

This especially becomes appearant when i delete a folder and want to add a new folder with the same filegroup. Although the old entry is no longer visible it still has effect on newly created folders.

**To Reproduce**
1. Create a new folder 
2. Leave the project page without savin
3. Return to the project page - created folder is there
4. Delete a folder
5. Leave the page without saving
6. Return to the page, folder is not deleted

**Expected behavior**
The behaviour should be coherent. I suggest to only save the folders on an actual save. If a folder is deleted in the UI (although not saved) it should be possible to add a new folder with the same file group.

Kitodo 3.7, 3.8, Master

PR #6946, opened 2026-03-29
Fix folder editing and deleting

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6940
Fixes https://github.com/kitodo/kitodo-production/issues/6337

Supersedes https://github.com/kitodo/kitodo-production/pull/6944

This Pull Request adresses the same issues outlined in https://github.com/kitodo/kitodo-production/pull/6944, but tries so without adding a full blown DTO setup. Working and comparing UI edits with potentially unitialized Hibernate entities leads to errors, which are not 100% reproducible, but i can trigger them many times when doing adding, editing and deleting on the folder form. 

The main concept is to create an in memory copy of the Hibernate entities once the form is loaded and compare UI edits against those in memory list. When saving the database state is consolidated against the memory copy.  So in detail:

- When the form is loaded, the project’s folders are copied into an in-memory list (workingFolders)
- All UI interactions operate on this in-memory state instead of directly modifying the persistent entities
- On save, the persistent entity state is synchronized with the in-memory list

I tried multiple chains of editing and deleting and could not reproduce the issues from https://github.com/kitodo/kitodo-production/issues/6940. I also added tests. The tests only cover the basic adding and deleting case, which is better than nothing i guess.

Edit: After having inspected again i noticed that my "copy" of the Hibernate data is not a real deep copy but puts the same hibernate entities on a different list. In my tests this seems to fix the issues i encountered but it is probably less strict than DTO approach and might still expose buggy behavior. (https://github.com/kitodo/kitodo-production/pull/6944) The current approach is for now the simplest change i was able to implement which appears to work.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
@@ -295,6 +295,7 @@ public class ProjectEditView extends BaseEditView {
      *
      */
     public void deleteFolder() {
+        // Identity comparison is intentional: remove the exact folder instance being edited.
         workingFolders.removeIf(folder -> folder == editingFolder);
     }
 


Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
```diff
@@ -857,10 +816,10 @@ public Boolean hasProcesses() {
      * @return the list of possible validation configurations that can be assigned to a folder
      */
     public List<LtpValidationConfiguration> getPossibleLtpValidationConfigurations() {
-        if (Objects.isNull(myFolder)) {
+        if (Objects.isNull(editingFolder)) { // Changed from myFolder
```
matthias-ronge: This comment will no longer help anyone in the future. Please delete it.

Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
```diff
@@ -276,33 +264,31 @@ public void delete(int projectId) {
...
      */
     public String addFolder() {
-        this.myFolder = new Folder();
-        this.myFolder.setProject(this.project);
-        this.generator = new FolderGenerator(myFolder);
-        this.newFolders.add(this.myFolder.getId());
+        this.editingFolder = new Folder();
+        this.editingFolder.setProject(this.project);
+        // Default values
+        this.editingFolder.setMimeType("image/jpeg");
```
matthias-ronge: The media types are specified in an XML file, and it's possible (is it possible?) that there isn't one, and I wouldn't hardcode it. If there's nothing there, maybe just set the first value from the list, or nothing at all.

BartChris: I removed the setter and it seems to work.

Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
```diff
@@ -311,11 +295,7 @@ public void saveFolder() {
      *
      */
     public void deleteFolder() {
-        if (Objects.isNull(myFolder.getId())) {
-            project.getFolders().remove(myFolder);
-        } else {
-            deletedFolders.add(this.myFolder.getId());
-        }
+        workingFolders.removeIf(folder -> folder == editingFolder);
```
matthias-ronge, line 299: Maybe leave a comment here that pointer comparison is intentional

BartChris: good idea, i added a comment.

solth:
@BartChris do you think it would be feasible to backport this fix to 3.9.x (and potentially even 3.8.x)?

BartChris:
Yes i did the backports with the necessary changes. So far it looks good.

PR #6947, opened 2026-03-31
Simplify Web driver handling

BartChris:
Pull Request https://github.com/kitodo/kitodo-production/pull/6872 introduced Selenium 4, which also introduced Selenium Manager for more straightforward handling of browser drivers:
https://www.selenium.dev/documentation/selenium_manager

I think this allows us to massively simplify our handling of the Chrome Driver, which seems to solve the CI issues for me  @solth.


Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/WebDriverProvider.java
solth:
```suggestion
```

PR #6961, opened 2026-04-09
[3.9] Fix metadata editor media cannot be deleted

thomaslow:
Backport of #6930

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
@@ -273,14 +273,18 @@ public class StructurePanel implements Serializable {
      * Delete all currently selected physical divisons.
      */
     public void deleteSelectedPhysicalDivisions() {
-        if (Objects.isNull(selectedPhysicalNodes)) {
-            // there is nothing to do
-            return;
+        if (Objects.nonNull(selectedPhysicalNodes)) {
+            // iterate over selected tree nodes in physical tree
+            for (TreeNode selectedPhysicalNode : selectedPhysicalNodes) {
+                deleteSelectedPhysicalDivision(selectedPhysicalNode);
+            }
         }
-        for (TreeNode selectedPhysicalNode : selectedPhysicalNodes) {
-            deleteSelectedPhysicalDivision(selectedPhysicalNode);
+        if (Objects.nonNull(selectedLogicalNodes)) {
+            // iterate over selected tree nodes in logical tree
+            for (TreeNode selectedLogicalNode : selectedLogicalNodes) {
+                deleteSelectedPhysicalDivision(selectedLogicalNode);
+            }
         }
-
         int i = 1;
         for (PhysicalDivision physicalDivision : dataEditor.getWorkpiece().getAllPhysicalDivisionChildrenSortedFilteredByPageAndTrack()) {
             physicalDivision.setOrder(i);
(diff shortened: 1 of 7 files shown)


solth (commented):
Deleting the media via the context menu in the structure tree removes it from the gallery, but not from the tree itself for me:

https://github.com/user-attachments/assets/656278c7-1285-4285-9390-38add765e460

Maybe some `update` attribute/statement needs to be adjusted?

thomaslow:
Can you describe in more detail how you ended up with this single image inside the process?

I cannot reproduce this the way it is shown in the video. I can delete unstructured media. However, I noticed that I cannot delete unstructured media after they become unstructured again by deleting the parent structure element they were previously a part of. Example:

https://github.com/user-attachments/assets/13647403-0614-4be5-96b0-4eaeeac9ec57

Is this the same scenario that you tested? If so, I think this is a different bug and probably should be handled in a different PR, not the backport of #6930.

solth:
> Can you describe in more detail how you ended up with this single image inside the process?
> 
> I cannot reproduce this the way it is shown in the video. I can delete unstructured media. However, I noticed that I cannot delete unstructured media after they become unstructured again by deleting the parent structure element they were previously a part of.

I don't remember the specifics of how I created this particular process, but this is the default state of this project, where posters are digitized, normally only contain _one_ image (two, if the backside is printed as well) and does not contain any structures.

I will try it with a new process and multiple images contained in logical structures.

solth:
@thomaslow I just tested it with another, "fresh" process and can confirm that - almost - all constellations worked:
- removing non-last structured media ✅ 
- removing last structured media ✅ 
- removing non-last unstructured media ✅ 
- removing last unstructued media ✅ 

The problem persisted in the process where I first observed it, though. In addition, I was able to reproduce it in other processes as well. 

When comparing the `meta.xml` files from those processes that worked with those that didn't, I found the difference: in the metadata file of processes where removed media was _not_ removed from the structure tree, the media wasn't linked via `<mets:smLink>` elements in the first place! This is only the case when processes are opened and new media was found in the images directory, but references to the new files haven't been persisted to the `meta.xml` file, yet. In other words, this is the situation - for example in a "Scanning" task - when the the process is opened but not yet _saved_ with the newly found images.

So the problem seems to be removing media from processes before the references to that media has been saved to the file. Since the media is still removed from the gallery, though, it seems to work in general, and there is only some small quirk in the update of the metadata tree, I guess.

solth (approved):
> So the problem seems to be removing media from processes before the references to that media has been saved to the file. Since the media is still removed from the gallery, though, it seems to work in general, and there is only some small quirk in the update of the metadata tree, I guess.

Ok, I just realised the same limitation holds true for the current `main` branch, where the corresponding fix #6930 has already been applied. So merging this pull request brings version 3.9 up to par with the upcoming version 4.0 in this regard. 

I will open a separate issue for the problem of unsaved media being unremovable.

thomaslow:
Yes. I think it makes sense to merge this PR and have a look at the remaining problem in a second PR. It's on my todo list. I'll look into it. Do you think its a blocker for the 4.0 release?

solth:
> Yes. I think it makes sense to merge this PR and have a look at the remaining problem in a second PR. It's on my todo list. I'll look into it. 

Thanks a lot!

> Do you think its a blocker for the 4.0 release?

No, I don't think so. The functionality has already been restored in general by your fix, so working out this last little issue with a later bugfix release is no problem in my opinion.

thomaslow:
Sorry for the confusion with the 3.9.1 release. I didn't realize there was a debate about whether to include this PR or not.

solth:
> Sorry for the confusion with the 3.9.1 release. I didn't realize there was a debate about whether to include this PR or not.

Don't worry, there was no debate (outside of my head, at least), and I still had to merge other PRs for the `3.9.x` branch before releasing 3.9.1 anyway.

PR #6964, opened 2026-04-09
Update to Spring security 7.0.x components

henning-gerhardt:
Update to most recent 7.0.x Spring security components which include:

- Replacing some deprecated method calls with more recent usage
- Using `PathPatternRequestMatcher` for URI with `**` patterns
- Remove unused / old request pattern

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
+++ b/Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
@@ -22,7 +22,8 @@ import org.springframework.security.config.annotation.web.configuration.EnableWe
 import org.springframework.security.core.session.SessionRegistry;
 import org.springframework.security.core.session.SessionRegistryImpl;
 import org.springframework.security.web.SecurityFilterChain;
-import org.springframework.security.web.access.intercept.FilterSecurityInterceptor;
+import org.springframework.security.web.access.intercept.AuthorizationFilter;
+import org.springframework.security.web.servlet.util.matcher.PathPatternRequestMatcher;
 import org.springframework.web.servlet.handler.HandlerMappingIntrospector;
 
 /**
@@ -124,7 +125,7 @@ public class SecurityConfig {
         // more general rules should be at end
         authorizeGeneralPages(http);
 
-        http.addFilterAfter(new SecurityObjectAccessFilter(), FilterSecurityInterceptor.class);
+        http.addFilterAfter(new SecurityObjectAccessFilter(), AuthorizationFilter.class);
 
         handleFormLogin(http);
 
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
```diff
@@ -163,8 +164,8 @@ private void authorizeSpecificPages(HttpSecurity http) throws Exception {
 
     private void authorizeGeneralPages(HttpSecurity http) throws Exception {
         http.authorizeHttpRequests(requests -> requests
-            .requestMatchers("/pages/images/**").permitAll()
-            .requestMatchers("/jakarta.faces.resource/**", "**/resources/**").permitAll()
+            .requestMatchers(PathPatternRequestMatcher.pathPattern("/pages/images/**")).permitAll()
+            .requestMatchers(PathPatternRequestMatcher.pathPattern("/jakarta.faces.resource/**")).permitAll()
```
thomaslow: Why did you add `PathPatternRequestMatcher.pathPattern` here? I couldn't find a reason. 

As far as I can tell, the purpose of `PathPatternRequestMatcher` is to allow customization of how matching works, e.g., to specify a specific base path (in our case `/kitodo`) as described [here](https://docs.spring.io/spring-security/reference/servlet/authorization/authorize-http-requests.html#match-by-mvc). Removing `PathPatternRequestMatcher.pathPattern` seems to work fine.

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
```diff
@@ -163,8 +164,8 @@ private void authorizeSpecificPages(HttpSecurity http) throws Exception {
 
     private void authorizeGeneralPages(HttpSecurity http) throws Exception {
         http.authorizeHttpRequests(requests -> requests
-            .requestMatchers("/pages/images/**").permitAll()
-            .requestMatchers("/jakarta.faces.resource/**", "**/resources/**").permitAll()
```
thomaslow, line 167: Nice catch. A generic rule like `**/resources/**` seems a bit dangerous.

thomaslow (commented):
The update to spring security 7 works for me.

The migration notes state that an upgrade to Jackson v3 is recommended. Kitodo.Production currently uses Jackson v2, and I remember that there were some problems with upgrading that. I think there are some older libraries in 
our dependencies that still use very old versions (e.g. JHove). Anyway, I think this doesn't apply to how we use spring security in Kitodo.Production. There isn't any XML serialization happening during login.

Besides, spring security v7 lists multi-factor authentication as a new feature. Maybe this is something that is relevant to increase security in the future, e.g., for admin accounts?

For  references:
-  [Migration notes to v7](https://docs.spring.io/spring-security/reference/migration/index.html)
- [What's new in v7](https://docs.spring.io/spring-security/reference/whats-new.html)

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
henning-gerhardt: I got errors on startup with the double `*` entries and the internet told me that this would be a solution. I can try it again and report back.

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
henning-gerhardt: Before removing it I looked into our log files and I did not found any request for `resources` inside and the git history told me that this `resources` is old and maybe only used in the past, so removing it should be a good way.

Kitodo/src/main/java/org/kitodo/production/security/SecurityConfig.java
henning-gerhardt: Okay, now it does work without an error message that double `*` are not allowed in this two lines. I will remove it and update the pull request.

thomaslow: Hmm. Maybe it had something to do with the `**/resources/**` pattern. Wildcards at the beginning of a pattern are always a bit more troublesome than wildcards at the end.

thomaslow (approved):
Works for me!

PR #6966, opened 2026-04-09
Prevent exception during catalog import

solth:
This pull request prevents an exception that could - and in my tests always would - occur during metadata import from remote catalog search interfaces (I tested with the [SRU interface of Kalliope](https://kalliope-verbund.info/de/support/sru.html))

I was unable to reproduce this error in Kitodo 3.9.0, but I am not sure where it was introduced afterwards. 

It seems the metadata nodes used to create a `TempProcess` contain _text_ nodes as well, which do not contain attributes and therefor provoke the exception mentioned above when trying to save a new process. Originally, only the `<kitodo:metadata name="...">` _element_ nodes where expected in the corresponding method. 

I now check for the node type before trying to retrieve the `name` attribute from it to prevent the exception.


BartChris (approved):
Does the error also not appear in the latest Snapshot of 3.9.x?

solth:
> Does the error also not appear in the latest Snapshot of 3.9.x?

It does indeed! Backport at #6967

BartChris:
It might be that my change has subtle effects on whitespaces or similar which expose those issue. https://github.com/kitodo/kitodo-production/pull/6888

PR #6969, opened 2026-04-10
Update and remove obsolete plexus dependencies

henning-gerhardt:
- Update `plexus-archiver` to version `4.11.0`
- Remove obsolete `plexus-container-default` dependency


BartChris (approved):
Both `poi-ooxml` and `plexus-archiver` bring in `commons-compress` as dependency, so Excel export has always to be checked.  `plexus-archiver` now brings in version 1.28, which is newer than the version brought in by `poi-ooxml`. It works for me and i cannot see any other problem. Looks good to me.

PR #6970, opened 2026-04-10
Fix CSV action

BartChris:
The CSV export action is referencing the wrong form and therefor does not work.


thomaslow (approved):
This is a mistake from #6903. CSV export works for me again after this.

PR #6974, opened 2026-04-12
Fix problems in workflow editor

### LINKED ISSUE #6290: Empty .svg worfklow file, if workflow saved without changes

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/6290
Adresses https://github.com/kitodo/kitodo-production/pull/6355#issuecomment-2627106319

Supersedes https://github.com/kitodo/kitodo-production/pull/6408

Working with the workflow editor already had some problems (see above) in the past. It seems like the switch to `ViewScoped` exposed additional problems:
When you create a new workflow, set a title but do not define anything else:

image

saving now leads to an error page:

> Oops!
> 
> A problem occurred during processing the ajax request. Subsequently, another problem occurred during processing the error page which should inform you about that problem.
> 
> If you are the responsible web developer, it's time to read the server logs about the bug in the error page itself.

This PR tries to adress this problem and the other problems in the following (linked) issues:

- Saving an existing workflow without any changes stores an empty SVG file (https://github.com/kitodo/kitodo-production/issues/6290)
- It is not prevented that a new workflow with an already used name is created. (https://github.com/kitodo/kitodo-production/pull/6355#issuecomment-2627106319)

In addition to that the PR also deletes old files when a workflow is renamed and stores the files under the new workflow title.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
@@ -312,15 +312,17 @@ public class WorkflowEditView extends BaseEditView {
      * @return true if save, false if not
      */
     private boolean saveFiles() throws IOException, WorkflowException {
-        Map<String, String> requestParameterMap = FacesContext.getCurrentInstance().getExternalContext()
-                .getRequestParameterMap();
-
+        FacesContext context = FacesContext.getCurrentInstance();
         Map<String, URI> diagramsUris = getDiagramUris();
 
         URI svgDiagramURI = diagramsUris.get(SVG_DIAGRAM_URI);
         URI xmlDiagramURI = diagramsUris.get(XML_DIAGRAM_URI);
 
-        xmlDiagram = requestParameterMap.get("editForm:workflowTabView:xmlDiagram");
+        if (Objects.nonNull(context)) {
+            xmlDiagram = context.getExternalContext()
+                    .getRequestParameterMap()
+                    .get("editForm:workflowTabView:xmlDiagram");
+        }
         if (Objects.nonNull(xmlDiagram)) {
             svgDiagram = StringUtils.substringAfter(xmlDiagram, "kitodo-diagram-separator");
             xmlDiagram = StringUtils.substringBefore(xmlDiagram, "kitodo-diagram-separator");
--- a/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
@@ -175,6 +175,20 @@ public class WorkflowEditView extends BaseEditView {
         }
     }
 
+    /**
+     * Handles renaming of diagram files when the workflow title changes.
+     *
+     * <p>XML files are treated as authoritative and will be replaced if a new version exists.</p>
+     *
+     * <p>SVG handling depends on whether a new (non empty) SVG-diagram was submitted:
+     * <ul>
+     *   <li>If no SVG was submitted (e.g. editor did not modify it), the existing SVG is moved to the new filename.</li>
+     *   <li>If a new SVG was submitted and successfully written, the old SVG is deleted.</li>
+     * </ul>
+     *
+     * <p>This logic avoids unnecessary data loss and ensures that existing diagrams are preserved
+     * when no changes were made in the editor.</p>
+     */
     private void handleDiagramFilesAfterTitleChange() {
         if (Objects.isNull(originalTitle) || originalTitle.equals(workflow.getTitle())) {
             return;
@@ -189,7 +203,6 @@ public class WorkflowEditView extends BaseEditView {
             if (fileService.fileExist(oldXml) && fileService.fileExist(newXml)) {
                 fileService.delete(oldXml);
             }
-            // SVG: move or delete depending on edit
             if (fileService.fileExist(oldSvg)) {
                 if (StringUtils.isBlank(svgDiagram)) {
                     fileService.moveFile(oldSvg, newSvg);
--- a/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
@@ -142,7 +142,7 @@ public class WorkflowEditView extends BaseEditView {
             return this.stayOnCurrentPage;
         }
         try {
-            if (workflow.getId() == null
+            if (!workflow.getTitle().equals(originalTitle)
                     && ServiceManager.getWorkflowService()
                     .workflowTitleExists(workflow.getTitle())) {
                 Helper.setErrorMessage(
@@ -160,9 +160,7 @@ public class WorkflowEditView extends BaseEditView {
                     migration = false;
                     return MIGRATION_FORM_PATH + "&workflowId=" + workflow.getId();
                 }
-                if (Objects.nonNull(originalTitle) && !originalTitle.equals(workflow.getTitle())) {
-                    handleDiagramFilesAfterTitleChange();
-                }
+                handleDiagramFilesAfterTitleChange();
                 return WorkflowListView.VIEW_PATH + "&" + getReferrerListOptions();
             } else {
                 return this.stayOnCurrentPage;
@@ -188,7 +186,6 @@ public class WorkflowEditView extends BaseEditView {
             URI oldSvg = oldUris.get(SVG_DIAGRAM_URI);
             URI newXml = newUris.get(XML_DIAGRAM_URI);
             URI newSvg = newUris.get(SVG_DIAGRAM_URI);
-            // XML: delete only if new one exists
             if (fileService.fileExist(oldXml) && fileService.fileExist(newXml)) {
                 fileService.delete(oldXml);
             }
@@ -326,8 +323,10 @@ public class WorkflowEditView extends BaseEditView {
         if (Objects.nonNull(xmlDiagram)) {
             svgDiagram = StringUtils.substringAfter(xmlDiagram, "kitodo-diagram-separator");
             xmlDiagram = StringUtils.substringBefore(xmlDiagram, "kitodo-diagram-separator");
-            if (Objects.isNull(xmlDiagram) || StringUtils.isBlank(xmlDiagram)) {
-                Helper.setErrorMessage("Workflow diagram is missing.");
+            if (StringUtils.isBlank(xmlDiagram)) {
+                Helper.setErrorMessage(
+                        Helper.getTranslation("errorWorkflowDiagramMissing")
+                );
                 return false;
             }
 
--- a/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
@@ -141,6 +141,14 @@ public class WorkflowEditView extends BaseEditView {
             return this.stayOnCurrentPage;
         }
         try {
+            if (workflow.getId() == null
+                    && ServiceManager.getWorkflowService()
+                    .workflowTitleExists(workflow.getTitle())) {
+                Helper.setErrorMessage(
+                        Helper.getTranslation("duplicateWorkflowTitle", workflow.getTitle())
+                );
+                return this.stayOnCurrentPage;
+            }
             if (saveFiles()) {
                 this.workflow.setStatus(this.workflowStatus);
                 saveWorkflow();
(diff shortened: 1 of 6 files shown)


solth:
@Erikmitk would you feel comfortable to review this pull request since you worked on the workflow editor?

Erikmitk:
> @Erikmitk would you feel comfortable to review this pull request since you worked on the workflow editor?

Yes, I can take a look!

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
```diff
@@ -141,6 +142,14 @@ public String saveAndRedirect() {
             return this.stayOnCurrentPage;
         }
         try {
+            if (workflow.getId() == null
```
Erikmitk, line 145: This would only check duplicate titles for new workflows. We also need to cover renames of existing titles. Doesn't matter if the workflow is new or old. The given title should not already exist. If the `originalTitle` is `null` that's fine.


```suggestion
            if (!workflow.getTitle().equals(originalTitle)
```

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
```diff
@@ -165,6 +177,35 @@ public String saveAndRedirect() {
         }
     }
 
+    private void handleDiagramFilesAfterTitleChange() {
+        if (Objects.isNull(originalTitle) || originalTitle.equals(workflow.getTitle())) {
+            return;
+        }
```
Erikmitk, line 183: See above. Same check but second time.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
```diff
@@ -151,6 +160,9 @@ public String saveAndRedirect() {
                     migration = false;
                     return MIGRATION_FORM_PATH + "&workflowId=" + workflow.getId();
                 }
+                if (Objects.nonNull(originalTitle) && !originalTitle.equals(workflow.getTitle())) {
+                    handleDiagramFilesAfterTitleChange();
+                }
```
Erikmitk, line 165: The if guard is the exact same as inside of the method. One is `&&` the other `||` but checks the same condition, so one of them can go. I'd say the check should happens inside `handleDiagramFilesAfterTitleChange()` so it's self-contained.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
Erikmitk: The files are saved before the workflow so a file loss would happen before the DB catches the title duplication even on renames. The problem exists for old and new workflows at this point.

A separate and new test case should prevent this behavior in the future.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
```diff
@@ -165,6 +177,35 @@ public String saveAndRedirect() {
...
+            return;
+        }
+        try {
+            Map<String, URI> oldUris = getDiagramUris(originalTitle);
+            Map<String, URI> newUris = getDiagramUris(workflow.getTitle());
+            URI oldXml = oldUris.get(XML_DIAGRAM_URI);
+            URI oldSvg = oldUris.get(SVG_DIAGRAM_URI);
+            URI newXml = newUris.get(XML_DIAGRAM_URI);
+            URI newSvg = newUris.get(SVG_DIAGRAM_URI);
+            // XML: delete only if new one exists
```
Erikmitk, line 191: The comment seems redundant with the condition. The if-guard also checks for more conditions (Delete only when new AND old one exist). I'd remove the comment.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
```diff
@@ -165,6 +177,35 @@ public String saveAndRedirect() {
...
+            if (fileService.fileExist(oldXml) && fileService.fileExist(newXml)) {
+                fileService.delete(oldXml);
+            }
+            // SVG: move or delete depending on edit
```
Erikmitk, line 195: This comment could be more elaborate to explain what the edit condition is exactly i.e. why would the `svgDiagram` be empty in the first place and needs to be checked here?

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
```diff
@@ -271,30 +312,37 @@ public void delete() {
...
+        if (Objects.nonNull(context)) {
+            xmlDiagram = context.getExternalContext()
+                    .getRequestParameterMap()
+                    .get("editForm:workflowTabView:xmlDiagram");
+        }
         if (Objects.nonNull(xmlDiagram)) {
             svgDiagram = StringUtils.substringAfter(xmlDiagram, "kitodo-diagram-separator");
             xmlDiagram = StringUtils.substringBefore(xmlDiagram, "kitodo-diagram-separator");
+            if (Objects.isNull(xmlDiagram) || StringUtils.isBlank(xmlDiagram)) {
+                Helper.setErrorMessage("Workflow diagram is missing.");
```
Erikmitk, line 330: Message needs translation.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
```diff
@@ -271,30 +312,37 @@ public void delete() {
...
-        xmlDiagram = requestParameterMap.get("editForm:workflowTabView:xmlDiagram");
+        if (Objects.nonNull(context)) {
+            xmlDiagram = context.getExternalContext()
+                    .getRequestParameterMap()
+                    .get("editForm:workflowTabView:xmlDiagram");
+        }
         if (Objects.nonNull(xmlDiagram)) {
             svgDiagram = StringUtils.substringAfter(xmlDiagram, "kitodo-diagram-separator");
             xmlDiagram = StringUtils.substringBefore(xmlDiagram, "kitodo-diagram-separator");
+            if (Objects.isNull(xmlDiagram) || StringUtils.isBlank(xmlDiagram)) {
```
Erikmitk, line 329: `Objects.isNull(xmlDiagram)` can be removed as it is already check in the prior if-guard. At this line `xmlDiagram` is always `nonNull`.
```suggestion
            if (StringUtils.isBlank(xmlDiagram)) {
```

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
Erikmitk, line 171: Side note: Should this title duplication case be more explicit for the users? There's an exception with the correct information on trying to save:

https://github.com/kitodo/kitodo-production/pull/6974/changes#diff-aabde4c0f43f9f2f9814025f301d777742e0d1d0aecbe0b50ea94b2861804a92L142

> `new DAOException(Helper.getTranslation("duplicateWorkflowTitle", workflow.getTitle()));
`

but it get's caught here with the others and the error becomes more generic.`IOException | SAXException | FileStructureValidationException` make sense a a group but I think the `DAOException` should be caught in its own block so the title duplication error message can propagate to the users.

Kitodo/src/test/java/org/kitodo/production/forms/WorkflowFormIT.java
```diff
@@ -139,4 +158,73 @@ private void createAndSaveDataEditorSetting(int templateTaskId) throws DAOExcept
...
+
+        Files.writeString(oldXml.toPath(), validXmlWorkflow, StandardCharsets.UTF_8);
+        currentWorkflowForm.setWorkflow(workflow);
+        currentWorkflowForm.getWorkflow().setTitle(newTitle);
+        currentWorkflowForm.setXmlDiagram(validXmlWorkflow + "kitodo-diagram-separator");
+        currentWorkflowForm.saveAndRedirect();
+
+        File newSvg = new File(diagramDir + newTitle + ".svg");
+        File newXML = new File(diagramDir + newTitle + ".bpmn20.xml");
+        assertFalse(oldSvg.exists(), "Old SVG should be moved");
```
Erikmitk, line 192: Removed?

```suggestion
        assertFalse(oldSvg.exists(), "Old SVG should have been removed");
```

Kitodo/src/test/java/org/kitodo/production/forms/WorkflowFormIT.java
```diff
@@ -139,4 +158,73 @@ private void createAndSaveDataEditorSetting(int templateTaskId) throws DAOExcept
...
+        assertTrue(newSvg.exists(), "New SVG should exist");
+        assertFalse(oldXml.exists(), "Old XML should be moved");
```
Erikmitk, line 194: Removed?

```suggestion
        assertFalse(oldXml.exists(), "Old XML should have been removed");
```

Erikmitk (changes requested):
The changes overall make sense but there is one important and some minor changes necessary. (I only looked at the code.)

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
BartChris: Good point, i added a long comment to the whole method trying to outline the intent.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
BartChris: The duplicate title check is part of the business logic and handled in the view layer as part of the normal control flow.

A DAOException may still occur during persistence, but it would indicate unexpected database issues (e.g. connection problems or concurrent modifications). So i am not sure if we should change this here.

BartChris:
@Erikmitk 
Thanks a lot for your in depth review. I tried to address your comments. I am still thinking about this comment:

>The files are saved before the workflow so a file loss would happen before the DB catches the title duplication even on renames. The problem exists for old and new workflows at this point.
A separate and new test case should prevent this behavior in the future.

Are you suggesting to do the DB save on the workflow before saving the files?

Erikmitk:
> @Erikmitk Thanks a lot for your in depth review. I tried to address your comments. I am still thinking about this comment:
> 
> > The files are saved before the workflow so a file loss would happen before the DB catches the title duplication even on renames. The problem exists for old and new workflows at this point.
> > A separate and new test case should prevent this behavior in the future.
> 
> Are you suggesting to do the DB save on the workflow before saving the files?

I don't think it makes a difference if we switch the save of files and DB around. In any case the first operation could work and the second fail, so we'd end up in an inconsistency anyway. As of right now the PR checks duplication of titles right at the beginning. Validate first, then save. I think that's fair. I would add an explicit test though:

1. Two Workflows A and B
2. Attempt to rename A to B (should not work)
3. Files from A should have kept the original filenames

If that scenario is tested for, the order doesn't matter because we prevented a potential file loss.

The main pain point here is that filenames are derived from the title in the first place. If they were sorta random on disk (random name, hash, id or whatever) we would not have to make such an elaborate effort of checking duplicates and existence of files. A new or old workflow could have a duplicate name and they wouldn't be coupled by this filename logic.

Might be a future task for improvement.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
Erikmitk: I'm unemotional about that. Renaming workflows is not an every day activity so it probably doesn't matter. 😅  What do you think, @solth?

Erikmitk (approved):
I'd add an additional test as I commented but the overall changes are good to go for me.

Kitodo/src/main/java/org/kitodo/production/forms/WorkflowEditView.java
solth: @Erikmitk I get your point, but since all exception types in the `catch` block are currently handled the same way (e.g. show the exception message in the global messages component on the workflow edit page) I don't see how separating the catch block of the `DAOException` from the others would improve the situation. As @BartChris mentioned, `DAOException`s can also be caused by something else, so dedicating a catch block to them and then assuming that they always represent "duplicate workflow titles" is probably not correct.

At any rate, when I just tested the changes in this PR, the duplicate check still worked as intended:

Image

And since the code segment in question is not part of the changes of this PR I would suggest to deal with with this separately, if required.

Kitodo/src/main/resources/messages/errors_de.properties
solth, line 1: @BartChris could you provide a Spanish translation as well, even if it's just generated with Google Translate?

solth:
@BartChris and please update with current `main` branch as well!

BartChris:
Will do on monday 👍

PR #6977, opened 2026-04-13
Update MariaDB java client to 3.5.8

henning-gerhardt:
Updating two year old version to latest version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
         <spotbugs-maven-plugin.version>4.9.8.2</spotbugs-maven-plugin.version>
         <spotbugs-annotations.version>4.9.8</spotbugs-annotations.version>
         <pitest.version>1.4.10</pitest.version>
-        <mariadb-java-client.version>2.7.12</mariadb-java-client.version>
+        <mariadb-java-client.version>3.5.8</mariadb-java-client.version>
         <maven-javadoc-plugin.version>3.3.0</maven-javadoc-plugin.version>
     </properties>
 


BartChris (approved):
Works for me!

BartChris:
I suppose this is not documented anywhere. It might be that the mariadb driver in older versions worked with the MySQLDialect. This definitely does not work anymore, so Hibernate config has to be.

```xml
<!-- SQL - Settings -->
<property name="hibernate.dialect">org.hibernate.dialect.MariaDBDialect</property>
<property name="hibernate.dialect.storage_engine">innodb</property>
<property name="hibernate.connection.driver_class">org.mariadb.jdbc.Driver</property>

<property name="hibernate.connection.url">
            jdbc:mariadb://localhost:3306/kitodo
</property>
```

henning-gerhardt:
@BartChris : 
- as we use MySQL as default DBMS I did not change the hibernate configuration. On my local and SLUB systems the hibernate configuration is adjusted to the MariaDB use. I had this documented a few years ago already but this documentation is lost or my memories are wrong.
- the `hibernate.dialect` configuration entry can be omitted with client version 3.5.8 - you got a warning message on application startup that this line is not needed.

BartChris:
@henning-gerhardt 

I just noticed that if i switch from

```xml
<property name="hibernate.connection.driver_class">com.mysql.cj.jdbc.Driver</property>
<property name="hibernate.connection.url">
    jdbc:mysql://localhost:3306/kitodo
</property>
```

to:
```xml
<property name="hibernate.connection.driver_class">org.mariadb.jdbc.Driver</property>
<property name="hibernate.connection.url">
    jdbc:mariadb://localhost:3306/kitodo
</property>
```

mass indexing does not work anymore. Can you confirm that?

```
ERROR] 2026-04-21 17:33:11.593 [Hibernate Search - Mass indexing - Process - ID loading - 0] LogFailureHandler - HSEARCH000058: Exception occurred org.hibernate.exception.SQLGrammarException: JDBC exception executing SQL [select p1_0.id from process p1_0] [(conn=984) invalid fetch size] [n/a]
Failing operation:
Fetching identifiers of entities to index for entity 'Process' during mass indexing
org.hibernate.exception.SQLGrammarException: JDBC exception executing SQL [select p1_0.id from process p1_0] [(conn=984) invalid fetch size] [n/a]
```

I think that the reason is the usage of 

`massIndexer.idFetchSize(Integer.MIN_VALUE).batchSizeToLoadObjects(1000);`

https://github.com/kitodo/kitodo-production/blob/1dc805cf7b795a498e8d1004b43a6ee281841594/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java#L123 

which is a a MySQL-specific "hack". I suppose this breaks with MariaDB.

BartChris:
I think we have to evaluate the used driver here:

```java
String driverName = ormSession
              .doReturningWork(c -> c.getMetaData().getDriverName())
              .toLowerCase();
if (driverName.contains("mysql") && !driverName.contains("mariadb")) {
    // MySQL → enable streaming
    massIndexer.idFetchSize(Integer.MIN_VALUE);
} else {
    // MariaDB, PostgreSQL, etc.
    massIndexer.idFetchSize(500);
}
```

henning-gerhardt:
@BartChris : I will look into this and reporting back.

henning-gerhardt:
@BartChris : I can confirm the issue and the changed value `Integer.MIN_VALUE` to `500` on method calling `idFetchSize(...)` did solve the issue.

PR #6980, opened 2026-04-13
Update JUnit to 6.0.3

henning-gerhardt:
Update from JUnit 5 to JUnit 6 is simple as major change was the JVM 8 to JVM 17 change. See https://github.com/junit-team/junit-framework/wiki/Upgrading-to-JUnit-6.0


thomaslow (approved):
Works for me!

Issue #6883, opened 2026-02-09
Selecting a catalog on a larger mass import list throws an Exception


BartChris:
**Describe the bug**
When uploading a bigger CSV file (here 493 entries) and selecting a catalog for the import afterwards:

<img width="410" height="781" alt="Image" src="https://github.com/user-attachments/assets/fa01f418-5891-46e2-af47-140dbed4b337" />

an exception is thrown:

<img width="1294" height="589" alt="Image" src="https://github.com/user-attachments/assets/0d3461da-18b6-42da-aa17-37290e739862" />

**To Reproduce**
1. Go to Massimport Page
2. Upload a big csv file (500 entries) with a column of type 'recordIdentifier'
3. Try to select a import configuration

**Release**
Kitodo 3.9

Strack trace

```
Es ist ein Fehler aufgetreten:
Date/time: 2026-02-09 19:16:35
User agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36
User IP: 127.0.0.1
Request URI: /kitodo/pages/massImport.jsf
Ajax request: Yes
Status code: 500
Exception type: class java.lang.NullPointerException
Exception message:
Exception UUID: d54dcaea-a339-4853-a4bc-8b5a880f7801
Stack trace:
java.lang.NullPointerException
	at org.kitodo.production.forms.massimport.MassImportForm.getRulesetTitle(MassImportForm.java:524)
	at jdk.internal.reflect.GeneratedMethodAccessor2476.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at javax.el.BeanELResolver.getValue(BeanELResolver.java:95)
	at javax.el.CompositeELResolver.getValue(CompositeELResolver.java:107)
	at org.apache.myfaces.el.unified.resolver.FacesCompositeELResolver.getValue(FacesCompositeELResolver.java:179)
	at org.apache.el.parser.AstValue.getValue(AstValue.java:163)
	at org.apache.el.parser.AstDeferredExpression.getValue(AstDeferredExpression.java:42)
	at org.apache.el.parser.AstCompositeExpression.getValue(AstCompositeExpression.java:47)
[...]

PR #6983, opened 2026-04-13
Refactor mass import page to use base list view instead of base edit view

### LINKED ISSUE #6805: Save button is not activated after change is made in edit mode

solth:
Fixes #6805 by refactoring the mass import page so that it is not based upon the `baseEditView.xhtml` template anymore and thus can continue to use the required `enctype="multipart/form-data"`

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml
@@ -17,181 +17,188 @@
         xmlns:h="http://xmlns.jcp.org/jsf/html"
         xmlns:p="http://primefaces.org/ui">
 
-    <p:panelGrid layout="grid"
-                 id="fileUploadButtonWrapper"
-                 columns="2">
-        <p:panel>
-            <h:panelGroup style="display: inline-block;">
-                <p:outputLabel for="csvFileUpload"
-                               value="#{msgs['massImport.csvFileUpload']}"/>
-                <p:fileUpload id="csvFileUpload"
-                              value="#{MassImportForm.file}"
-                              listener="#{MassImportForm.handleFileUpload}"
-                              allowTypes="/(\.|\/)(csv)$/"
-                              sizeLimit="100000"
-                              skinSimple="true"
-                              update="editForm"
-                              styleClass="input"
-                              label="#{msgs.select}"
-                              uploadLabel="#{msgs.upload}"
-                              cancelLabel="#{msgs.cancel}"
-                              chooseIcon="fa fa-plus fa-lg"
-                              uploadIcon="fa fa-upload fa-lg"
-                              cancelIcon="fa fa-times fa-lg"/>
-            </h:panelGroup>
-            <h:panelGroup id="skipEmptyColumnsWrapper">
-                <p:outputLabel for="skipEmptyColumns"
-                               value="#{msgs['massImport.skipEmptyColumns']}"/>
-                <p:selectBooleanCheckbox id="skipEmptyColumns"
-                                         disabled="#{empty MassImportForm.records}"
-                                         value="#{MassImportForm.skipEmptyColumns}"
-                                         styleClass="switch input">
-                    <p:ajax event="change"
-                            listener="#{MassImportForm.parseCsvLines()}"
-                            update="editForm:recordsTableWrapper"/>
-                </p:selectBooleanCheckbox>
-            </h:panelGroup>
-        </p:panel>
+    <h:form id="fileUploadForm"
+            enctype="multipart/form-data">
+        <p:panelGrid layout="grid"
+                     id="fileUploadButtonWrapper"
+                     columns="2">
+            <p:panel>
+                <h:panelGroup style="display: inline-block;">
+                    <p:outputLabel for="csvFileUpload"
+                                   value="#{msgs['massImport.csvFileUpload']}"/>
+                    <p:fileUpload id="csvFileUpload"
+                                  value="#{MassImportForm.file}"
+                                  listener="#{MassImportForm.handleFileUpload}"
+                                  allowTypes="/(\.|\/)(csv)$/"
+                                  sizeLimit="100000"
+                                  skinSimple="true"
+                                  update="recordsForm fileUploadForm"
+                                  styleClass="input"
+                                  label="#{msgs.select}"
+                                  uploadLabel="#{msgs.upload}"
+                                  cancelLabel="#{msgs.cancel}"
+                                  chooseIcon="fa fa-plus fa-lg"
+                                  uploadIcon="fa fa-upload fa-lg"
+                                  cancelIcon="fa fa-times fa-lg"/>
 
-        <p:panel>
-            <!-- separator option to parse metadata groups inside individual CSV cells;
-                 IMPORTANT: does not support nested groups, yet! -->
-            <h:panelGroup id="separatorCharacterWrapper"
-                          layout="block"
-                          rendered="#{MassImportForm.file ne null}">
-                <h:panelGroup styleClass="separator-selection">
-                    <p:outputLabel for="csvSeparator"
-                                   value="#{msgs['massImport.csvSeparator']}"/>
-                    <p:selectOneMenu id="csvSeparator"
-                                     styleClass="input"
-                                     maxlength="1"
-                                     value="#{MassImportForm.csvSeparator}">
-                        <f:selectItems value="#{MassImportForm.csvSeparatorCharacters}"/>
+                </h:panelGroup>
+                <h:panelGroup id="skipEmptyColumnsWrapper">
+                    <p:outputLabel for="skipEmptyColumns"
+                                   value="#{msgs['massImport.skipEmptyColumns']}"/>
+                    <p:selectBooleanCheckbox id="skipEmptyColumns"
+                                             disabled="#{empty MassImportForm.records}"
+                                             value="#{MassImportForm.skipEmptyColumns}"
+                                             styleClass="switch input">
                         <p:ajax event="change"
                                 listener="#{MassImportForm.parseCsvLines()}"
-                                update="editForm:recordsTableWrapper editForm:catalogSelectionWrapper editForm:groupEntrySeparator"/>
-                    </p:selectOneMenu>
+                                update="recordsForm:recordsTableWrapper"/>
+                    </p:selectBooleanCheckbox>
                 </h:panelGroup>
+            </p:panel>
 
-                <h:panelGroup styleClass="separator-selection">
-                    <p:outputLabel for="groupEntrySeparator"
-                                   value="#{msgs['massImport.metadataGroupSeparator']}"/>
-                    <p:selectOneMenu id="groupEntrySeparator"
-                                     styleClass="input"
-                                     maxlength="1"
-                                     autoWidth="false"
-                                     value="#{MassImportForm.metadataGroupEntrySeparator}">
-                        <f:selectItem noSelectionOption="true"
-                                      itemValue="#{null}"
-                                      itemLabel="-- Select metadata group entry separator --"
-                                      itemDescription="Please select a separator character to parse CSV cell contents into Kitodo metadata groups"/>
-                        <f:selectItems value="#{MassImportForm.metadataGroupEntrySeparators}"/>
-                        <p:ajax event="change"
-                                update="editForm:recordsTableWrapper editForm:catalogSelectionWrapper"/>
-                    </p:selectOneMenu>
+            <p:panel>
+                <!-- separator option to parse metadata groups inside individual CSV cells;
+                     IMPORTANT: does not support nested groups, yet! -->
+                <h:panelGroup id="separatorCharacterWrapper"
+                              layout="block"
+                              rendered="#{MassImportForm.file ne null}">
+                    <h:panelGroup styleClass="separator-selection">
+                        <p:outputLabel for="csvSeparator"
+                                       value="#{msgs['massImport.csvSeparator']}"/>
+                        <p:selectOneMenu id="csvSeparator"
+                                         styleClass="input"
+                                         maxlength="1"
+                                         value="#{MassImportForm.csvSeparator}">
+                            <f:selectItems value="#{MassImportForm.csvSeparatorCharacters}"/>
+                            <p:ajax event="change"
+                                    listener="#{MassImportForm.parseCsvLines()}"
+                                    update="recordsForm:recordsTableWrapper recordsForm:catalogSelectionWrapper fileUploadForm:groupEntrySeparator"/>
+                        </p:selectOneMenu>
+                    </h:panelGroup>
+
+                    <h:panelGroup styleClass="separator-selection">
+                        <p:outputLabel for="groupEntrySeparator"
+                                       value="#{msgs['massImport.metadataGroupSeparator']}"/>
+                        <p:selectOneMenu id="groupEntrySeparator"
+                                         styleClass="input"
+                                         maxlength="1"
+                                         autoWidth="false"
+                                         value="#{MassImportForm.metadataGroupEntrySeparator}">
+                            <f:selectItem noSelectionOption="true"
+                                          itemValue="#{null}"
+                                          itemLabel="-- Select metadata group entry separator --"
+                                          itemDescription="Please select a separator character to parse CSV cell contents into Kitodo metadata groups"/>
+                            <f:selectItems value="#{MassImportForm.metadataGroupEntrySeparators}"/>
+                            <p:ajax event="change"
+                                    update="recordsForm:recordsTableWrapper recordsForm:catalogSelectionWrapper"/>
+                        </p:selectOneMenu>
+                    </h:panelGroup>
                 </h:panelGroup>
-            </h:panelGroup>
-        </p:panel>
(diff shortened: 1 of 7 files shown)


BartChris:
also to check: https://github.com/kitodo/kitodo-production/issues/6883

solth:
> also to check: #6883

@BartChris would you mind reviewing this pull request, with that particular issue in mind, as you worked on the mass import quite a lot yourself?

Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml
```diff
@@ -17,181 +17,188 @@
...
+                     columns="2">
+            <p:panel>
+                <h:panelGroup style="display: inline-block;">
+                    <p:outputLabel for="csvFileUpload"
+                                   value="#{msgs['massImport.csvFileUpload']}"/>
+                    <p:fileUpload id="csvFileUpload"
+                                  value="#{MassImportForm.file}"
+                                  listener="#{MassImportForm.handleFileUpload}"
+                                  allowTypes="/(\.|\/)(csv)$/"
+                                  sizeLimit="100000"
```
BartChris: The `sizeLimit` was introduced in an earlier PR (https://github.com/kitodo/kitodo-production/commit/dd7dd28dca179e83bd809be8501d0e6564f9a655#diff-7bacb843a9271945e9d61dfbf7df11f5abb077fc36a6e50e8ecc6b2331a0ee35), but i just stumbled over it. 100kb seems way too small, i already failed with uploading a 101 kb large CSV file. Or do we want to impose a hard limit here?

BartChris:
When the uploaded csv file gets bigger i run into problems:

I uploaded a csv file with 453 rows:

image

When removing one of the red marked pseudo columns or when trying to switch the catalog the table is emptied:

image

image

Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml
```diff
@@ -17,181 +17,188 @@
...
+                    <f:facet name="header">
+                        <h:outputText value="#{MassImportForm.getColumnHeader(columnIndex)}"
+                                      title="#{MassImportForm.getColumnHeader(columnIndex)}"
+                                      styleClass="#{MassImportForm.getFunctionalMetadataStyleClass(columnIndex)}"/>
+                        <p:commandButton id="removeMetadataColumn"
+                                         styleClass="secondary remove-metadata-column"
+                                         action="#{MassImportForm.removeMetadata(columnIndex)}"
+                                         title="#{msgs.delete}"
+                                         icon="fa fa-trash-o"
+                                         update="recordsForm:recordsTableWrapper"/>
```
BartChris, line 136:
```suggestion
<p:commandButton id="removeMetadataColumn"
                                         styleClass="secondary remove-metadata-column"
                                         action="#{MassImportForm.removeMetadata(columnIndex)}"
                                         process="@this"
                                         partialSubmit="true"
                                         title="#{msgs.delete}"
                                         icon="fa fa-trash-o"
                                         update="recordsForm:recordsTableWrapper"/>
```

This seems to fix the issue when removing columns from a tables based on a big csv file (450 rows 5 colums)

Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml
```diff
@@ -17,181 +17,188 @@
...
+                                     converter="#{importConfigurationConverter}"
+                                     value="#{MassImportForm.importConfigurationId}">
+                        <f:selectItem itemValue="#{null}"
+                                      itemLabel="-- #{msgs.selectCatalog} --"
+                                      noSelectionOption="true"/>
+                        <f:selectItems value="#{CreateProcessForm.catalogImportDialog.importConfigurations}"
+                                       var="configuration"
+                                       itemLabel="#{configuration.title}"
+                                       itemValue="#{configuration}"/>
+                        <p:ajax update="recordsForm"/>
```
BartChris:
```suggestion
<p:ajax process="@this"
      partialSubmit="true"
      update="recordsForm"/>
```

This seems to fix the issue when changing the import catlog for a table based on a big csv file (450 rows 5 colums)

Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml
```diff
@@ -17,181 +17,188 @@
...
+                                     update="recordsForm:recordsTableWrapper massImportButtonForm:importButton  fileUploadForm:skipEmptyColumns"/>
+                </p:column>
+            </p:dataTable>
+            <div>
+                <p:commandButton id="addCsvRecord"
+                                 action="#{MassImportForm.addRecord}"
+                                 title="#{msgs['massImport.addRow']}"
+                                 value="#{msgs['massImport.addRow']}"
+                                 icon="fa fa-plus"
+                                 iconPos="right"
```
BartChris, line 182: Similar problem to the others, what works for me is:

```xml
<p:commandButton id="addCsvRecord"
                 action="#{MassImportForm.addRecord}"
                 process="@this"
                 partialSubmit="true"
                 title="#{msgs['massImport.addRow']}"
                 value="#{msgs['massImport.addRow']}"
                 icon="fa fa-plus"
                 iconPos="right"
                 styleClass="secondary"
                 update="recordsForm:recordsTableWrapper 
                         massImportButtonForm:importButton 
                         fileUploadForm:skipEmptyColumns"/>
``

**Comment by BartChris** [comment]
When uploading a bigger csv file the "Add metadata" dialog is just an empty dialog:

image

vs

image

Edit: OK the fix for this button is the same:

```xml
   <p:commandButton id="addMetadataColumn"
                                         title="#{msgs['dataEditor.addMetadata.newMetadata']}"
                                         disabled="#{MassImportForm.records.size() eq 0}"
                                         process="@this"
                                         partialSubmit="true"
                                         style="margin: 3px"
                                         icon="fa fa-plus"
                                         styleClass="secondary"
                                         action="#{MassImportForm.addMetadataDialog.prepareMetadataTypes()}"
                                         update="addMetadataDialog"
                                         oncomplete="PF('addMetadataDialog').show();"/>
```

BartChris:
Maybe @henning-gerhardt can also check if his issue with the role settings is fixed, because the error always comes down to the number of elements edited.

BartChris:
I just noticed that the problems i encountered also appear in 3.9.0.

henning-gerhardt:
As we at SLUB did not use the mass import feature and all changes are only focused on mass import I don't think that this pull request is fixing #6805 in first place. The comments in #6805 are switching fast from the original issue to an other maybe related issue and so I don't think that the main issue is fixed with this changes here.

henning-gerhardt:
I tried it again and #6805 is already fixed by some changes between December 2025 and April 2026. So this pull request is not really related to the original issue of #6805 but maybe still to the comments in #6805.

solth:
> I tried it again and #6805 is already fixed by some changes between December 2025 and April 2026. So this pull request is not really related to the original issue of #6805 but maybe still to the comments in #6805.

Yes, exactly. The linked issue was already fixed by #6810, I think, and the necessity to refactor the mass import form resulted from the changes in that pull request.

solth:
@BartChris thanks a lot for your review and suggestions, which fixed the problems you reported. (I commited them manually instead of applying the suggestions via GitHub because some indentations seemed to be lost in the suggestions).

I also increased the `sizeLimit` to a hopefully more reasonble value (and rebased against the current `main` branch). 

Would you mind checking the pull request again?

Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml
```diff
@@ -17,181 +17,196 @@
...
                                      styleClass="secondary"
-                                     action="#{MassImportForm.addMetadataDialog.prepareMetadataTypes()}"
-                                     update="addMetadataDialog"
-                                     oncomplete="PF('addMetadataDialog').show();"/>
-                </f:facet>
-                <p:commandButton id="removeLine"
-                                 action="#{MassImportForm.removeLine(record)}"
+                                     title="#{msgs.delete}"
+                                     icon="fa fa-trash-o"
+                                     update="recordsForm:recordsTableWrapper massImportButtonForm:importButton  fileUploadForm:skipEmptyColumns"/>
```
BartChris:
```xml
<p:commandButton id="removeLine"
                                     action="#{MassImportForm.removeLine(record)}"
                                     styleClass="secondary"
                                     title="#{msgs.delete}"
                                     icon="fa fa-trash-o"
                                     process="@this"
                                     partialSubmit="true"
                                     update="recordsForm:recordsTableWrapper massImportButtonForm:importButton  fileUploadForm:skipEmptyColumns"/>
```

The delete button of a row needs the same fix.

BartChris:
I noticed that the MassImport button to trigger the import is never activated. (Maybe because of the changes i suggested)

image

I suppose we have to update it explicitely whenever we execute an action. I tested it with the `catalogueSelect` box and added `massImportButtonForm` to the list of to be updated components. Now the button is enabled after switching the catalog.

```xml
  <p:selectOneMenu id="catalogueSelect"
                                     disabled="#{not MassImportForm.firstColumnContainsRecordsIdentifier()}"
                                     required="#{not empty param['massImportButtonForm:importButton']}"
                                     converter="#{importConfigurationConverter}"
                                     value="#{MassImportForm.importConfigurationId}">
                        <f:selectItem itemValue="#{null}"
                                      itemLabel="-- #{msgs.selectCatalog} --"
                                      noSelectionOption="true"/>
                        <f:selectItems value="#{CreateProcessForm.catalogImportDialog.importConfigurations}"
                                       var="configuration"
                                       itemLabel="#{configuration.title}"
                                       itemValue="#{configuration}"/>
                        <p:ajax process="@this"
                                partialSubmit="true"
                                update="recordsForm massImportButtonForm"/>
                    </p:selectOneMenu>
```

BartChris:
This is not directly related to the PR and can maybe be fixed later, but i am wondering if we should spam the screen with tons of messages for every imported row:

image

solth:
> This is not directly related to the PR and can maybe be fixed later, but i am wondering if we should spam the screen with tons of messages for every imported row:

Yes, that's suboptimal, and yes, we should address that in a separate pull request.

solth:
@BartChris I get a `FacesException` when trying to edit cell contents from slightly larger CSV files (about 100 rows), which I cannot explain, yet. Maybe it has something to do with lazy loading additional entries, but I don't now for sure.

BartChris:
> @BartChris I get a `FacesException` when trying to edit cell contents from slightly larger CSV files (about 100 rows), which I cannot explain, yet. Maybe it has something to do with lazy loading additional entries, but I don't now for sure.

I do not get an error here (maybe you fixed that) but i also cannot edit and save the values (it keeps the old value). I would have to check if this ever worked on large tables.

solth:
> I do not get an error here (maybe you fixed that) but i also cannot edit and save the values (it keeps the old value). I would have to check if this ever worked on large tables.

Yes, I was able to fix the problem with larger files by using the `enctype="multipart/form-data"` in the `recordsForm`, but editing CSV cells is currently not possible.

I tried updating the current row in 

https://github.com/solth/kitodo-production/blob/c078685857381483bfff1233f0d19009c8d177da/Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/massImportTab.xhtml#L148

but that did not work, either.

BartChris:
> > I do not get an error here (maybe you fixed that) but i also cannot edit and save the values (it keeps the old value). I would have to check if this ever worked on large tables.
> 
> Yes, I was able to fix the problem with larger files by using the `enctype="multipart/form-data"` in the `recordsForm`, but editing CSV cells is currently not possible.
> 

Did this work in 3.8? I would say this is a Mass Import and values should be provided by the CSV in general, so i do not see it as a critical problem, if values cannot be edited directly in the UI.

solth:
> Did this work in 3.8? I would say this is a Mass Import and values should be provided by the CSV in general, so i do not see it as a critical problem, if values cannot be edited directly in the UI.

Yes, it did work in 3.8. And in 3.9 it works as well. In `main`, the mass import is pretty broken right now, though, to the point where it is not usable at all. The pull request does improve things and restores usabilty in general, but not being able to edit uploaded data (in contrast editing manually entered rows without a CSV upload seems to work fine) still is quite a regression when compared to previous releases, in my opinion.

solth:
@BartChris as a temporary step I deactivated the option to edit datatable cells after uploading a CSV file, so that this feature is not offered in a broken state. 

I would suggest to go ahead and merge this pull request if you don't find any other problems with it, to at least restore the basic functionality of the mass import again. Once we get cell editing working again we can reactivate it afterwards in a separate pull request. What do you think?

BartChris (approved):
Yes, i agree. The rest of the functionality works for me

BartChris:
Hmm, right now i got a problem when starting the import i have not seen before although i suppose this is not connected to the PR:

image

solth:
> Hmm, right now i got a problem when starting the import i have not seen before although i suppose this is not connected to the PR:

We should probably make sure it isn't. Perhaps you could test with the same case in the `main` branch?

I just noticed that the "Mass import" button still isn't activated immediately after uploading a CSV files that already contains all data (thus metadata of type `documentType` in the first column) and doesn't require catalog queries.

BartChris:
I will check. Right now i suspect that this is again about external schema files i cannot resolve...

BartChris:
Ok, after commenting out

```java
validationService.validateMappingFiles(importConfiguration.getMappingFiles());
```

and deactivating validation here

```java
private String importProcessAndReturnParentID(String recordId, LinkedList<TempProcess> allProcesses,
                                                  ImportConfiguration importConfiguration, int projectID,
                                                  int templateID, boolean isParentInRecord, String parentIdMetadata,
                                                  boolean validateAgainstXmlSchema)
```

it works. So this is not directly connected and about some validation issue.

solth:
It should be possible to deactivate XML validation of external data for an import configuration in general, so you don't have to tamper with the code to achieve that:

Bildschirmfoto 2026-04-17 um 17 07 58

BartChris:
Same error on main, so definitely not connected!

BartChris:
> It should be possible to deactivate XML validation of external data for an import configuration in general, so you don't have to tamper with the code to achieve that:

 In `importProcessForMassImport` the validation flag seems hardcoded and not influenced by the setting.

Edit:It seems to be additionally checked in `convertDataRecordToInternal`, sorry. But `validateExternal` is always true.

```java
// validate external record against corresponding XML metadata schema(ta)
if (validateExternal && importConfiguration.getValidateExternalData()) {
           validationService.validateExternalRecord(xmlContent, importConfiguration, identifier);
}
```

But it seems like 

```java
// validate mapping files against xslt schema
validationService.validateMappingFiles(importConfiguration.getMappingFiles()); 
```

is always called. In my case i suppose the mapping file validation let the process fail.

solth:
> In `importProcessForMassImport` the validation flag seems hardcoded and not influenced by the setting.

Good that we found this out so it can be fixed as well!

solth:
> > In `importProcessForMassImport` the validation flag seems hardcoded and not influenced by the setting.
> 
> Good that we found this out so it can be fixed as well!

Interestingly, the setting in the import configuration actually seemed to work for me. 

When schema validation was **activated** in the configuration, I got the following result in mass import:

Bildschirmfoto 2026-04-17 um 17 40 59


After **deactivating** the schema validation in the configuration, though, the import of the same IDs from the same catalog interface was succesful:

Bildschirmfoto 2026-04-17 um 17 43 44

It seems the static parameter `true` in https://github.com/solth/kitodo-production/blob/e1fdd90482a14ce073dac77044db74c8a9e1884c/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L1577 does not actually control whether schema validation is applied during mass import or not 🤔 

Maybe your import didn't really fail because of _schema_ invalid metadata, but because the imported XML file or the import mapping XSLT did not contain valid XML to begin with?

BartChris:
> Maybe your import didn't really fail because of _schema_ invalid metadata, but because the imported XML file or the import mapping XSLT did not contain valid XML to begin with?

I have to check deeper, but my investigations right now seem to indicate that is not the validation of the catalog entries which fails for me, but the validation of the mapping files. And the latter might come down to the well known problem of external schema files, which are maybe blocked for me. 

https://github.com/solth/kitodo-production/blob/e1fdd90482a14ce073dac77044db74c8a9e1884c/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L962

https://github.com/kitodo/kitodo-production/blob/e1fdd90482a14ce073dac77044db74c8a9e1884c/Kitodo/src/main/resources/schemata/xslt20.xsd#L57-L64

Edit: OK, i get the same error in "normal" catalog import... will make an issue next week.

PR #6985, opened 2026-04-14
Export only selected processes

### LINKED ISSUE #6925: Actions from dropdown are executed for all processes instead of selected ones
**Describe the bug**
Not all actions are executed only for the selected processes. Some actions are executed for all processes instead of only the selected ones.

**To Reproduce**
1. Go to Processes
2. Select multiple specific processes
3. Open the Actions dropdown menu
4. Execute an action (e.g. create PDF, etc.)
5. Observe that the action is executed for all processes instead of only the selected ones.

**Expected behavior**
The selected action should only be executed for the processes that were explicitly selected via the checkbox.

**Screenshots**

**Release**
Kitodo.Production Version: 4.0.0-SNAPSHOT

BartChris:
Fixes: https://github.com/kitodo/kitodo-production/issues/6925

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/process/ProcessListBaseView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/process/ProcessListBaseView.java
@@ -141,6 +141,13 @@ public class ProcessListBaseView extends BaseListView {
         return stopwatch.stop(selectedProcesses);
     }
 
+    private List<Integer> getSelectedProcessIds() {
+        return selectedProcesses.stream()
+                .map(Process::getId)
+                .filter(Objects::nonNull)
+                .toList();
+    }
+
     private LazyProcessModel getLazyProcessModel() {
         return (LazyProcessModel) this.lazyBeanModel;
     }
@@ -328,15 +335,20 @@ public class ProcessListBaseView extends BaseListView {
     }
 
     /**
-     * Generates the current search result as an Excel file.
+     * Generates and downloads the current search result as an Excel file using the active filter,
+     * visibility flags (closed/inactive), and selection state (all selected, selected IDs, excluded IDs).
      */
     public void generateExcel() {
         Stopwatch stopwatch = new Stopwatch(this, "generateExcel");
         try {
-            ServiceManager.getProcessService().generateExcel(
+            ServiceManager.getProcessService().export(
                     this.filter,
                     this.isShowClosedProcesses(),
-                    this.isShowInactiveProjects()
+                    this.isShowInactiveProjects(),
+                    org.kitodo.production.enums.ExportFormat.EXCEL,
+                    this.allSelected,
+                    getSelectedProcessIds(),
+                    this.excludedProcessIds
             );
         } catch (IOException | DocumentException e) {
             Helper.setErrorMessage(ERROR_CREATING,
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -1145,40 +1145,14 @@ public void downloadDocket(List<Process> processes) throws IOException {
...
-    }
-
     /**
      * Generate result set.
      *
      * @param filter
      *            for generating search results
      */
-    public void export(String filter, boolean showClosedProcesses, boolean showInactiveProjects, ExportFormat format)
+    public void export(String filter, boolean showClosedProcesses, boolean showInactiveProjects, ExportFormat format,
```
henning-gerhardt, line 1168: Please adjust the JavaDoc as new and even "old" parameters are not listed.

Kitodo/src/main/java/org/kitodo/production/forms/process/ProcessListBaseView.java
```diff
@@ -359,15 +377,24 @@ public void generateCsv() {
...
+                    this.isShowClosedProcesses(),
+                    this.isShowInactiveProjects(),
+                    ExportFormat.PDF,
+                    this.allSelected,
+                    getSelectedProcessIds(),
+                    this.excludedProcessIds
+            );
         } catch (IOException | DocumentException e) {
-            Helper.setErrorMessage(ERROR_CREATING, new Object[] {Helper.getTranslation("resultPDF") }, logger, e);
+            Helper.setErrorMessage(ERROR_CREATING,
```
henning-gerhardt: Not needed line wrapping.

henning-gerhardt (changes requested):
Only two small things in the code should be adjusted.

General question: if no process is selected and you are generate an excel, csv or pdf file than "empty" excel / csv / pdf file with only the header column is created. Is this expected? Other actions do nothing if not at least one process is selected.

BartChris:
> General question: if no process is selected and you are generate an excel, csv or pdf file than "empty" excel / csv / pdf file with only the header column is created. Is this expected? Other actions do nothing if not at least one process is selected.

Good point. I changed that. Now nothing gets executed when nothing is selected.

henning-gerhardt (approved):

solth:
> Fixes: #6925

Are we certain that the problem described in #6925 only affects the _export_ actions, and thus the issue is really resolved by this pull request? It would be quite unfortunate if actions like calling a Kitodo-Script (potentially something like `deleteProcess`) is performend on more than just the selected processes...

BartChris:
> > Fixes: #6925
> 
> Are we certain that the problem described in #6925 only affects the _export_ actions, and thus that the issue is really resolved by this pull request? It would be quite unfortunate if actions like calling a Kitodo-Script (potentially something like `deleteProcess`) is performend on more than just the selected processes...

I do not know honestly. But i suppose this would then also apply to 3.8 and 3.9, or? The current behavior probably has been the same since quite some time, i think.

solth:
> I do not know honestly. But i suppose this would then also apply to 3.8 and 3.9, or? The current behavior probably has been the same since quite some time, i think.

Yes, but I think it's better to make sure of that instead of just guessing.

I just checked most of the other list functions (especially the critical ones like executing Kitodo scripts, renaming media or setting import configurations) for selected processes and it seems indeed that they are all not affected by #6925 and are only executed on the processes actually selected, so it looks good.

aetherfaerber:
> > I do not know honestly. But i suppose this would then also apply to 3.8 and 3.9, or? The current behavior probably has been the same since quite some time, i think.
> 
> Yes, but I think it's better to make sure of that instead of just guessing.
> 
> I just checked most of the other list functions (especially the critical ones like executing Kitodo scripts, renaming media or setting import configurations) for selected processes and it seems indeed that they are all not affected by #6925 and are only executed on the processes actually selected, so it looks good.

For the record: we have been experiencing this when setting processes a few steps up and down (which we do a lot for a lot of processes at once as you all surely remember). We also saw Kitodo acting like all result list pages after the first page - where only a few entries were selected - were fully selected.

It has been a while since this has been reported to me. So I will have to check if this has been resolved, did not occur again or if only our users adopted some kind of workaround strategy to avoid the problem.

BartChris:
> We also saw Kitodo acting like all result list pages after the first page - where only a few entries were selected - were fully selected.

I tested this with current 4.0.x (basically the last released version) and for me this works correct. But it would be good if you also check.

> which we do a lot for a lot of processes at once as you all surely remember

At least for the "down" option i made some perfomance improvments. Those improvements are only in the current main branch and not released yet.
https://github.com/kitodo/kitodo-production/pull/7174

PR #6986, opened 2026-04-14
[3.9] Fix folder editing and deleting

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6946
Includes improvements from: https://github.com/kitodo/kitodo-production/pull/6993


solth:
I still get the following uncaught error when changing the FILEGROUP of an existing folder to a value that is already used in another folder:

Bildschirmfoto 2026-04-16 um 16 26 57


[Your fix in `main`](https://github.com/kitodo/kitodo-production/pull/6946) handled this error better by just displaying the error message (even though the content is a little questionable: "Dateigruppe wird bereits verwendet: .folders") and not the error page including the whole stack trace (which is still documented in the log, though):

Bildschirmfoto 2026-04-16 um 16 30 43

BartChris:
@solth I pushed a fix for that. I suppose it would also be better to integrate that in Main, although it does not lead to an exception there. The switch to ViewScoped leads to behavior differences, so that this does not crash there.

BartChris:
I also backported the changes here:
https://github.com/kitodo/kitodo-production/pull/6993

BartChris:
This will not get backported to 3.9.

PR #6987, opened 2026-04-14
[3.8] Fix folder editing and deleting

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/6946


BartChris:
This will not get backported to 3.8

PR #6988, opened 2026-04-15
Fix two EL variable comments

solth:
This PR fixes the package path of the classes in two EL variable comments, that falsely marked `ProcessListView.createProcessAsChild` as unavailable in `processActionsColumn.xhtml` and the following nine methods in `ProcessListBaseView.java` as "unused" in IntelliJ:
- `createProcessesWithCalendar`
- `createProcessAsChildPossible`
- `downloadToHome`
- `createXML`
- `exportMets`
- `downloadDocket`
- `uploadFromHome`
- `getDeleteProcessDialog`
- `canBeExported`


henning-gerhardt (approved):
I can confirm that IntelliJ IDEA this mentioned methods now marks as used.

PR #6991, opened 2026-04-16
Primefaces 13

### LINKED ISSUE #5926: Allow to move multiple page at once in the Structure tree
Right now it is possible to select multiple pages in the gallery view using shift+click. One can then drag those pages to structures in the Gallery.
The selection however is not reflected in the structure tree.  It is also not possible to select multiple pages in the structure tree and drag them to structural elements in the structure tree.
image

It should be possible to select multiple pages in the structure tree and to drag multiple pages to a structural element. The selections should be reflected in the Gallery. Selecting multiple pages in the gallery should be reflected in the structure tree.

### LINKED ISSUE #6681: Drag'n'drop error in PrimeFaces tree component used in metadata editor
**Describe the bug**
The version of PrimeFaces used in Kitodo.Production was recently updated from 8 to 12. This new release contains a bug in the `Tree` component where the resulting position of a tree node moved via drag'n'drop from an earlier to a later location inside the same structure element is incorrect. 

**Note**: the error does _not_ seem to occur when moving a node in the other direction, e.g. from a later to an earlier position within the same structure element

**To Reproduce**
1. Open a process in the metadata editor
2. Drag a page in a structure element from one position and drop it at a later position in the same element
3. See wrong position after dropping the element

**Expected behavior**
Tree nodes should be inserted at the correct location where they were moved via drag'n'drop actions.

**Screenshots**
See clip in review by @thomaslow in https://github.com/kitodo/kitodo-production/pull/6591#pullrequestreview-3210799423
(Left part of that video shows bug in Kitodo with PrimeFaces 12, right part shows the same bug did not appear in the previously used version 8 of PF)

**Release**
Current `main` branch **after** the merge of #6591

**Additional context**
This bug is **fixed** in PrimeFaces 13. Updating to this version does break quite a few layouts and designs in Kitodo, though, which makes it a non-trivial task.

### LINKED ISSUE #6702: Wrong checkbox styling in task list filtering
**Describe the bug**
After upgrading Primefaces to version 12 the check boxes in the filtering dialog of tasks are not or wrong styled:

Image

In 3.8.4 it looks like this

Image

The check boxes in the filtering list of processes looks like before

**Expected behavior**
Used checkboxes should be styled overall identical in the whole application

**Release**
main (2025-10-06)

 - Browser: Chrome(ium) 141.0.7390.54 & Firefox 143.0.4

solth:
Update PrimeFaces from version 12 to 13 to fix some metadata editor related issues:

Fixes #6681 
Fixes #5926 
Fixes #6702

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryMediaContent.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryMediaContent.java
@@ -263,7 +263,7 @@ public class GalleryMediaContent {
         try {
             InputStream viewData = ServiceManager.getFileService().read(uri);
             return DefaultStreamedContent.builder().stream(() -> viewData).contentType(mimeType)
-                    .name(Paths.get(uri.getPath()).getFileName().toString()).contentLength(viewData.available())
+                    .name(Paths.get(uri.getPath()).getFileName().toString()).contentLength((long) viewData.available())
                     .build();
         } catch (IOException e) {
             logger.catching(e);
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -2762,6 +2762,10 @@ form#metadata div label,
     vertical-align: top;
 }
 
+#editFolderDialog .ui-panelgrid-cell {
+    width: 100%;
+}
+
 #uploadFileDialog {
     max-height: 550px;
 }
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -959,7 +959,8 @@ body #filterInputForm\:filterfield {
     padding-bottom: var(--input-padding);
 }
 
-body #filterOptionsFormWrapper .ui-chkbox.switch .ui-chkbox-icon {
+body #filterOptionsFormWrapper .ui-chkbox.switch .ui-chkbox-icon,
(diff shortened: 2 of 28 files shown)


thomaslow:
I had a general look at the user interface. I couldn't find any major layout issues. I noticed two minor things:

The "gallery" and "detail view" buttons in the metadata editor are a bit misaligned:

Main Branch: 
image

Your PR:
image

Besides, some elements in the "Add folder" dialog on the project edit page are a bit misaligned, too:

Main Branch:
main

Your PR:
pr

I'll test the multi drag & drop in more detail soon, but at first glance, it seems to work. I also didn't look at your code changes yet.

thomaslow (changes requested):
The new structure element dialog of the metadata editor misses a bit of spacing between the first two rows:

Main Branch:
main

Your PR:
pr

Also, I quickly checked the metadata editor in "extended mode". The physical structure tree is a bit cut off and the input boxes for physical metadata have some unnecessary large spacing. However, this is also the case in the current main branch:

image

\
Besides, code changes are alright.

solth:
@thomaslow thanks for the review. I tried to fix the issues you pointed out. Would you mind taking another look?

> Also, I quickly checked the metadata editor in "extended mode". The physical structure tree is a bit cut off and the input boxes for physical metadata have some unnecessary large spacing. However, this is also the case in the current main branch

I would prefer to tackle this problem separately if it's not caused by the update to PrimeFaces 13. There are still many other small and large issues with the styles and layouts in the application, and we can't fix them all in the context of this update. I think it would be better to get this update out of the door and then deal with the remaining issues when we have the ressources to do so.

thomaslow (approved):
The relevant layout problems are fixed now.

PR #6993, opened 2026-04-16
Improve folder editing form

BartChris:
This PR improves upon https://github.com/kitodo/kitodo-production/pull/6946.
When i edit folders and in the same UI interaction session also edit the folder usage (Which folder is used to show thumbnails etc.) sometimes the screen gets stuck on Save. It seems to me as if using "parallel stream" is not safe here and has unwanted effects. After my changes i can edit both the general folder settings as well as their usage.

I also improve the file group handling insofar as when a filegroup of an existing folder is changed to the filegroup of another folder and an error indicates a duplicate file group we do not switch to the rejected file group in the UI.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
@@ -476,7 +476,9 @@ public class ProjectEditView extends BaseEditView {
      * @return modified ArrayList
      */
     public List<SelectItem> getSelectableFolders() {
-        return getFolderList().stream().map(folder -> new SelectItem(folder.getFileGroup(), folder.toString()))
+        return getFolderList().stream()
+                .filter(folder -> Objects.nonNull(folder.getId()))   // exclude transient
+                .map(folder -> new SelectItem(folder.getFileGroup(), folder.toString()))
                 .collect(Collectors.toList());
     }
 
@@ -501,6 +503,7 @@ public class ProjectEditView extends BaseEditView {
     private Map<String, Folder> getFolderMap() {
         return getFolderList().stream()
                 .filter(folder -> StringUtils.isNotBlank(folder.getFileGroup()))
+                .filter(folder -> Objects.nonNull(folder.getId())) //do not include transient folders
                 .collect(Collectors.toMap(
                         Folder::getFileGroup,
                         Function.identity(),
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
@@ -494,7 +494,13 @@ public class ProjectEditView extends BaseEditView {
     }
 
     private Map<String, Folder> getFolderMap() {
-        return getFolderList().parallelStream().collect(Collectors.toMap(Folder::getFileGroup, Function.identity()));
+        return getFolderList().stream()
+                .filter(folder -> StringUtils.isNotBlank(folder.getFileGroup()))
+                .collect(Collectors.toMap(
+                        Folder::getFileGroup,
+                        Function.identity(),
+                        (existing, replacement) -> existing
+                ));
     }
 
 
--- a/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
@@ -184,6 +184,11 @@ public class ProjectEditView extends BaseEditView {
         }
     }
 
+    /**
+     * Sync project folders with workingFolders using fileGroup as key.
+     * Removes folders deleted in UI and replaces existing ones with same fileGroup.
+     * Ensures exactly one folder per fileGroup is persisted.
+     */
     private void syncFoldersToProject() {
         // remove folders whose fileGroup is no longer present
         project.getFolders().removeIf(projectFolder ->
@@ -512,7 +517,7 @@ public class ProjectEditView extends BaseEditView {
                         Folder::getFileGroup,
                         Function.identity(),
                         (existing, replacement) ->
-                                existing.getId() != null ? existing : replacement
+                                Objects.nonNull(existing.getId()) ? existing : replacement
                 ));
     }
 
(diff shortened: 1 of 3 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
```diff
@@ -494,7 +494,9 @@ public boolean hasVideoFolder() {
     }
 
     private Map<String, Folder> getFolderMap() {
-        return getFolderList().parallelStream().collect(Collectors.toMap(Folder::getFileGroup, Function.identity()));
+        return getFolderList().stream()
+                .filter(folder -> StringUtils.isNotBlank(folder.getFileGroup()))
```
BartChris, line 515: This is a change in behavior. I suppose it is safer to only allow folders with a file group as a functional folder. Empty file groups are not forbidden by the UI in general. It might be that they lead to problems downstream.

BartChris:
I further refined this to address the edge case where a folder is deleted and a new one with the same fileGroup is added in the same edit session and afterwards assigned to a usage group. This can create duplicates.

Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
solth: This does behave a little strange right now. Folders _without_ a filegroup are not just still possible, but they are also offered as an option for specific usages like "preview" etc:
Bildschirmfoto 2026-05-06 um 14 42 46

But selecting such a folder without filegroup and then saving the project does not trigger an error, thus does not inform the user about an incorrect configuration, but rather accepts the setting, but then deactivates the "preview" functionality altogether on its own:
Bildschirmfoto 2026-05-06 um 14 43 12

It should either be prevented to set a folder without METS filegroup in the "Folder use" section of the project settings, if we really want to introduce this change, or keep the existing, more liberal behaviour where filegroups are not required for using the contents of a folder for some specific usages.

Since I am not a user I cannot really judge wether this change is going in the right direction, though, so it would be great if @andre-hohmann , @subhhwendt or @henning-gerhardt could comment on this.

BartChris: Thanks for looking into it; i will also check in more detail.

BartChris: I think it might be better to revert this change and return to the more liberal behaviour as my PR aims at making the existing functionality work, so it should probably not change the existing functionality. It can be discussed seperately whether we want to introduce a behaviour change here.

BartChris: I removed this change, which was incomplete as you pointed out.

BartChris:
Still some buggy behavior with the file groups... will check again..

Kitodo/src/main/java/org/kitodo/production/forms/ProjectEditView.java
BartChris: Allowing for empty filegroups leads to bugs in the logic suppose, which i will try to fix to keep the existing functionality.

BartChris: > It should either be prevented to set a folder without METS filegroup in the "Folder use" section of the project settings, if we really want to introduce this change, or keep the existing, more liberal behaviour where filegroups are not required for using the contents of a folder for some specific usages.
> 
> Since I am not a user I cannot really judge wether this change is going in the right direction, though, so it would be great if @andre-hohmann , @subhhwendt or @henning-gerhardt could comment on this.

I want to iterate again on this question brought up by @solth because i fought a little with empty file groups as selectable options. To extend the questions and in light of the fact that we became more strict on validating things:

What is the reason to allow for custom file groups at all (custom here means file groups which are not defined in the kitodo_config.properties)? Is it actually necessary to allow for folders with file group "abc" (which is not defined by the config)? I mostly stumbled over it, because the empty file group "" is also allowed, but only once. (As duplicates are not allowed in general). And it (an empty filegroup value)  leads to problems as the whole code treats the file group as something meaningful.
I know we would produce problems with backwards compatibility if we become more strict here, but a general assesment might help.  @andre-hohmann , @subhhwendt or @henning-gerhardt

henning-gerhardt: So far I know the values in `kitodo_config.properties` are hints or easy selectable values for the used METS file group names but you should be able to add even other names here which are not listed in the `kitodo_config.properties` file as interacting in the UI should be possible without changing the file on the filesystem first. 

That the empty value is allowed (at least on adding new folder entries) must be an error as you can not empty a file group name later (at least in 3.8.4).

BartChris: > That the empty value is allowed (at least on adding new folder entries) must be an error as you can not empty a file group name later (at least in 3.8.4).

Thanks that would simplify things. I would in addition to that consider adding a DB constraint to enforce that filegroups in a project are unique and non NULL.

solth: I am not sure that is what @henning-gerhardt meant. IIRC leaving file groups empty **in the first place** when defining a new folder has always been possible. It just wasn't allowed to _remove_ an already existing file group from an already existing folder.

BartChris: The problem is, when you want to do ANY change to the folder with the empty file group later, as then it is not even possible to make the file group empty. So changing anything in the folder automatically applies a file group.

BartChris: If the possibility for empty filegroups has to be preserved and this file groups has to be selectable as preview folder for example i will probably define some sentinel value (for empty strings) which i use in the UI to make non-selected state differentiable from the selection of an empty string. Because that right now leads to problems as it seems. 
I am not sure if using the empty filegroup as preview leads to downstream problems later.

BartChris: > The problem is, when you want to do ANY change to the folder with the empty file group later, as then it is not even possible to make the file group empty. So changing anything in the folder automatically applies a file group.

OK, in 3.9 this works, so i suppose then this became stricter. Selecting the empty filegroup as preview also leads to some UI-problems in 3.9 though, but appearantly works. I will try what i can to preserve the 3.9 behavior, as backward compatibility is probably critical here.

BartChris:
@solth Maybe you can also test again to see if it still possible to break the form.

matthias-ronge (approved):
I think I understand why it works, and the code looks correct.

solth (approved):
Works for me as well! Code looks good, too. I do think retaining the current behaviour was the right decision. Thanks for the changes, @BartChris !

PR #7000, opened 2026-04-20
Add fullscreen mode to metadata editor

BartChris:
This PR adds a button to switch to "fullscreen" mode in the metadata editor, by hiding the header and the footer. The goal is to use more from the screen when working in the editor. This can be used in conjunction with normal browser fullscreen mode.

Normal view:

image

Fullscreen view:

image

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -39,6 +39,37 @@ body.first-visit {
     animation-name: slidein;
 }
 
+/* Hide header + footer */
+body.fullscreen-mode #portal-header,
+body.fullscreen-mode footer {
+    display: none !important;
+}
+
+/* Remove layout constraints */
+body.fullscreen-mode .flex-container {
+    height: 100vh;
+}
+
+/* Make main take full space */
+body.fullscreen-mode main.flex-item {
+    height: 100vh;
+    width: 100vw;
+    margin: 0 !important;
+    padding: 0 !important;
+}
+
+/* Expand editor to full screen */
+body.fullscreen-mode #metadataEditorContainer {
+    margin: 0 !important;
+    padding: 0 !important;
+    height: 100%;
+}
+
+/* Remove wrapper spacing */
+body.fullscreen-mode #metadataEditorWrapper {
+    height: 100%;
+}
+
 .flex-container {
     height: 100%;
     overflow: auto;
--- a/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
@@ -54,15 +54,12 @@ body.fullscreen-mode .flex-container {
 body.fullscreen-mode main.flex-item {
     height: 100vh;
     width: 100vw;
-    margin: 0 !important;
-    padding: 0 !important;
 }
 
 /* Expand editor to full screen */
 body.fullscreen-mode #metadataEditorContainer {
-    margin: 0 !important;
-    padding: 0 !important;
     height: 100%;
+    margin: 0;
 }
 
 /* Remove wrapper spacing */
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_editor.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_editor.js
@@ -16,6 +16,19 @@
 
 var metadataEditor = metadataEditor || {};
 
+metadataEditor.global = {
+    toggleFullscreenMode() {
+    document.body.classList.toggle('fullscreen-mode');
+    // Force layout recalculation
+    setTimeout(function () {
+        if (typeof window.setHeight === 'function') {
+            window.setHeight();
+        }
+        window.dispatchEvent(new Event('resize'));
+    }, 50);
+    }
+};
+
 metadataEditor.metadataTree = {
 
     /**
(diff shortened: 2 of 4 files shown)


solth:
Judging by the screenshots this doesn't look like much additional "screen real estate" is gained by the fullscreen mode. Maybe screenshots taken at a lower resolution would sell it more convincingly? 😛 

But IIRC this feature is something that @oliver-stoehr may appreciate.

BartChris:
> Judging by the screenshots this doesn't look like much additional "screen real estate" is gained by the fullscreen mode. Maybe screenshots taken in a lower resolution would sell it more convincingly? 😛
> 
> But IIRC this feature is something that @oliver-stoehr may appreciate.

Yeah, the difference is not gigantic ;-), but mostly visible in thumbnail view:

normal
image

full screen:
image

You see one additional row of images. When extending the gallery panel the difference increases.

Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
```diff
@@ -39,6 +39,37 @@ body.first-visit {
...
+body.fullscreen-mode .flex-container {
+    height: 100vh;
+}
+
+/* Make main take full space */
+body.fullscreen-mode main.flex-item {
+    height: 100vh;
+    width: 100vw;
+    margin: 0 !important;
+    padding: 0 !important;
```
oliver-stoehr:
```suggestion
```
It looks like these rule are not necessary.

Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
```diff
@@ -39,6 +39,37 @@ body.first-visit {
...
+}
+
+/* Expand editor to full screen */
+body.fullscreen-mode #metadataEditorContainer {
+    margin: 0 !important;
+    padding: 0 !important;
```
oliver-stoehr:
```suggestion
    margin: 0;
```
I discovered that the margin and padding was already defined twice for the `#metadataEditorContainer`.
You can remove the `style="margin: auto 16px; padding: 0;"` from metadataEditor.xhtml:60 (as the same values are defined in kitodo.css:3063) and then simplify the marked lines with the suggested change.

Unfortunately, I cannot add the metadataEditor.xhtml change to the suggestion by myself.

Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_editor.js
```diff
@@ -16,6 +16,19 @@
...
+    toggleFullscreenMode() {
+    document.body.classList.toggle('fullscreen-mode');
+    // Force layout recalculation
+    setTimeout(function () {
+        if (typeof window.setHeight === 'function') {
+            window.setHeight();
+        }
+        window.dispatchEvent(new Event('resize'));
+    }, 50);
+    }
```
oliver-stoehr:
```suggestion
metadataEditor.toggleFullscreenMode = function () {
    document.body.classList.toggle('fullscreen-mode');
    // Force layout recalculation
    setTimeout(function () {
        if (typeof window.setHeight === 'function') {
            window.setHeight();
        }
        window.dispatchEvent(new Event('resize'));
    }, 50);
```
Is there a specific reason for the extra layer with the `global` object? I think it would be sufficient to add the function to the `metadataEditor` object itself.

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/header.xhtml
```diff
@@ -92,6 +92,10 @@
                              rendered="#{mayWrite}"
                              onclick="deactivateButtons();setUnsavedChanges(false);PF('sticky-notifications').renderMessage({'summary':'#{msgs.metadataSaving}','detail':'#{msgs.youWillBeRedirected}','severity':'info'});"
                              update="notifications"/>
+            <p:commandButton type="button"
+                             icon="pi pi-window-maximize"
+                             styleClass="fullscreen-btn"
+                             onclick="metadataEditor.global.toggleFullscreenMode()" />
```
oliver-stoehr:
```suggestion
                             onclick="metadataEditor.toggleFullscreenMode()" />
```
If the function is added directly to the `metadataEditor` object, this call must be changed.

oliver-stoehr (changes requested):
I like the new feature! It works fine, I found some ways to simplify the code.

Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_editor.js
BartChris: No specific reason, you are right!

Kitodo/src/main/webapp/WEB-INF/resources/js/metadata_editor.js
```diff
@@ -16,6 +16,17 @@
...
+metadataEditor.toggleFullscreenMode = function () {
+    document.body.classList.toggle('fullscreen-mode');
+    // Force layout recalculation
+    setTimeout(function () {
+        if (typeof window.setHeight === 'function') {
+            window.setHeight();
+        }
+        window.dispatchEvent(new Event('resize'));
+    }, 50);
+}
```
oliver-stoehr:
```suggestion
};
```
Variable assignment should have a semicolon at the end.

oliver-stoehr (changes requested):
Only one semicolon is missing, everything else looks good.

oliver-stoehr (approved):

PR #6999, opened 2026-04-20
Add missing 'referrer' parameter to 'takeOverTask' method calls

solth:
Fixes #6990

PR #7007, opened 2026-04-21
Allow for serial processing of tasks in process list

### LINKED ISSUE #6327: Allow for serial accepting of tasks in process list
It is possible to accept a task through the process list. (Show details, press bell-button.) Once you accept the task, Kitodo opens the task, i.e. moves to the task area with the task opened. If you now mark the task as done you get forwarded to the task list (Aufgaben). If you want to accept another task from the process list, you will first have to move back to the process list. But oh no, you get a fresh process list and all your filters have been removed and you have to re-enter them one by one.

When I open a task from the process list I should get back to the process list after closing the task. Filters on the process list should be preserved as they are when opening and closing a task from the task list.

Using the task list seems like a workable alternative but doing it through the process list should be possible as well because some users just prefer the process list (also filtering is more effective here).

**Additional context**
Could also be split into a bug (filter gets lost) and a feature request (forward back to process list).

BartChris:
Fixes: https://github.com/kitodo/kitodo-production/issues/6327
Builds upon: https://github.com/kitodo/kitodo-production/pull/6999

When a task is taken over from the process list and finished, the UI should navigate back to the process list.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/task/TaskWorkView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/task/TaskWorkView.java
@@ -36,6 +36,7 @@ import org.kitodo.export.TiffHeader;
 import org.kitodo.production.enums.GenerationMode;
 import org.kitodo.production.enums.ObjectType;
 import org.kitodo.production.forms.ValidatableForm;
+import org.kitodo.production.forms.process.ProcessListView;
 import org.kitodo.production.helper.Helper;
 import org.kitodo.production.helper.tasks.TaskManager;
 import org.kitodo.production.metadata.MetadataLock;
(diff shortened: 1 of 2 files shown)


thomaslow (commented):
Unfortunately, this doesn't fully work yet.

After clicking on the bell icon for a task in the process list, the "TaskWorkView" is opened without the URL parameter "referrerListOptions". When closing or rejecting the task, the user is returned to the correct referrer (the process list), but previous filter options of the process list are not preserved. 

If you'd like, I can look into that myself, since I have implemented this "referrerListOption" thingy.

BartChris:
> Unfortunately, this doesn't fully work yet.
> 
> After clicking on the bell icon for a task in the process list, the "TaskWorkView" is opened without the URL parameter "referrerListOptions". When closing or rejecting the task, the user is returned to the correct referrer (the process list), but previous filter options of the process list are not preserved.
> 
> If you'd like, I can look into that myself, since I have implemented this "referrerListOption" thingy.

Yes, would be good, if you can take over, since you are way deeper into it than i am. Thanks a lot.

BartChris:
Superseded by https://github.com/kitodo/kitodo-production/pull/7012

PR #7009, opened 2026-04-22
Fix MassIndexing error in MariaDB

BartChris:
As elaborated in https://github.com/kitodo/kitodo-production/pull/6977#issuecomment-4289865450

When using newer driver versions of MariaDB JDBC driver the current MassIndexer settings in Kitodo fail because they are operating on a MySQL specific setting. I therefor now inspect the used driver and use a different strategy when the MySQL driver is used. 

I am not sure about the exact values here (500 idFetchSize, 1000 Batch Size). They seem to work and so far there have not be any complaints that the Kitodo MassIndexer fails in 3.9. One can however think of making those values configurable and have lower defaults in place.


BartChris:
I have not tested it with a real MySQL database.

henning-gerhardt (approved):
Works for me with a MariaDB server. I can not prove if there are any problems with MySQL servers.

PR #7011, opened 2026-04-22
Make CommentForm ViewScoped

thomaslow:
This PR improves the CommentForm and makes it view scoped. 

I am not exactly sure why the method `refreshProcess` was necessary before. Maybe it had something to do with the task list itself being session scoped and not correctly showing the updated comment list. However, this cannot be a problem any more, since the task list is now view scoped, too.

I couldn't observe any issues creating and editing new correction comments or regular comments. New comments are immediately visible in the task list comment tooltip.


solth (approved):
Looks good and works.

PR #7012, opened 2026-04-22
Fix page navigation clicking on task bell in process list

### LINKED ISSUE #6327: Allow for serial accepting of tasks in process list

thomaslow:
This PR extends #7007 and fixes #6327.

The problem with #7007 is that non-static methods from the `TaskListView` are being referenced inside the `processList.xhtml` template. However, the task list is not being initialized when viewing the process list, and any list state changes (filter / sorting) are only available in `ProcessListView`. 

Because of that, I had to essentially copy the methods `takeOverTask` and `workOnTask` from `TaskListView` to `ProcessListView`. Unfortuantely, there doesn't really seem to be a reasonable way to refactor them into a single static method or a common workflow function.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/process/ProcessListView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/process/ProcessListView.java
@@ -13,6 +13,7 @@ package org.kitodo.production.forms.process;
 
 import static java.util.Map.entry;
 
+import java.io.IOException;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


BartChris (approved):
Works good for me, thanks a lot!

BartChris:
The workflow described in https://github.com/kitodo/kitodo-production/issues/6327 is now functional. I tested a little bit more. When you open the metada editor from the task page your original process list referrer gets of course lost.    

So if you accept a task from a process list and then on the taskEdit page open the metadata editor and afterwards finish the task, you do not go back to the process page but the task page.  I have no idea if it is even possible to preserve the original referrer in that constellation or whether we should. I am happy with the state of the PR.

thomaslow:
Yes, preserving list parameters in these multi-stage navigation scenarios is not really supported right now. 

There is one exception, which is the new "open metadata" context menu entry in the metadata editor tree view, which forwards all parameters (referrer & list options) to the next metadata editor when opening a parent process. Maybe we can adapt this solution also for the "work on task" page and forward referrer and list options as well.

PR #7015, opened 2026-04-24
Rename misleading class variable

henning-gerhardt:
Instance of class `ProcessProgressHelper` was used with `progressService` as class variable name. Both classes had different objects in use and variable naming should respect this.


BartChris (approved):
Yes, this is of course better.

PR #7021, opened 2026-04-29
Show process ID on task details page

solth:
There have been quite a few times when I needed the ID of a specific process when self-assigning a task of that process, but the process ID wasn't displayed on the "current task details" page (where you are forwarded when clicking the "bell" button in the task list), which made things cumbersome. So I decided to add it:

Bildschirmfoto 2026-04-29 um 10 15 17


henning-gerhardt (approved):
Change looks good and did work.

It is maybe possible to add or extend an existing UI test so that this new output value got not lost in future?

solth:
> It is maybe possible to add or extend an existing UI test so that this new output value got not lost in future?

@henning-gerhardt Test added

henning-gerhardt (approved):

PR #7024, opened 2026-05-04
[3.9] Catch Count exception during indexing

BartChris:
When indexing using the index page it might happen that an exception is thrown in the UI before the actual indexing begins. This is because before doing the indexing the destructive method `massIndexer.dropAndCreateSchemaOnStart(true);` is called which drops the index. When Kitodo tries to count in that moment, we get the Exception:

Request: GET /kitodo-process-read/_count with parameters {}
Response: 503 'Service Unavailable' from 'http://localhost:9200' with body 
{
  "error": {
    "root_cause": [
      {
        "type": "no_shard_available_action_exception",
        "reason": "[christoph-laptop][127.0.0.1:9300][indices:data/read/search[phase/query]]",
        "index_uuid": "AAMQB5gMQQO4L5tdDuXhAA",
        "shard": "0",
        "index": "kitodo-process-000001"
      }
    ],

When this exception is returned we can just return 0.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
@@ -175,6 +175,7 @@ public class IndexingService {
                     .where(f -> f.matchAll())
                     .fetchTotalHitCount();
         } catch (SearchException e) {
+            logger.debug("Search index temporarily unavailable during indexing initialization/rebuild.", e);
             // Index temporarily not available, just return 0
             return 0;
         }
--- a/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/index/IndexingService.java
@@ -25,6 +25,7 @@ import org.hibernate.search.mapper.orm.Search;
 import org.hibernate.search.mapper.orm.massindexing.MassIndexer;
 import org.hibernate.search.mapper.orm.session.SearchSession;
 import org.hibernate.search.mapper.pojo.massindexing.MassIndexingMonitor;
+import org.hibernate.search.util.common.SearchException;
 import org.kitodo.data.database.beans.BaseBean;
 import org.kitodo.data.database.beans.Process;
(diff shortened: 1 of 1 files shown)


solth:
@BartChris Since the response mentions a `no_shard_available_action_exception`, I wonder if there is a more specific exception that could be caught here instead of `SearchException`, which might occur for other, more legitimate reasons that would warrant a different handling?

BartChris:
Sometimes i also get another exception (see below). Problem is that HibernateSearch does not seem to expose the specific error type directly as an error class, so we would have to inspect the error message's body. Not sure if there is a satisfying solution here.

Also see the subclasses of the `SearchException` class: https://docs.hibernate.org/search/6.1/api/org/hibernate/search/util/common/SearchException.html

org.hibernate.search.util.common.SearchException: HSEARCH400007: Elasticsearch request failed: HSEARCH400090: Elasticsearch response indicates a failure.
Request: GET /kitodo-process-read/_count with parameters {}
Response: 404 'Not Found' from 'http://localhost:9200' with body 
{
  "error": {
    "root_cause": [
      {
        "type": "index_not_found_exception",
        "reason": "no such index [kitodo-process-read]",
        "resource.type": "index_or_alias",
        "resource.id": "kitodo-process-read",
        "index_uuid": "_na_",
        "index": "kitodo-process-read"
      }
    ],
  "status": 404
}

henning-gerhardt (commented):
It is an improvement to not to see the caught exception by the application in the intial phase (dropping and creating index) if you run the first time or you doing a re-index.

This exception must not displayed to the user as we can interact with them but I had add a logging statement (debug or trace level) when an exception is caught. It is easier then if you have issues on indexing to "see" at least the messages in the log file if you enable this logging.

It is up to you (@BartChris & @solth ) if an exception logging should made or not.

henning-gerhardt (approved):

solth:
> It is up to you (@BartChris & @solth ) if an exception logging should made or not.

I agree that would be a useful small change

BartChris:
I added a debug message.

PR #7031, opened 2026-05-06
[3.9] Update to version 3.9.1

solth:
Update Kitodo.Production `3.9.x` to version `3.9.1`

### DIFF
--- a/Kitodo-API/pom.xml
+++ b/Kitodo-API/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <artifactId>kitodo-production</artifactId>
         <groupId>org.kitodo</groupId>
-        <version>3.9.1-SNAPSHOT</version>
+        <version>3.9.1</version>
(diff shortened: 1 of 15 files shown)


henning-gerhardt (approved):

BartChris:
Do you see too many question unanswered here to include this PR in the release?
https://github.com/kitodo/kitodo-production/pull/6961

solth:
> Do you see too many question unanswered here to include this PR in the release?
> #6961

As pointed out in that pull request, even with that fix the option to delete media does not seem to work when that media has not yet been linked in the METS file. That's the reason I haven't included it in 3.9.1.

BartChris:
I understand, not being able to delete files when your workflow relies on that is a pain though.

solth:
> I understand, not being able to delete files when your workflow relies on that is a pain though.

Absolutely agree.

PR #7040, opened 2026-05-18
Fix layout of 'Add metadata' popup dialog on mass import page

### LINKED ISSUE #7039: Broken layout in "Add metadata" dialog of mass import
**Describe the bug**
When adding a new metadata column to the mass import record table, the layout of the popup dialog is slightly broken when at least one metadata label in the list is longer than the pulldown menus width.

**To Reproduce**
1. Go to 'Mass import'
2. Add some rows, either via uploading a CSV file or by adding them manually
3. Click the "+" button in the last column to add a new metadata column
4. If the ruleset contains a metadatum with a long label that can be added here the broken layout becomes visible

**Expected behavior**
The layout of the dialog should not break when metadata with long labels exist

**Screenshots**

Image

**Release**
3.9.x, 4.0.0.RC1

solth:
Fixes #7039

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/dialogs/addMetadata.xhtml
+++ b/Kitodo/src/main/webapp/WEB-INF/templates/includes/massImport/dialogs/addMetadata.xhtml
@@ -39,6 +39,7 @@
                             <p:selectOneMenu id="metadataType"
                                              converter="#{processDetailConverter}"
                                              value="#{MassImportForm.addMetadataDialog.metadataDetail}"
+                                             autoWidth="false"
                                              title="#{msgs.type}">
                                 <f:selectItems value="#{MassImportForm.addMetadataDialog.allMetadataTypes}"
                                                var="metadataType"


henning-gerhardt (commented):
Sorry, I can not decide if this solution solves the issue or not as I can not reproduce the original issue itself :-( 

The only issue which I found is that the text input field of this dialog is align to top instead of middle so the input itself not right.

Image

solth:
> Sorry, I can not decide if this solution solves the issue or not as I can not reproduce the original issue itself :-(
> 

It should be easily reproducable by temporarily changing the label of a metadata that appears in this list - for example "Haupttitel (Sortierung)" - to a much longer value that would exceed the available horizontal space.

> The only issue which I found is that the text input field of this dialog is align to top instead of middle so the input itself not right.
> Image

I can't reproduce this problem with the incorrect vertical alignement of the text content:

Bildschirmfoto 2026-05-21 um 12 57 21

As far as I can tell your screenshot shows an example where the text input field is actually a "text area" input field, which would correspond to a `multiline="true"` setting in the ruleset. If I am not mistaken, these fields are always rendered with "vertical-align: top" in Kitodo, be it the metadata editor, process creation mask or mass import.

henning-gerhardt:
Thanks for hinting me that the meta data title must be long. I did not notice this on the screenshot of the issue itself - or at least I did see the reason why this happen as the blank spaces are part of the title. Even that the settings from the ruleset are applied here.

I found an other issue for which I will open a separate issue: if a meta data field is defined with `maxOccurs="1"` you can add it two times which would result in invalid data (or so - as I did not test is further).

henning-gerhardt (approved):

PR #7043, opened 2026-05-19
Hide processes and tasks of inactive projects

### LINKED ISSUE #5295: Processes and tasks of disabled projects do not disappear
**Describe the bug**
Disabling a project in the project settings should hide all processes and tasks related to that project. This doesn’t happen, they remain visible.

**To Reproduce**
1. Have a project with at least one process in work.
2. Go to project settings of that project, disable it, and save it.
3. Check list of processes or list of tasks.
4. The processes and tasks of the project are still visible.

**Expected behavior**
The processes and tasks of the project which was disabled should no longer be showing.

**Release**
3.4.3

solth:
Fixes #5295

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -294,9 +294,17 @@ public class ProcessService extends BaseBeanService<Process, ProcessDAO> {
         if (!showClosedProcesses) {
             query.restrictToNotCompletedProcesses();
         }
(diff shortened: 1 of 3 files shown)


BartChris (approved):
Works good and code looks good. 

The tasks of the disabled projects are however still visible. This could be adressed in this PR or in a seperate PR.

https://github.com/kitodo/kitodo-production/blob/28f75a895678d22947335d97bb64556e2e21ed38/Kitodo/src/main/java/org/kitodo/production/services/data/TaskService.java#L269-L271

solth:
> The tasks of the disabled projects are however still visible. This could be adressed in this PR or in a seperate PR.

I updated the query for tasks as well. Should now also work as intended.

BartChris (approved):

PR #7050, opened 2026-05-21
Update commons-configuration2 to 2.15.0

henning-gerhardt:
Commons-configuration2 2.15.0 need at least commons-io 2.17.0 and commons-io switched from optional to required dependency for commons-configuration2. This should be fixed in commons-configuration2 version 2.15.1 (or later).

Supersede https://github.com/kitodo/kitodo-production/pull/7045

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
         <awaitility.version>4.0.3</awaitility.version>
         <com.h2database.h2.version>2.2.224</com.h2database.h2.version>
         <commons-codec.version>1.15</commons-codec.version>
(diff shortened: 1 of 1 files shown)


solth (approved):

solth:
@henning-gerhardt thanks for the quick fix! Would you mind checking if it is compatible to 3.9.x as well?

henning-gerhardt:
> @henning-gerhardt thanks for the quick fix! Would you mind checking if it is compatible to 3.9.x as well?

I will check it and open a pull request.

PR #7053, opened 2026-05-22
Use commons-collections4

henning-gerhardt:
Used dependency `org.apache.poi:poi-ooxml` already used `commons-collections4` but our code was still using old `commons-collections`. Switching to new `commons-collection4` is more or less an in-place change.

I removed the exclusion entry for `myfaces-impl` as both exclusions are not valid anymore and one of them was the old `commons-collections

### DIFF
--- a/Kitodo-DataManagement/pom.xml
+++ b/Kitodo-DataManagement/pom.xml
@@ -40,8 +40,8 @@
             <artifactId>commons-codec</artifactId>
         </dependency>
         <dependency>
-            <groupId>commons-collections</groupId>
-            <artifactId>commons-collections</artifactId>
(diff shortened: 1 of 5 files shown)


thomaslow (approved):
Works for me. I checked the impact on the resolved dependency tree:

```diff
238c238
< |  +- commons-collections:commons-collections:3.2.2:compile
---
> |  +- org.apache.commons:commons-collections4:4.5.0:compile
426,427c426
< |  +- com.github.virtuald:curvesapi:1.08:compile
< |  \- org.apache.commons:commons-collections4:4.4:compile
---
> |  \- com.github.virtuald:curvesapi:1.08:compile
```

Looks good to me.

PR #7059, opened 2026-05-28
Add tests for URL query parameters for task and process list

thomaslow:
This PR adds two new selenium tests, which verify that both the process and task list is correctly filtered and sorted based on the new URL query parameters. Also, they check that navigating from the list to a task or process edit page and back to the list will preserve all sort and filter criteria.

Related PRs and discussions are:
- #6903
- #6906
- #6963


henning-gerhardt (approved):
Code changes looks good and selenium based tests are successful in CI environment.

PR #7060, opened 2026-05-28
Fix exception when sorting list views by invisible column

### LINKED ISSUE #7046: `FacesException` when navigating to task, project or user list of new client
**Describe the bug**
I created a new client in my Kitodo installation and copied the roles from an existing client. When navigating to the task list, project list or user list afterwards, I get a `FacesException` with the error messages "**No column with field 'title' or columnKey 'null' has been found**" (task list or project list) or "**No column with field 'surname' or columnKey 'null' has been found**" (user list) respectively. The process list can be opened without error.

**To Reproduce**
1. Create a new client and copy the roles from an existing client
2. Self-assign this new client to your user
3. Self-assign a role - for example "Administrator" - of this new client to your user that grants access to task, user and project lists
4. Change to the new client (potentially needs a logout/re-login) 
5. navigate to the task, project or user lists
6. See error

**Expected behavior**
It should be possible to navigate to the user, task and project lists for a new client without error

**Screenshots**

_User list:_

Image

_Task list and project list:_

Image

**Release**
4.0.0.RC1

**Additional context**
Maybe connected to the new refactored filter introduced in #6770 
Error does **not** occur in Kitodo 3.9.x!

thomaslow:
Fixes #7046

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/BaseListView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/BaseListView.java
@@ -14,6 +14,7 @@ package org.kitodo.production.forms;
 import static java.util.Map.entry;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -39,7 +40,10 @@ public class BaseListView extends BaseForm {
      * @return the field name as URL query parameter (e.g. "title")
      */
     public String getSortByField() {
-        return sortBy.getField();
+        if (Objects.nonNull(sortBy)) {
+            return sortBy.getField();
+        }
+        return null;
     }
 
     /**
@@ -48,7 +52,13 @@ public class BaseListView extends BaseForm {
      * @return the sort order as URL query parameter (either "asc" or "desc")
      */
     public String getSortByOrder() {
-        return Map.ofEntries(entry(SortOrder.ASCENDING, "asc"), entry(SortOrder.DESCENDING, "desc")).getOrDefault(sortBy.getOrder(), "");
+        if (Objects.nonNull(sortBy)) {
+            return Map.ofEntries(
+                entry(SortOrder.ASCENDING, "asc"), 
+                entry(SortOrder.DESCENDING, "desc")
+            ).getOrDefault(sortBy.getOrder(), null);
+        }
+        return null;
     }
 
     /**
@@ -105,12 +115,13 @@ public class BaseListView extends BaseForm {
      * @return the combined list view options (URL query parameters)
      */
     public String getCombinedListOptions() {
-        return Map.ofEntries(
-            entry("firstRow",  String.valueOf(getFirstRow())),
-            entry("sortField", getSortByField()), 
-            entry("sortOrder", getSortByOrder()), 
-            entry("filter", getFilter())
-        ).entrySet().stream()
+        Map<String, String> parameterMap = new HashMap<>();
+        parameterMap.put("firstRow", String.valueOf(getFirstRow()));
+        parameterMap.put("sortField", getSortByField());
+        parameterMap.put("sortOrder", getSortByOrder()); 
+        parameterMap.put("filter", getFilter());
+
+        return parameterMap.entrySet().stream()
             .filter(entry -> Objects.nonNull(entry.getValue()))
             .map(entry -> entry.getKey() + "=" + entry.getValue().replace("&", "%26"))
             .collect(Collectors.joining("&"));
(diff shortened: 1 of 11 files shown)


Kitodo/src/main/java/org/kitodo/production/forms/process/ProcessListView.java
```diff
@@ -494,10 +495,23 @@ public String getCombinedListOptions() {
...
+            Map.entry("correctionCommentStatus", "process.comments"),
+            Map.entry("project.title", "process.project"),
+            Map.entry("creationDate", "process.duration")
         );
+
+        return Stream.concat(
+            Stream.of("id"), // id column is always allowed and can't be hidden
+            fieldsToColumns.entrySet().stream()
+                .filter(entry ->  selectedColumns.stream().anyMatch(column -> entry.getValue().equals(column.getTitle())))
+                .map(entry -> entry.getKey())
```
solth:
```suggestion
                .map(Map.Entry::getKey)
```
Can be simplified by using method reference instead of lambda function, but that is just a matter of taste (applies to other list views as well)

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
```diff
@@ -129,7 +129,10 @@ public void init() {
         selectedTaskStatus.add(TaskStatus.OPEN);
         selectedTaskStatus.add(TaskStatus.INWORK);
 
-        sortBy = SortMeta.builder().field("title").order(SortOrder.ASCENDING).build();
+        if (selectedColumns.stream().anyMatch(column -> "task.title".equals(column.getTitle()))) {
```
solth, line 132: Since `sortBy` is not initialized (only declared) in `BaseForm`, what happens if this condition is not met? The `taskList.xhtml` will still reference `sortBy="#{TaskListView.sortBy}"` in its datatable, which could then cause a NullPointerException if the condition is not met and thus `sortBy` is not initialized, couldn't it? 

In other words: don't we have to make sure, `sortBy` is always set to something as the default case? (same applies to other list views - Docket, LdapGroup, Project, Role, Ruleset, Template and Workflow - as well)

Kitodo/src/main/java/org/kitodo/production/forms/BaseListView.java
```diff
@@ -48,7 +52,13 @@ public String getSortByField() {
      * @return the sort order as URL query parameter (either "asc" or "desc")
      */
     public String getSortByOrder() {
-        return Map.ofEntries(entry(SortOrder.ASCENDING, "asc"), entry(SortOrder.DESCENDING, "desc")).getOrDefault(sortBy.getOrder(), "");
+        if (Objects.nonNull(sortBy)) {
+            return Map.ofEntries(
+                entry(SortOrder.ASCENDING, "asc"), 
+                entry(SortOrder.DESCENDING, "desc")
+            ).getOrDefault(sortBy.getOrder(), null);
```
solth: My IDE tells me: "Passing 'null' argument to parameter annotated as @NotNull", but I guess this is just a false positive

solth (commented):
The exception is gone, which is a big improvement. 

I did observe some strange behaviour on the user page, though. When switching to the new client, the list contains rows for all existing users, but shows _no_ columns apart from the "actions" column, which looks very wierd:

Image

The columns can be activated manually via the menu, but shouldn't they all be visible by default?

thomaslow:
> The columns can be activated manually via the menu, but shouldn't they all be visible by default?

Yes, apparently I missed copying a line while refactoring `ClientForm` to `ClientEditView`, see #7062.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
thomaslow: Yes, I considered and tested this. There is no exception when `sortBy` is null. You can test this by hiding the column that is currently sorted.

I agree that it would be nice if there was always some kind of sorting. However, in many list views you can hide all sortable columns, leaving no alternatives. In this case, `sortBy` will definitely be null. Also, I am not sure what is supposed to happen when hiding a column that is currently used for sorting. How do we pick the next best "default" column? There would need to be a priority list for every list view.

Kitodo/src/main/java/org/kitodo/production/forms/BaseListView.java
thomaslow: Yes, changed it to `.get(sortBy.getOrder())`. Somehow I assumed an exception would be thrown if a particular key is not available, but it is not.

Kitodo/src/main/java/org/kitodo/production/forms/process/ProcessListView.java
thomaslow: Updated all list views.

Kitodo/src/main/java/org/kitodo/production/forms/task/TaskListView.java
solth: Yes, thanks for explaining, I understand now. For me the most important aspect was to prevent the exception, so I am happy with the pull request now!

solth (approved):

PR #7061, opened 2026-05-28
Local only schemas for xslt

### LINKED ISSUE #6855: Schema files with local and external schema references

### LINKED ISSUE #7028: Mapping file validation cannot be deactivated and files are validated against an external schema
**Describe the bug**
In contrast to the imported source documents for which validation can be activated or deactivated mapping file validation cannot be deactivated.

The mapping files are always validated against standard schema files. 

https://github.com/kitodo/kitodo-production/blob/e1fdd90482a14ce073dac77044db74c8a9e1884c/Kitodo/src/main/java/org/kitodo/production/services/data/ImportService.java#L962

https://github.com/kitodo/kitodo-production/blob/a022f704dbf7c569fc0251b3b86bfbe0bf40a426/Kitodo/src/main/java/org/kitodo/production/services/validation/FileStructureValidationService.java#L156-L164

The problem is that those schema files reference external schema files

https://github.com/kitodo/kitodo-production/blob/a022f704dbf7c569fc0251b3b86bfbe0bf40a426/Kitodo/src/main/resources/schemata/xslt20.xsd#L57-L64

which might not be resolvable. 

**To Reproduce**
1. Create a new process via an import configuration which uses mapping files
2. When the external schema files cannot be resolved import fails.

**Expected behavior**
There should be no references to external schema files so the import at least does not fail on some machines. It could be discussed whether mapping file validation should be disabled as well when the ImportConfiguration is setup with no validation.

**Release**
Kitodo Main

**Additional context**
First noticed during MassImport but also during normal import. https://github.com/kitodo/kitodo-production/pull/6983#issuecomment-4269059355

BartChris:
Fixes: https://github.com/kitodo/kitodo-production/issues/6855
Fixes: https://github.com/kitodo/kitodo-production/issues/7028

### DIFF
--- a/Kitodo/src/main/resources/schemata/w3-xml.xsd
+++ b/Kitodo/src/main/resources/schemata/w3-xml.xsd
@@ -1,0 +1,286 @@
+<?xml version='1.0'?>
+<?xml-stylesheet href="../2008/09/xsd.xsl" type="text/xsl"?>
+<xs:schema targetNamespace="http://www.w3.org/XML/1998/namespace"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns   ="http://www.w3.org/1999/xhtml"
+           xml:lang="en">
+
+    <xs:annotation>
+        <xs:documentation>
+            <div>
+                <h1>About the XML namespace</h1>
+
+                <div class="bodytext">
+                    <p>
+                        This schema document describes the XML namespace, in a form
+                        suitable for import by other schema documents.
+                    </p>
+                    <p>
+                        See <a href="http://www.w3.org/XML/1998/namespace.html">
+                        http://www.w3.org/XML/1998/namespace.html</a> and
+                        <a href="http://www.w3.org/TR/REC-xml">
+                            http://www.w3.org/TR/REC-xml</a> for information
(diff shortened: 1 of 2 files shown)


BartChris:
I checked what DNS lookups are happening with my changes during catalog import and it seems like no lookup to W3C is done. (In Main branch DNS lookup and data retrieval for W3C is done).

henning-gerhardt (approved):

solth (approved):

PR #7062, opened 2026-05-29
Fix empty list columns when creating new client

thomaslow:
This PR fixes the problem that no list columns are assigned when creating a new client. This problem was probably introduced by #6770, which misses this line in the new `ClientEditView.java`: https://github.com/kitodo/kitodo-production/blob/ee7cb3fe62ab49f926dd88dd637c32de25113d5c/Kitodo/src/main/java/org/kitodo/production/forms/ClientForm.java#L95

This PR is related to:
- #7046
- #7060

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/production/forms/ClientEditViewIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/forms/ClientEditViewIT.java
@@ -79,7 +79,7 @@ public class ClientEditViewIT {
      * Test that when creating new clients, standard list columns are assigned.
      */
     @Test
-    public void testListColumnsForNewClients() throws Exception {
+    public void testListColumnsForNewClients() {
         ClientEditView clientEditView = new ClientEditView();
         clientEditView.init();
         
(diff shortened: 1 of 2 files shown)


Kitodo/src/test/java/org/kitodo/production/forms/ClientEditViewIT.java
```diff
@@ -72,4 +74,16 @@ public void testRoleAdding() throws DAOException {
         assertEquals(numberOfOldAuthorities, numberOfNewAuthorities, "Authorities were not added");
         assertEquals(numberOfAuthoritiesToCopy, numberOfOldAuthorities, "Authorities were removed from second client");
     }
+
+    /**
+     * Test that when creating new clients, standard list columns are assigned.
+     */
+    @Test
+    public void testListColumnsForNewClients() throws Exception {
```
solth:
```suggestion
    public void testListColumnsForNewClients() {
```
Is the `throws` required here? None of the methods called in this tests is declared to throw an exception...

solth (commented):
Resolves the issue, thank you, @thomaslow ! 

Just one tiny thing I wondered about the test you thankfully added as well.

solth (approved):

Issue #6680, opened 2025-09-18
Incorrectly configured image generation deletes originals


matthias-ronge:
**Describe the bug**
If image generation is enabled in the folder configured as the source for generating derivatives, starting the derivative generation will delete the original images instead.

**To Reproduce**
1. Go to the project page
2. Configure the `images/(processtitle)_tif, LOCAL` folder as the source for generating derivatives
3. In the `images/(processtitle)_tif, LOCAL` folder, configure Generate Images: Create Derivative
4. In the workflow, execute the Generate Images step.
5. See: All TIFF files are gone.

**Expected behavior**
- [ ] If the target is the source, the image generator should ignore it.
- [ ] optional: In the project settings, there should be no incorrect configuration options (it will be grayed out).
- [ ] optional: Warn if an incorrect configuration is already configured in the display (orange bar).

**Screenshots**
<img width="848" height="539" alt="Image" src="https://github.com/user-attachments/assets/f18942f2-c6eb-47b5-acdf-0e60ce348447" />

**Release**
3.6.2, 3.9.0.RC1 (probably versions inbetween as well)

**Desktop (please complete the following information):**
 - OS: Window 11
 - Browser Firefox
 - Version 142.0.1

PR #7063, opened 2026-05-29
Prevent invalid folder settings in the UI

BartChris:
This Pull Request adresses the UI related parts of the problems described in https://github.com/kitodo/kitodo-production/issues/6680 by preventing invalid values.

Two things are prevented:

a) A folder configured as source for image generation can not at the same time have its images generated:

image

image

b) A folder which images are generated from another source cannot be setup to be used as a source for image generation. 

image

image

Configurations which were invalid before can still be edited, but if they are changed they can only be changed to a valid configuration.


thomaslow (approved):
Looks good to me. I can confirm that images are deleted with an incorrect folder configuration in the current main branch, and this PR doesn't allow this incorrect configuration any more.

solth:
@BartChris please rebase this pull request against `main` and provide backports for `4.0.x` and `3.9.x`, if feasible!

PR #7068, opened 2026-06-05
[3.9] Add Kitodo script to set ImportConfiguration for selected processes

solth:
This pull request adds a new Kitodo script allowing users to set the import configuration for selected processes in the process list. There already is a dedicated list action for the same purpose, but by adding a Kitodo script this function can now also be used for script tasks in the workflow.


Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -787,4 +791,18 @@ private void saveTask(String processTitle, Task task) {
             Helper.setErrorMessage("Error while saving - " + processTitle, logger, e);
         }
     }
+
+    private void setImportConfiguration(List<Process> processes) throws DAOException {
+        Object id = this.parameters.get("id");
```
henning-gerhardt: `get()` method is returning type `String` so `Object` for the method variable is as the `parameters` variable is defined as `Map` of `String` key with a `String` value.

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -787,4 +791,18 @@ private void saveTask(String processTitle, Task task) {
...
+        if (Objects.nonNull(id)) {
+            Integer configurationId = Integer.valueOf(id.toString());
```
henning-gerhardt: - No need for `id.toString()` call if the `id` variable is declared as `String` class type. 
- The possible thrown `NumberFormatException` is caught outside of this method. How this exception (or even others) are displayed to the user is an other question.

Kitodo/src/test/java/org/kitodo/production/services/command/KitodoScriptServiceIT.java
```diff
@@ -40,6 +41,7 @@
 import org.kitodo.config.enums.ParameterCore;
 import org.kitodo.data.database.beans.Folder;
 import org.kitodo.data.database.beans.Process;
+import org.kitodo.data.database.beans.ImportConfiguration;
```
henning-gerhardt, line 43: Wrong import statement order: should be moved before `Process`.

henning-gerhardt (changes requested):
Changes are working and are good in generell - see my comments for changes.

I wonder how an user can find out the import configuration id which is only available in the database (or: not visible in the UI)? In the list action there is an dialog where the user is choosing an import configuration by import configuration title. 

So using this KitodoScript needs the knowledge of database entries or a few iterations of try and error until you find out which is the correct id. Using the import configuration title could be an option as this title is easy available in the UI?

Kitodo/src/test/java/org/kitodo/production/services/command/KitodoScriptServiceIT.java
solth: Done

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
solth: Done

solth:
> I wonder how an user can find out the import configuration id which is only available in the database (or: not visible in the UI)? In the list action there is an dialog where the user is choosing an import configuration by import configuration title.
> 

Yes, I understand what you are saying. The ID of an import configuration is more or less hidden, so it may not seem as the best choice as a parameter for a script. It is, however, visible in the URL when editing a configuration, and also in the link when hovering over the "edit" symbol of an import configuration in the list view. 

Bildschirmfoto 2026-06-05 um 20 20 41

Also, there is also a precedence where IDs are used for a Kitodo script, which is not really displayed in any UI element. That is the script `importProcesses` and it used the ID of a template _and_ of a project:

Bildschirmfoto 2026-06-05 um 20 12 05

> So using this KitodoScript needs the knowledge of database entries or a few iterations of try and error until you find out which is the correct id. Using the import configuration title could be an option as this title is easy available in the UI?


Unfortunately, we currently do not have a function in place to retrieve import configuration by title from the database, and implementing such a method is not as straight forward as it may seem, as we do not have a restriction in place for unique import configuration titles right now, e.g. multiple configurations can have the same title, and thus uniquely identifying them by title title is not possible.

Bildschirmfoto 2026-06-05 um 20 19 57

Maybe an alternative solution would be to just display the ID of import configurations in another column of the import configuration list? Than they would be readily available for users and could be used in the Kitodo script?

solth:
@henning-gerhardt I now opted for the least intrusive solution and just added the ID as a read only field to the import configuration edit mask (not rendered when creating new configurations, of course):

Bildschirmfoto 2026-06-06 um 12 23 47

Would this work for you as a low tech minimal approach to tackle the issue you mentioned about unknown IDs?

henning-gerhardt:
@solth:
* I did not know that the title of import configurations is not unique as we use unique titles for the import configurations. As title of import configurations are no option as not unique, the import configurations id is the only possible unique identifier.
* I had in mind that displaying import configuration id should be identical to the display of processes in the process list view: the id as first column and even added in edit mode as part of the import configuration title similar / equal to the edit mode of processes. This should be not part of this pull request but for a follow up pull request.
* I will look into your change and maybe this is already good.

solth:
> * I had in mind that displaying import configuration id should be identical to the display of processes in the process list view: the id as first column and even added in edit mode as part of the import configuration title similar / equal to the edit mode of processes. This should be not part of this pull request but for a follow up pull request.

@henning-gerhardt good idea, I think it's better to stick to one style of displaying IDs instead of showing them in a different style on every page. I will adjust that, so you can wait with your review, ok?

henning-gerhardt:
I will wait.

solth:
> I will wait.

I updated the import configuration list and edit pages to display the ID in a way that is more in line with the established page layouts in Kitodo:

_List view:_
Bildschirmfoto 2026-06-08 um 09 36 17

_Edit view:_
Bildschirmfoto 2026-06-08 um 09 36 56

Please note that sorting the import configuration list currently is bugged. That is not related to the changes in this pull requerst and does not work in the `3.9.x` branch either! (fixed in `main`, however)

Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
```diff
@@ -787,4 +791,18 @@ private void saveTask(String processTitle, Task task) {
...
     }
+
+    private void setImportConfiguration(List<Process> processes) throws DAOException, NumberFormatException {
+        String id = this.parameters.get("id");
+        if (Objects.nonNull(id)) {
+            Integer configurationId = Integer.valueOf(id);
+            ImportConfiguration importConfiguration = ServiceManager.getImportConfigurationService().getById(configurationId);
+            for (Process process : processes) {
+                process.setImportConfiguration(importConfiguration);
+                ServiceManager.getProcessService().save(process);
```
henning-gerhardt: Why not re-use the local private method `saveProcess` which store the changed process and handle the `DAOException` (which then not be needed to be declared)? There is a similar change possible in line 549 (changing the ruleset for processes) which can maybe changed too.

henning-gerhardt (changes requested):
Only one (or two) small change(s) which I discovered. But everything else is working.

henning-gerhardt (approved):

PR #7069, opened 2026-06-05
Fix deleting media after deleting parent logical division

thomaslow:
This PR fixes the issue discovered in #6961, see [comment](https://github.com/kitodo/kitodo-production/pull/6961#issuecomment-4250731683). 

Demo of the problem:

https://github.com/user-attachments/assets/13647403-0614-4be5-96b0-4eaeeac9ec57


solth:
@thomaslow please rebase against current `main` branch where Selenium test problems should be resolved!

solth (approved):
This does not only fix the linked issue, but also the problem I documented in https://github.com/kitodo/kitodo-production/pull/6961#issuecomment-4286389480, so great value all around!

PR #7072, opened 2026-06-08
[3.9] Harden XMLUtils EAD/XML upload parsers against XXE

solth:
The parseXMLString and getNumberOfEADElements helpers built their JAXP factories with the default configuration, which resolves external entities and external DTDs. These helpers parse XML uploaded through the Create-process file-upload dialog, so a crafted document allowed local file disclosure and outbound SSRF (XXE / CWE-611).

Disable DOCTYPE declarations and external entity resolution on both factories, reusing the same secure configuration the load() and newDocument() siblings and the external-catalog response parsers (XmlResponseHandler, QueryURLImport) already apply. Add regression tests that confirm the two entry points reject a DOCTYPE / external-entity payload instead of expanding it.

Credit: @tonghuaroot

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/helper/XMLUtils.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/XMLUtils.java
@@ -60,6 +60,10 @@ import org.xml.sax.SAXException;
  */
 public class XMLUtils {
 
+    private static final String DISALLOW_DOCTYPE_DECL = "http://apache.org/xml/features/disallow-doctype-decl";
+    private static final String EXTERNAL_GENERAL_ENTITIES = "http://xml.org/sax/features/external-general-entities";
+    private static final String EXTERNAL_PARAMETER_ENTITIES = "http://xml.org/sax/features/external-parameter-entities";
+
     /**
      * Private constructor to hide the implicit public one.
      */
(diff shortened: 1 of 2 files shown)


Kitodo/src/test/java/org/kitodo/production/helper/XMLUtilsTest.java
```diff
@@ -84,6 +89,37 @@ public void shouldParseXMLString() {
         Assertions.assertDoesNotThrow(() -> XMLUtils.parseXMLString(XML_STRING));
     }
 
+    @Test
+    public void parseXMLStringShouldNotResolveExternalEntities() throws Exception {
+        Path secretPath = Files.createTempFile("xxe-canary", ".txt");
```
henning-gerhardt: As we use Junit5 we can use feature `TempDir` annotation of Junit5 to create a "secure" temporary test directory. See https://www.heise.de/hintergrund/Secure-Coding-CWE-377-Unsichere-temporaere-Dateien-und-wie-man-sie-vermeidet-10028888.html?seite=2 (Junit5 Jupiter paragraph). Could be used in both places.

Kitodo/src/main/java/org/kitodo/production/helper/XMLUtils.java
```diff
@@ -289,4 +292,25 @@ public static int getNumberOfEADElements(String xmlString, String eadLevel) thro
         return count;
     }
 
+    private static final String DISALLOW_DOCTYPE_DECL = "http://apache.org/xml/features/disallow-doctype-decl";
+    private static final String EXTERNAL_GENERAL_ENTITIES = "http://xml.org/sax/features/external-general-entities";
+    private static final String EXTERNAL_PARAMETER_ENTITIES = "http://xml.org/sax/features/external-parameter-entities";
```
henning-gerhardt: Maybe move this static declaration to the top of the class.

henning-gerhardt (approved):
Looks good and should be forward ported to main branch.

PR #7081, opened 2026-06-10
Submit search terms in Selenium Test using Enter Key and harden further Selenium tests

BartChris:
The Searching Selenium Tests (`SearchingST`) are still flaky in CI  (Locally not reproducable). Instrumentation showed several failing runs where Selenium successfully entered text into the search field and clicked the search button, but the backing JSF action `SearchResultForm.searchForProcessesBySearchQuery()` was never invoked. (Especially when CI uses Chrome 149 - always fails -, with Chrome 148 this is more stable - mostly succeeds -. Github sometimes uses 148 and sometimes 149)

I suspect that this might be caused by a failed execution of the search form (combining an `disabled` attribute and an update action on "keyup") which might be affected by race conditions, although i could not really nail down the root cause. 

https://github.com/kitodo/kitodo-production/blob/d8e68f13d45445262ef3a99c6f4506a95b6492b5/Kitodo/src/main/webapp/WEB-INF/templates/includes/header/search.xhtml#L28-L37

This PR tries to adress this by using the Enter key to submit the form. In addition to that another Selenium accessor method is made more resilient.

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/Page.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/Page.java
@@ -23,8 +23,8 @@ import java.util.function.Predicate;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.kitodo.selenium.testframework.Browser;
-import org.kitodo.selenium.testframework.Pages;
 import org.openqa.selenium.By;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.NotFoundException;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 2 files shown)


BartChris:
Run 1 - Chrome 148.0.7778.178 - success
Run 2 - Chrome 148.0.7778.178 - success (Fork 149.0.7827.53 - success)
Run 3  - Chrome 148.0.7778.178 - success (Fork 149.0.7827.53 - success)
Run 4  - Chrome 148.0.7778.178 - success (Fork 149.0.7827.53 - success)
Run 5 - Chrome 149.0.7827.53 - success (Fork 148.0.7778.178 - success) (!)
Run 6 - Chrome 148.0.7778.178 - success (Fork 149.0.7827.53 - success)

The Kitodo Repo (almost) always uses Chrome 148, which is a little annoying when you want to check 149. My fork on the other hand always uses 149.

solth (approved):

PR #7088, opened 2026-06-19
Update supported versions

solth:
Update supported versions: 
- added support for upcoming release `4.1.x`
- removed support for old version `3.8.x`

### DIFF
--- a/SUPPORTED_VERSIONS.md
+++ b/SUPPORTED_VERSIONS.md
@@ -4,9 +4,10 @@ The following versions of Kitodo.Production are currently being supported with o
 
 | Version | Active Development       |      Security Fixes      |
 |-----| :----------------------: |:------------------------:|
-| 4.0 | :heavy_check_mark:       |    :heavy_check_mark:    |
-| 3.9 | :heavy_check_mark:       |    :heavy_check_mark:    |
-| 3.8 | :heavy_multiplication_x: |    :heavy_check_mark:    |
+| 4.1 | :heavy_check_mark:       |    :heavy_check_mark:    |
+| 4.0 | :heavy_multiplication_x: |    :heavy_check_mark:    |
+| 3.9 | :heavy_multiplication_x: |    :heavy_check_mark:    |
+| 3.8 | :heavy_multiplication_x: | :heavy_multiplication_x: |
 | 3.7 | :heavy_multiplication_x: | :heavy_multiplication_x: |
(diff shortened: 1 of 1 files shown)


henning-gerhardt (approved):
Suggestion: maybe the list can be compacted by using something like `< 3.9` for all not supported versions?

solth:
> Suggestion: maybe the list can be compacted by using something like `< 3.9` for all not supported versions?

Good idea, I was thinking about that as well. Maybe we can compress the section about support for `3.x` the same way as is already done for `1.x` and `2.x`, like this:

Bildschirmfoto 2026-06-19 um 10 17 37

I don't think we still have to specify that potential security fixes will only be released for `3.9.x`, and not for `3.8.x` anymore. @henning-gerhardt what do you think? Would that work for you?

henning-gerhardt:
I had the same idea with `3.x` but on fast read it could be interpreted that even all `3.x` version get security fixes. 

So I changed my idea to `< 3.9` to collapse them together. Maybe the `1.x` and `2.x` can stand for even clarified statement. If all `3.x` versions are out of support then we can collapse them to a single `3.x` entry.

solth:
> I had the same idea with `3.x` but on fast read it could be interpreted that even all `3.x` version get security fixes.

Yes, at first I was afraid this could be misinterpreted, too, but on second thought, that is then just a conjecture by the reader, and I don't think we are still obliged to document every `3.x` version individually, now that `4.x` is the new main version.

If we still supply security updates for `3.9.x` when needed, the promised support for `3.x` as a whole is fulfilled, in my opinion; no need to specify _which_ version of `3.x` gets those updates. (IMHO, it should be common sense that the latest version, thus `3.9.x`, is the one that may still be updated). 

Additionally, by compressing the support for `3.x` versions into one line and instead documenting support for `4.x` versions individually, we can help shifting focus to the newer version. 

> So I changed my idea to < 3.9 to collapse them together. Maybe the 1.x and 2.x can stand for even clarified statement. If all 3.x versions are out of support then we can collapse them to a single 3.x entry.

I don't like the idea of introducing a new notation (e.g. `< 3.9`) here, to be honest. I would prefer to keep the existing format and collapse `3.x` as described above.

henning-gerhardt:
If there are misunderstandings on reading and understanding which version is supported then we can change it.

henning-gerhardt (approved):

PR #7089, opened 2026-06-19
Explicitly insert semicolon instead of relying on automatic insertion

solth:
Insert missing semicolon as suggested in https://github.com/kitodo/kitodo-production/pull/7000#pullrequestreview-4361585541


BartChris (approved):
I thought i have added it. Thanks for fixing.

Issue #5997, opened 2024-03-07
 Investigate and fix issues reported by Coverity Scan


stweil:
The static code analyzer Coverity Scan reports 330 open issues (after #5995 is applied) for the current git master branch. A quick review of those issues shows that at least some of them are real bugs which must be fixed.

Therefore all issues should be reviewed and classified (Coverity Scan has classifications for bug, false positive, intentional and pending). Ideally bugs should also be fixed.

stweil:
The estimated costs for the classification of all reported issues are a few days (low).

Fixing bugs ranges from low costs for trivial fixes, but can also require medium or even high costs for more complex bugs. Therefore I'd restrict this task to fixing trivial bugs and reporting all other bugs as issues on GitHub.

Ideally all remaining issues in Coverity Scan which were classified as bugs should contain references to GitHub issues.

solth:
Votes: 8

stweil:
[Coverity Scan status](https://scan.coverity.com/projects/kitodo-kitodo-production) update for Git master branch:
2024-08-05: 393 outstanding issues
2024-08-22: 354 outstanding issues
2025-03-13: 358 outstanding issues
2026-06-24: 296 outstanding issues (with #7094 applied)

All but one resource leaks are fixed in #7094. About 28 issues are null pointer dereferences. 2 issues are known security issues. The rest are medium or low priority issues.

stweil:
@solth, among the low priority issues are more than 30 of this kind:
> PI: Do not reuse public identifiers from Java Standard Library.
> It's essential to avoid reusing public identifiers from the Java Standard Library as class names.
> It is a good practice to avoid reusing public identifiers from the Java Standard Library as field names in your code.
> [...]

Examples of such identifiers are `FileFormat`, `Option`, `Namespace`, all very generic names, so the conflicts are not surprising.

Do we want to ignore such issues, or should we try to use more specific names, perhaps by prepending "Kitodo" to the name?

PR #7094, opened 2026-06-24
Fix memory leaks (reported by Coverity Scan)

stweil:
This pull request fixes all memory leaks which were reported by Coverity Scan, no matter whether they are real leaks or false positives.

### DIFF
--- a/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
+++ b/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
@@ -80,8 +80,7 @@ public class KitodoServiceLoader<T> {
     private static final String JAR = "*.jar";
     private static final String ERROR = "Classpath could not be accessed";
 
-    private static final Path SYSTEM_TEMP_FOLDER = FileSystems.getDefault()
-            .getPath(System.getProperty("java.io.tmpdir"));
+    private static final Path SYSTEM_TEMP_FOLDER = Paths.get(System.getProperty("java.io.tmpdir"));
 
     private static final Logger logger = LogManager.getLogger(KitodoServiceLoader.class);
 
--- a/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
+++ b/Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
@@ -20,7 +20,6 @@ import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.file.DirectoryStream;
-import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -141,7 +140,7 @@ public class KitodoServiceLoader<T> {
      * they can be used in all frontend files
      */
     private void loadBeans() {
-        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+        Path moduleFolder = Paths.get(modulePath);
         try (DirectoryStream<Path> stream = Files.newDirectoryStream(moduleFolder, JAR)) {
             for (Path f : stream) {
                 try (JarFile jarFile = new JarFile(f.toString())) {
@@ -196,7 +195,7 @@ public class KitodoServiceLoader<T> {
      */
     private void loadFrontendFilesIntoCore() {
 
-        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+        Path moduleFolder = Paths.get(modulePath);
 
         try (DirectoryStream<Path> stream = Files.newDirectoryStream(moduleFolder, JAR)) {
 
@@ -381,7 +380,7 @@ public class KitodoServiceLoader<T> {
      * earlier class loader created at an earlier time.</p>
      */
     private void loadModulesIntoClasspath() {
-        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+        Path moduleFolder = Paths.get(modulePath);
 
         try (DirectoryStream<Path> stream = Files.newDirectoryStream(moduleFolder, JAR)) {
 
--- a/Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
+++ b/Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
@@ -22,13 +22,16 @@ import static org.mockito.Mockito.when;
 
 import java.util.NoSuchElementException;
 
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.kitodo.config.enums.ParameterAPI;
+import org.mockito.MockedStatic;
 
 public class KitodoConfigTest {
 
     private static ParameterAPI NONE;
+    private static MockedStatic<ParameterAPI> mockedParameterAPI;
 
     /**
      * Init once before tests.
@@ -40,10 +43,15 @@ public class KitodoConfigTest {
         NONE = mock(ParameterAPI.class);
         doReturn(3).when(NONE).ordinal();
 
-        mockStatic(ParameterAPI.class);
+        mockedParameterAPI = mockStatic(ParameterAPI.class);
         when(ParameterAPI.values())
                 .thenReturn(new ParameterAPI[] {ParameterAPI.DIR_MODULES, ParameterAPI.DIR_PROCESSES,
                                                 ParameterAPI.DIR_XML_CONFIG, NONE });
+    }
+
+    @AfterAll
+    public static void tearDown() {
+        mockedParameterAPI.close();
     }
 
     @Test
--- a/Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
+++ b/Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
@@ -21,6 +21,7 @@ import static org.mockito.Mockito.mockStatic;
 import static org.mockito.Mockito.when;
 
 import java.util.NoSuchElementException;
+import java.util.Objects;
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -51,7 +52,9 @@ public class KitodoConfigTest {
 
     @AfterAll
     public static void tearDown() {
-        mockedParameterAPI.close();
+        if (Objects.nonNull(mockedParameterAPI)) {
+            mockedParameterAPI.close();
+        }
     }
 
     @Test
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -141,8 +141,7 @@ public class FileService {
             logger.info("Metadata directory: {} already existed! No new directory was created", directoryName);
         } else {
             CommandService commandService = ServiceManager.getCommandService();
-            String path = FileSystems.getDefault()
-                    .getPath(ConfigCore.getKitodoDataDirectory(), parentFolderUri.getRawPath(), directoryName)
+            String path = Paths.get(ConfigCore.getKitodoDataDirectory(), parentFolderUri.getRawPath(), directoryName)
                     .normalize().toAbsolutePath().toString();
             List<String> commandParameter = Collections.singletonList(path);
             File script = new File(ConfigCore.getParameter(ParameterCore.SCRIPT_CREATE_DIR_META));
--- a/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
@@ -18,7 +18,6 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.nio.file.FileSystems;
 import java.nio.file.Paths;
 import java.text.MessageFormat;
 import java.util.Arrays;
--- a/Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
+++ b/Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
@@ -50,6 +50,7 @@ public class MediaPartialFormTest {
     PhysicalDivision physicalStructure;
 
     private static MockedStatic<Ajax> ajaxMockedStatic;
+    private static MockedStatic<PrimeFaces> primefacesSingleton;
 
     /**
      * Initialize test class.
@@ -59,7 +60,7 @@ public class MediaPartialFormTest {
         // mock frontend update calls
         ajaxMockedStatic = Mockito.mockStatic(Ajax.class);
         PrimeFaces primeFaces = mock(PrimeFaces.class);
-        MockedStatic<PrimeFaces> primefacesSingleton = Mockito.mockStatic(PrimeFaces.class);
+        primefacesSingleton = Mockito.mockStatic(PrimeFaces.class);
         primefacesSingleton.when(PrimeFaces::current).thenReturn(primeFaces);
     }
 
@@ -70,6 +71,9 @@ public class MediaPartialFormTest {
     public static void cleanupTestClass() {
         if (ajaxMockedStatic != null) {
             ajaxMockedStatic.close();
+        }
+        if (primefacesSingleton != null) {
+            primefacesSingleton.close();
         }
     }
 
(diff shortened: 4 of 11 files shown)


solth:
Please always provide at least a short description as I will not accept pull requests without any description in the future.

stweil:
Each commit has a lengthy description. If you prefer I can make a PR for each commit. But it is easier to either "create a merge commit" or to "rebase and merge" this PR. Then all detailled commit messages are preserved and will be part of the git history. Don't use "squash and merge" (unless in a few special cases) because it will mix unrelated changes into a single one. That's bad for `git bisect`, for example.

stweil:
> Please always provide at least a short description as I will not accept pull requests without any description in the future.

Relevant discussion and documentation:

- https://github.com/kitodo/kitodo-production/discussions/6190
- https://github.com/kitodo/kitodo-production/wiki/Hinweise-f%C3%BCr-Entwickler

solth:
I prefer having the information about the pull request available in the pull request description itself when reviewing it instead of having to click each and every commit of a pull request individually to see the information. In this repository we also use "squash and merge" to keep the commit history from becoming to cluttered, where possible, so please try to group similar changes accordingly. 

I accept this pull request as an exception but normally it is not necessary to make separate commits in one pull request for the same adjustment in different files (for example wrapping `Files.newOutputStream` in `try-with-resources` in `MetsKitodoValidatorTest`, `MetsKitodoWrapperTest` and several other files in this pull request.)

stweil:
Related issue: #5997.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -80,8 +79,7 @@ public class KitodoServiceLoader<T> {
     private static final String JAR = "*.jar";
     private static final String ERROR = "Classpath could not be accessed";
 
-    private static final Path SYSTEM_TEMP_FOLDER = FileSystems.getDefault()
-            .getPath(System.getProperty("java.io.tmpdir"));
+    private static final Path SYSTEM_TEMP_FOLDER = Paths.get(System.getProperty("java.io.tmpdir"));
```
henning-gerhardt, line 82: According to https://stackoverflow.com/a/70943708 it is a false-positive. There is nothing wrong with the calls itself. The replacement is maybe smaller but calls indirect the same code as already done.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -142,7 +140,7 @@ public List<T> loadModules() {
      * they can be used in all frontend files
      */
     private void loadBeans() {
-        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+        Path moduleFolder = Paths.get(modulePath);
```
henning-gerhardt, line 143: See comment above.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -197,7 +195,7 @@ private void loadBeans() {
      */
     private void loadFrontendFilesIntoCore() {
 
-        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+        Path moduleFolder = Paths.get(modulePath);
```
henning-gerhardt, line 198: See comment above.

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
```diff
@@ -382,7 +380,7 @@ private File findFile(String name, File folder) throws FileNotFoundException {
      * earlier class loader created at an earlier time.</p>
      */
     private void loadModulesIntoClasspath() {
-        Path moduleFolder = FileSystems.getDefault().getPath(modulePath);
+        Path moduleFolder = Paths.get(modulePath);
```
henning-gerhardt, line 383: See comment above.

Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
```diff
@@ -40,12 +43,17 @@ public static void init() {
...
-        mockStatic(ParameterAPI.class);
+        mockedParameterAPI = mockStatic(ParameterAPI.class);
         when(ParameterAPI.values())
                 .thenReturn(new ParameterAPI[] {ParameterAPI.DIR_MODULES, ParameterAPI.DIR_PROCESSES,
                                                 ParameterAPI.DIR_XML_CONFIG, NONE });
     }
 
+    @AfterAll
+    public static void tearDown() {
+        mockedParameterAPI.close();
```
henning-gerhardt: In your change in class `MediaPartialFormTest` you are guarding this class with a null pointer check. Why not here or why there? Both use the same calling (creating in `@BeforeAll` and closing in `@AfterAll`).

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
```diff
@@ -141,8 +141,7 @@ URI createMetaDirectory(URI parentFolderUri, String directoryName) throws IOExce
             logger.info("Metadata directory: {} already existed! No new directory was created", directoryName);
         } else {
             CommandService commandService = ServiceManager.getCommandService();
-            String path = FileSystems.getDefault()
-                    .getPath(ConfigCore.getKitodoDataDirectory(), parentFolderUri.getRawPath(), directoryName)
+            String path = Paths.get(ConfigCore.getKitodoDataDirectory(), parentFolderUri.getRawPath(), directoryName)
```
henning-gerhardt, line 143: * See comment above regarding the replacement `FileSystem.getDefault().getPath()` with `Paths.get()`.
* Replacing the let the import statement for `java.nio.file.FileSystems` unused and should be removed too.

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
```diff
@@ -48,19 +49,30 @@ public class MediaPartialFormTest {
...
         MockedStatic<PrimeFaces> primefacesSingleton = Mockito.mockStatic(PrimeFaces.class);
         primefacesSingleton.when(PrimeFaces::current).thenReturn(primeFaces);
     }
 
+    /**
+     * Clean up static mocks.
+     */
+    @AfterAll
+    public static void cleanupTestClass() {
+        if (ajaxMockedStatic != null) {
```
henning-gerhardt: * We are using `Objects.nonNull()` calls in favour for the `null` comparing.
* Your change in `KitodoConfigTest` is without a null pointer guard. Why you are adding this guard here?

henning-gerhardt (changes requested):
* Most / all fixed "memory leaks" are inside tests which are executed only while tests are executed and not while the application itself is running. So the benefit is small or not really noticeable.
* The replacement of `FileSystems.getDefault().getPath()` with `Paths.get()` is a replacement without any benefit nor did fix it some kind of memory leaks. `Paths.get()` itself is calling `FileSystems.getDefault().getPath()` and according to https://stackoverflow.com/a/70943708 it is false positive by scanning with Coverity or Sonar.

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
```diff
@@ -48,19 +49,30 @@ public class MediaPartialFormTest {
...
     /**
      * Initialize test class.
      */
     @BeforeAll
     public static void initTestClass() {
         // mock frontend update calls
-        Mockito.mockStatic(Ajax.class);
+        ajaxMockedStatic = Mockito.mockStatic(Ajax.class);
         PrimeFaces primeFaces = mock(PrimeFaces.class);
         MockedStatic<PrimeFaces> primefacesSingleton = Mockito.mockStatic(PrimeFaces.class);
```
henning-gerhardt: Coverity did not found this `autocloseable` like the other one?

```java
MockedStatic<PrimeFaces> primefacesSingleton = Mockito.mockStatic(PrimeFaces.class);
```

can be replaced with

```java
try (MockedStatic<PrimeFaces> primefacesSingleton = Mockito.mockStatic(PrimeFaces.class)) {
    primefacesSingleton.when(PrimeFaces::current).thenReturn(primeFaces);
}
```

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
stweil: The replacement is significantly smaller and very common in the code (224 occurrences before this PR), and it satisfies the static code analyzers.

The old call `FileSystems.getDefault` only occurs 5 times in the whole code, and it triggers a warning message which is nagging. Even false positives are nagging and must be handled somehow. I think the handling in this PR is the best choice.

stweil: Why:

- FileSystems.getDefault() returns a FileSystem instance that represents the default file system
- This FileSystem is a resource that should be properly managed (closed after use in some contexts)
- Codacy/Coverity flags this as CWE-404 (resource leak) because the FileSystem object could be lingering
- Paths.get() internally calls FileSystems.getDefault().getPath() but does not expose the FileSystem reference, so the linter doesn't complain
- Both produce the identical Path object — the change is purely to silence the resource-leak warning

Summary: Same result, but Paths.get() is the cleaner API that avoids triggering the linter's resource-leak detection on the intermediate FileSystem object.

henning-gerhardt: It is a false positive detected and described in my link. I'm not discussing it. Feel free to get an other person to review this pull request. I'm not doing any reviews on your pull request yet and in future.

Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
stweil: The null check could be added here, too. Should I add it?

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
stweil: The null check is a defensive pattern for two reasons:

1. If @BeforeAll fails partway through: initTestClass() could throw an exception before ajaxMockedStatic = Mockito.mockStatic(Ajax.class) completes, leaving the field null. Without the null check, close() would throw NullPointerException.
2. JUnit guarantees @AfterAll runs even if @BeforeAll fails: This means cleanup could run when the mock was never initialized.

stweil:
```suggestion
        if (Objects.nonNull(ajaxMockedStatic)) {
```

stweil: This change was suggested in the review. The pattern `Objects.nonNull` creates a longer code line and requires an additional import statement, but is nevertheless used about 8 times more often in the whole code than the shorter form with `!= null`.

Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
stweil:
```suggestion
        if (Objects.nonNull(mockedParameterAPI)) {
            mockedParameterAPI.close();
        }
```

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
solth: Why don't you add another commit that implements this change request instead of just adding a suggestion? It is your own pull request, after all.

Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
```diff
@@ -22,13 +22,16 @@
 
 import java.util.NoSuchElementException;
```
stweil, line 25:
```suggestion
import java.util.Objects;

```

Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
stweil: Note that using `Objects.nonNull` also requires an import statement for `Objects`.

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
```diff
@@ -23,6 +23,7 @@
 import java.util.LinkedList;
```
stweil, line 25:
```suggestion
import java.util.Objects;

```

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
stweil: The try-with-resources would close primefacesSingleton before any tests run, not after. That's the opposite of what we need.

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
stweil: I did not do this because I wanted to hear another opinion. After all, the variant which I used initially occurs more than 100 times in the code, and the proposed alternative occurs more often, but has several disadvantages. Just say that you want it nevertheless, and I'll add the commit.

Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
solth: Yes, please!

Kitodo-API/src/main/java/org/kitodo/serviceloader/KitodoServiceLoader.java
solth: Even if it's a false positive, if the replacement does not break anything and is functionally identical (because it calls the same code internally) I think the reasons @stweil presents here are convincing. Shorter calls and unified approaches in the code to implement the same tasks (number of occurrences mentioned above) may be _small_ improvements, but they are improvements nontheless, IMHO.

Kitodo-API/src/test/java/org/kitodo/config/KitodoConfigTest.java
solth: Noted, that's not a real problem, IMO.

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
solth: Good point! That unused `import` statement should be removed as well.

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
solth: Please use `Objects.nonNull` instead of `!=`. That is established in our code base and I would prefer future changes to adhere this.

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
```diff
@@ -48,19 +49,34 @@ public class MediaPartialFormTest {
...
+        primefacesSingleton = Mockito.mockStatic(PrimeFaces.class);
         primefacesSingleton.when(PrimeFaces::current).thenReturn(primeFaces);
     }
 
+    /**
+     * Clean up static mocks.
+     */
+    @AfterAll
+    public static void cleanupTestClass() {
+        if (ajaxMockedStatic != null) {
```
solth:
```suggestion
        if (Objects.nonNull(ajaxMockedStatic)) {
```

solth (changes requested):

Kitodo/src/main/java/org/kitodo/production/services/file/FileService.java
stweil:
```suggestion
```

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
stweil: This suggestion was a duplicate of my previous suggestion which was applied meanwhile.

Kitodo/src/test/java/org/kitodo/production/forms/dataeditor/MediaPartialFormTest.java
```diff
@@ -48,19 +50,34 @@ public class MediaPartialFormTest {
...
 
+    /**
+     * Clean up static mocks.
+     */
+    @AfterAll
+    public static void cleanupTestClass() {
+        if (Objects.nonNull(ajaxMockedStatic)) {
+            ajaxMockedStatic.close();
+        }
+        if (primefacesSingleton != null) {
```
solth:
```suggestion
        if (Objects.nonNull(primefacesSingleton)) {
```

solth (changes requested):

solth (approved):

Issue #6208, opened 2024-08-30
Fix security issues and warnings


stefanCCS:
## Description

For KITODO.PRODUCTION the security issues which are reported [on the security tab](https://github.com/kitodo/kitodo-production/security/code-scanning) should be removed.
This is a similar task as we have done already for KITODO.PRESENTATION (see [https://github.com/kitodo/kitodo-presentation/issues/893](https://github.com/kitodo/kitodo-presentation/issues/893).

At this moment this issue here is just prepared with the aim to have it available for next development fund round.
It might be needed to update this content here, when the next round of development fund is started.

stefanCCS:
@solth : Please consider to create label "development fund 2025", change the label on this issue to that (2025), and assign this new label for the issue type "task for development fund".

henning-gerhardt:
> ## Description
> 
> For KITODO.PRODUCTION the security issues which are reported [on the security tab](https://github.com/kitodo/kitodo-production/security/code-scanning) should be removed.

Your mentioned link is only available for certain persons which have the right to see them. But independent of this link: you can see all the issues if you fork the repository and enable the security scanning from GitHub. The security issues are not solved by removing them from the list they are only "hidden" and still existing and should be solved instead of ignored. Hiding them until and presenting them only on the next round of the development found is in my opinion a bad move.

stweil:
Related: #5997. Some issues are reported by both static analyzers.

stefanCCS:
> > For KITODO.PRODUCTION the security issues which are reported [on the security tab](https://github.com/kitodo/kitodo-production/security/code-scanning) should be removed.
> 
> Your mentioned link is only available for certain persons which have the right to see them. But independent of this link: you can see all the issues if you fork the repository and enable the security scanning from GitHub. The security issues are not solved by removing them from the list they are only "hidden" and still existing and should be solved instead of ignored. Hiding them until and presenting them only on the next round of the development found is in my opinion a bad move.

The purpose of this issue here, is not not intended to "hide" the issue, but to solve these (I have updated the description accordingly.

solth:
12 votes

stweil:
Related fixes for Coverity issues: #7169, #7170.

PR #7097, opened 2026-06-24
Fix regular expressions with DoS vulnerability

thomaslow:
This PR fixes two regular expressions that are somewhat vulnerable to denial-of-service attacks. Users could enter input values that could cause high cpu load for these regular expressions due to `O(n²)` search complexity.

Related issue:
- #6208

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/filters/FilterMenu.java
+++ b/Kitodo/src/main/java/org/kitodo/production/filters/FilterMenu.java
@@ -200,7 +200,6 @@ public class FilterMenu {
      * @return the last category in this filter string
      */
     private String findSuggestionCategory(String input) {
-        // 
         String[] colonSplit = (input + " ").split(":"); 
         String secondLastSplit = colonSplit.length > 1 ? colonSplit[colonSplit.length - 2] : "";
         // find the last word by reversing the string and finding the first word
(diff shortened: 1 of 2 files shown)


BartChris (approved):
Took me some time to wrap my head around the string reversal strategy, but it looks good to me. Filtering still works.

Kitodo/src/main/java/org/kitodo/production/filters/FilterMenu.java
```diff
@@ -187,6 +187,29 @@ public void updateSuggestions(String input) {
...
+     * denial of service attack from user input. Instead, split input at colon symbol by adding 
+     * space at the end such that the potential category is always split as the second last 
+     * array element. Finally, finds last word in string by revesering it and finding first 
+     * word instead, which has linear search complexity.
+     * 
+     * @param input the filter string
+     * @return the last category in this filter string
+     */
+    private String findSuggestionCategory(String input) {
+        //
```
BartChris: Is a comment missing here? Otherwise "//" can go.

solth:
@thomaslow Since the code in the `main` branch and `4.0.x` are still virtually identical and the changed code parts also do not differ between `main` and `3.9.x` I took the liberty to backport your pull request to those both branches as well, since this is a security related improvement.

PR #7100, opened 2026-06-30
Do not abort resaving metadata files when one file cannot be loaded

### LINKED ISSUE #7090: Resaving of metadata files stops if there is a corrupt meta.xml
**Describe the bug**
PR https://github.com/kitodo/kitodo-production/pull/6815 added validation to Kitodo`s meta.xml files. This made all old metadata files invalid. To fix this a Kitodoscript was introduced:

> However, since fixing the METS structure of potentially millions of existing processes manually is not feasible, this pull request also adds a new Kitodo-Script called resaveMetadataFile, which can be applied to multiple selected processes in the process list. It resaves the meta.xml file with the fixed program logic mentioned above, thus fixing the METS structure of existing legacy process and therefore allowing to open them in the metadata editor again.

This works in general, but can silently break (action is stopped) if there is a single broken meta.xml file. In that case it seems like all selected processes are succesfully re-saved, but they are not. The failure with the single XML file is only visible in the log. 

**To Reproduce**
1. Have a broken XML file as meta.xml, e.g. without matching XML tags
2. Try to resave some XML file, including the broken one
3. Process stops with the broken XML file and an error message appears in the log.

**Expected behavior**
I would consider continue processing the non broken files, and skip the broken one. (Error is still shown in the log). 

**Release**
Kitodo 4.0, Main

solth:
As suggested resaving metadata files with the KitodoScript `resaveMetadataFile` should not be aborted when encountering a process whose metadata file cannot be loaded. This PR adjusts the functionality accordingly and adds corresponding tests.

Fixes #7090

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/command/KitodoScriptService.java
@@ -55,6 +55,7 @@ import org.kitodo.production.services.file.FileService;
 import org.kitodo.production.services.image.ImageGenerator;
 import org.kitodo.production.thread.TaskImageGeneratorThread;
 import org.xml.sax.SAXException;
+import org.xml.sax.SAXParseException;
 
 public class KitodoScriptService {
     private static volatile KitodoScriptService instance = null;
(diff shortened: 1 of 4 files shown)


BartChris (approved):
This fixes the issue! Thanks a lot for adding the tests as well.

Issue #4646, opened 2021-09-07
Replace LDAP for Samba authentication


matthias-ronge:
## Description

I asked [a question on Ask Ubuntu](https://askubuntu.com/q/1315481). As a side note, I was told:

> you should be aware that Samba is actively working on removing SMBv1, this will mean that you will no longer be able to use openldap with Samba. This will not happen at once, it may be a year or so, but it will happen, so I suggest you start planning to upgrade to Samba AD or similar

This will mess up our whole system of using LDAP, with LDAP groups, for creating Samba users.

## TODO:

It seems to me, that the simplest solution is: to do without LDAP authentication for file upload access. Instead, the `script_createUserHome` should be passed the user password in clear text so that the script can configure file upload access using system commands. A second `script_changeUserPassword` would need to be added, which is run when a user changes their password so that the script can catch up with the password change for file upload access.

This does **not affect the options for authentication** against LDAP, which will still be possible. But, on the other hand, it would also make it possible for Production to manage file upload access, even if the LDAP is a non-writable, company-wide directory (that does not support a Samba scheme, for example Microsoft Active Directory), which is currently not supported.

## Estimated Costs and Complexity

This is a **mid-ranged** project with 5 PT.

henning-gerhardt:
Thank you for sharing this information. 

- A samba AD for the Kitodo.Production environment could be difficult if you have already a Windows AD environment. I don't know if there a problems in a local network with two AD environments.
- Moving current users to a Windows AD could be possible but then must you must provide the access to the image files with a samba configuration which is using the Windows AD. I don't know if this is possible and how a good and clean integration should look like.
- Providing the image files through a WebDAV system which is using the LDAP for authentification could be an useful solution. In SLUB we are using ownCloud (maybe in future Nextcloud) to provide access for some partners from outside to up- and download image files. But: Windows is more and more removing the support for WebDAV :-( If you want to use a fast WebDAV client you should not relay on the Windows WebDAV client.

matthias-ronge:
It seems that the best solution would be to get all of the writable LDAP stuff out of Production. Instead, the `script_createUserHome` invokes `useradd` and `samba-tool user create` to create the user locally and create the file share. A second `script_updateUserPassword` could handle password changes. This would save us a lot of complexity in the Production server installation process.

[...]

PR #7111, opened 2026-07-06
Support better algorithms for password hashing and add encryption with salt

stweil:
This fixes the weak password hashing which is reported by Codacy.

Test cases for password hashing were missing and are added, too.


stweil:
This is currently a draft pull request until it was tested with existing and new installations.

solth:
As far as I know @thomaslow is working on this as part of #6208. (@stweil I think you were in the development fund meeting where this was commisioned) Maybe you could coordinate your efforts with him so this work isn't done twice.

stweil:
I know, and that's one of the reasons why I marked it as draft and did not invest more efforts. @thomaslow, feel free to re-use this PR or ignore it, whichever fits best for you.

Kitodo/src/main/java/org/kitodo/production/ldap/LdapUser.java
```diff
@@ -81,49 +84,64 @@ public void configure(User user, String inPassword, String inUidNumber)
...
-             * Encryption of password und Base64-Enconding
-             */
-
-            String passwordEncrytion = ldapGroup.getLdapServer().getPasswordEncryption().getTitle();
-
-            MessageDigest md = MessageDigest.getInstance(passwordEncrytion);
-            md.update(inPassword.getBytes(StandardCharsets.UTF_8));
-            String encodedDigest = new String(Base64.encodeBase64(md.digest()), StandardCharsets.UTF_8);
-            this.attributes.put("userPassword", "{" + passwordEncrytion + "}" + encodedDigest);
+    private void setSambaPasswords(String inPassword, MD4Digest digester) throws NoSuchAlgorithmException {
```
BartChris, line 102: If i understand correctly, the long term goal is to remove write access to the LDAP from Kitodo production, so enhancing password security would benefit an writable-LDAP implementation, which we eventually want to replace

https://github.com/kitodo/kitodo-production/issues/4646#issuecomment-915043102

stweil: Thank you, this is an important information. So the question remains what "long term" means. Maybe removing write access can be done early to fix the issue?

stweil:
I noticed that checkstyle complains: `Method name 'hashPbkdf2WithHmac' must match pattern '^[a-z][a-zA-Z]*$'`. Why are digits forbidden in method names? Of course it's easy to avoid them, but in some cases this can result in names which are worse.

Kitodo/src/main/java/org/kitodo/production/ldap/LdapUser.java
BartChris: https://github.com/kitodo/kitodo-production/issues/4646 has been assessed in multiple development fund rounds but has not been funded yet. Newer Debian/Ubuntu versions are not only moving to newer Samba versions but keep on deprecating related functionality in newer Kernels. So i do not know if this is already urgent or if Debian 13/Ubuntu 26 still support the current architecture.

stweil:
This PR was superseded by #7188, so it is no longer needed.

thomaslow:
@stweil No, pull request #7188 does not change how passwords are encoded for LDAP. It only affects how passwords are stored in the database. This PR is still relevant. 

I also thought about your changes. I think a good approach to achieve both security and backwards-compatibilty with samba is to introduce a new LDAP server setting "save credentials for samba", which enables or disables storing user credentials via the old legacy LM/NTLM hashing methods. This way, admins that want to use LDAP without Samba can disable these insecure hash methods, but still use LDAP. Only when samba support is needed, the old methods are used to store user credentials in LDAP. 

What do you think?

stweil:
@thomaslow, thank you for correcting me. I reopened the PR now. It depends on the timeline for the "long term goal" mentioned above whether more efforts should be made here or not. So this is a question which I cannot answer, but which might be addressed by the release management (@solth).

thomaslow:
@stweil I successfully tested our current setup of Kitodo.Production / LDAP / Samba with Ubuntu 26.04. So, deployments that use Samba should be able to do so until 2031. In my opinion, that makes it worth it.

Also, from what I heard from @henning-gerhardt, he is using LDAP with WebDav over Apache, which would also benefit from this PR.

I will have a look at your changes soon and probably make some changes myself, if this is ok with you.

stweil:
> I will have a look at your changes soon and probably make some changes myself, if this is ok with you.

Yes, sure. Thank you for all your work.

thomaslow:
@stweil I added a new PR #7265. In my opinion, this PR can now be closed. Thank you for looking into this and providing a starting point.

PR #7112, opened 2026-07-06
ci: Replace Java distribution adopt by temurin

stweil:
This fixes build notices:

    AdoptOpenJDK has moved to Eclipse Temurin https://github.com/actions/setup-java#supported-distributions
    please consider changing to the 'temurin' distribution type in your setup-java configuration.


stweil:
See https://github.com/actions/setup-java#supported-distributions and https://adoptium.net/.

solth (approved):

PR #7117, opened 2026-07-07
Fix project deletion

### LINKED ISSUE #7116: Projects cannot be deleted
**Describe the bug**
When trying to delete a simple project (with just a name and not other settings), we already get an error on delete. 
The UI shows `Error deleting Project`.

The Hibernat exception is:
```
jakarta.persistence.EntityNotFoundException: deleted object would be re-saved by cascade (remove deleted object from associations): [org.kitodo.data.database.beans.Project#28]
```

This probably has to do with failed cleanup of related entities. 

**To Reproduce**
1. Go to project page
2. Create a simple project with just a a project title
3. Try to delete it

**Expected behavior**
Project should be deleted.

**Release**
Kitodo 4.0, main

BartChris:
Fixes: https://github.com/kitodo/kitodo-production/issues/7116

Deleting the project by ID and not by passing the object fixes the issue for me. The problem seem to stem from the usage of the `session.merge(baseBean)` call in the `remove-`method, 

https://github.com/kitodo/kitodo-production/blob/d4c3c4f67fd3fae0ee99a069822b1d9025393bb4/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java#L130

which leads to an Hibernate exception because it tries to resave already deleted entities. The usage of the `merge`-pattern has been discussed in general already (https://github.com/kitodo/kitodo-production/issues/5813#issuecomment-3628433955) and i think Hibernate 6 became more strict here in some regard. 

Deletion by ID uses another code path which does not rely on `session.merge()`:

https://github.com/kitodo/kitodo-production/blob/d4c3c4f67fd3fae0ee99a069822b1d9025393bb4/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java#L380

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/ProjectServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/ProjectServiceIT.java
@@ -17,7 +17,6 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-
 import java.util.Collections;
 import java.util.List;
 
--- a/Kitodo/src/test/java/org/kitodo/production/services/data/ProjectServiceIT.java
+++ b/Kitodo/src/test/java/org/kitodo/production/services/data/ProjectServiceIT.java
@@ -17,6 +17,7 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


henning-gerhardt (changes requested):
Changes looks good and it is now possible again to delete projects without any processes attached.

Why did you choose `ProcessListViewIT` as base for the added tests instead of `ProjectServiceIT`? `ProjectService.delete()` call is used in three places `DesktopForm`, `ProjectEditView` and `ProjectListView` and is used very similar in all three places. I would move the tests to the more general `ProjectServiceIT` class as they are not specific for `ProcessListView` usage.

Kitodo/src/test/java/org/kitodo/production/services/data/ProjectServiceIT.java
```diff
@@ -17,6 +17,7 @@
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+
```
henning-gerhardt: Unnecessary empty line

henning-gerhardt (changes requested):
Looks good and is still working. Only a not needed empty single line was added.

henning-gerhardt (approved):

solth:
@BartChris can you guess what I am going to ask you?

PR #7118, opened 2026-07-07
Filter out inactive projects in desktop view

### LINKED ISSUE #7108: Deactivated projects visible in desktop project widget
**Describe the bug**
We created a new project and set it to `deactivated` as some properties are not final and assigned already this deactivated project to some users which will interact with this new project. As a few new users are created too we checked this users (has right to create processes in this new project) and discovered that the could create new processes for this deactivated project over the project desktop widget. Creating new processes for this deactivated project over the project page itself is not possible.

**To Reproduce**
1. Create a new project as `deactivated` or deactivate an existing project
2. Re-log as an user which can create processes in this deactivated project
3. Click on the `+` symbol in the project desktop widget and see the create new process dialog
4. It is not possible for this user to create a new process over the project list as the `+` symbol is grayed out

**Expected behavior**
Deactivated projects should not be visible / listed in the project desktop widget as this widget is used by regular users without administrative or project management assignment.

**Release**
3.8.4, 3.8.6, 3.9.2, 3.9.3-SNAPSHOT (3.9.x development), 4.0.0, 4.0.1-SNAPSHOT (4.0.x development), main branch

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/7108

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/DesktopForm.java
@@ -161,7 +161,7 @@ public class DesktopForm extends BaseForm {
     public List<Project> getProjects() {
         try {
             if (ServiceManager.getSecurityAccessService().hasAuthorityToViewProjectList() && projectList.isEmpty()) {
-                projectList = ServiceManager.getProjectService().loadData(0, 10, "title", SortOrder.ASCENDING, null);
+                projectList = ServiceManager.getProjectService().loadActiveProjects(0, 10, "title", SortOrder.ASCENDING);
             }
         } catch (DAOException | JsonException e) {
             Helper.setErrorMessage(ERROR_LOADING_MANY, new Object[] {ObjectType.PROJECT.getTranslationPlural() },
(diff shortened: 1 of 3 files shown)


henning-gerhardt (changes requested):
Change is working and code looks good but there are no tests to cover this behaviour even in future.

BartChris:
I added an integration test.

henning-gerhardt (approved):

solth:
Just a remark: I think we should discuss how deactivated objects in Kitodo should be marked in general. I have opened a corresponding discussion [here](https://github.com/kitodo/kitodo-production/discussions/7126).

@BartChris this should also be backported to `4.0.x` and, if feasible without extra effort, even to `3.9.x`!

henning-gerhardt:
Just a notice - no change needed and I did not see it on my review: The added integration test `DesktopFormIT` class is fine as the used method `loadActiveProjects` of `ProjectService` class is indirect used only in `DesktopForm` class but the method is a more general method which could be used everywhere in the project and the test then should be moved to the `ProjectServiceIT` test class.

solth:
@BartChris maybe you could rebase just one more time against `main`, since that branch now contains your fix for project deletion? That would show us if both changes concerning projects work well together.

BartChris:
> @BartChris maybe you could rebase just one more time against `main`, since that branch now contains your fix for project deletion? That would show us if both changes concerning projects work well together.

sure, i did the rebase.

PR #7119, opened 2026-07-08
Duplicate missing settings and add missing tests

henning-gerhardt:
Fixes #7114 

All existing settings of project and folder should now be clone / duplicated / copied.

Backport to 4.0.x and 3.9.x branches should be possible.

PR #7134, opened 2026-07-29
[3.9] Duplicate missing settings and add missing tests

henning-gerhardt:
Back port of #7119 with test method for duplicated project already moved to `ProjectServiceIT`.


BartChris (approved):
Works for me and loos good as well!

PR #7137, opened 2026-07-30
[3.9] Update OmniFaces to latest bugfix version 3.14.21

solth:
Updates OmniFaces version to latest bugfix version for Kitodo `3.9.x`. As that Kitodo version does not yet use OmniFaces 4, I updated to the latest bugfix release of OmniFaces 3 instead, for which the known vulnerabilites are [fixed](https://mvnrepository.com/artifact/org.omnifaces/omnifaces/versions) as well. I checked the parts of the application using OmniFaces and they all continue to work as they did before this update.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -612,7 +612,7 @@ from system library in Java 11+ -->
             <dependency>
                 <groupId>org.omnifaces</groupId>
                 <artifactId>omnifaces</artifactId>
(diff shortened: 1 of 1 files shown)


thomaslow (approved):
Works, too. No regression with regards to previous Omnifaces issue.

PR #7152, opened 2026-08-03
[3.9] Reload user entity before attaching it to the project

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/7135


solth:
@BartChris if backporting the fix to `3.9.x` proofs to be more complicated than expected I think we can skip it!

BartChris:
Will skip the 3.9 fix!

Issue #6466, opened 2025-03-21
Can not assign import configuration to ten thousands of processes


henning-gerhardt:
**Describe the bug**
We must assign the import configuration to hundred thousand of processes but this is not possible. We can select the to be used import configuration and waiting  a long time but the finish dialog never appear nor where the values set in database.

**To Reproduce**
1. Select a few ten thousand processes (for 4,000 processes is working)
2. Select "Set import configuration"
3. Select the import configuration which should be set
4. Wait for ever to see the finish dialog

**Expected behavior**
Setting the import configuration even to ten thousand of processes should work

**Release**
3.8.2-SNAPSHOT

PR #7156, opened 2026-08-04
Use bulk operation for updating import configuration

BartChris:
Adresses: https://github.com/kitodo/kitodo-production/issues/6466

@henning-gerhardt I tried to adress it, maybe you can check if this improves the situation.

### DIFF
--- a/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
+++ b/Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
@@ -509,6 +509,30 @@ public abstract class BaseDAO<T extends BaseBean> implements Serializable {
     }
 
     /**
+     * Executes an update or delete HQL query.
+     *
+     * @param query
+     *            HQL query to execute
+     * @param parameters
+     *            named query parameters
+     * @return number of affected database rows
+     */
+    public int executeUpdate(String query, Map<String, Object> parameters) throws DAOException {
+        try (Session session = HibernateUtil.getSession()) {
+            debugLogQuery(query, parameters);
+            Transaction transaction = session.beginTransaction();
+            Query<?> updateQuery = session.createQuery(query);
+            addParameters(updateQuery, parameters);
+            int updatedRows = updateQuery.executeUpdate();
+            session.flush();
+            transaction.commit();
+            return updatedRows;
+        } catch (PersistenceException e) {
+            throw new DAOException(e);
+        }
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


henning-gerhardt:
@BartChris : I tried your changes and now setting the import configuration is fast. I tried it only up to 10.000 processes so I did not hit the search limit.

Small note to your changes itself: it looks like that your editor / IDE is setting the indent on many places to 8 spaces instead of only 4. My IDE IntelliJ IDEA had added a new option (I don't know how long this option already was there but it was new to me) called "continuation indent" which was set by default to 8 spaces. After I changed it to 4 spaces the indents is now set correct in all cases so far as I could discover.

BartChris:
Thanks a lot @henning-gerhardt, i found the setting to change it to 4.

solth:
@henning-gerhardt would you mind reviewing this pull request?

henning-gerhardt:
@solth : I can review it as I already test it.

solth:
> @solth : I can review it as I already test it.

Thank you!

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
```diff
@@ -508,6 +510,29 @@ List<T> retrieveAllObjects(Class<T> cls) throws DAOException {
...
+    /**
+     * Executes an update or delete HQL query.
+     *
+     * @param query
+     *            HQL query to execute
+     * @param parameters
+     *            named query parameters
+     * @return number of affected database rows
+     */
+    public int executeUpdate(String query, Map<String, Object> parameters) throws DAOException {
```
henning-gerhardt: Return value of this method is (currently) not used. Maybe we should remove it until we need it? If we stay with this return value than the description should be changed as the "MutationQuery.executeUpdate()" describe the return value as `the number of affected entity instances (may differ from the number of affected rows)`.

henning-gerhardt (commented):
The change improve the setting of import configuration to really fast in compare to the time consuming implementation before.

Only a maybe small change is needed as a method is returning a value which is currently not in use.

Kitodo-DataManagement/src/main/java/org/kitodo/data/database/persistence/BaseDAO.java
BartChris: You are right, i changed the method to void.

henning-gerhardt (approved):

solth:
@BartChris would you mind providing a backport for Kitodo `4.0.x`?

PR #7157, opened 2026-08-05
Add missing explicit semicolons

solth:
Add missing explicit semicolons in JavaScript files.

(Codestyle issue detected by Codacy.)

**Issue description**:
"Automatic semicolon insertion can lead to unexpected behavior or bugs because JavaScript may interpret line breaks in ways that differ from developer intent. Consistently using or omitting semicolons reduces the risk of such errors and improves code clarity."


Kitodo/src/main/webapp/WEB-INF/resources/js/tabView.js
```diff
@@ -31,4 +31,4 @@ kitodo.tabView.onTabChange = function(tabId) {
     kitodo.removeQueryParameter('sortField');
     kitodo.removeQueryParameter('sortOrder');
     kitodo.removeQueryParameter('filter');
-}
\ No newline at end of file
+};
```
stweil: Each source file should terminate with a LF which is missing here. I suggest to add it as line 33 is already touched by this PR.

stweil (changes requested):

stweil (approved):
Looks good, thank you!

PR #7164, opened 2026-08-07
Avoid unnecessary case change

solth:
Fixes code style issue detected by Codacy.

**Issue description from Codacy**:
"Using toUpperCase() or toLowerCase() before equals() is less efficient and less readable compared to using equalsIgnoreCase(). The latter is optimized for case-insensitive comparison and leads to cleaner code."

**Fix by**:
"Replace expressions that use toUpperCase() or toLowerCase() followed by equals() with a direct call to equalsIgnoreCase()."


stweil (approved):
Looks good, thanks. Similar cases which could use `compareToIgnoreCase` exist and could be handled in a separate PR.

PR #7166, opened 2026-08-07
Fix exclusion query

### LINKED ISSUE #7165: 4.0.0 "Alle Treffer" funktioniert nicht, sobald man einen Treffer abwählt
Wenn man in einer Vorgangsliste "Alle Treffer" wählt, dann händisch 1-n Vorgänge abwählt und ein Skript ausführt, wird dieses nicht angewendet. Es passiert gar nichts. Die Skripte funktionieren nur, wenn tatsächlich "Alle Treffer" angewählt sind oder alle einzeln ausgewählt werden.
"Alle Treffer dieser Seite" funktioniert hingegen wie erwartet. 
In Kitodo 3.9.X lief noch alles wie erwartet.

----

If you select ‘All results’ in a task list, then manually deselect 1–n tasks and run a script, the script is not executed. Nothing happens at all. The scripts only work if ‘All results’ is actually selected or if all tasks are selected individually.
‘All results on this page’, on the other hand, works as expected.
In Kitodo 3.9.X, everything still worked as expected.

~Kitodo-Team Staatsbibliothek zu Berlin

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/7165

This PR tries to adress the problem, that when all processes in the list are selected and then a part of them are deselected the constructed query does not deliver the correct result. The reason is, that we apply both a `NOT IN` as well as a `IN` restriction and both receive the same placeholder names:

```sql
process.id NOT IN (:id)
...
AND process.id IN (:id)
```

Since both restrictions refer to the same placeholder, one parameter value overwrites the other, resulting in contradictory query conditions and no matching processes. I tried to fix that by naming the `NOT IN` placeholder different.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java
@@ -79,6 +79,18 @@ public class BeanQuery {
         sorting = Pair.of(varName + ".id", "ASC");
     }
 
+    private String uniqueParameterName(String field) {
+        String baseName = varName(field);
+        String parameterName = baseName;
+        int count = 2;
+
+        while (parameters.containsKey(parameterName)) {
+            parameterName = baseName + count++;
+        }
+
+        return parameterName;
+    }
+
     /**
      * Requires that the hits in a specific field must have a specific value.
      * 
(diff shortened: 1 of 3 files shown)


BartChris:
I think the better aproach is to make sure that all clauses added as "IN" or "NOT IN" get unique parameter naming. I added a private helper which enforces that.

matthias-ronge (approved):
Reads correct and IMHO solves the described bug.

solth:
@BartChris as this was reported for "4.0.0", would you mind providing a backport for that version as well?

PR #7167, opened 2026-08-07
Use parameterized log messages instead of string concatenations

solth:
Replaces string concatenation in log4j logs with parametrizd log messages.

**Issue description from IDE:**
"Non-constant concatenations are evaluated at runtime even when the logging message is not logged; this can negatively impact performance. It is recommended to use a parameterized log message instead, which will not be evaluated when logging is disabled."


thomaslow (approved):
Although I don't think performance is a problem for these log statements, I agree that it is a good pattern to submit parameters as arguments instead of doing string concatenation. I checked, but unfortunately, checkstyle doesn't seem to support this scenario. Otherwise it would have been nice to add a checkstyle rule for this.

All changes look good to me. Kitodo.Production compiles and runs. I didn't test each log statement, though.

PR #7169, opened 2026-08-10
Prevent XML External Entity injection and DTD expansion (Coverity)

stweil:
Introduce a shared XMLSecurity utility class providing hardened DocumentBuilder, Transformer, Schema, SAXParser and XMLInputFactory instances that reject DOCTYPE declarations and disable external entity resolution to prevent XML External Entity (XXE) and entity expansion attacks.

Use the hardened factories in all XML parsing and transformation code paths and null-check the Transformer returned by the factory.

CIDs: 486979, 486985, 486986, 486976, 486990, 487007, 454260, 454251, 431632, 431623, 431616, 431601, 431588, 431586, 431585, 431567, 501288, 501287, 501299, 501300, 501305, 501296, 501292, 501285, 415127, 415247, 415087, 501286, 501294, 501303, 503907, 503911


Kitodo-Validation/src/main/java/org/kitodo/validation/filestructure/FileStructureValidation.java
```diff
@@ -87,13 +88,18 @@ private ValidationResult validateStreamSource(StreamSource source, Validator val
...
         FileStructureValidationErrorHandler xmlValidationErrorHandler = new FileStructureValidationErrorHandler();
-        SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+        SchemaFactory schemaFactory = XMLSecurity.newSchemaFactory();
         Source[] sources = new Source[xsdFilePaths.size()];
         for (int i = 0; i < xsdFilePaths.size(); i++) {
             sources[i] = new StreamSource(new File(xsdFilePaths.toArray(new URI[0])[i]));
         }
         Schema schema = schemaFactory.newSchema(sources);
         Validator xmlValidator = schema.newValidator();
+        try {
```
solth: Would it make sense to move the instantiation of the validator to the new `XMLSecurity` class as well?

solth (changes requested):
Thanks for this pull request. I think it is a good idea to introduce a central class providing secure instances of different XML transformer factories. 

Due to the security relevance of these factories I think you should add tests ensuring the hardening works as intended. Maybe you can take the tests added by @tonghuaroot in #7072 as a template.

Also, please rebase the branch of this pull request against the current `main` branch.

Kitodo-Validation/src/main/java/org/kitodo/validation/filestructure/FileStructureValidation.java
stweil: Yes, this is now implemented.

stweil:
> Due to the security relevance of these factories I think you should add tests ensuring the hardening works as intended. Maybe you can take the tests added by @tonghuaroot in #7072 as a template.
> 
> Also, please rebase the branch of this pull request against the current `main` branch.

Done now. I also started a review by Copilot.

Kitodo-API/src/main/java/org/kitodo/utils/XMLSecurity.java
```diff
@@ -0,0 +1,207 @@
...
+     *
+     * <p>DOCTYPE declarations must be rejected by the caller, e.g. by feeding the
+     * input through newSecureSource(), since the JAXP Validator does not
+     * universally support disallow-doctype-decl.
+     *
+     * @param schema compiled XML schema
+     * @return hardened Validator
+     * @throws SAXException if the Validator cannot be created
+     */
+    public static Validator newSecureValidator(Schema schema) throws SAXException {
```
solth: Since you are already catching the possible `SAXException` sub types in the method you don't need to declare it in the signature anymore

```suggestion
    public static Validator newSecureValidator(Schema schema) {
```

When you apply this suggestion, please adjust the JavaDoc accordingly (instead of a `SAXException` an `IllegalStateException` is thrown if the validator cannot be created)

solth (commented):
Thanks for implementing the change requests and for adding tests. I have just one small remark regarding the code but want to test some basic functions like metadata import, EAD upload and schema validation in a running system to see if the changes actually break any of these functions.

Kitodo-API/src/main/java/org/kitodo/utils/XMLSecurity.java
stweil: Good idea – I added a commit which fixes the function's documentation and signature.

Kitodo-API/src/main/java/org/kitodo/utils/XMLSecurity.java
```diff
@@ -0,0 +1,175 @@
...
+import org.xml.sax.SAXException;
+import org.xml.sax.SAXNotRecognizedException;
+import org.xml.sax.SAXNotSupportedException;
+import org.xml.sax.XMLReader;
+
+/**
+ * Provides factory instances that are hardened against XML External Entity
+ * (XXE) injection and unrestricted document type definitions.
+ */
+public final class XMLSecurity {
```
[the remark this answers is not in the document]
solth: @stweil are you planning on adressing these points as well? If I am not mistatken, I think you can adjust the `SAXBuilder` by calling `.setExpandEntities(false)` on it and by using the hardened `TransformerFactory` from your new `XMLSecurty` class in `XsltHelper`.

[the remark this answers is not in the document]
stweil: The latest commits addressed this and new review results from Copilot.

Kitodo-XML-SchemaConverter/src/test/java/org/kitodo/xmlschemaconverter/XmlSchemaConverterTest.java
```diff
@@ -154,6 +157,27 @@
         assertEquals("Test-Shelflocator", shelfmarksource, "shelfmarksource after conversion is wrong!");
     }
 
+    @Test
+    public void shouldRejectExternalEntitiesInSourceRecord() throws IOException {
+        String canary = "XXE-CANARY-12345";
+        File secret = File.createTempFile("xxe-canary", ".txt");
```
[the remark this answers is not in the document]
stweil: This one is only in test code. Nevertheless commit 02714c87f4c26eae4d428a247dd1f1f6c1844bdc should fix it.

BartChris:
A general comment: Maybe it is also useful in the future to evaluate

https://commons.apache.org/proper/commons-secure-xml/
which just had the 1.0 release.

Kitodo-API/src/test/java/org/kitodo/utils/XMLSecurityTest.java
```diff
@@ -0,0 +1,234 @@
...
+import java.nio.file.Path;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.parsers.SAXParserFactory;
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import javax.xml.transform.Source;
```
solth: This import is now unused and should be removed.
```suggestion
```

Kitodo-API/src/test/java/org/kitodo/utils/XMLSecurityTest.java
```diff
@@ -0,0 +1,234 @@
...
+                "local xs:import must still resolve under ACCESS_EXTERNAL_SCHEMA=file");
+    }
+
+    private void parseWithDocumentBuilder(DocumentBuilderFactory factory, String xml)
+            throws IOException, SAXException, ParserConfigurationException {
+        DocumentBuilder builder = factory.newDocumentBuilder();
+        builder.parse(new InputSource(new ByteArrayInputStream(xml.getBytes(StandardCharsets.UTF_8))));
+    }
+
+    private String readWithStax(XMLInputFactory factory, String xml) throws XMLStreamException {
```
solth: The return value of this method is never used. Thus the signature should be changed to return `void` instead of `String`.

Kitodo-Validation/src/main/java/org/kitodo/validation/filestructure/FileStructureValidation.java
```diff
@@ -47,23 +52,41 @@ public class FileStructureValidation implements FileStructureValidationInterface
...
         Validator xmlValidator = initializeXmlValidator(xsdFiles);
-        return validateStreamSource(new StreamSource(new StringReader(xmlContent)), xmlValidator, "N/A", xsdFiles);
+        return validateStreamSource(createSecureSource(new StringReader(xmlContent)), xmlValidator, "N/A", xsdFiles);
+    }
+
+    private SAXSource createSecureSource(InputStream xmlInput) throws SAXException {
+        try {
+            return XMLSecurity.newSecureSource(xmlInput);
+        } catch (ParserConfigurationException e) {
+            throw new SAXException("Unable to create hardened SAXSource", e);
```
solth: I don't like wrapping the `ParserConfigurationException` inside a `SAXException` here (and in line 85) because the original cause does not have anything to do with the actual XML, but if we do this, we should have unique exception messages to better distinguish between multiple cases
```suggestion
            throw new SAXException("Unable to create hardened SAXSource from InputStream", e);
```

Kitodo-Validation/src/main/java/org/kitodo/validation/filestructure/FileStructureValidation.java
```diff
@@ -47,23 +52,41 @@ public class FileStructureValidation implements FileStructureValidationInterface
...
+        }
+    }
+
+    private SAXSource createSecureSource(Reader xmlInput) throws SAXException {
+        try {
+            return XMLSecurity.newSecureSource(xmlInput);
+        } catch (ParserConfigurationException e) {
+            throw new SAXException("Unable to create hardened SAXSource", e);
```
solth: See above.
```suggestion
            throw new SAXException("Unable to create hardened SAXSource from Reader", e);
```

Kitodo-Validation/src/test/java/org/kitodo/validation/filestructure/FileStructureValidationTest.java
```diff
@@ -72,7 +76,67 @@ public void shouldFailToValidateInvalidXmlString() throws IOException, SAXExcept
...
+            assertFalse(message.contains("XXE-CANARY-12345"),
+                    "Internal entity content must not be resolved or leaked");
+        }
+    }
+
+    @Test
+    public void shouldValidateXmlStringWithNonUtf8EncodingDeclaration() throws IOException, SAXException {
+        Path schema = createPermissiveSchema();
+        String xmlContent = "<?xml version=\"1.0\" encoding=\"UTF-16\"?>\n"
+                + "<root><value>caf\u00e9</value></root>";
```
solth: I think all Java source code files are now guaranteed to be encoded in UTF-8, so we do not need Unicode escape sequences in the code anymore, do we?
```suggestion
                + "<root><value>café</value></root>";
```

Kitodo-XML-SchemaConverter/src/test/resources/identity.xsl
solth, line 1: Until the license issue has been resolved, I think we should include our obligatory license header in all source code files. Please add it here as well.

Kitodo/src/main/java/org/kitodo/export/XsltHelper.java
```diff
@@ -12,53 +12,98 @@
...
 
+    private static final Logger logger = LogManager.getLogger(XsltHelper.class);
+
     private XsltHelper() {
         // private constructor to hide implicit one
     }
 
     /**
-     * Transforms a xml file by xslt and returns the result as string.
+     * Transforms a xml file by xslt and returns the result as string. The input
```
solth:
```suggestion
     * Transforms an xml file by xslt and returns the result as string. The input
```

Kitodo/src/test/resources/xslt/identity.xsl
solth, line 1: Please add obligatory Kitodo license header (see above)

Kitodo-Docket/src/main/java/org/kitodo/docket/ExportXmlLog.java
```diff
@@ -413,8 +413,11 @@ private void prepareMetadataElements(List<Element> metadataElements, boolean use
             Namespace[] namespaces, Namespace xmlns)
             throws IOException, JDOMException {
         HashMap<String, String> fields = getMetsFieldsFromConfig(useAnchor);
+        SAXBuilder builder = new SAXBuilder();
```
solth: Maybe this `SAXBuilder` could also be instantiated in the new `XMLSecurity` to make it reusable, should any future functionalities require a `SAXBuiler` again?

solth (changes requested):

Kitodo-API/src/main/java/org/kitodo/api/externaldatamanagement/XmlResponseHandler.java
```diff
@@ -42,17 +42,17 @@
...
-        documentBuilderFactory.setNamespaceAware(true);
         try {
-            documentBuilderFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
-        } catch (ParserConfigurationException parserConfigurationException) {
-            throw new UndeclaredThrowableException(parserConfigurationException);
+            documentBuilderFactory = XMLSecurity.newDocumentBuilderFactory();
+        } catch (ParserConfigurationException e) {
+            throw new ExceptionInInitializerError(e);
         }
+        documentBuilderFactory.setNamespaceAware(true);
```
solth: Why did you change this? This does not relate to the security aspect and may change intended behaviour.

Kitodo-Docket/src/main/java/org/kitodo/docket/ExportXmlLog.java
stweil: Done now.

Kitodo/src/test/resources/xslt/identity.xsl
stweil: Done now.

Kitodo-XML-SchemaConverter/src/test/resources/identity.xsl
stweil: Done now.

Kitodo-API/src/main/java/org/kitodo/api/externaldatamanagement/XmlResponseHandler.java
stweil: I now reverted the changes in this file.

Kitodo-API/src/main/java/org/kitodo/utils/XMLSecurity.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private static void applyAttribute(TransformerFactory factory, String name, Object value) {
+        try {
+            factory.setAttribute(name, value);
+        } catch (IllegalArgumentException e) {
+            logger.warn("TransformerFactory {} does not support the JAXP property '{}'; "
+                    + "external DTD/stylesheet access cannot be restricted for this vendor. "
+                    + "Input must be parsed via a hardened source.", factory.getClass().getName(), name);
+        }
```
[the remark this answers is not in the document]
stweil: @solth, after the reverted changes in this file, Copilot now reports rightfully that there remains a potential security issue here. It also suggested to upgrade to Saxon 10.6 where a clean fix could be done. I suggest to address this Saxon upgrade and this special fix in a separate pull request. Then the PR here could be merged now.

Kitodo-API/src/main/java/org/kitodo/api/externaldatamanagement/XmlResponseHandler.java
stweil: This revert re-introduced a potential security issue. See my comment below on this.

Kitodo-API/src/main/java/org/kitodo/utils/XMLSecurity.java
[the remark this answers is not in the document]
stweil: Instead of using Saxon 10.6 (or an ever newer release), commons-secure-xml which was mentioned by @BartChris could also be considered as an alternative in the follow-up pull request.

Kitodo-API/src/main/java/org/kitodo/api/externaldatamanagement/XmlResponseHandler.java
solth: So it _was_ related to the security aspect of this pull request after all? In that case we should probably check whether the change does indeed have any negative impact on the behaviour, as I mentioned above, or if it isn't a problem in the general use case of Kitodo.

stweil: Yes, but maybe later. Even the current changes already address a large number of security issues (see CID list above). So merging it should improve the security, and together with the merged #7188 new security scans should show the improvements, but also remaining issues, so these can be addressed in a new PR.

Kitodo-API/src/main/java/org/kitodo/utils/XMLSecurity.java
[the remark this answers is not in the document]
solth: > @solth, after the reverted changes in this file, Copilot now reports rightfully that there remains a potential security issue here.

Which "reverted changes in this file" do you refer to? The reverted change was in `XmlResponseHandler.java`, not in `XMLSecurity.java`. Please explain how you think the namespaces awarness in the former causes the potential issue in the later, described by Copilot above.

solth:
Please address the open point mentioned above in `XMLSecurityTest.java` and rebase one more time against the current `main` branch; if no further issues arise, I will then merge this pull request.

solth (approved):
I think this looks good now, but I am really curious why the deactivated namespace awareness should be related to the new security concern Copilot has stated above. @stweil Maybe you can explain why you think both are connected?

PR #7170, opened 2026-08-10
Avoid null pointer dereferences (Coverity)

stweil:
Check return values that may be null before dereferencing them:

- GalleryPanel.select(): null-check the GalleryMediaContent before resolving its logical structure instead of after
- GalleryPanel.isSelected(): null-check the GalleryStripe returned by getLogicalStructureOfMedia()
- MediaPartialsPanel.getMediaPartialChildDivisionsOfSelection(): null-check the GalleryMediaContent returned by getGalleryMediaContent()
- MassImportForm.handleFileUpload(): null-check the uploaded file
- ProcessService.downloadToHome(): null-check the Process returned by getById()

CIDs: 487003, 486993, 486994, 486980, 486983, 415161


BartChris (approved):
Looks good to me and does not introduce a behavior change as far as i can see.

PR #7171, opened 2026-08-10
Update ActiveMQ client to 6.3.0

henning-gerhardt:
- Update ActiveMQ from 5.19.4 to 6.3.0 which requires a change from JavaX to Jakarta namespace
- Did some code cleanup in the changed ActiveMQ use classes
- Did some tests (closing tasks in most cases) with ActiveMQ server 5.19.x and 6.3.0 without any issue. So updating ActiveMQ server to 6.3.0 is not necessary.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/ActiveMQDirector.java
+++ b/Kitodo/src/main/java/org/kitodo/production/interfaces/activemq/ActiveMQDirector.java
@@ -16,13 +16,13 @@ import java.util.Collection;
 import java.util.Objects;
 import java.util.Optional;
 
-import javax.jms.Connection;
-import javax.jms.DeliveryMode;
-import javax.jms.Destination;
-import javax.jms.JMSException;
-import javax.jms.MessageConsumer;
-import javax.jms.MessageProducer;
-import javax.jms.Session;
+import jakarta.jms.Connection;
+import jakarta.jms.DeliveryMode;
+import jakarta.jms.Destination;
+import jakarta.jms.JMSException;
+import jakarta.jms.MessageConsumer;
+import jakarta.jms.MessageProducer;
+import jakarta.jms.Session;
 
 import org.apache.activemq.ActiveMQConnectionFactory;
 import org.apache.activemq.ActiveMQSslConnectionFactory;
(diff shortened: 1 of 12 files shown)


stweil (approved):
Looks good, but maybe a small documentation update should be added:

Docs are now stale: docs/developer/api/activemq_jms_api.md still documents javax.jms.MapMessage and shows import javax.jms.*; (lines 110, 145, 217).

solth:
> Docs are now stale: docs/developer/api/activemq_jms_api.md still documents javax.jms.MapMessage and shows import javax.jms.*; (lines 110, 145, 217).

Good point. @henning-gerhardt would you mind updating those lines as well?

henning-gerhardt:
@solth: so far as I know the wiki is the source of any documentation and not was is in the source repo or did this change again? On comparing the wiki page did contain a lot more information than the out dated documentation file in the source code repo. The source code examples on both placs are so out dated that is it not only the mentioned lines should be updated.

solth:
@henning-gerhardt I get what you are saying, and I think you are correct, the wiki is still the main source of documentation. Many of those documentation files in the repo are probably very outdated and should be thoroughly updated. 

But until someone finds the time and resources to actually do that, I think it is still an improvement, albeit a small one, to update outdated information selectively whenever we stumble upon it, wouldn't you agree? I am a fan of incremental improvements, and these files are still part of the repository, after all, so if we detect an error or outdated information, it makes sense to correct that.

henning-gerhardt:
@solth: Why must we maintain 2 sources if only one - the wiki - is the real one? Why not delete the out dated documentation from the source code repo what I suggested a long time ago but was not done by release or documentation manager?

solth:
Ok, I will merge this pull request without the suggested changes to the (outdated) documentation files. Feel free to suggest deleting the old documentation files to the @kitodo/kitodo-community-board or @matthiaskissler as our documentation manager. I don't have an overview over how many old documentation files would be affected by this as I have never really worked on them. 

I have to admit, though, that it is a little frustrating that we have to spend so much time on these fundamental discussions in a pull request where just three little lines would have to be changed slightly.

henning-gerhardt:
https://github.com/kitodo/kitodo-production/wiki/Developer_3.x-Active-MQ get an update but without any warranty.

PR #7173, opened 2026-08-10
Allow digits in method, member, parameter and variable names

stweil:
Checkstyle's MethodName format disallowed digits, so names like getFile2() or X509_getSerialNumber() failed the check even though digits are perfectly legal Java identifiers (JLS 3.8: identifiers consist of Java letters followed by Java letters or Java digits).

All mainstream Java style guides allow digits in identifier names:
- checkstyle's default format is ^[a-z][a-zA-Z0-9]*$ for MethodName, MemberName, ParameterName and LocalVariableName
- Google Java Style camelCase regex includes digits
- Sun/Oracle Code Conventions only require verbs in mixed case

Update MethodName, MemberName, ParameterName, LambdaParameterName and LocalVariableName from ^[a-z][a-zA-Z]*$ to ^[a-z][a-zA-Z0-9]*$ to match the checkstyle default and common conventions.

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -121,26 +121,26 @@
         </module>
 
         <module name="MemberName">
-            <property name="format" value="^[a-z][a-zA-Z]*$"/>
+            <property name="format" value="^[a-z][a-zA-Z0-9]*$"/>
             <message key="name.invalidPattern"
                      value="Member name ''{0}'' must match pattern ''{1}''."/>
         </module>
 
         <module name="ParameterName">
-            <property name="format" value="^[a-z][a-zA-Z]*$"/>
+            <property name="format" value="^[a-z][a-zA-Z0-9]*$"/>
             <message key="name.invalidPattern"
                      value="Parameter name ''{0}'' must match pattern ''{1}''."/>
         </module>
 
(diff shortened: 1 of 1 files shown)


stweil:
The current restriction which forbids digits in names existed right from the beginning (commit 152bf3863fa6b6457428adffcf293dd6fbb5c262). Maybe @beatrycze-volk knows the reason for it.

solth:
Even though I understand your reasoning, changing the Checkstyle rules is something that should be discussed in a...discussion, not changed without first agreeing upon it in the community, especially among developers. Therefor I close this pull request and encourage you to open a discussion about the topic, first.

stweil:
I am rather sure that there was no discussion when this rule was introduced nearly ten years ago. And I started a [discussion about coding style and rules](https://github.com/kitodo/kitodo-production/discussions/6190) nearly two years ago. Up to now I got no feedback.

beatrycze-volk:
There was a discussion at that time. I have introduced those rules with the agreement from other developers which were working on the project at that time. Of course you are free to go back to the topic, discuss it again with the current developers (I don't work on Production since around 8 years) and adjust it according to your current needs.

stweil:
Thank you for this feedback!

PR #7174, opened 2026-08-10
Reduce memory consumption on task state down and improve performance

BartChris:
This Pull Request significantly improves the performance of the "Set Task state down" operation, especially if done in a batch via the process list. In my tests, when setting 3500 processes "down" runtime went down from around a minute to 16 seconds. 
The main change is to use more targeted persistence operations for the affected tasks and for `sortHelperStatus`, instead of calling `ServiceManager.getProcessService().save(processForStatus)`. Saving the whole process causes Hibernate to allocate and process large object graphs, resulting in significantly higher runtime and memory usage.

Adresses https://github.com/kitodo/kitodo-production/issues/5813

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -923,6 +923,11 @@ public class ProcessService extends BaseBeanService<Process, ProcessDAO> {
                 .filter(t -> TaskStatus.INWORK.equals(t.getProcessingStatus())).collect(Collectors.toList());
     }
 
+    public void updateSortHelperStatus(Integer processId, String sortHelperStatus)
+        throws DAOException {
+        dao.updateSortHelperStatus(processId, sortHelperStatus);
+    }
+
     /**
      * Create and return String used as progress tooltip for a given process. Tooltip contains OPEN tasks and tasks
      * INWORK.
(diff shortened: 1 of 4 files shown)


Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
```diff
@@ -923,6 +923,11 @@ private List<Task> getTasksInWork(Process process) {
                 .filter(t -> TaskStatus.INWORK.equals(t.getProcessingStatus())).collect(Collectors.toList());
     }
 
+    public void updateSortHelperStatus(Integer processId, String sortHelperStatus)
```
henning-gerhardt, line 932: Public method without JavaDoc.

henning-gerhardt (changes requested):
Except of the missing JavaDoc the changes are good and good improvement regarding execution time and memory usage.

henning-gerhardt (approved):

solth:
@BartChris did you intentionally use "adresses" instead of an issue linking keyword like "fixes" or "resolves"? 
@henning-gerhardt with this pull request merged, is #5813 sufficiently resolved for you to close it?

henning-gerhardt:
@solth: I can not yet judge if #5813 is solved or not as I has only verified the optimization on my local development system where one user was active and the reported issue is from our productive environment where a lot more users was parallel active and even more other actions where executed. I see this as a big improvement beside other changes in the past but I will not judge that #5813 is now solved.

solth:
@henning-gerhardt ok, but could you monitor the situation a little bit and then re-evaluate the mentioned issue after some time?

henning-gerhardt:
> @henning-gerhardt ok, but could you monitor the situation a little bit and then re-evaluate the mentioned issue after some time?

I can not tell you when the upcoming version 4.1.x (or any later version) go live in our productive environment as we struggle with the search introduced in 3.9 and we need upcoming development of the database only search to have a good base to go live.

But if you want I can close the issue with a remark and when I discover in future this behaviour again I can re-open this issue or making a new issue. Would be this okay?

solth:
No, that's alright, no hurry to close that ticket. I just didn't want to forget it. We can check it again later.

PR #7175, opened 2026-08-13
Remove out dated documentation as documentation is moved to wiki

henning-gerhardt:
Documentation from the source code repo was moved a long time ago to the GitHub based wiki and should be maintained there.


henning-gerhardt:
@kitodo/kitodo-community-board and @matthiaskissler : please decide if the old out dated documentation in the source code repo should stay and be maintained beside the documentation in the wiki so that two documenation sources must be maintained.

matthiaskissler (approved):
It is better to maintain the documentation in one place; I approve the removal of the outdated documentation in the code and propose that the wiki documentation is the leading documentation source.

stweil:
My personal opinion: Technical documentation that is closely linked to the code (API, configuration, setup) belongs in the repository; user-oriented guides and community content are often better suited to the wiki.

With code in the repository and technical documentation in the Wiki there would be two places to maintain, and it would be difficult or even impossible to handle different code branches which might need different documentation in the Wiki. In addition, such splitting would make the increasing use of coding agents much more complicated because they would need the Wiki repository in addition to the code repository to handle their tasks.

stweil:
@matthiaskissler, maybe we should discuss the different alternatives for the documentation at [GitHub discussions](https://github.com/kitodo/kitodo-production/discussions). I created a [prototype for improved documentation](https://stweil.github.io/kitodo-production/).

PR #7176, opened 2026-08-13
[4.0]  Fix NPE opening Metadata Editor with other Error

thomaslow:
Backport of #7172


solth (approved):
Works, too, thank you!

PR #7179, opened 2026-08-14
Update flyway to 12.11.0

henning-gerhardt:
- Update to current latest 12.x version
- Fix message "[WARNING] Using MariaDB 11.8 which is newer than the version Flyway has been verified with. The latest verified version of MariaDB is 11.7." on Debian Trixie (Debian 13) as 12.3.0 did not official support MariaDB 11.8 (but works too).


thomaslow (approved):
Works for me with:
- current database
- migration of old backup from `2.135` until now

PR #7182, opened 2026-08-14
[4.0] Fix exclusion query (#7166)

solth:
Backport of #7166 (by @BartChris) for Kitodo `4.0.x`

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/BeanQuery.java
@@ -79,6 +79,18 @@ public class BeanQuery {
         sorting = Pair.of(varName + ".id", "ASC");
     }
 
+    private String uniqueParameterName(String field) {
+        String baseName = varName(field);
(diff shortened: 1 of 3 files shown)


thomaslow (approved):
I struggled reproducing this issue for an hour or so until I realized the problem only occurs in combination with a full-text filter query, which was not mentioned in #7165 or #7166. 

Without a full-text filter query, only one restriction (either `IN` or `NOT IN`) is added and there is no duplicate parameter and KitodoScripts and other actions are correctly applied to the selected processes. Only when adding a full-text filter the additional `IN` clause is added, because search results are iterated over in multiple batches and batches are encoded with an additonal `IN (:id)` restriction.

Anyway, the PR works and correctly solves the problem.

BartChris:
> until I realized the problem only occurs in combination with a full-text filter query, which was not mentioned in https://github.com/kitodo/kitodo-production/issues/7165 or https://github.com/kitodo/kitodo-production/pull/7166.

Ah sorry, i should have made this clearer. Thanks for digging in!

PR #7146, opened 2026-08-03
Fix typos in Spanish texts (found by aspell)

stweil:
Fix various typos in Spanish UI and error messages:
- messages_es.properties: dataEditor.mediaNotFound, dataEditor.unableToMoveError, numProcesses, search, taskUpdateInternalMetaInformation, userEdit.metadataEditorSettings.defaultGalleryView, userEdit.metadataEditorSettings.showPaginationByDefault, viewAllProcesses
- errors_es.properties: divisiòn -> división (2x), específicado -> especificado (2x), occurre -> ocurre (2x), mediaPartialFormStartLessThanMediaDuration, noProcessTitleRule
- password_es.properties: inválida -> inválido

Assisted-by: OpenCode / big-pickle (opencode)

PR #7183, opened 2026-08-16
Fix flaky ImportingST tests in CI

stweil:
ImportingST.checkOptionalMetadataValidationDuringProcessCreation sometimes failed with NoSuchElementException for #linkProjects right after login, because the browser was still on the login page when the navigation link was looked up. The fixed delay of 2 seconds after login is not always enough on slow CI servers. All navigation methods in TopNavigationPage now wait until the corresponding link is present in the DOM before clicking it.

ImportingST.checkOrderOfImportConfigurations sometimes failed because the titles of the import configurations in the dropdown were read before the list items were populated. getImportConfigurationsTitles now waits until all list items have a non-empty title.

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/ProcessFromTemplatePage.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/ProcessFromTemplatePage.java
@@ -171,6 +171,15 @@ public class ProcessFromTemplatePage extends EditPage<ProcessFromTemplatePage> {
      */
     public List<String> getImportConfigurationsTitles() {
         clickElement(catalogSelect.findElement(By.cssSelector(CSS_SELECTOR_DROPDOWN_TRIGGER)));
+        await("Wait for import configuration titles to be loaded")
+                .pollDelay(500, TimeUnit.MILLISECONDS)
+                .atMost(30, TimeUnit.SECONDS).ignoreExceptions()
+                .until(() -> {
+                    List<WebElement> items = Browser.getDriver()
+                            .findElement(By.id("catalogSearchForm:catalogueSelectMenu_items"))
+                            .findElements(By.className("ui-selectonemenu-list-item"));
+                    return !items.isEmpty() && items.stream().allMatch(item -> !item.getText().isEmpty());
+                });
         WebElement selectMenuItems = Browser.getDriver().findElement(By.id("catalogSearchForm:catalogueSelectMenu_items"));
         return selectMenuItems.findElements(By.className("ui-selectonemenu-list-item"))
                 .stream().map(WebElement::getText).collect(Collectors.toList());
@@ -383,7 +392,7 @@ public class ProcessFromTemplatePage extends EditPage<ProcessFromTemplatePage> {
      * @param index index of checkbox to click
      */
     public void selectCheckBox(int index) {
-        metadataTable.findElements(By.className("ui-chkbox-icon")).get(index).click();
+        clickElement(metadataTable.findElements(By.className("ui-chkbox-box")).get(index));
     }
 
     /**
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/ProcessFromTemplatePage.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/ProcessFromTemplatePage.java
@@ -173,12 +173,12 @@ public class ProcessFromTemplatePage extends EditPage<ProcessFromTemplatePage> {
         clickElement(catalogSelect.findElement(By.cssSelector(CSS_SELECTOR_DROPDOWN_TRIGGER)));
         await("Wait for import configuration titles to be loaded")
                 .pollDelay(500, TimeUnit.MILLISECONDS)
-                .atMost(30, TimeUnit.SECONDS).ignoreExceptions()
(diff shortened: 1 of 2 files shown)


stweil:
I noticed the flakiness in the CI for PR #7146. It occurs also for main: https://github.com/kitodo/kitodo-production/actions/runs/31694173962.

stweil:
The CI tests are more flaky than expected with random failures at different tests. Therefore it was necessary to add a whole bunch of changes.

stweil:
The [CI now passed](https://github.com/kitodo/kitodo-production/actions/runs/31953830568). I cleaned the commit history and created a new pull requests #7185 for changes which were part of this one but which are unrelated to the flaky ImportingST. Both PRs together fix the CI.

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/TopNavigationPage.java
```diff
@@ -145,49 +145,55 @@ public void cancelClientSelection() {
...
+
+    /**
+     * Waits until the navigation link with the given ID is present in the DOM and clicks it.
+     *
+     * @param linkId ID of the navigation link
+     */
+    private void clickNavigationLink(String linkId) {
         RemoteWebDriver driver = Browser.getDriver();
-        ((JavascriptExecutor) driver).executeScript(ARGUMENTS_CLICK, driver.findElement(By.id("linkSystem")));
+        WebDriverWait webDriverWait = new WebDriverWait(driver, Duration.ofSeconds(90));
```
solth: Please change to some reasonable value here.
```suggestion
        WebDriverWait webDriverWait = new WebDriverWait(driver, Duration.ofSeconds(5));
```

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/TopNavigationPage.java
```diff
@@ -145,49 +145,55 @@ public void cancelClientSelection() {
...
+        webDriverWait.until(ExpectedConditions.presenceOfElementLocated(By.id(linkId)));
+        ((JavascriptExecutor) driver).executeScript(ARGUMENTS_CLICK, driver.findElement(By.id(linkId)));
```
solth: Casting 'driver' to 'JavascriptExecutor' is redundant
```suggestion
        driver.executeScript(ARGUMENTS_CLICK, driver.findElement(By.id(linkId)));
```

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/ProcessFromTemplatePage.java
```diff
@@ -171,6 +171,15 @@ public void selectKalliope() throws InterruptedException {
...
     public List<String> getImportConfigurationsTitles() {
         clickElement(catalogSelect.findElement(By.cssSelector(CSS_SELECTOR_DROPDOWN_TRIGGER)));
+        await("Wait for import configuration titles to be loaded")
+                .pollDelay(500, TimeUnit.MILLISECONDS)
+                .atMost(30, TimeUnit.SECONDS).ignoreExceptions()
+                .until(() -> {
+                    List<WebElement> items = Browser.getDriver()
+                            .findElement(By.id("catalogSearchForm:catalogueSelectMenu_items"))
+                            .findElements(By.className("ui-selectonemenu-list-item"));
+                    return !items.isEmpty() && items.stream().allMatch(item -> !item.getText().isEmpty());
```
solth: Simplify steam API call chain to make code more readable.
```suggestion
                    return !items.isEmpty() && items.stream().noneMatch(item -> item.getText().isEmpty());
```

Kitodo/src/test/java/org/kitodo/selenium/testframework/pages/ProcessFromTemplatePage.java
```diff
@@ -171,6 +171,15 @@ public void selectKalliope() throws InterruptedException {
      */
     public List<String> getImportConfigurationsTitles() {
         clickElement(catalogSelect.findElement(By.cssSelector(CSS_SELECTOR_DROPDOWN_TRIGGER)));
+        await("Wait for import configuration titles to be loaded")
+                .pollDelay(500, TimeUnit.MILLISECONDS)
+                .atMost(30, TimeUnit.SECONDS).ignoreExceptions()
```
solth: Please change to some more reasonable value like 5 or at most 10 seconds here.
```suggestion
                .atMost(5, TimeUnit.SECONDS).ignoreExceptions()
```

solth (changes requested):
Thanks for taking a look at this unreliable test. Code looks good (I have just a few small suggestions) and the changes do indeed seem to stabilize the tests in question. 

I disagree with the lengths of the introduced timeouts, though.  Even on a slow CI populating the list of import configuration titles will not take anything near 30 seconds, and if it did, I would rather consider that a bug in which case the test _should_ fail. 

Additionally, I do not want to wait for some failing builds even longer than is already the case just because of some inappropriate timeouts. If the list is not populated in 5 seconds, it will probably not be populated in 30 seconds either. (and yes, I know, the same class contains other `.atMost(30, TimeUnit.SECONDS)` calls, but these are outdated as well and should be adjusted in the future, too)

Please change the timeout values as suggested below.

solth (approved):

PR #7184, opened 2026-08-16
Enforce zero checkstyle violations

stweil:
The Kitodo module allowed up to two checkstyle violations. With a non-zero maximum, the checkstyle check would not fail when new violations were introduced, as long as their number stayed below the limit, so partial test runs would not detect them. Reduce maxAllowedViolations to 0 so that every violation fails the build.

The two remaining MethodLength violations are in Paginator.parse() and ImportEadProcessesThread.run(), both intricate state machines that are hard to split without hurting readability. Exempt these two methods from the MethodLength check via
@SuppressWarnings("checkstyle:MethodLength"), enabled by the newly added SuppressWarningsHolder and SuppressWarningsFilter in the checkstyle configuration.

### DIFF
--- a/Kitodo/pom.xml
+++ b/Kitodo/pom.xml
@@ -27,7 +27,7 @@
 
     <properties>
         <main.basedir>${project.parent.basedir}</main.basedir>
-        <maxAllowedViolations>2</maxAllowedViolations>
+        <maxAllowedViolations>0</maxAllowedViolations>
         <cargo.plugin.version>1.10.22</cargo.plugin.version>
         <jsf.projectStage>Production</jsf.projectStage>
         <myfaces.resourceMaxTimeExpires>604800000</myfaces.resourceMaxTimeExpires>
--- a/Kitodo/src/main/java/org/kitodo/production/helper/metadata/pagination/Paginator.java
+++ b/Kitodo/src/main/java/org/kitodo/production/helper/metadata/pagination/Paginator.java
@@ -35,6 +35,7 @@ public class Paginator implements Iterator<String> {
      */
     private HalfInteger value;
 
+    @SuppressWarnings("checkstyle:MethodLength")
     private void parse(String initializer) {
 
         StringBuilder stringBuilder = new StringBuilder();
--- a/Kitodo/src/main/java/org/kitodo/production/thread/ImportEadProcessesThread.java
+++ b/Kitodo/src/main/java/org/kitodo/production/thread/ImportEadProcessesThread.java
@@ -125,6 +125,7 @@ public class ImportEadProcessesThread extends EmptyTask {
     }
 
     @Override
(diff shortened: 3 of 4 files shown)


stweil:
This PR eliminates the two nasty annotations in each CI GitHub action.

stweil:
> This PR eliminates the two nasty annotations in each CI GitHub action.

@solth, it would be really nice to get rid of these annoying annotations ...

stweil:
@solth, I see that you tried several times to restart the failing CI test. I now asked Copilot for the reason of the [latest failure](https://github.com/kitodo/kitodo-production/actions/runs/31950314890/job/98599528937). Here is the answer:

> Solutions
> 1. Increase WebElement Interaction Timeout
> 
> The 5-second timeout for clearing the username field is too short. Modify the test framework's helper to increase the timeout.

I tried this solution some days ago, but had to abandon it because even with timeouts above a minute the failures did not disappear. Therefore I asked Copilot whether PR #7185 would help. Its answer is very clear:

> Yes, PR #7185 would directly help solve this issue.
> 
> The PR addresses the root cause of the timeout failures you're experiencing. Here's why:
> The Problem (Your Current Job Failure)
> 
>     Multiple selenium tests are timing out when trying to interact with the login form (clearing the username field)
>     The timeouts are preceded by stale HTTP connections to the search server
>     The logs show the application experiencing long delays on index queries after the OpenSearch node restarts
> 
> What PR #7185 Fixes
> 
> The PR changes the OpenSearch node lifecycle management:
> 
>     Before: The node was started/stopped for every selenium test class (in setUp/tearDown)
>         Each restart caused the Tomcat application to maintain stale connections to the killed node
>         The first index query after each restart (typically the login POST's isIndexCorrupted check) would stall for 43-307 seconds while the client retried those connections
>         These stalls exceeded the WebDriver wait timeouts (5 seconds)
> 
>     After: The node is started once per surefire execution and kept alive across all test classes
>         No more restarts → no stale connection issues
>         Eliminates the unpredictable stalls that were causing timeouts
>         Per-test-class isolation is preserved because setUp already drops and recreates the Hibernate Search index
> 
> Implementation Details
> 
>     Adds a null check in MockDatabase.startNode() to prevent restarting if already running
>     Removes MockDatabase.stopNode() from BaseTestSelenium.tearDown()
>     The node is cleaned up automatically when the JVM terminates
> 
> This matches exactly the symptoms in your failing job where tests timeout on login after connection stalls to the search server.

Should I add the fix from #7185 here to convince you? Or do you want to merge #7185? Then I can rebase the PR here, and we will see that CI passes.

solth (approved):
Good idea. I tried a few times refactoring those two long methods in question, but it wasn't trivial and excluding them from the checkstyle rule is justifiable.

Issue #6873, opened 2026-02-04
Enforce password change


henning-gerhardt:
**Is your feature request related to a problem? Please describe.**
If you - as an administrator - add a new user to a Kitodo.Production system and provide this person an initial password then it is currently possible that this initial password is not changed after the first login.

**Describe the solution you'd like**
After a new user is created or a password is changed with administrative power after a user request it would be really nice that this new password must be changed by this user after the first time the user has been log in the first time successful. The new entered password must not be the old provided password and must match the configured password guidelines of this system.

If an user or administrator is changing the password itself then this dialog should not appear. If an administrator is changing the password of an other administrator then this must change you password dialog should appear too.

PR #7188, opened 2026-08-17
Migrate user passwords to hash-based encoding

### LINKED ISSUE #456: Replace current mechanism to store encrypted passwords

### LINKED ISSUE #626: User passwords visible for administrators
Administrators can enter the "Edit user" page for any registered user.
There they can see the password of individual users (with minimum efforts).

### LINKED ISSUE #4576: Change the creation of a user in LDAP to upon login

thomaslow:
This PR improves security by migrating user passwords from the old symmetric encryption strategy (`PBEWithMD5AndDES`) to a modern hash-based password encoding strategy called [`argon2id`](https://en.wikipedia.org/wiki/Argon2).

### Todos

- [x] First draft implementation
- [x] Discuss approach with community
- [x] Implement tests
- [x] Cleanup code

### Related Issues

- Fixes #456
- Fixes #626 
- Fixes #4576
- Contributes to #6208

### How It Works

To keep everything as easy as possible in terms of migration, password encodings are replaced upon the next user login automatically via Spring's [`DelegatingPasswordEncoder`](https://docs.spring.io/spring-security/reference/features/authentication/password-storage.html#authentication-password-storage-dpe). This means the database will contain both old password encodings and new password encodings for some time. New password encodings are distinguished from old ones based on a prefix, e.g. `{argon2}password`. Old passwords do not have any prefix.

Old password encodings are still considered valid for now. This should reduce any issues when, e.g., loading a database backup with old passwords. We should discuss when to remove support for old password encodings in the future. Because of that, the corresponding security warning "Use of a broken or risky cryptographic algorithm" will remain in effect.

#### LDAP Login Task

The change from a symmetric encryption strategy to a hash-based encoding strategy also means that plaintext passwords of users cannot be decrypted anymore. Currently, passwords are decrypted to register the plaintext password with LDAP when clicking the "Write LDAP configuration" button on the user edit page. This PR changes this behavior and introduces "login tasks". Instead of writing the plaintext password immediately upon clicking the button, a login task is created for that particular user. When this user logs in, the plaintext password is available again shortly during the login and used to write the LDAP entry, see `CustomLoginSuccessHandler`. Any error messages are stored for the login task and shown on the user edit page. The process is as follows:

1. Register login task for user

Screenshot From 2026-08-17 14-41-03<br/>

2. If not needed anymore, the login task can be canceled

Screenshot From 2026-08-17 14-41-12<br/>

3. User logs in

4. Successful execution is reported

Screenshot From 2026-08-17 14-41-35<br/>

5. Or, execution error is reported

Screenshot From 2026-08-17 14-42-14<br/>

When changing passwords via the "password change/reset dialog", the LDAP configuration is written immediately (as before).

@solth @henning-gerhardt Do you think this process is acceptable for administrators when synchronizing user passwords with LDAP? Do you see any issues? Please add anyone that you think might be interested/affected to this PR.

### Release / Migration Notes

- Your hibernate configuration `hibernate.cfg.xml` needs to be modified by adding:

```
<mapping class="org.kitodo.data.database.beans.LoginTask"/>
```

### Developer Notes

Upgraded users passwords are not valid in earlier branches of Kitodo.Production. You may reset user passwords in your database to the old test password (cleartext `test` encoded as `OvEJ00yyYZQ=`) with something like:

```
mysql --skip_ssl --host=database --user=kitodo --password=kitodo -D kitodo --execute="UPDATE user SET password = 'OvEJ00yyYZQ=';"
```

### Future Ideas

The mechanism of "login tasks" may be extended in the future, e.g.:

- to force users to reset passwords, see https://github.com/kitodo/kitodo-production/issues/6873
- to force users to set up 2FA
- to ask users to accept terms and conditions


henning-gerhardt:
@thomaslow : Did I understand it correct that the LDAP entry is written the first time as the user logged in the first time and no LDAP entry exists until then?

thomaslow:
> @thomaslow : Did I understand it correct that the LDAP entry is written the first time as the user logged in the first time and no LDAP entry exists until then?

No. If no LDAP entry exists for a user, but an LDAP group is configured for the user, the login will fail with "bad credentials". An LDAP entry is only created if explicitly requested by clicking the button on the user edit page and then asking the user to log in with the existing database password.

henning-gerhardt:
> No. If no LDAP entry exists for a user, but an LDAP group is configured for the user, the login will fail with "bad credentials". An LDAP entry is only created if explicitly requested by clicking the button on the user edit page and then asking the user to log in with the existing database password.

Maybe there is only a little misunderstanding. Even now the LDAP entry is not created until the administrator is clicking on the button to create the LDAP entry. After this the entry exists in the LDAP and can be used (f.e. using the login credentials on a WebDAV/Samba file access system). With your change the LDAP entry is delayed until the user did the first login, so until the user did not made his / her first login associated systems like WebDAV or Samba did not "work" for this users as no LDAP entry exists. Correct? I'm asking only for a right understanding of the change.

thomaslow:
> With your change the LDAP entry is delayed until the user did the first login, so until the user did not made his / her first login associated systems like WebDAV or Samba did not "work" for this users as no LDAP entry exists. Correct? I'm asking only for a right understanding of the change.

Yes, with this PR there is a new delay introduced. A User needs to log in at least once after an admin has clicked the button before the LDAP entry is created and the user may use WebDav/Samba that is configured with this LDAP server.

The problem is that when clicking the "write LDAP entry" button, with this PR, it will no longer be possible to retrieve the plaintext password of a user from the database. The following strategies would be possible:

- show a dialog and ask for the plaintext user password again  (but the admin that clicks the button shouldn't know the password)
- reset the password to a new password defined by the admin or a random password (meaning the new password needs to be handed over the user again)
- wait for the user to log in again so we have access to the plaintext user password again

henning-gerhardt:
@thomaslow Thank you for the clarification. I like your approach as this is much better then the current solution or many others except for the delay but this should be okay and can be communicated.

solth:
@thomaslow ~~does this also resolve the ancient issue #456 ?~~ Sorry, just realized that you already mentioned the issue above!

henning-gerhardt:
I tried your changes with the following scenarios 
- first login
- password change administrativ
- password change by user
- using credentials in external application if the user has a ldap password


1. Existing user with database only password
- After first login the password get changed to the new system and get stored.
- Changing password by administrator and the user itself is working and every time changing the password works as expected.

2. Existing user with LDAP group
- After first login the stored password in database is untouched and login in application and external was possible.
- Changing password by administrator and user did the following: 
  - the password get changing in LDAP and database. That the password is changed in the database for a LDAP user is new and a different behaviour then before.
  - if the user is changing his password it must be match with the stored password in the database and could not be different than the password in LDAP system. If database and LDAP password did not match the user can not change his password.

3. New user with database only password
- see existing user notes

4. New user with LDAP group
- Without the new task the user is unable to login like before if you forget to write the LDAP configuration on creating the user. This is / was expected by me and works
- If the new user is logged in the first time the LDAP entry is created. The user can log in afterwards inside the Kitodo.Production without issue. But authentication with this credentials on external applications are not possible as the password is wrong. I created a new user with a well known password and as the passwords in LDAP for samba are created every time in the same way without any salt or other dynamic information the same password will create the same hash. So the new user has a complete different hash values in LDAP for the samba passwords and even for the stored user password. After the user is changing the password twice (first to a random value than back to the well known password which was used before) the entries inside LDAP for samba and user password are created correct and external applications are able to verify / authenticate the user with his given credentials. Its look for me that on creating the LDAP entry with the new task a wrong password is submitted.
- The user can only change his password in the password change dialog if this password it the same as the password stored inside the database. An administrator can change the password in any case.


I'm hoping this is written well and good to understand. If not, please contact me outside GitHub and we can discuss this in German.

thomaslow:
@henning-gerhardt Thank you for testing this.

My first thoughts are:

- In the scenario "Existing user with LDAP group", the database password is currently not migrated, because Spring-Security uses the LDAP authentication module when logging in this user, not the database authentication module. This means, the migration logic of Spring Security is not triggered in this case. I think the most sensible solution would be to add a "Nutzer-Passwörter migrieren" button to the "System -> Migrieren" page, which migrates all user passwords in the database (without any changes to Ldap).

- You describe that you did not expect user passwords to change in the database if a user is set up with an Ldap group. However, there is explicit code that does this in the current main branch, see:

https://github.com/kitodo/kitodo-production/blob/1607347a07aaeb48796697b34dc62be4c3d61793/Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java#L226-L227

I don't know what this "bug" is/was. In theory, we could adapt this line such that the changed password is only submitted to Ldap.

- You describe that users cannot change passwords if the database password differs from the Ldap password. This is also the case in the current main branch, see:

https://github.com/kitodo/kitodo-production/blob/1607347a07aaeb48796697b34dc62be4c3d61793/Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java#L279

I'll check how difficult it would be to check the "oldPassword" against Ldap if an Ldap group is configured for this user.

- You describe that the new Login-Task seems to submit a wrong password to Ldap and that you cannot login over samba with this password. I'll investigate this. But, I think I tested this successfully on my system.

henning-gerhardt:
> In the scenario "Existing user with LDAP group", the database password is currently not migrated, because Spring-Security uses the LDAP authentication module when logging in this user, not the database authentication module. This means, the migration logic of Spring Security is not triggered in this case.

I did not wrote that this is an issue or did I? I even did not expect that the stored password for the LDAP bases user get migrated. I'm fine with this no-change as long as they are not needed for other use cases.

>  I think the most sensible solution would be to add a "Nutzer-Passwörter migrieren" button to the "System -> Migrieren" page, which migrates all user passwords in the database (without any changes to Ldap).

This could be a way if this way is needed. If this done in a secure way than is the risk low to expose them.

> You describe that you did not expect user passwords to change in the database if a user is set up with an Ldap group.

Maybe I run into a caching / display issue on comparing the changed values in the database. I can not reproduce is anymore.

> I don't know what this "bug" is/was. 

Bug is described in #5787 and fixed by #5807 (UserForm class in 3.9.x and before). So until version 3.6.2 or 3.7.0 passwords which were changed are changed only in LDAP but not inside the database. If this passwords are not changed since them with a fixed version they are still differ between database and LDAP. If users are "hitted" by this issue only administrative persons can change password to a one-time password and the user can change it itself afterwords (if he want). Later I created #6873 as you can not "see" that this provided one-time password is changed.

> You describe that users cannot change passwords if the database password differs from the Ldap password. This is also the case in the current main branch

Even this I can not reproduce yet anymore :-( I tried it two times to verify it before I wrote it down. Maybe I mixed up this password changes or did not use right or wrong passwords in wrong or right places.

> I'll check how difficult it would be to check the "oldPassword" against Ldap if an Ldap group is configured for this user.

This could be helpful for all the passwords which are created before using version 3.6.2 / 3.7.0. But this is maybe not at a high solution rate.

> You describe that the new Login-Task seems to submit a wrong password to Ldap and that you cannot login over samba with this password.

Not over Samba - I used an Apache WebDAV configuration which is using the LDAP `userPassword` field. I compared the LDAP ldif output of this user with the first time set password and the expected ones from an existing user with the same password. LDAP keys was `sambaLMPassword`, `sambaNTPassword` and `userPassword` for comparing.

solth:
> If not, please contact me outside GitHub and we can discuss this in German.

Englisch ist nicht Pflicht auf GitHub! Wenn die Diskussion besser auf Deutsch zu führen ist, spricht nichts dagegen, um Missverständnisse zu vermeiden!

thomaslow:
@henning-gerhardt I investigated what went wrong. Spring removes the user's raw password from the authentication object before triggering the `AuthenticationSuccessHandler` for security reasons. I think the idea is to clear the user's raw password from memory as soon as it is not needed any more. The incorrect password that was submitted for the user was the string 'null'.

To fix this, I moved the login task processing into `DynamicAuthenticationProvider`. At this point in time (when credentials are checked against the database or ldap), the raw user password is still available inside the authentication object (`authentication.getCredentials()`).

I don't know why I thought it worked before. It should not have worked in any of my previous commits. Maybe I mixed up what user information and passwords were already stored in the ldap server.

In order to track down the problem, I also set up an Apache with webdav, because I first thought it has something to do with Apache. Anyway, I can confirm now that I can access both webdav and samba over Ldap as soon as the user logs in once and the login task is executed.

Let me know in case there are other changes that you think are useful.

Kitodo-DataManagement/src/main/resources/db/migration/V2_146__Add_logintask_table.sql
```diff
@@ -0,0 +1,27 @@
...
+    createdAt DATETIME NOT NULL COMMENT 'creation date',
+    executedAt DATETIME DEFAULT NULL COMMENT 'time and date of task execution',
+    error VARCHAR(255) DEFAULT NULL COMMENT 'error message in case execution failed',
+    user_id INT(11) NOT NULL COMMENT 'user id of associated user',
+    PRIMARY KEY(id),
+    CONSTRAINT FK_logintask_user_id 
+        FOREIGN KEY (user_id) REFERENCES user (id),
+    KEY index_pending_tasks_for_user (user_id, status) COMMENT 'index for finding tasks for a specific user'
+) DEFAULT CHARACTER SET = utf8mb4
+  COLLATE utf8mb4_unicode_ci;
```
henning-gerhardt: Missing new line at end of file.

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
```diff
@@ -54,13 +55,15 @@ public class UserEditViewDetailsTab extends BaseTabEditView<User> {
...
 
-    private final transient SecurityPasswordEncoder passwordEncoder = new SecurityPasswordEncoder();
     private final transient UserService userService = ServiceManager.getUserService();
+    private final transient LoginTaskService loginTaskService = ServiceManager.getLoginTaskService();
 
     private String passwordToEncrypt;
     private String oldPassword;
     private List<Client> clientsOfUser;
 
+    private Optional<LoginTask> lastLdapLoginTask;
```
henning-gerhardt: IntelliJ is telling me that this field is not serializable. Maybe add a `transient` to it?

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
```diff
@@ -194,18 +198,66 @@ public boolean isLdapServerReadOnly() {
...
+    /**
+     * User requests to add a new ldap login task to write user credentials to the ldap server.
+     */
+    public void addLdapLoginTask() {
+        loginTaskService.addLoginTask(this.userObject, LoginTaskType.SAVE_USER_TO_LDAP);
+        loadLastLdapLoginTask();
+    }
+
+    /**
+     * User cancels the the last pending ldap login task.
```
henning-gerhardt: One `the` can be removed.

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
```diff
@@ -194,18 +198,66 @@ public boolean isLdapServerReadOnly() {
...
+        if (Objects.nonNull(this.lastLdapLoginTask) && this.lastLdapLoginTask.isPresent()) {
+            loginTaskService.cancelLoginTask(this.lastLdapLoginTask.get());
+        }
+        loadLastLdapLoginTask();
+    }
+
+    /**
+     * Retrieve whether the current user already has a pending login task to save user credentials to the ldap server.
+     * 
+     * @return true if pending login task exists
```
henning-gerhardt: `@return` statement can be removed as the method is not returning anything.

Kitodo/src/main/java/org/kitodo/production/security/password/KitodoUserDetailsPasswordService.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Migrate user passwords to the newest password encoding strategy using Spring's UserDetailsPasswordService.
+ * 
+ * <p>Spring will call "updatePassword" with a new encoding of the user's password after login if this password neeeds to migrated.</p>
+ */
+public class KitodoUserDetailsPasswordService implements UserDetailsPasswordService {
+
+    private static final Logger logger = LogManager.getLogger(KitodoUserDetailsPasswordService.class);
+
+    private static UserService userService = ServiceManager.getUserService();
```
henning-gerhardt: Field `userService` can be final.

Kitodo/src/main/java/org/kitodo/production/security/password/KitodoUserDetailsPasswordService.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public UserDetails updatePassword(UserDetails userDetails, String newEncodedPassword) {
+        if (userDetails instanceof SecurityUserDetails) {
+            try {
+                User user = new User((SecurityUserDetails)userDetails);
+                user.setPassword(newEncodedPassword);
+                userService.save(user);
+                logger.debug("Password for user '{}' migrated to newest encoding strategy", userDetails.getUsername());
+            } catch (DAOException e) {
+                logger.error(String.format("Failed to migrate user password for user '%s'", userDetails.getUsername()), e);
```
henning-gerhardt: Line can be replaced by 

```
logger.error("Failed to migrate user password for user '{}'", userDetails.getUsername(), e);
```

as `String.format()` call is not needed - or did I miss something why it is needed here?

Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
```diff
@@ -11,21 +11,28 @@
 
 package org.kitodo.production.security;
 
+import java.io.IOException;
+import java.security.NoSuchAlgorithmException;
```
henning-gerhardt: Not needed import statement.

Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
```diff
@@ -11,21 +11,28 @@
...
 import java.util.Objects;
 
+import javax.naming.NameAlreadyBoundException;
+import javax.naming.NamingException;
+
```
henning-gerhardt: Not needed import statement.

Kitodo/src/main/java/org/kitodo/production/services/data/LoginTaskService.java
```diff
@@ -0,0 +1,200 @@
...
+import org.kitodo.data.database.beans.User;
+import org.kitodo.data.database.exceptions.DAOException;
+import org.kitodo.data.database.persistence.LoginTaskDAO;
+import org.kitodo.production.helper.Helper;
+import org.kitodo.production.services.ServiceManager;
+import org.springframework.security.core.Authentication;
+
+public class LoginTaskService extends BaseBeanService<LoginTask, LoginTaskDAO> {
+
+    private static List<LoginTaskType> PRIORITY_SEQUENCE = List.of(LoginTaskType.SAVE_USER_TO_LDAP);
```
henning-gerhardt: Field could be `final`.

Kitodo/src/main/java/org/kitodo/production/services/data/LoginTaskService.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private static final Logger logger = LogManager.getLogger(UserService.class);
```
henning-gerhardt: Wrong logging class. Instead of `UserService.class` use `LoginTaskService.class`.

Kitodo/src/main/java/org/kitodo/production/services/data/LoginTaskService.java
```diff
@@ -0,0 +1,200 @@
...
+     */
+    public Optional<LoginTask> getLastLoginTaskForUserAndType(User user, LoginTaskType type) {
+        return dao.getByUserAndType(user, type).stream().findFirst();
+    }
+
+    /**
+     * Return the highest priority login task for a user that is pending.
+     * 
+     * @param user the user
+     * @return optional containing highest priority login task if this user has a pending task
```
henning-gerhardt: Maybe add a `the` in front of `highest priority` (suggestion by spell / grammar checker)

henning-gerhardt (changes requested):
Changes are working for me:
- old stored passwords in database get migrated for database only users
- new user passwords are using new algorithm
- new users which are using LDAP are created correct in database and LDAP after first login
- existing LDAP based users get database password updated on changing there LDAP password

There are some needed changes in the code but they are small.

Kitodo-DataManagement/src/main/resources/db/migration/V2_146__Add_logintask_table.sql
thomaslow: Added a new line.

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
thomaslow: Hmm. A transient would mean that `lastLdapLoginTask` is not serialized at all. I refactored the code such that the login task is not stored as `Optional` any more.

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
thomaslow: Done.

Kitodo/src/main/java/org/kitodo/production/forms/user/UserEditViewDetailsTab.java
thomaslow: Done.

Kitodo/src/main/java/org/kitodo/production/security/password/KitodoUserDetailsPasswordService.java
thomaslow: Done.

Kitodo/src/main/java/org/kitodo/production/security/password/KitodoUserDetailsPasswordService.java
thomaslow: My IDE describes the call signature of your suggestion as `logging.error(String msg, Object param1, Object param2)`, which is why I assumed that the second parameter (the exception) would not be included in the log output, since there is only one parameter reference (`{}`) in the message.

 I tested your variant and the exception message and stack-trace is still printed in this case. I changed the line to your suggestion.

Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
thomaslow: Done.

Kitodo/src/main/java/org/kitodo/production/security/DynamicAuthenticationProvider.java
thomaslow: Done.

Kitodo/src/main/java/org/kitodo/production/services/data/LoginTaskService.java
thomaslow: Done.

Kitodo/src/main/java/org/kitodo/production/services/data/LoginTaskService.java
thomaslow: Done.

Kitodo/src/main/java/org/kitodo/production/services/data/LoginTaskService.java
thomaslow: Done.

henning-gerhardt (approved):

solth:
> * [Replace current mechanism to store encrypted passwords #456](https://github.com/kitodo/kitodo-production/issues/456)
> 
> * [User passwords visible for administrators #626](https://github.com/kitodo/kitodo-production/issues/626)
> 
> * [Change the creation of a user in LDAP to upon login #4576](https://github.com/kitodo/kitodo-production/issues/4576)
> 
> * [Fix security issues and warnings #6208](https://github.com/kitodo/kitodo-production/issues/6208)

Should any of these issues be marked as "fixed" by this pull request?

thomaslow:
Yes, in my opinion:
- Fixes #456
- Fixes #626 
- Fixes #4576
- Contributes to #6208

PR #7189, opened 2026-08-18
Processes of projects a user is not assigned to can be exported

thomaslow:
This PR fixes the issue that processes can be exported for projects even if a user is not assigned to said projects. You can reproduce this issue by:

1. Create a user that is not assigned to all projects of a client
2. Go to the process list
3. Select all processes by clicking on "Alle Treffer" 
4. Export processes via Excel/CSV export

The CSV/Excel file contains processes for projects the user has no access to.

@BartChris You are more involved in this export code. I simply copied the "project restriction" from the `ProcessService.findSelectedProcesses()` method and added it to the `ProcessService.createExportQuery()` method. That seems to fix the issue.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/ProcessService.java
@@ -2561,10 +2561,6 @@ public class ProcessService extends BaseBeanService<Process, ProcessDAO> {
         if (!includeClosed) {
             query.restrictToNotCompletedProcesses();
         }
-        if (!includeInactiveProjects) {
-            query.addBooleanRestriction("project.active", Boolean.TRUE);
-        }
-
         if (allSelected) {
             if (Objects.nonNull(excludedProcessIds) && !excludedProcessIds.isEmpty()) {
                 query.addNotInCollectionRestriction("id", excludedProcessIds);
@@ -2574,6 +2570,11 @@ public class ProcessService extends BaseBeanService<Process, ProcessDAO> {
         }
 
         query.restrictToClient(sessionClientId);
+
+        Collection<Integer> projectIDs = ServiceManager.getUserService().getCurrentUser().getProjects().stream()
+                .filter(project -> includeInactiveProjects || project.isActive()).map(project -> project.getId())
+                .collect(Collectors.toList());
+        query.restrictToProjects(projectIDs);
         query.applyIndexRestriction(FIELD_ID);
         query.addInnerJoin("project proj");
         query.defineSorting("id", SortOrder.ASCENDING);


solth:
@thomaslow would you mind backporting this fix for Kitodo `4.0.x` when @BartChris approves the pull request?

BartChris (approved):
Good catch, this fixes the issue. 👍

BartChris:
Edit: Since the related query is using the same filter condition it is good to harmonize those queries. 
I remember however that i worked on another issue where the projects hold by the user object were stale, so we probably have to check whether projects which were set to inactive shortly before might still be included in the exported list. 

Related issue/PR: 
https://github.com/kitodo/kitodo-production/pull/7043
https://github.com/kitodo/kitodo-production/issues/5295

BartChris:
Both `ProcessService.findSelectedProcesses()` and `ProcessService.createExportQuery()` unfortunately do include processes which belong to projects which were set to inactive in the current user session. (Even when the inactive projects are not included via the ui:

image
)



This would be prevented by applying the same technique as in https://github.com/kitodo/kitodo-production/pull/7043:

```java
 Collection<Integer> projectIDs = ServiceManager.getUserService()
                .getCurrentUser()
                .getProjects().stream()
                .map(Project::getId)

        if (!showInactiveProjects) {
            query.addBooleanRestriction("project.active", Boolean.TRUE);
        }
```

The `ProcessService` class unfortunately has a lot of duplicated code, so the best would be if we extract the filter logic to a common helper. Maybe you could change the project filter logic in  `ProcessService.findSelectedProcesses()` and `ProcessService.createExportQuery()` to the logic above. And then we can make another PR which refactors the class to have less code duplication.

thomaslow:
I think the additional `project.active` filter addresses only one part of the problem. If a user is assigned or unassigned to a project (from a different user) while being logged in, the list of project ids will also be incorrect.

The more general problem is that the current user object is retrieved from spring security, and therefore cached for the current user session. This causes that every call to `ServiceManager.getUserService().getCurrentUser().getSomethingElse()` is cached for the whole user session, too, including user roles, user permissions, assigned projects, assigned clients and potentially more.

Maybe we can also figure out a solution for this general design flaw. I think it should be acceptable - from a performance perspective - to retrieve a fresh user object and related information (permissions, etc.) once for each view or for each request. However, there needs to be request-scoped or view-scoped cache to avoid hundreds of database queries retrieving the current user information when checking for individual permissions.

BartChris:
> I think the additional `project.active` filter addresses only one part of the problem. If a user is assigned or unassigned to a project (from a different user) while being logged in, the list of project ids will also be incorrect.

Good point, i have not tested that scenario. I only tested removing myself from a project. When i do that my Excel export is correct and does not contain the project. Does it make a difference here, when i am not removing myself but a different user from the project? Would this user's export be wrong?

solth:
Would it make sense to take an incremental approach here? By that I mean separating the bug originally fixed by this pull request from the issue of user objects being cached during a session and not reflecting changes made during the time period a user is logged in? Maybe we can first just fix the original problem, since it is a smaller adjustment that could also be backported to `4.0.x` and later tackle the problem with outdated user objects in a separate follow-up pull request?

thomaslow:
> When i do that my Excel export is correct and does not contain the project. Does it make a difference here, when i am not removing myself but a different user from the project? Would this user's export be wrong?

Yes, I have two users: `testAdmin` (all permissions, all clients, all projects) and `testMetadata` (assigned to a subset of projects of a specific client, can only edit metadata). Adding or removing projects for `testMetadata` from the admin account does not have an effect for `testMetadata` until that user logs out and logs in again. This problem is not limited to the export function, but affects every view.

I think this has been an somewhat accepted imperfection for Kitodo.Production for a long time.

> Would it make sense to take an incremental approach here? By that I mean separating the bug originally fixed by this pull request from the issue of user objects being cached during a session and not reflecting changes made during the time period a user is logged in? Maybe we can first just fix the original problem, since it is a smaller adjustment that could also be backported to 4.0.x and later tackle the problem with outdated user objects in a separate follow-up pull request?

@solth Yes, of course. I'll open a new PR for the general problem.

BartChris:
> Would it make sense to take an incremental approach here? By that I mean separating the bug originally fixed by this pull request from the issue of user objects being cached during a session and not reflecting changes made during the time period a user is logged in? Maybe we can first just fix the original problem, since it is a smaller adjustment that could also be backported to `4.0.x` and later tackle the problem with outdated user objects in a separate follow-up pull request?

Sounds good to me. I consider the stale data during a current user session not that critical and would approve the current Pull Request.

BartChris (approved):

solth:
@BartChris thanks! 

@thomaslow if you provide a backport for `4.0.x` of this fix I will include it in the upcoming bugfix release `4.0.1`!

PR #7194, opened 2026-08-19
Refresh current user object from the database during active user session

thomaslow:
This PR addresses the issue that some views are rendered on outdated information if user details or associated beans were updated by another user during an active user session, e.g., as discussed in #7189 for projects that are activated or deactivated while the user is logged in.

Previously, the user object was loaded from the spring security context. This meant that the user object and associated beans (e.g. assigned clients, assigned projects) were cached for the full duration of the user session. If user details were changed by other users (e.g. administrators) while a user is logged in, these would not be reflected immediately for this user, because the user object was never updated from the database during the user session.

Instead, the user object is now loaded exactly once from the database for each request. Any updates to user details or associated beans (e.g. assigned clients, assigned projects) are immediately reflected in views even while the user is logged in.

This PR only updates the user object and associated beans. It does not update permissions (spring security authorities) for the user. Users still need to logout and login again after roles or permissions were changed.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java
+++ b/Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java
@@ -48,6 +48,7 @@ import org.kitodo.data.database.exceptions.DAOException;
 import org.kitodo.data.database.persistence.UserDAO;
 import org.kitodo.exceptions.FilterException;
 import org.kitodo.production.helper.Helper;
+import org.kitodo.production.helper.cache.RequestScopeCacheHelper;
 import org.kitodo.production.security.SecurityUserDetails;
 import org.kitodo.production.security.password.SecurityPasswordEncoder;
 import org.kitodo.production.services.ServiceManager;
@@ -64,6 +65,7 @@ public class UserService extends BaseBeanService<User, UserDAO> implements UserD
     private final SecurityPasswordEncoder passwordEncoder = new SecurityPasswordEncoder();
     private static final int DEFAULT_CLIENT_ID =
             ConfigCore.getIntParameterOrDefaultValue(ParameterCore.DEFAULT_CLIENT_ID);
+    private static final String CURRENT_USER_CACHE_KEY = "current_user";
 
     /**
      * Constructor.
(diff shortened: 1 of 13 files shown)


BartChris (commented):
Works flawlessly 👍.  While checking the remaining calls to `getAuthenticatedUser()` i was wondering whether we should use `getCurrentUser()` here as well:

https://github.com/kitodo/kitodo-production/blob/2426895cd2363c530a1bf1e2b88dd81cfb27a186/Kitodo/src/main/java/org/kitodo/production/forms/createprocess/CreateProcessForm.java#L1122-L1131

Edit: I suppose `getAuthenticatedUser()` can be used here as `ImportEadProcessesThread` mostly needs the Spring Security Context which is not changed before logout.

Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java
```diff
@@ -204,6 +206,9 @@ private User uniqueResult(List<User> users, String login) {
 
     /**
      * Gets the current authenticated user of current threads security context.
```
BartChris:
```suggestion
     * Gets the currently authenticated user from the current thread's security context.
```

Kitodo/src/main/java/org/kitodo/production/services/data/UserService.java
```diff
@@ -212,17 +217,36 @@ public SecurityUserDetails getAuthenticatedUser() {
     }
 
     /**
-     * Get the current authenticated user as User bean.
+     * Return the current User object after loading it from the database (cached for each request).
+     * 
+     * <p>Previously, the user object was loaded from the spring security context. This meant that the user
```
BartChris:
```suggestion
     * <p>Previously, the user object was loaded from the Spring Security context. This meant that the user 
```

BartChris (approved):

thomaslow:
> While checking the remaining calls to `getAuthenticatedUser()` i was wondering whether we should use `getCurrentUser()` here as well:
> Edit: I suppose getAuthenticatedUser() can be used here as ImportEadProcessesThread mostly needs the Spring Security Context which is not changed before logout.

Yes, I investigated this part as well and decided to leave it unchanged because of the same reason.

I also looked into the possibility of caching the user object in view-scope. Unfortunately that would require a lot more refactoring. There would have to be an `@ViewScoped` bean that provides access to the current user object. And then we would have to carefully dissect which calls of `getAuthenticatedUser()` and `getCurrentUser()` are only ever executed inside the JSF lifecycle, such that they can be replaced by a getter of the `@ViewScoped` bean. Unfortunately, lots of the `getCurrentUser()` calls are scattered around in global service classes, which are also used outside of the JSF lifecycle. Outside of the JSF lifecycle, there is no access to view-scoped beans.

PR #7199, opened 2026-08-21
[3.9] ci: Replace Java distribution 'adopt' by 'temurin'

solth:
Backport of #7112 for Kitodo `3.9.x`


stweil (approved):

stweil:
Using a cherry-pick would preserve the author ...

solth:
> Using a cherry-pick would preserve the author ...

Yes, you are right. I normally do that for backports (for example #7181) but forgot to do it here, sorry.

PR #7209, opened 2026-08-27
Do not expand structure tree or gallery column when adding new structure

### LINKED ISSUE #4344: MetadataEditor - collapsed "Structure Panel" is opened when structure is added via gallery
When the user collapses the "Structure Panel" in the Metadata Editor, it is unexpectedly expanded again when creating a structure element via the gallery, so the user needs to re-collapse it every time after creating a new structure. Presumably this happens because the structure is selected in the structure tree after creating it, which leads to the panels expansion.

**Goal**: if the user collapsed the structure panel it should not be automatically re-opened when creating a new structure via the gallery.

**_Edit_**: the same is true the other way around: if the gallery is collapsed and the user adds a structure element via the structure tree, the gallery is re-expanded. This leads to equally awkward UI behaviour as the case described above.

solth:
Fixes #4344

_Note:_ I will open backports of this fix for Kitodo `3.9.x` and `4.0.x` once this pull request is approved

### DIFF
--- a/Kitodo/src/main/webapp/WEB-INF/resources/js/resize.js
+++ b/Kitodo/src/main/webapp/WEB-INF/resources/js/resize.js
@@ -504,11 +504,9 @@ function expandThirdColumn() {
 function updateMetadataEditorView(showMetadataColumn) {
     PF('dialogAddDocStrucType').hide();
     PF('dialogEditDocStrucType').hide();
-    expandFirstColumn();
     if (showMetadataColumn) {
         expandSecondColumn();
     }
-    expandThirdColumn();
     scrollToSelectedThumbnail();
     metadataEditor.detailMap.update();
     metadataEditor.gallery.mediaView.update();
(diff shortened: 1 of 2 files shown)


henning-gerhardt (changes requested):
Your changes did only work for one case: adding / creating a new structure element from the structure panel did not re-open the gallery. 

The case that collapsed structure elements are not "opened" again did not work as they opened again on adding / creating new structure element from the gallery.

I tried this changes with Chromium and Firefox and did even cleaned the cache directories on the file system.

solth:
That's very strange. I just tried it both ways and verified that the collapsed panels wouldn't be re-expanded when adding new structure elements.

Collapsed gallery, adding structure via structure tree:

https://github.com/user-attachments/assets/a30e322d-a343-4f24-bea5-39f928616b67


Collapsed structure tree, adding structure via gallery:

https://github.com/user-attachments/assets/aadae030-15c8-40d5-afd0-185af5d6b40e

@henning-gerhardt Did you use a different way to create the structure?

henning-gerhardt:
Ah, I see now what the expecting is: The structure panel itself should not be reopened and not the entries inside the structure panel. I had read and expect this wrong. I will re-test this again after I know what the real meaning is.

henning-gerhardt (approved):

solth:
In hindsight I should probably have added a video or screenshot to the original issue description to better explain the problem. Sorry for the confusion.

PR #7210, opened 2026-08-27
Let mouse wheel zoom snap to view resolution levels

### LINKED ISSUE #4244: Mousewheel zoom steps to small
In the single page view of the metadata editor, the mouse wheel can be used to zoom the page in and out. However, the zoom change is very small and it needs a lot of turning the mouse wheel to zoom.

**Goal:** the mouse zoom steps should be bigger so that it needs less mousewheel turning

solth:
Fixes #4244 by using option [constrainResolution](https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#setConstrainResolution) to link mouse wheel "steps" to zoom levels in OpenLayer map of gallery detail view.

_Note:_ I will open backports of this fix for Kitodo `3.9.x` and `4.0.x` once this pull request is approved


Kitodo/src/main/webapp/WEB-INF/resources/js/ol_custom.js
```diff
@@ -358,7 +358,8 @@ class KitodoDetailMap {
                 constrainOnlyCenter: true,
                 smoothExtentConstraint: true,
                 showFullExtent: true,
-                padding: [20, 20, 20, 20]
+                padding: [20, 20, 20, 20],
+                constrainResolution: true
```
BartChris, line 362: I tested this and did not notice a real difference with regard to the necessary mouse scrolls. Setting `zoomFactor` to 1.3 had more effect on the aspects @matthias-ronge seems to talk about in the issue.

solth: > I tested this and did not notice a real difference with regard to the necessary mouse scrolls. Setting `zoomFactor` to 1.3 had more effect on the aspects @matthias-ronge seems to talk about in the issue.

Really? That's interesting - and exactly the opposite of what I have observed!

_Zooming 10 "ticks" in the current `main` branch looks like this:_

https://github.com/user-attachments/assets/d58ce4cf-de49-4e20-b581-79da735dc72e


_Zooming 10 "ticks" with `zoomFactor: 1.3`:_

https://github.com/user-attachments/assets/8dcec796-1cca-495d-ac06-c80d0d8d3384


_Zooming 10 "ticks" with `constrainResolution: true`:_

https://github.com/user-attachments/assets/25f1f32b-d259-41f8-b917-4044a86fd6cb


I am using Firefox, version 154 on MacOS 26.6.3 (Tahoe), btw.

@BartChris maybe have to reload your browsers cache to see the real difference with using `contrainResolution`?

oliver-stoehr (approved):
Works fine with Chrome, Firefox, Edge and Safari (macOS).

BartChris:
I will try to check if i just made a setup error. You are on MacOS? Are you using mousewheel or touchpad?

solth:
> I will try to check if i just made a setup error. You are on MacOS? Are you using mouswheel or touchpad?

I am using the mouse wheel, @oliver-stoehr is probably using the touchpad, I guess.

oliver-stoehr:
Yes, I'm using macOS with touchpad.

BartChris:
Hmm, both of you can confirm an improvement, so this is strange. I am on Ubuntu Linux and what makes the difference there is setting to `zoomFactor: 1.3` independent of the browser. It would be interesting to see what happens on Windows. Would it maybe be an option to do both settings? Or maybe @thomaslow or @henning-gerhardt can test what happens on their systems.

henning-gerhardt:
I know that Firefox at least has an option called "Use smooth scrolling". I had disabled this as I want a fast scrolling. Maybe this setting has some kind of influence. I can try this changes tomorrow.

solth:
> Would it maybe be an option to do both settings?

I think that's a good idea to cover all bases. I will add both settings and check whether it has any adverse effects on how the zoom feature works for me. Maybe you can do the same for your setup and confirm that it works for you, too.

BartChris:
There seems to be some device specific stuff going on, especially in OpenLayers version 6 which we are appearantly using:
https://github.com/openlayers/openlayers/issues/15423

BartChris:
Setting it to to 1.2 is already good enough for me


https://github.com/user-attachments/assets/844bb689-fd58-4b3e-aeec-3f3435dbe006

compared to main (1.1)

https://github.com/user-attachments/assets/aebf4261-bb96-48b7-b3c4-c78da52d4987

solth:
> Setting it to to 1.2 is already good enough for me

And that is with setting `constrainResolution` as well?

BartChris:
> > Setting it to to 1.2 is already good enough for me
> 
> And that is with setting `constrainResolution` as well?

Yes, `constrainResolution` has no visible effect for me, but i suppose it is important what happens on Windows. I will try to set those things on our production systems and let my colleagues test in Windows.

Edit: 
i have to correct myself. When adding `constrainResolution` (in combination with zoom 1.2) things seem even faster

https://github.com/user-attachments/assets/929cc42e-b01f-41b0-8b71-b35a22d00256

Changing only `constrainResolution` has no visible effect over `main` as far as i can see:


https://github.com/user-attachments/assets/622348e2-e27d-4675-a206-39c76851106b

Edit: Given the evidence from only my machine i would vote for `zoomFactor: 1.2` and `constrainResolution: true`.

thomaslow:
For me, zooming is a bit slow in the current main branch, but still usable. I would call it "precise". It is definitely way faster than @solth video. It looks to be the same as @BartChris video. Increasing zoomFactor to 1.2 would be an improvement for my devices (Fedora, Windows, Firefox, Chrome with basic Logitech Mouse).

Looking at the [documentation](https://openlayers.org/en/latest/apidoc/module-ol_View-View.html) of OpenLayers, the purpose of the setting `constrainResolution: true` is to enable auto-zooming to the closest pre-defined resolution level (or zoom level). When setting it to `true`, we should also specify a list of predefined `resolutions`. I guess, by default, if none are provided, some reasonable resolutions are auto-calculated somehow, which is why zooming may be different for some people (or not) with this setting (jumping from one auto-calculated resolution level to the next).

I think adjusting the `zoomFactor` would be better. Given that there are many different devices out there, maybe this can be a user setting in the future, e.g. with a interactive slider ranging from 1.05 to 1.5? Then, everybody can decide on their own.

solth:
> I think adjusting the `zoomFactor` would be better. Given that there are many different devices out there, maybe this can be a user setting in the future, e.g. with a interactive slider ranging from 1.05 to 1.5? Then, everybody can decide on their own.

Sounds like a good plan for the future. But I take it that you don't opose merging the current pull request as it is as a first step then? This could also be backported to `3.9.x` and `4.0.x`, while adding a new setting to the individual metadata editor configuration (which I support in general!) is something that would require a database migration and thus would not be eligible for a backport, and could therefor be added to the next major version (e.g. `4.1.0`, for example)

thomaslow:
> But I take it that you don't opose merging the current pull request

I am not opposed. Hopefully, this PR should improve things. Apparently, my system is not really affected by this. Both zoomFactor 1.1 and 1.2 work fine for me. One is a bit faster than the other. With `contrainResolution: true` zooming is even faster, but still perfectly fine. Personally I would prefer 1.2 with `contrainResolution: false` (see below). But I guess, maybe, there are other systems where this change is perceived by users as making things worse, I don't know.

One additional observation: In Windows, there is a setting at "Einstellungen -> Bluetooth und Geräte -> Maus -> Zeilen, die gleichzeitig gescrollt werden sollen" which controls how fast pages are scrolled with a mouse wheel. With `constrainResolution: true`, it does not matter whether you choose 1 (minimum) or 100 (maximum). The zoom behavior in OpenLayers is the same. With `constrainResolution: false`, this setting also influences the zoom speed. Again, this could be an advantage or disadvantage for some users, depending on their preferences.

solth:
> Personally I would prefer 1.2 with `contrainResolution: false` (see below). But I guess, maybe, there are other systems where this change is perceived by users as making things worse, I don't know.

Leaving `constrainResolution: false` results in a larger `zoomFactor` having next to no effect at all for me, so I am going to merge this pull request with both settings adjusted. It should be an improvement on the current situtation for most users and can be the start of further, incremental enhancements.

henning-gerhardt:
I tried this current changes and there are no big differences for me. It is maybe a little bit faster but no big difference. I did not use this feature very often - until this pull request only once or twice, so my impression could be wrong.

Issue #4718, opened 2021-10-04
Rulesets selection displays namespace declarations


matthias-ronge:
In the selection menu for rulesets, there are also namespace declaration files listed.

![Screenshot 2021-10-04 092359](https://user-images.githubusercontent.com/3040657/135810211-3b3acab3-225e-4f96-850a-af6bc8ca53a5.png)

Yes, these are in the same directory as rulesets files, but look differently:

Ruleset file:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<ruleset xmlns="http://names.kitodo.org/ruleset/v2">
    <declaration>
        <division id="section">
            <label>Section</label>
        </division>
        <!-- ... -->
```

Namespace file:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<namespace about="http://id.loc.gov/vocabulary/relators/" xmlns="http://names.kitodo.org/ruleset/v2">
    <option value="http://id.loc.gov/vocabulary/relators/abr">
        <label>Abridger</label>
    </option>
    <option value="http://id.loc.gov/vocabulary/relators/acp">
        <label>Art copyist</label>
    </option>
    <!-- ... -->
```

**Goal:** These files shouldn’t be listed here.

solth:
In my opinion we shouldn't mix files with different purposes in the same directory like this, just because they both happen to be XML files. Instead, we should have two separate, dedicated folders where the first contains _just_ rulesets and the second contains _just_ namespace files, if possible. Having to open all files in a folder just in order to analyse their contents and to be able to filter them for a frontend view seems highly ineffective and unnecessary to me.

The same is true for XSLT files, by the way. Kitodo just defines one `XSLT` folder, which contains not just mapping files for import and export, but also for docket creation. I think it's wrong to group files by their type or extension. Instead, they should be grouped by intended purpose. Therefore we should replace the `XSLT` folder by two folders `metadataMappings` and `dockets`, IMHO.

PR #7214, opened 2026-08-27
Handle 'IllegalArgumentException' when opening ruleset with missing ruleset file

### LINKED ISSUE #7207: Uncaught `IllegalArgumentException` when opening a ruleset with missing ruleset file
**Describe the bug**
When trying to open a ruleset whose ruleset file cannot be found, an unhandled `IllegalArgumentException` is thrown and its stack trace is displayed on the general error page. Thus, not only is the error handling missing, but it also prevents fixing a ruleset with a missing file by selecting a different file, since the corresponding form elements are not displayed.

**To Reproduce**
1. Create a ruleset from an existing ruleset file
2. Change the name of the ruleset file on the file system
3. Try to edit the ruleset via the Kitodo UI

**Expected behavior**
The error message about the missing ruleset file should be displayed in the corresponding messages component and the other form elements should still be available and usable, especially to fix the ruleset.

**Screenshots**

Image

**Release**
`4.0.1`, current `main` branch

**Additional context**
`3.9.x` releases are not affected because the exception is thrown when trying to load the functional metadata keys defined in the ruleset (see #6629), which was first introduced in Kitodo [4.0.0](https://github.com/kitodo/kitodo-production/releases/tag/kitodo-production-4.0.0)

solth:
Fixes #7207 by catching the `IllegalArgumentException` and displaying the error message in the corresponding message component.  

Bildschirmfoto 2026-08-27 um 16 31 54

Since said `error-messages` component is only available to display error messages _after_ the  `RENDER_RESPONSE` is finished, we need to check for the correct JSF lifecycle phase before setting the error message with the `Helper.setErrorMessage` method.

Distinguishing between these cases is necessary because `IOException`s can occur
1. when the page is initially loaded with a specific ruleset file (`RENDER_RESPONSE` phase is **not** yet finished)
2. when selecting a different file from the list of available ruleset files (`RENDER_RESPONSE` phase **is** finished)

_Note:_ I will open backports of this fix for Kitodo `4.0.x` once this pull request is approved (no fix required for `3.9.x` since the bug does not occur there: the problem was originally caused by loading and displaying functional metadata in the ruleset edit page, which is not available in Kitodo `3.9.x`)

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/forms/RulesetEditView.java
+++ b/Kitodo/src/main/java/org/kitodo/production/forms/RulesetEditView.java
@@ -24,6 +24,8 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import jakarta.faces.context.FacesContext;
+import jakarta.faces.event.PhaseId;
 import jakarta.faces.view.ViewScoped;
 import jakarta.inject.Named;
 
@@ -56,7 +58,7 @@ public class RulesetEditView extends ValidatableForm {
     private static final String AT_MARK = "@";
 
     /**
-     * Initialize Rulset form.
+     * Initialize Ruleset form.
      */
     public RulesetEditView() {
         super();
@@ -148,6 +150,9 @@ public class RulesetEditView extends ValidatableForm {
         if (Objects.nonNull(id) && !Objects.equals(id, 0)) {
             try {
                 this.ruleset = ServiceManager.getRulesetService().getById(id);
+                if (!hasValidRulesetFilePath(this.ruleset, ConfigCore.getParameter(ParameterCore.DIR_RULESETS))) {
+                    Helper.setErrorMessage("rulesetNotFound", new Object[] {this.ruleset.getFile()});
+                }
             } catch (DAOException e) {
                 Helper.setErrorMessage(ERROR_LOADING_ONE, new Object[] {ObjectType.RULESET.getTranslationSingular(), id }, logger, e);
             }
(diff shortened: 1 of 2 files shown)


BartChris (commented):
I tested the functionality and it works good. I can recover from a broken state, where the currently selected ruleset file does not exist. 

I stepped through with the debugger but that condition never evaluated to true if i see correctly:

` (Objects.nonNull(facesContext) && PhaseId.UPDATE_MODEL_VALUES.equals(facesContext.getCurrentPhaseId()))`

What do i have to do to actually test that condition? Value is always `PhaseId.RENDER_RESPONSE` for me.

solth:
@BartChris is there something strange going on with GitHub right now or did you delete your last comment? Didn't you write "Edit: I checked again. I have missed to check with an invalid file. I could now trigger the exception."? Or did you encounter some other problem afterwards?

BartChris:
@solth I thought i could trigger it, but i was not able to when checking again. So i deleted the message. Before further fumbling i wanted to wait for your explanation here. Sorry for the confusion.

solth:
@BartChris This can only be triggered by some edge cases like symbolic links pointing to non-existent files. It shouldn't be necessary very often, but has to be covered nonetheless, IMO.

solth:
As a side note: since ruleset file validation is very fast, we could consider to pre-filter the options in the pulldown menu accordingly to fix #4718. That would have the downside to loose informatin about the validation errors, though. 

Anyway, that is not part of this pull request here right now.

BartChris (approved):

PR #7216, opened 2026-08-28
Fix security warnings about local information disclosure in temporary directories

thomaslow:
This PR fixes 4 instances of the CodeQL security warning "Local information disclosure in a temporary directory".

The problem is that the old Java API `File.createTempFile` does not necessarily apply very restrictive permissions to these files such that other non-tomcat users may get access to them. Instead, it is now recommended to use the newer API `Files.createTempFile` (with an additional `s`). However, there is still no guarantee that files are assigned the most restrictive permissions available. The official [documentation](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Files.html) of `Files.createTempFile` only states:

> When no file attributes are specified, then the resulting file may have more restrictive access permissions to files created by the [File.createTempFile(String,String,File)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/File.html#createTempFile(java.lang.String,java.lang.String,java.io.File)) method.

We could also try to set specific permissions, but I couldn't find an OS-independent way to do this.

Testing these changes in my Ubuntu 26.04 docker container (my tomcat runs as root user inside the container) reveals the following access permissions, which seem reasonable to me. 

For a temporary directory (created by the ImageManagement class):

```
drwx------+  1 root   root      0 Aug 28 10:39 ImageManagement.tmpDir17890399326253926715
```

For a temporary file within that directory:

```
-rw-------+   1 root root   0 Aug 28 11:02 sizedWebImage-6578674863344965441.jpeg
```

Previously, the global `/tmp` directory was used (which can be accessed by everyone), and files had the following permissions:

```
-rw-r-----+  1 root root   0 Aug 28 10:39 sizedWebImage-12077786332429147037.jpeg
```


Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ImageManagement.java
```diff
@@ -46,7 +46,14 @@ public class ImageManagement implements ImageManagementInterface {
      * Temporary directory location.
      */
     private static final File TMPDIR = new File(
-            KitodoConfig.getParameter(ParameterImageManagement.DIR_TMP, System.getProperty("java.io.tmpdir")));
+            KitodoConfig.getParameter(ParameterImageManagement.DIR_TMP, () -> {
+                try {
+                    return Files.createTempDirectory(ParameterImageManagement.DIR_TMP.getName()).toString();
+                } catch (IOException e) {
+                    logger.error("could not create temporary directory for image management", e);
+                    return null;
```
henning-gerhardt: Would this `null` return value not cause null pointer exception on later access (`supplier.get()` call in `KitodoConfig` class) or other places?

Kitodo-API/src/main/java/org/kitodo/config/KitodoConfig.java
```diff
@@ -114,6 +115,30 @@ public static String getParameter(ParameterInterface key, String defaultValue) {
...
+     *
+     * @param key
+     *            as ParameterInterface enum implementation whose value is to be
+     *            returned
+     * @param supplier
+     *            the supplier function that is called in case the parameter taken from 
+     *            the config file does not exist or an exception occurred
+     * @return value for the requested key, or the supplied default value if not found
+     */
+    public static String getParameter(ParameterInterface key, Supplier<String> supplier) {
```
henning-gerhardt, line 130: Can you provide tests for this new method?

Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ImageManagement.java
```diff
@@ -46,7 +46,14 @@ public class ImageManagement implements ImageManagementInterface {
      * Temporary directory location.
      */
     private static final File TMPDIR = new File(
-            KitodoConfig.getParameter(ParameterImageManagement.DIR_TMP, System.getProperty("java.io.tmpdir")));
+            KitodoConfig.getParameter(ParameterImageManagement.DIR_TMP, () -> {
+                try {
+                    return Files.createTempDirectory(ParameterImageManagement.DIR_TMP.getName()).toString();
+                } catch (IOException e) {
+                    logger.error("could not create temporary directory for image management", e);
```
henning-gerhardt: Suggest replace "could not create ..." to "Could not create ..."

henning-gerhardt (changes requested):
Works in general and can be a big improvement on systems with not strict default system settings to little / non improvement on systems with strict / very strict settings.

henning-gerhardt:
> We could also try to set specific permissions, but I couldn't find an OS-independent way to do this.

As we support official only Unix/Linux based systems we could implement this and this need in every case a OS dependent way as every OS has other values for this permissions.

You current solution is in any way a good step on this way to improve the security of temporary files.

Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ImageManagement.java
thomaslow: Yes, you are correct, this is a bit problematic, because `null` would cause a NPE during class initialization because of `new File(null)`. I changed it to store the path as a string instead and check the string for `null` when it is used in `summarize()`.

Kitodo-API/src/main/java/org/kitodo/config/KitodoConfig.java
thomaslow: Added simple tests.

Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ImageManagement.java
thomaslow: Done.

Kitodo-ImageManagement/src/main/java/org/kitodo/imagemanagement/ImageManagement.java
```diff
@@ -137,8 +145,10 @@ public Image getSizedWebImage(URI sourceUri, int width) throws IOException {
      */
     private static Image summarize(String prefix, String suffix, URI sourceUri, Function<FutureDerivative, ?> lambda,
             String message, Object pTwo) throws IOException {
-
-        File tempFile = File.createTempFile(prefix, suffix, TMPDIR);
+        if (Objects.isNull(TEMPORARY_DIRECTORY_PATH)) {
+            throw new IOException("temporary directory for image generation not available");
```
henning-gerhardt: Suggested change:  From `temporary directory ...` to `Temporary directory ...`

henning-gerhardt (approved):

PR #7222, opened 2026-08-31
Exclude source folder from derivative generation

BartChris:
This Pull Request adresses the loigc related parts of the problems described in https://github.com/kitodo/kitodo-production/issues/6680 by preventing the deletion of source files when creating derivatives.


matthias-ronge (approved):
I think that solves the problem.

PR #7223, opened 2026-08-31
[3.9] Translate column selection

solth:
Backport of #7221 for Kitodo `3.9.x`


henning-gerhardt (approved):
Works but if you change your display language to an other language column selection did not get translated to the new used language like in 4.x where language switch is happen on all elements. I think that current solution is better than mixed / no translations.

solth:
> Works but if you change your display language to an other language column selection did not get translated to the new used language like in 4.x where language switch is happen on all elements. I think that current solution is better than mixed / no translations.

I fear this might be due to the fundamental refactoring of backing beans and scopes for Kitodo 4.0 (see #6456) and cannot be resolved for `3.9.x` without substantial changes.

PR #7229, opened 2026-09-01
Generate selenium screenshots if tests fail

thomaslow:
This PR generates a screenshot of the current browser window in the directory `Kitodo/target/selenium-screenshots` for every selenium test that failed. Hopefully, this will help to better understand why certain tests fail sometimes. 

Keep in mind that a screenshot might not truly represent what happened in a specific error scenario because of timing problems. Maybe the browser was still loading at the the time of the error, but was fully loaded at the time of the screenshot. Hopefully, it still provides some additional insight.

Screenshots are also stored as a downloadable build artifact for failed GitHub CI workflows (only if there are any failed tests).

### DIFF
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
@@ -27,7 +27,7 @@ import org.apache.logging.log4j.Logger;
 import org.jspecify.annotations.Nullable;
 
 /**
- * Junit TestWatcher implementation that monitors failed tests and generates a screenshot of the current browser window.
+ * JUnit TestWatcher implementation that monitors failed tests and generates a screenshot of the current browser window.
  */
 public class ScreenshotTestWatcher implements TestWatcher {
 
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
@@ -26,6 +26,9 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jspecify.annotations.Nullable;
 
+/**
+ * Junit TestWatcher implementation that monitors failed tests and generates a screenshot of the current browser window.
+ */
 public class ScreenshotTestWatcher implements TestWatcher {
 
     private static final Logger logger = LogManager.getLogger(ScreenshotTestWatcher.class);
@@ -35,6 +38,9 @@ public class ScreenshotTestWatcher implements TestWatcher {
         "target/selenium-screenshots"
     );
     
+    /**
+     * Is executed when a test case fails and saves a screenshot of the current browser window.
+     */
     public void testFailed(ExtensionContext context, @Nullable Throwable cause) {
         WebDriver driver = Browser.getDriver();
         if (!(driver instanceof TakesScreenshot)) {
@@ -42,22 +48,22 @@ public class ScreenshotTestWatcher implements TestWatcher {
             return;
         }
 
+        String className = context.getRequiredTestClass().getSimpleName();
+        String methodName = context.getTestMethod()
+                .map(method -> method.getName())
+                .orElse("unknown");
+
+        Path directory = Path.of(SCREENSHOT_DIRECTORY).toAbsolutePath().normalize();
+        Path filepath = directory.resolve(String.format("%s-%s-%d.png", className, methodName, System.currentTimeMillis()));
+
+        logger.debug("generating selenium screenshot at {} for failed test case {}#{}", filepath, className, methodName);
+        byte[] screenshot = ((TakesScreenshot) driver).getScreenshotAs(OutputType.BYTES);       
+
         try {
-            byte[] screenshot = ((TakesScreenshot) driver).getScreenshotAs(OutputType.BYTES);
-
-            Path directory = Path.of(SCREENSHOT_DIRECTORY).toAbsolutePath().normalize();
             Files.createDirectories(directory);
-
-            String className = context.getRequiredTestClass().getSimpleName();
-            String methodName = context.getTestMethod()
-                    .map(method -> method.getName())
-                    .orElse("unknown");
-
-            Path filepath = directory.resolve(String.format("%s-%s-%d.png", className, methodName, System.currentTimeMillis()));
-            logger.debug("writing selenium screenshot to {}", filepath);
             Files.write(filepath, screenshot);
         } catch (IOException e) {
-            logger.error("error generating screenshot after test failure", e);
+            logger.error("error saving screenshot after test failure", e);
         }
 	}
 }
--- a/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
+++ b/Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
@@ -41,6 +41,7 @@ public class ScreenshotTestWatcher implements TestWatcher {
     /**
      * Is executed when a test case fails and saves a screenshot of the current browser window.
      */
+    @Override
     public void testFailed(ExtensionContext context, @Nullable Throwable cause) {
         WebDriver driver = Browser.getDriver();
         if (!(driver instanceof TakesScreenshot)) {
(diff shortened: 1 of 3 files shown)


thomaslow:
Screenshots can be downloaded from the "[summary](https://github.com/kitodo/kitodo-production/actions/runs/33524587021?pr=7229)" page of the failed "CI / build" workflow or by opening the "[upload selenium screenshots](https://github.com/kitodo/kitodo-production/actions/runs/33524587021/job/99912053591)" log protocol via the drop-down menu and following the download link shown at the end of the log protocol.

thomaslow:
@stweil Have a look at the latest [failed pipeline](https://github.com/kitodo/kitodo-production/actions/runs/33524587021?pr=7229) and the associated screenshot for the failed `MetadataImagePreviewST.zoomLevelTest`. Based on the screenshot, for some reason, the browser navigated to the indexing page instead of the dashboard page. As far as I know, the indexing page is automatically opened after login if `isIndexCorrupted` is true. Maybe there is still some kind issue when resetting the index for each test case.

MetadataImagePreviewST-zoomLevelTest-1788288447058

solth:
This might proof useful if it works. 

We did have some similar functionality that would even send screenshots of failed Selenium tests to the user (implemented way back in #1303, I think), but it didn't really work. Since it depended on Travis it was later removed when we switched from Travis to GitHub CI in #5534

thomaslow:
> We did have some similar functionality that would even send screenshots of failed Selenium tests to the user (implemented way back in #1303, I think), but it didn't really work. Since it depended on Travis it was later removed when we switched from Travis to GitHub CI in #5534

Ok, I can't promise that it will work. In one pipeline I noticed a generic screenshot containing only Kitodo's background image, no login form, no navigation bar, nothing. I am not sure whether this is caused by the failing test or by the way screenshots are taken:

ImportingST-checkOptionalMetadataValidationDuringProcessCreation-1788276761376

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
```diff
@@ -0,0 +1,73 @@
...
+    public void testFailed(ExtensionContext context, @Nullable Throwable cause) {
+        WebDriver driver = Browser.getDriver();
+        if (!(driver instanceof TakesScreenshot)) {
+            logger.warn("cannot take screenshot with driver that doesn't support screenshots: {}", driver);
+            return;
+        }
+
+        String className = context.getRequiredTestClass().getSimpleName();
+        String methodName = context.getTestMethod()
+                .map(method -> method.getName())
```
solth, line 57:
```suggestion
                .map(Method::getName)
```

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * It is licensed under GNU General Public License version 3 or later.
+ *
+ * For the full copyright and license information, please read the
+ * GPL3-License.txt file that was distributed with this source code.
+ */
+
+package org.kitodo.selenium.testframework.helper;
+
+import org.junit.jupiter.api.extension.ExtensionContext;
```
solth, line 14: The following import is required when applying the suggestion in line 48
```suggestion

import org.jspecify.annotations.NonNull;
import org.junit.jupiter.api.extension.ExtensionContext;
```

solth (commented):
Code looks good, just some very minor suggestions.

Lets hope this helps finding the cause of those Selenium test failures! 👍

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
thomaslow: What do you mean by suggestion in line 48? Do you want to add "NonNull" or "Nullable" annotations everywhere in the future?

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
```diff
@@ -0,0 +1,73 @@
...
+    );
+    
+    /**
+     * Is executed when a test case fails and saves a screenshot of the current browser window.
+     * 
+     * @param context the current extension context
+	 * @param cause the throwable that caused test failure
+     */
+    @Override
+    public void testFailed(ExtensionContext context, @Nullable Throwable cause) {
```
solth, line 48: My IDE tells me the first parameter should be `@NonNull` annotated (don't forget the additional import when applying this suggestions)
```suggestion
    public void testFailed(@NonNull ExtensionContext context, @Nullable Throwable cause) {
```

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
solth: Sorry, it seems I forgot to add one review comment I had prepared concerning this suggestion (done now, see below)

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
thomaslow: My IDE doesn't complain about `context` missing a `@NonNull` annotation, but reports a problem about `Method::getName`:

> Null type safety: parameter 'this' provided via method descriptor Function<Method,String>.apply(Method) needs unchecked conversion to conform to '`@Nonnull` Method'Java(67109822)

Not sure what the problem is exactly. Seems like some internal thing to me (combination of `Optional.map` and the static reference of `Method::getName`). A minimal example is this:

```Java
Optional<String> text = Optional.of("test");
text.map(String::toLowerCase).orElse("null"); // null type safety problem
text.map(t -> t.toLowerCase()).orElse("null"); // no complaints
```

thomaslow: It seems that our IDEs use a different linter implementation for this whole "Null Analysis" topic. And in this special scenario, fixing your IDE warning introduces a warning on my side. 

Also, after reading about this, it seems to me that this null analysis is not very stable at the moment, mainly because not all code is annotated (even Java libraries), there are many different annotations for this, and linters sometimes try to (incorrectly) guess whether something is nullable or nonnull. 

My suggestion would be to find a common ground for this such that every developer has the same warnings or errors, either by

1. forcing an IDE (not my favorite approach)
2. force null analysis to be part of the maven compiler step, so warnings and errors show up during compilation (independent of the IDE) and trigger a compilation failure (also in the GitHub CI)
3. globally ignore null analysis

I would prefer 2 or 3. What do you think?

Kitodo/src/test/java/org/kitodo/selenium/testframework/helper/ScreenshotTestWatcher.java
thomaslow: Done

thomaslow:
I removed the `@Nullable` annotation since it is not something we use in Kitodo.Production anyway at the moment.

solth:
Sorry for the delay in my reply. It was mainly because I didn't have an immediate answer, but after reading your explanation I agree with you and would say we shouldn't focus on null analysis and IDE settings right now. I think IDE suggestions often hint to sensible ideas to improve the code, so I tend to heed their advice, but you might have found an exception here. 

Since this particular case does not seem to do any harm, though, I think it was okay to apply the suggestion and move forward to avoid wasting too much time on this topic.

I have to admit I do not understand what the potential problem about the method reference in the lambda expression is, though, but since you applied that suggestion as well I suppose you don't deem this as problematic, either, do you?

thomaslow:
> I have to admit I do not understand what the potential problem about the method reference in the lambda expression is, though, but since you applied that suggestion as well I suppose you don't deem this as problematic, either, do you?

It seems to be a result of the problem that a lot of Java libraries are not yet fully annotated, and there is an open issue for VSCode see https://github.com/redhat-developer/vscode-java/issues/2731.

I wasn't really aware of this null analysis topic before. I also installed IntelliJ to figure out the origin of your suggestion. At the moment, I still don't really know why `@NonNull ExtensionContext context` is suggested by IntelliJ. In my opinion, either `@Nullable` or `@NonNull` should be chosen to be the default interpretation. Annotating everything with both seems excessive to me. Also the interface class `TestWatcher` doesn't claim `context` to be not null with an annotation, only with a Javadoc comment. But I doubt this was parsed from the Javadoc. I don't know.

PR #7231, opened 2026-09-01
Add loading symbol to metadata editor gallery detail view

### LINKED ISSUE #7057: Signal loading in progress for large images in detail view
When loading large image files in the detail view of the metadata editors gallery it can sometimes seem as though nothing is happening or like an error occured because the image does not appear for a long time and no indication is given that the image is actually being loaded:

https://github.com/user-attachments/assets/4b71a5db-453b-4552-bc79-c94413f52eb1

The progress of loading images should be visible to the user. If possible, the images should be [rendered incrementally](https://en.wikipedia.org/wiki/Incremental_rendering), like in classic HTML tags.

Alternatively, a progess bar could signal the progess of the image loading process.

The simplest and cheapest solution would probably be to just show a loading symbol in the view, from the time the user clicked a thumbnail until the loading of the images is finished and the images is actually displayed.

To determine the time in which the progress bar or loading symbol should be displayed, the [image load events](https://openlayers.org/en/latest/examples/image-load-events.html) of OpenLayers can probably be used.

solth:
This pull request adds a small SVG loader graphic that is displayed when an image starts loading and removed/hidden when the rendering of the image/media finished, to proivide better feedback to the user. Thus fixes #7057 

https://github.com/user-attachments/assets/948ef689-c46e-4017-82da-3236beb6918a

_Note:_ I will provide backports of this fix for Kitodo `3.9.x` and `4.0.x` once this pull request has been approved


Kitodo/src/main/webapp/WEB-INF/resources/css/kitodo.css
```diff
@@ -4089,6 +4089,15 @@ div[id$='metadataTable'].ui-treetable tr.focusedRow {
     background-color: var(--blue);
 }
 
+#imagePreviewForm\:loadingSymbol {
+    display: flex;
+    height: 100px;
+    margin: auto;
+    position: absolute;
+    left: calc(50% - 100px);
+    top: calc(50% - 100px);
```
oliver-stoehr:
```suggestion
    top: calc(50% - 50px);
```
The icon should be place a little bit lower to be visually centered.

oliver-stoehr (changes requested):
The loading icon does not hide when loading an audio file and keeps spinning in the background:
Image

The loading icon should also be added to the `externalView.xhtml`.

And there is still an old loading icon present. I think it is broken quite some time now and not visible, anyway. It should be removed, now that we add a new loading icon:
- `<i class="fa fa-spinner fa-spin"/>` in `externalView.xhtml`
- `<i class="fa fa-spinner fa-spin"/>` in `mediaDetail.xhtml`
- `loadingIcon` in `ol_custom.js::showCanvas`
- `loadingIcon` in `ol_custom.js::hideCanvas`
- `#map > i` in `kitodo.css`
- `#loader` in `media_detail_audio_waveform.js` (Line 88 is probably the place where the new loading icon should be hidden, when loading an audio file.)

solth:
@oliver-stoehr would you mind checking the changes again? I hope I covered all points you mentioned.

oliver-stoehr (approved):

PR #7235, opened 2026-09-03
Prevent NPE when deleting a process

### LINKED ISSUE #7234: NPE when trying to delete a process
**Describe the bug**
When trying to delete a process from a system where an `ImportConfiguration` of type **Process template** was configured, a `NullPointerException` was thrown. The `ImportConfiguration` in question did not contain an actual process, so maybe that was the problem.

**To Reproduce**
1. Have a broken import configuration of type **Process template** in the system with missing process (if necessary, set field `default_templateprocess_id` in table `importconfiguration` to `<null>` to simulate)
2. Try to delete any process
3. See error

**Expected behavior**
It should be possible to delete a process independent of whether a broken `ImportConfiguration` exists in the system or not.

**Screenshots**

Image

**Release**
`4.1.0-SNAPSHOT`, `4.0.1`, `3.9.3`

**Additional context**
It shouldn't be possible to end up with an `ImportConfiguration` of type **Process template**, but in case this does happen, the system should have a corresponding check in place.

solth:
Fixes #7234


BartChris (approved):
Tested and works.

PR #7239, opened 2026-09-04
Prevent potential NPEs in 'Block.java' reported by CodeQL

solth:
This adds some null checks in `Block.java` to prevent potential `NullPointerException`s found by CodeQL.

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Block.java
+++ b/Kitodo/src/main/java/org/kitodo/production/model/bibliography/course/Block.java
@@ -488,7 +488,7 @@ public class Block {
             Helper.setErrorMessage(e.getMessage());
         }
         try {
-            if (!this.firstAppearance.equals(firstAppearance)) {
+            if (Objects.nonNull(this.firstAppearance) && !this.firstAppearance.equals(firstAppearance)) {
                 course.clearProcesses();
             }
         } catch (NullPointerException e) {
@@ -533,7 +533,7 @@ public class Block {
             Helper.setErrorMessage(e.getMessage());
         }
         try {
-            if (!this.lastAppearance.equals(lastAppearance)) {
+            if (Objects.nonNull(this.lastAppearance) && !this.lastAppearance.equals(lastAppearance)) {
                 course.clearProcesses();
             }
         } catch (NullPointerException e) {
@@ -576,7 +576,7 @@ public class Block {
     public void setPublicationPeriod(LocalDate firstAppearance, LocalDate lastAppearance) {
         prohibitOverlaps(firstAppearance, lastAppearance);
         try {
-            if (!this.firstAppearance.equals(firstAppearance)) {
+            if (Objects.nonNull(this.firstAppearance) && !this.firstAppearance.equals(firstAppearance)) {
                 course.clearProcesses();
             }
         } catch (NullPointerException e) {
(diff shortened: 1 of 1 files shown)


henning-gerhardt:
I'm unsure to judge and review this change. CodeQL is reporting possible NPEs but a few lines later the possible causing NPEs are caught and additional code is executed. So I think that null values and NPEs are excepted at this code execution and with your change it is possible that this code path are not executed.

With this change maybe the expected behaviour is changed or maybe not as there is no tests for this class at all which is a general big issue with the class. An other - especially with the first and last date settings - issue is that invalid dates are detected, an error message is thrown but there are additional code is executed which try to set class values and is manipulation an injected class object. This is very confusing and could be more understandable if there are tests which are showing what is expected and what not.

It is not up to you in this pull request to add tests and show that the expected behaviour is not affected by this change.

I will look into this a little bit more (I was still doing this on opening this pull request and not only after set review request to me) but it will be good if somebody else is analysing and reviewing this change too.

solth:
@henning-gerhardt I understand what you are saying and I agree that this class should be better covered by tests. 

The code is confusing in parts, but I think the changes suggested here will not have any effect on the behaviour - they just ensure the `.equals` method can be called in the same lines (!) without causing a `NullPointerException` (_if_ `this.firstAppearance` or `this.lastAppearance` is indeed `null`, calling `null.equals` in the same `if` check will definitely trigger the exception):

For example: 

`if (!this.firstAppearance.equals(firstAppearance))`

vs

`if (Objects.nonNull(this.firstAppearance) && !this.firstAppearance.equals(firstAppearance))`

solth:
Or did you mean it was possibly _expected_ that a `NullPointerException` is thrown here (and dealt with later) and that is how my changes would change the behavior of the application?

henning-gerhardt:
> Or did you mean it was possibly _expected_ that a `NullPointerException` is thrown here (and dealt with later) and that is how my changes would change the behavior of the application?

Yes, after long thinking (and maybe overthinking without going crazy) your change will change the behaviour in case that both dates are null and there was processes added to the `course` object.

My focus was: under which conditions is `course.clearProcesses()` is executed and when not.

Without your change:
- Call is executed if
  - `this.firstAppearance` is not null and `firstAppearance` is not equal to `this.firstAppearance` (code execution in line 492)
  - a NPE is thrown an caught by `this.firstAppearance` is null or `course` is null
    - in case that `this.firstAppearance` is null and has thrown the NPE than `firstAppearance` must be not null that the code in line 496 get executed
    - in case that `course` object is null and has thrown the NPE than `this.firstAppearance` exclusive or `firstAppearance` must be null that the code in line 496 get executed
- in any other ways the code will not be executed and the possible attached processes stay inside this list

With your change:
- Call is executed if
  - `this.firstAppearance` is not null and `firstAppearance` is not equal to `this.firstAppearance` (code execution in line 492)
  - `course` object is null and conditions from line above are true as code execution in line 492 can not be executed and no NPE can be thrown
    - After NPE is caught and we know that `this.firstAppearance` is not null only `firstAppearance` can be null that the code in line 496 can be executed which will thrown a new NPE which will be caught outside of this method
- in any other ways the code will not be executed and the possible attached processes stay inside this list

I'm hoping I did read and interpret the code right and did the right conclusions.

solth:
@henning-gerhardt I think you are correct, so I will close this pull request. I do have a feeling however that this condition check is way more complicated than it needs to be, so maybe this should be refactored if possible.

Issue #7238, opened 2026-09-03
Improve performance of interactions in the metadata editor


BartChris:
**Is your feature request related to a problem? Please describe.**
Especially when structuring large digital objects, adding e.g. chapters to a book via the metadata editor sometimes takes multiple seconds, especially applying a new "docstruct"

<img width="649" height="371" alt="Image" src="https://github.com/user-attachments/assets/765d6aa9-57a0-43bf-b38d-f59d8c656c8c" />

Large objects also take long to open up in the editor initially.

I timed a lot of structuring operations and also initially opening the editor. When structuring the document the runtime is not dominated by Java logic but by rendering. And the problem with rendering is not the logical tree but the gallery.

Having hundreds of thumbnails on the page forces JSF to render a lot of primefaces components on load. The same applies for structuring a doc, because when adding a page to the chapter the gallery tree has to be consolidated. After the action the whole chunk of pages in the "unstructured" section, which might contain 800 pages, gets rerendered, which takes seconds.

This all comes down to rerendering the gallery (`galleryWrapperPanel`), when updating the structure.

https://github.com/kitodo/kitodo-production/blob/dc7bb2d029c45e8b99d962cea88f83a732cc28a3/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/dialogs/addDocStrucType.xhtml#L205-L218

In case of mostly unstructured documents, restricting the update command down to

```xml
 update="logicalTree
          imagePreviewForm:structuredPages"
```
brings massive performance improvements, but is of course no option because the gallery panel is out of sync.

The fact that large documents take very long to open in the editor has basically the same reason: Huge rendering tasks in the gallery.

In a test with approximately 800 media items, updating the gallery produced a 2.7 MB JSF partial response. Approximately 2.2 MB of that response consisted solely of the unstructured media list. The underlying Java methods responsible for rebuilding the gallery model took only tens of milliseconds, while rendering the corresponding JSF component tree took seconds. This suggests that reducing the number of rendered/updated thumbnail components is likely to provide substantially larger gains than optimizing the backing Java logic.

This is why i think making substantial performance improvements at both fronts would come down to bringing down rendering costs. I am not sure whether a large thumbnail list would benefit from not being rendered as primefaces components with massive overhead(?) and whether we should rely on more plain HTML here (if this would even be possible)

[...]

PR #7240, opened 2026-09-04
Improve metadata editor gallery performance for large processes

BartChris:
Addresses https://github.com/kitodo/kitodo-production/issues/7238

This Pull Request tries to optimize the performance of the metadata editor with regard to initialization and interaction. I noticed that when structuring large documents there is a large delay to initialize the editor and doing structuring often suffers from slow interactions. While working on that i also stumbled over the discussions and optimizations already done on that topic (https://github.com/kitodo/kitodo-production/discussions/5010) and i agree with the assessment that the current software stack limits the possible optimizations.

While profiling the editor i noticed that many of the performance problems can be explained by large rendering times. This led me to the conclusion that the huge number of components inside the editor makes patterns problematic which are usually the standard solution.

I noticed the following things and adressed them in my changes:

1.  Complex EL expressions have a significant overhead when compared to precalculating the necessary values in Java. 
2. We can make more use of caching and lookup maps to improve the performance of the Java code.
3. Components like `<p:graphicImage>` are a nice convenience, but have a lot of overhead when hundreds of them are rendered. 
4. Repeated work and unnecessarily broad rebuilding or rerendering become very expensive for large documents. Operations should therefore only update the parts of the editor whose state actually changed.

In more detail:

1. While logging runtime it became apparent that EL expressions with function call like `#{DataEditorForm.consecutivePagesSelected()` have a considerable effect on runtime. In therefore extended the data model of `GalleryMediaContent` to precalculate those values before rendering and store them on the view model. This avoid the repeated evaluation of more expensive expressions.

2. Rendering time can also be reduced by using lookup maps. One example is `mediaByPhysicalDivision`, which is built together with the media list and then reused while constructing the gallery stripes. Previously, the corresponding `GalleryMediaContent` was found by repeatedly scanning the complete media list for every View. The lookup map replaces these repeated linear searches with direct access by PhysicalDivision.

3. Huge numbers of PrimeFaces components can slow down rendering. Plain html tags can be rendered faster. I therefore replaced the  `<p:graphicImage>` component for every thumbnail with an ``-tag. This improves performance but has some drawbacks.
One problem is that we have to provide the image with the correct URL ourselves and cannot rely on PrimeFaces doing parts of the work behind the scenes. This means that the image URL now has to be prepared explicitly. 

PrimeFaces normally creates this dynamic resource URL internally while rendering <p:graphicImage>. This URL points back to the application and is associated with the EL expression that provides the dynamic content. In our case this expression is `#{mediaProvider.previewData}`. The code backing this expression then provides the dynamic content when thebrowser retreives the image.

This mechanism is described here:

https://primefaces.github.io/primefaces/13_0_0/#/core/dynamiccontent
Also see:
https://jsfcorner.blogspot.com/2012/11/advanced-primefaces-graphic-image.html

https://stackoverflow.com/a/54984985

 The new implementation uses the same DynamicContentSrcBuilder mechanism once to create the base URL, appends the thumbnail-specific parameters in Java and stores the resulting URL in GalleryMediaContent. The  tag can then use this prepared URL directly.

```java
private String buildPreviewBaseUrl(FacesContext context) {
        ValueExpression expression = context.getApplication()
            .getExpressionFactory()
            .createValueExpression(
                context.getELContext(),
                "#{mediaProvider.previewData}",
                Object.class);

        return DynamicContentSrcBuilder.buildStreaming(
            context,
            null,
            expression,
            true);
}
```

Parts of the implementation, especially around the PrimeFaces dynamic resource URL generation, were developed with the help of AI. The resulting code was reviewed, tested and profiled as part of this change.. While a lot of stuff can still be optimized i already achieved a significantly better startup time. (See videos at the bottom).

Comparison with 1800 pages, most unstructured:

Main:

https://github.com/user-attachments/assets/545ff241-24d0-43f9-9744-d029a564c1e7

PR:

https://github.com/user-attachments/assets/286b7280-e413-4152-933f-d388a0e9b10b


thomaslow:
@BartChris Are you still working on this PR or do you just rebase your commits with main? 

In my opinion, it is easier and more comprehensible if branches are not force-pushed as soon as other people are involved. I simply can't see what you are doing with a force push and have to re-checkout your branch to my local repository.

BartChris:
>  I simply can't see what you are doing with a force push and have to re-checkout your branch to my local repository.

No, this is ready for review, i just wanted to bring the latest changes from main in to see if everything still works. I can stop the merges from main until you have finished your review of course.

thomaslow:
@BartChris From my point of view, you don't have to stop merging with main. Just don't force-push your changes. You can still merge with main and add that as a separate commit. I would see the merge-commit and know what happened and what was changed (aka the purpose of version control systems).

BartChris:
@thomaslow I copied the practice of other committers who also do force pushes after a rebase on main. But your argument is of course a good one. cc @solth

solth:
> @BartChris From my point of view, you don't have to stop merging with main. Just don't force-push your changes. You can still merge with main and add that as a separate commit. I would see the merge-commit and know what happened and what was changed (aka the purpose of version control systems).

I think this becomes a problem when merging with new commits from `main` is mixed with rebasing against an updated `main` branch. Since most of us follow the strategy of rebasing instead of merging (to see the latests merged changes on top, no matter when the pr originated) this maybe difficult for the current pr to keep apart.

thomaslow:
> I think this becomes a problem when merging with new commits from main is mixed with rebasing against an updated main branch. Since most of us follow the strategy of rebasing instead of merging (to see the latests merged changes on top, no matter when the pr originated) this maybe difficult for the current pr to keep apart.

Ah, yes, rebasing a merge-commit might be problematic. I wasn't aware of this yet. I don't rebase very often, mostly on tiny local branches.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -620,6 +627,37 @@ public void show() {
...
+        this.previouslySelectedLogicalNodes = getSelectedLogicalNodes();
+        dataEditor.checkForChanges();
+    }
+
+    /**
+     * Reloads the logical structure tree.
+     *
+     * @param keepSelection whether the current logical tree selection should be restored
+     */
+    public void showLogical(boolean keepSelection) {
```
thomaslow: Why did you add these `showLogical` methods. It looks to me they are copies of the `show` methods, but without updating the physical tree. In "combined" mode, there is no physical tree and updating it should not have any impact on performance. To me, it looks like this change just adds additional duplicate code. Are you sure this change is necessary?

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java
```diff
@@ -417,8 +428,72 @@ private void updateMedia() {
      */
     public void updateStripes() {
         updateMedia();
+
         stripes = new ArrayList<>();
         addStripesRecursive(dataEditor.getWorkpiece().getLogicalStructure());
+
+        updateSelectionState();
+
+        boolean discontinuous = !dataEditor.consecutivePagesSelected();
```
thomaslow: I understand that calling `DataEditorForm::consecutivePagesSelected()` multiple times from various places might negatively affect performance, because it calculates this information again and again, especially for large selections. 

However, I don't really agree with the way this information is now distributed to every single `GalleryMediaContent` instance. From an object oriented point of view, it feels wrong to say `media.setDiscontinuous(true/false)`. The media itself is not discontinuous. The selection state is discontinuous. Wouldn't it be easier to refactor `DataEditorForm::consecutivePagesSelected()` to a class property that is only updated if the selection state is changed? This way, other calls to it would also benefit from better performance, not only the GalleryPanel.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java
```diff
@@ -540,16 +613,42 @@ private GalleryMediaContent buildGalleryMediaContent(View view, String stripeTre
...
+            mediaContentType,
+            view,
+            canonical,
+            Objects.nonNull(previewMediaVariant) ? previewMediaVariant.getMimeType() : null,
+            resourceListUri,
+            Objects.nonNull(mediaViewMediaVariant) ? mediaViewMediaVariant.getMimeType() : null,
+            resourceMediaViewUri,
+            treeNodeId);
+
+        setPreviewUrlSupplier(media);
```
thomaslow: Why do you set the property `previewUrlSupplier` of `media` directly after its constructor via an additional private setter method. This seems convoluted. Could you not just have added the `previewUrlSupplier` to the constructor above?

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryStripe.java
```diff
@@ -115,4 +120,17 @@ public LogicalDivision getStructure() {
     public String getLogicalTreeNodeId() {
         return logicalTreeNodeId;
     }
+
+    public boolean isSelected(String mediaId) {
```
thomaslow: Public methods should have JavaDoc. Also, the name of the method `isSelected` feels wrong here. A method called `GalleryStripe::isSelected` implies to me, that this is related to the selection state of the gallery stripe. I would call this method `isMediaSelected`.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
```diff
@@ -149,6 +150,12 @@ public class StructurePanel implements Serializable {
      */
     private boolean showHierarchyLevel = false;
 
+    /**
+     * Tracks the physical division instances already added while rebuilding the physical structure.
+     * An identity-based set is used because PhysicalDivision has mutable structural equality/hash state.
+     */
+    private Set<PhysicalDivision> addedPhysicalDivisions;
```
thomaslow: I would prefer if this set is not added as a class property, but kept as a local variable in `preserveLogicalAndPhysical` and passed as parameter to `preserveLogicalAndPhysicalRecursive`. 

It doesn't need to be a class property here and having it as class property means that it could be re-used somehow in other unrelated methods introducing some kind of side-effects. I generally prefer code that keeps all of its different aspects as tight together as possible (similar to functional-style programming).

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/partials/media-list-content.xhtml
```diff
@@ -14,39 +14,46 @@
...
+                    #{msgs.image} #{media.shortId}, #{msgs.page} #{media.orderlabel}
+                </span>
+            </ui:fragment>
+        </div>
+
     </ui:fragment>
 
-    <ui:fragment rendered="#{isPreviewTooltip}">
-        <p:outputPanel id="mediaListImageTooltipTrigger" class="mediaListImageTooltipTrigger thumbnail-overlay">
-            <h:panelGroup>
```
thomaslow, line 44: Did you really see a performance improvement replacing `h:panelGroup` with a `span`?

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/gallery.xhtml
```diff
@@ -157,12 +156,17 @@
...
                                     </p:panel>
-                                    <!-- add one last drop area after the last page in a stripe -->
-                                    <ui:fragment rendered="#{editImages and DataEditorForm.galleryPanel.stripes.getFirst().medias.size() ne 0 and DataEditorForm.galleryPanel.stripes.getFirst().medias.size() - 1 eq currentMedia.rowIndex}">
-                                        <p:panel id="unstructuredPageLastDropArea"
-                                                 styleClass="page-drop-area"/>
-                                    </ui:fragment>
                                 </p:dataList>
+
+                                <!-- Render once instead of checking every row. -->
+                                <ui:fragment rendered="#{
```
thomaslow: This change makes sense from a logical perspective, but it doesn't seem to work. I don't have a drop area at the end of the unstructured images stripe any more.

thomaslow (changes requested):
Your PR seems to work and significantly improve performance of the metadata editor. These are my measurements (based on the initial HTTP request, not actual loading time of the full UI):

- opening a large process (with 2000 images) has improved from 5.59 to 2.21 seconds
- selecting a gallery image has improved from 3.3 to 0.4 seconds
- adding a new structure element has improved from 6.37 seconds to 1.75 seconds
- opening the context menu (right click) has improved from 3.1 seconds to 0.38 seconds

These are very good results and certainly not easy to achieve considering the code complexity of the metadata editor.

As you can see from my code suggestions, I think some parts of your changes are not ideal and could be improved. In general, I think performance improvements often lead to more complicated code, because things that are otherwise easy to calculate in a nice clean function suddenly needs to be split up into additional cache state that needs to be updated in various scenarios and so on. I also realize that this is not a PR intended to refactor the metadata editor and improve its code quality. 

I have tested basic functions, like adding a new structure elements and drag-and-drop, but not everything. Maybe there are scenarios that I missed.

@BartChris Did you check and/or consider the "extended mode" (separate logical and physical tree visualization)? It seems to work for me, but again, it is very time-consuming to test every single function and scenario of the metadata editor.

I haven't really checked the implications of replacing `p:graphicImage` with a simple `img` tag. Images are shown and it seems that the same image URLs are generated. But I don't know whether there are other aspects that are consider by PrimeFaces with `p:graphicImage`, e.g., in terms of security or caching.

@BartChris Overall, I think this PR is an improvement. Please have a look whether it is possible to consider some of my suggestions with reasonable time investment, and fix the superficial things (like missing JavaDoc).

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java
BartChris: Sounds good. I did that.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryStripe.java
BartChris: You are right, i did that.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
BartChris: Good idea, i implemented that!

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/partials/media-list-content.xhtml
BartChris: At one point i experimented with over 5000 pages to challenge performance and this change might stem from that. I reverted it to do not inflate the changes.

Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/gallery.xhtml
BartChris: I tried a different solution which preserves the drag and drop area.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/StructurePanel.java
BartChris: I removed those. This also was a result from my tests with 4000 or more pages, but i think it is better to keep it out for now.

Kitodo/src/main/java/org/kitodo/production/forms/dataeditor/GalleryPanel.java
BartChris: You are right. I tried to address the modelling issue by making `isSelectionDiscontinuous()` a property of `GalleryPanel` instead. Your suggested refactoring sounds good, but I think it would require broader changes.

BartChris:
@thomaslow Thanks a lot for your extensive review and critical assesment of the code. I tried to adress most of your comments, which are as always spot on.

> @BartChris Did you check and/or consider the "extended mode" (separate logical and physical tree visualization)? It seems to work for me, but again, it is very time-consuming to test every single function and scenario of the metadata editor.

This is of course a very good point. My colleagues work exclusively in the combined structure tree and i have only looked once at the "extended mode" during the PR. It looked OK, but you are right that this should be checked more in depth. I know there is a lot of (not always obvious) feature disparity between the two modes so checking the "extendend" tree is some work, but i will do my best.

I agree that performance optimizations are a risk. So i already removed a lot of changes before which seem to bring render times down a little bit more but were not worth an even larger diff. Thanks for catching some unnecessary optimizations as weill. My hope was that some extensive testing by different people would uncover regressions

BartChris:
I currently try to validate the behaviour in the "extended" Editor mode. It looks good on first sight, but after cross-checking on main some features in general seem strange as i have never used it before.

In my tests on current `main`:
When i click "Add Element" in the physical structure it lets me insert logical elements in the physical structure

image

Changing the order of the pages on the left side is impossible. Trying it it the gallery sometimes leads to an exception:

image

(OK this seems to happen for me after first trying to move pages around in the tree on the left, which does not work. Moving in the gallery afterwards then sometimes lead to the exception).

While clicking around i was able to produce that message (i do not yet have an reproducer):

image

I am then forwarded to this error message (seems to be a consequence of `corruptDataWarning.xhtml` referencing `DataEditorForm.templateTaskId`):

https://github.com/kitodo/kitodo-production/blob/ff87578166bd38a7bb0976c30a40ba93e287986d/Kitodo/src/main/webapp/WEB-INF/templates/includes/metadataEditor/dialogs/corruptDataWarning.xhtml#L40

image

Linking elements is not possible also on main, so i do not have to test a regression here. 

Edit:

See: 

https://github.com/kitodo/kitodo-production/issues/7262
https://github.com/kitodo/kitodo-production/issues/7263

thomaslow:
I am not really an expert on how the "extended mode" is supposed to work, but I think the "physical structure" is a completely separate hierarchy, e.g. in case pages are physically stored in different folders (dt. "Mappen"). The gallery still assigns images to logical sections or chapters, the same way as in the "combined mode".

Yes, this is likely a bug in main. The "extended mode" is often forgotten, not really tested, and not used by many (if at all). And so, there are bugs.

PR #7211, opened 2026-08-27
Fix CodeQL Log-related Security Warnings

thomaslow:
This PR addresses the CodeQL security warnings:

- Insertion of sensitive information into log files
- Log Injection

Affected log statements have been removed or secured by escaping parameters.

PR #7242, opened 2026-09-08
Hide login error details from user

### LINKED ISSUE #4769: Reduce the error information in the login form to improve security
Information "user is disabled" give the attacker the information that the login is existing. So some error messages should been logged and change to more general message.

I think that mainly concerns all exceptions which are inheriting of AccountStatusException.
https://docs.spring.io/spring-security/site/docs/4.2.13.RELEASE/apidocs/org/springframework/security/authentication/AccountStatusException.html

Default security error messages in spring:
https://github.com/spring-projects/spring-security/blob/main/core/src/main/resources/org/springframework/security/messages.properties

solth:
Fixes #4769 by writing the original exception message upon login to the log files and presenting the user just with a generic "Bad credentials" message.

Bildschirmfoto 2026-09-08 um 11 31 58


henning-gerhardt (approved):
I tried this changes in the following scenarios:
1. invalid login / login did not exist
2. login with a user name from the deny list
3. login with a disabled user
4. using wrong password for an existing user

In all cases the message was "Bad credentials" in the UI and a detailed error log message was created.

For scenario 1) and 2) the log message itself was clear and useful but for 3) and 4) the log message itself is not helpful with `org.kitodo.production.forms.LoginForm - User is disabled` or `org.kitodo.production.forms.LoginForm - Bad credentials` as you did not know which login / user name was used. In bad case you will now see thousands of login tries in your login but you did not know which login was used. Maybe this can be changed in this pull request or should be opened as a follow up issue with a separate fix.

In general I will approve this changes and suggest improvement can be discussed.

Kitodo/src/main/resources/messages/errors_en.properties
```diff
@@ -116,6 +116,7 @@ errorLoadingMany=Error loading ''{0}''
 errorLoadingOne=Error loading ''{0}'' with ID {1}
 errorLoadingMetadataFile=Error loading metadata file
 ldapGroupInUse=The LDAP group is still associated with users and could not be deleted.
+loginBadCredentials=Bad credentials.
```
stweil, line 119: Technically this message is okay, but I think a more user friendly message would be better. Suggestions:

- "Invalid username or password"
- "The username or password you entered is incorrect"
- "We couldn't sign you in. Check your username and password, or reset your password."

stweil (changes requested):

solth:
> For scenario 1) and 2) the log message itself was clear and useful but for 3) and 4) the log message itself is not helpful with `org.kitodo.production.forms.LoginForm - User is disabled` or `org.kitodo.production.forms.LoginForm - Bad credentials` as you did not know which login / user name was used. In bad case you will now see thousands of login tries in your login but you did not know which login was used. Maybe this can be changed in this pull request or should be opened as a follow up issue with a separate fix.

Yes, I agree that the detailed log message may not be very useful when it does not contain the actual username entered. I wasn't sure, though, if it was better to add it or if that would count as "Insertion of sensitive information into log files", which was one of the security issue reported by CodeQL and resolved by @thomaslow (see #7211)...

Kitodo/src/main/resources/messages/errors_en.properties
solth: I updated the message according to your suggestion.

henning-gerhardt:
@solth : In the other two scenarios the user name will get logged too only in this two mentioned scenarios the user name is missing.

henning-gerhardt:
Both not so helpful messages (`User is disabled` and `Bad credentials`) are created inside the spring security framework. This would take a bigger change (I think so) than needed here in this pull request. This change - if wanted / needed - should be done in a separate pull request.

solth:
> Both not so helpful messages (`User is disabled` and `Bad credentials`) are created inside the spring security framework. This would take a bigger change (I think so) than needed here in this pull request. This change - if wanted / needed - should be done in a separate pull request.

Alright, sounds good. I just found a relatively simple way to add the required information to the log message, but I will open a separate pull request for that then!

Issue #7123, opened 2026-07-23
3.9.2 Snapshot: Editor: Dual image display after having assigned the first structure element


anspallek:
**Describe the bug**
In the "Gallery" of the editor the image(s) of the first structure element is/are shown in the structure view as well as in "unstructured media". Only after the second structure element is assigned, the image(s) of the first structure element is/are substracted from the "unstructured media"

**To Reproduce**
1. Go to Gallery and assign the first structure element
2. (Bug) View the Gallery:  dual image display of the first structure element in the structure view as well as in "unstructured media"
3. Assign the second structure element
4. View the updated "unstructured media": images of the first and second structure element have been substracted

**Expected behavior**
The image(s) of the first structure element is/are immediately substracted from the "unstructured media" like the following structure elements when assigned.

**Screenshots**
2. 
<img width="250" height="373" alt="Image" src="https://github.com/user-attachments/assets/6c0f81a3-c074-4f2d-abed-9f6ba0277c11" />

4. 
<img width="251" height="536" alt="Image" src="https://github.com/user-attachments/assets/2f226d64-5871-447a-964c-08f5b424ea76" />

**Release**
3.9.2 Snapshot:

**Desktop (please complete the following information):**
 - macOS, Windows
 - Chrome, Firefoxe

solth:
Unfortunately, I am unable to reproduce this error in Kitodo 3.9:

https://github.com/user-attachments/assets/a9966504-93c4-455d-9231-22d7d6c16bb0

There are, however, several slighly different ways to assign media to a structure. 

@anspallek maybe you could be more specific in how you created the structure and assigned the media to it? Did you use the context menu of the gallery or of the structure tree? Did you create the structure beforehand or by first selecting the corresponding media?

anspallek:
@solth : I tried both, gallery and structure tree, using the context menu and different browsers and OS. The unstructured media was not updated automatically, only if the browser was manually reloaded or a second structure element was added. Saving didn't work.

Here is a screencast of the gallery way:
1) assigning the structure element: unstructured media not updated
2) saving: unstructured media not updated
3) refreshing the browser: unstructured media **updated**

https://github.com/user-attachments/assets/05a66cb4-7a65-4ecd-a8ed-d69dc5ec2f1d

solth:
[...]

PR #7245, opened 2026-09-09
Handle addition of structure elements in unsaved processes

BartChris:
Fixes https://github.com/kitodo/kitodo-production/issues/7123

PR #7247, opened 2026-09-09
[3.9] Handle addition of structure elements in unsaved processes

BartChris:
Backport of https://github.com/kitodo/kitodo-production/pull/7245
Fixes https://github.com/kitodo/kitodo-production/issues/7123

### DIFF
--- a/Kitodo/src/main/java/org/kitodo/production/metadata/MetadataEditor.java
+++ b/Kitodo/src/main/java/org/kitodo/production/metadata/MetadataEditor.java
@@ -356,7 +356,7 @@ public class MetadataEditor {
                                          LogicalDivision rootStructure) {
         for (View viewToAdd : viewsToAdd) {
             List<LogicalDivision> logicalDivisions = viewToAdd.getPhysicalDivision().getLogicalDivisions();
-            if (logicalDivisions.isEmpty()) {
+            if (logicalDivisions.isEmpty() && Objects.nonNull(rootStructure)) {
                 // In an unsaved process, unstructured media may not yet have
                 // the root logical division registered explicitly.
                 rootStructure.getViews().remove(viewToAdd);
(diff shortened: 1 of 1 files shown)


Kitodo/src/main/java/org/kitodo/production/metadata/MetadataEditor.java
```diff
@@ -352,11 +352,18 @@ private static void handlePosition(Workpiece workpiece, LogicalDivision logicalD
...
     }
 
-    private static void handleViewsToAdd(List<View> viewsToAdd, LogicalDivision newStructure) {
+    private static void handleViewsToAdd(List<View> viewsToAdd, LogicalDivision newStructure,
+                                         LogicalDivision rootStructure) {
         for (View viewToAdd : viewsToAdd) {
             List<LogicalDivision> logicalDivisions = viewToAdd.getPhysicalDivision().getLogicalDivisions();
-            for (LogicalDivision elementToUnassign : logicalDivisions) {
-                elementToUnassign.getViews().remove(viewToAdd);
+            if (logicalDivisions.isEmpty()) {
```
solth: I wonder whether we should have a null check for `rootStructure` and `rootStructure.getViews` here, based on the complex nature of a workpiece and the many NPEs we had to handle lately.

solth (commented):
Works and fixes the problem reported in the linked issue. Just some small remark on the code, see below.

Kitodo/src/main/java/org/kitodo/production/metadata/MetadataEditor.java
BartChris: Yes, probably a good idea to check the `rootStructure` here. A `LogicalDivision` is always initialized with an empty list for `Views`, so i suppose `getViews()` cannot be null.

solth (approved):

Issue #7101, opened 2026-07-01
Custom Columns in process list and task list not working anymore


gerlingsm:
**Describe the bug**
Hello everybody,
Today I noticed that adding ‘Custom Columns’ (processPropertyColumns and taskProcessPropertyColumns) via the kitodo_config.properties file no longer works.
No matter what value I enter there, it is never populated with content in the interface.
A new column with the corresponding name does appear in the interface, but it always remains empty, even though the corresponding metadata field in the metadata editor is filled in correctly.
I’ve tried using the “id”, the “label” and the “label lang=de” names, but in all cases the field in the interface remains empty.
If I remember correctly, it still worked as intended in version 3.8.x.
(Although, if I remember correctly, even back then I found it a bit odd that I’d entered the German translation (label lang=de) of the metadata there rather than the ‘id’.)

**To Reproduce**
1. Create new ruleset with "ruleset_default.xml"
2. Create new docket with "ruleset_default.xsl"
3. Create new workflow (very basic, begin, task1, end should work) "empty_workflow_2026_07_01"
4. Create new process template "test_template_2026_07_01"
5. Create new project "testing_columns"
6. Create new process (I choosed document type: "band", writing "Test123" in all mandatory fields and writing "GermanTest" to the "Sprache" (language) field (see screenshot 1)
7. Open the new created process in the metadata editor and verify that the "Sprache" (language) field contains data (see screenshot 2)
8. Lookup "id" and "label" of the field in the "ruleset_default.xml" file (on the kitodo server): 
"id" is "DocLanguage" and "label" is "Sprache"
9. Open the kitodo_config.properties file (on the kitodo server) and add processPropertyColumns=DocLanguage and taskProcessPropertyColumns=DocLanguage (also tried with "Sprache" instead of "DocLanguage" without success) and restart the tomcat service
10. Go to the process view and open the "Configure columns" menu
11. Choose "DocLanguage"
12. A new column is added, but the content of the column is empty. No error shown. (see screenshot 3 and screenshot 4)

**Expected behavior**
Showing the additional metadata of the field "DocLanguage" in the custom column in processes and tasks view

**Screenshots**
screenshot1:
<img width="1912" height="921" alt="Image" src="https://github.com/user-attachments/assets/6ae98611-d091-421f-b0e3-1f423e81530d" />

screenshot2:
[...]

Issue #4317, opened 2021-03-30
Complete the conversion from properties to metadata


matthias-ronge:
In Production 2, so-called _properties_ were displayed and editable in each process:

![Screenshot 2021-03-30 140530](https://user-images.githubusercontent.com/3040657/112985840-019f3b80-9161-11eb-89b4-68fc79928432.jpg)

They could also be edited in each task:

![Screenshot 2021-03-30 140235](https://user-images.githubusercontent.com/3040657/112985571-a1a89500-9160-11eb-80df-39384d710ba7.jpg)

They could also be edited for a whole batch:

![Screenshot 2021-03-30 141657](https://user-images.githubusercontent.com/3040657/112987096-a1a99480-9162-11eb-99be-9a2503961dea.jpg)

These properties could be configured using a separate configuration file `goobi_processProperties.xml`. The properties were essentially metadata, but process-specific properties could also be noted. In Production 2, the properties could be searched, while search for metadata (in file `meta.xml`) was not possible. Properties gave access to three categories of metadata, namely `from="vorlage"`, `from="werk"` and `from="prozess"`, which correspond to `sourceMD`, `dmdSec` and `techMD` in METS format.

In Production 3, all of this should be summarized as part of a standardization. The aim was to keep all metadata in file `meta.xml`, and this is described in a ruleset file. In it, there are so-called `<acquisitionStage>`s, when which metadata is to be recorded. The aim was, in every workflow step, an acquisition stage can be performed. The search for metadata in `meta.xml` is possible in Production 3 using the search engine function. However, the acquisition stages feature is incomplete because the development time has expired and development is not yet complete. Currently, there is no replacement for the feature function.

The generation of dockets and TIFF-header files still depend on _properties_ and have to be changed.

## To-do

* It should be possible to configure a acquisition stage for each workflow step in the workflow editor (#3155)
* In every task, the metadata of the root element should be visible and editable (as specified by the rule set for this acquisition level) (#3156)
* The old properties should be removed as far as possible. It is not entirely possible because special functions in the Swiss project are pending. However, everything else is unnecessarily complicated and can be removed (#3335)
  * Remove `template` properties.
  * Change the generation of runnotes so that they are based on metadata (requires a different type of docket.xsl).
  * Change generation of TIFF headers
  * Remove properties of the type `workpiece`.
  * Change the relationship of the `properties` to 1:n.
* This answers question #3482.

## Estimated Cost and Complexity

This is a mid-range project for about 10 PT.

[...]

PR #7258, opened 2026-09-18
[3.9] Search all properties for custom column values

solth:
Fixes #7101 by searching _all_ properties - that is template, workpiece and regular process properties - for the contents to display in the custom columns. 

Alternatively, this could have been changed to only searching in _workpiece_ properties, where metadata has been copied to ever since the integration of HibernateSearch, but that would then not consider processes and properties created _before_ an installation was updated to `3.9.x`. Thus searching _all_ properties seems like the best way to find all relevant data and shouldn't impact performance since they are all readily loaded with the corresponding process object.

Although properties are planned to be removed in the future (see for example #4317), I think it is important to keep them and related functionalities working until then.


solth:
@gerlingsm would you mind checking if this patch fixes the reported problem for you?

solth:
@BartChris would you mind reviewing this pull request?

BartChris:
I could test it succesfully, after using the german label from my ruleset, as properties seem to be stored under the german label. I approve the PR, but am quite sceptical of forward porting.

The reason is, that this functionality might be broken and this just fixes it. But by fixing it we also give room to the expectation that all processes generated (independent of the way in which they are created) correctly generate properties. Which i think is not the case.  It also increases confusion when values in the ruleset are changed (and the properties stay the same) and leads to inconsistency as the search index is using the metadata and not the properties.

In my view a lot of pain in discussions comes from features which i consider "ghost features": They exist, but it is unclear whether they are actively used. And sometimes they conflict with other goals.

BartChris (approved):

solth:
@BartChris thank you for your quick response and review!

I think it would be wise not to deem any feature that your institution isn't using a "Ghost feature", though. I personally know of at least 3 institutions that are relying on these additional custom columns. 

As long as we offer a feature, it should work, so not fixing it in later versions seems rather illogical to me. The only alternative would be to _remove_ the feature entirely, but since - as I mentioned above - it _is_ used by institutions removing it is not an option, IMHO. 

I am aware of the problem that properties are not updated when metadata values change, so removing properties should remain our goal. For this particular feature, however, this means once we remove properties, custom columns should not be removed with them, but be refactored to display metadata values rather than property values.

BartChris:
I agree with you that we should preserve that feature as it is very useful. And my colleagues also want this precise feature, as the standard information in the table is too thin. Unfortunately it might be hard to replicate that based only on the meta.xml. So i would LOVE if we find a way to actually make this work with the new infrastructure and focus on how to make that work in a performant way.

> I think it would be wise not to deem any feature that your institution isn't using a "Ghost feature", though.

So my assesment there was not really about the feature, but about the properties in general (which are also a "feature") and it is not about my institution's usage. Me and my colleagues in the @kitodo/kitodo-community-board  spend a LOT of time hunting down ACTUAL usage of features and trying to clarify confusion. Properties are only one example. I think we have the same goal: Keeping necessary features while a the same time work on more clarity, which sometimes includes cutting down on unused features. I think we will tackle the properties again in one of our next meetings.

gerlingsm:
> @gerlingsm would you mind checking if this patch fixes the reported problem for you?

tested on a local build and working as expected / as before with using german label of ruleset 👍
Thanks for the quick fix!

solth:
> So my assesment there was not really about the feature, but about the properties in general (which are also a "feature") and it is not about my institution's usage. Me and my colleagues in the @kitodo/kitodo-community-board spend a LOT of time hunting down ACTUAL usage of features and trying to clarify confusion. Properties are only one example. I think we have the same goal: Keeping necessary features while a the same time work on more clarity, which sometimes includes cutting down on unused features. I think we will tackle the properties again in one of our next meetings.

I think my comment was not phrased very well, sorry for that. I absolutely agree that broken or abandoned features should really be removed and the work you and the Community Board are doing to identify those features is invaluable, but I do think not every institution that uses Kitodo is active in the community and thus may be difficult for you to reach, so it can be difficult to discern whether a feature can really be removed or whether it is still used by someone not interacting with the community. 

Some might even be using this feature with the custom columns without being aware of how it works - they see additional, valuable information in some columns in the task and process lists but do not know the background and configuration of this. Of course, this can lead to misunderstandings when changes to metadata are not reflected in the corresponding extra columns, but that is how the feature always worked and something that should be adressed when the feature is adapted to use actual metadata instead of properties.

It would be great if the @kitodo/kitodo-community-board could talk about the properties again, but I do still believe we should keep properties and related functionalities and features functional as long as they are available, meaning I plan to forward-port this patch for `4.0.x` and the `main` branch (where properties have not yet been replaced with metadata).

BartChris:
> It would be great if the @kitodo/kitodo-community-board could talk about the properties again, but I do still believe we should keep properties and related functionalities and features functional as long as they are available, meaning I plan to forward-port this patch for `4.0.x` and the `main` branch (where properties have not yet been replaced with metadata).

I will of course support the forward ports. The main thing we will strive for in the @kitodo/kitodo-community-board is probably to get a more up to date view on the properties.  https://github.com/kitodo/kitodo-production/issues/4317 is massively outdated and does e.g. not include the feature of this PR as a reason why properties are still needed.

PR #7261, opened 2026-09-21
Use static import statements for assertions

henning-gerhardt:
In all other test classes are static assertions are used.


solth (approved):
Looks good, but just out of curiosity: what's your main intention with this change? Just to bring the imports in line with all the rest of the test cases? Or do the static imports bring any other advantage? Replacing `import org.junit.jupiter.api.Assertions;` with multiple individual imports does require more lines after all...

henning-gerhardt:
I just want to bring up the changed tests in line with all others.

Edit: we can go even the other way but this would require at least once more changes. I have no preference except that we should use one way to keep maintenance easy.

